PR #3, opened 2016-11-21
First take at vault 

SchwarzM:
unclear: 
- global region/domain values ? ( since this is a non region chart )
- pvc ( we need to adopt the existing pv / pvc )


majewsky:
Closing because of old age. Please re-open if you disagree.

PR #8, opened 2016-12-08
HA Postgres with Patroni

BugRoger:
This adds the official patroni chart with slight adaptations...


majewsky:
Closing because of old age. Please re-open if you disagree.

PR #9, opened 2016-12-08
Prometheus alertmanager

auhlig:
small enhancements


auhlig:
That was quick. Thanks.

PR #12, opened 2016-12-13
initial Helm charts for Swift

majewsky:
I have verified through `helm install --dry-run` that this is reasonably close to what prepare_deployment renders. I would appreciate some comments from more experienced Helm users on the structure of the chart.

So far, this is mostly a single big chart; only `swift-drive-autopilot` is split into its own subcharts. I would like to split this further, e.g. into separate subcharts for the storage nodes and the proxy servers, but I need to share values between these, and I cannot see how I could do that in a clean manner (spewing everything into `.Values.global` is anything but clean to me).

Open issues:
- [ ] `sha256sum` of configmaps does not change when values inside the configmap change


openstack/swift/templates/account-caretaker-collect-daemonset.yaml
```diff
@@ -0,0 +1,73 @@
...
+            - account-caretaker-collect
+          # privileged access required for /swift-bin/unmount-helper (TODO: should be removed by https://github.wdf.sap.corp/monsoon/dataservices/issues/186)
+          securityContext:
+            privileged: true
+          env:
+            - name: DEBUG_CONTAINER
+              value: "false"
+            - name: OS_AUTH_URL
+              value: "{{.Values.keystone_auth_url}}"
+            - name: OS_PROJECT_NAME
```
reimannf: If this is for general use, we should make values project  and domain

openstack/swift/templates/account-caretaker-mergify-deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+            - /bin/bash
+            - /swift-bin/swift-start
+            - account-caretaker-mergify
+          env:
+            - name: DEBUG_CONTAINER
+              value: "false"
+            - name: OS_AUTH_URL
+              value: "{{.Values.keystone_auth_url}}"
+            - name: OS_PROJECT_NAME
+              value: "swift_test"
```
reimannf: If this is for general use, we should make values project and domain

openstack/swift/templates/etc/_account-server.conf
```diff
@@ -0,0 +1,40 @@
...
+
+workers = 4
+max_clients = 1024
+backlog = 4096
+log_statsd_host = localhost
+log_statsd_port = 9125
+log_statsd_default_sample_rate = 1.0
+log_statsd_sample_rate_factor = 1.0
+log_statsd_metric_prefix = swift
+log_level = INFO
```
reimannf: Can we make this configurable as in proxy?

openstack/swift/templates/etc/_container-reconciler.conf
```diff
@@ -0,0 +1,52 @@
+[DEFAULT]
+# swift_dir = /etc/swift
+# user = swift
+# You can specify default log routing here if you want:
+# log_name = swift
+# log_facility = LOG_LOCAL0
+# log_level = INFO
```
reimannf: Can we make this configurable as in proxy?

openstack/swift/templates/etc/_object-expirer.conf
```diff
@@ -0,0 +1,19 @@
+[DEFAULT]
+log_level = INFO
```
reimannf: Can we make this configurable as in proxy?

openstack/swift/templates/etc/_object-server.conf
```diff
@@ -0,0 +1,43 @@
...
+
+workers = 12
+max_clients = 1024
+backlog = 8192
+log_statsd_host = localhost
+log_statsd_port = 9125
+log_statsd_default_sample_rate = 1.0
+log_statsd_sample_rate_factor = 1.0
+log_statsd_metric_prefix = swift
+log_level = INFO
```
reimannf: Can we make this configurable as in proxy?

thgrs:
hi stefan, i just changed ceilometer from subcharts to one big one, as the config file checksumming will not work (at least with current knowledge and code) properly with subchart - but i'm still checking, if it will even work with one chart :)

openstack/swift/templates/account-caretaker-mergify-deployment.yaml
majewsky: I'm not sure. We depend on the keystone chart, but since it's not there, I cannot judge how general that one is. If that one has `swift_test` hard-coded into its seed, I don't see a problem with relying on that name over here. If it uses a variable there, I would rather rely on that variable (however that will work).

openstack/swift/templates/etc/_dispersion.conf
```diff
@@ -0,0 +1,10 @@
+[dispersion]
+auth_user = test_admin
+user_domain_name = cc3test
+project_name = swift_dispersion
```
majewsky: See my comment about `swift_test` project name above.

openstack/swift/templates/etc/_object-server.conf
majewsky: done

openstack/swift/templates/etc/_object-expirer.conf
majewsky: done

openstack/swift/templates/etc/_container-server.conf
```diff
@@ -0,0 +1,45 @@
...
+
+workers = 8
+max_clients = 1024
+backlog = 4096
+log_statsd_host = localhost
+log_statsd_port = 9125
+log_statsd_default_sample_rate = 1.0
+log_statsd_sample_rate_factor = 1.0
+log_statsd_metric_prefix = swift
+log_level = INFO
```
majewsky: done

openstack/swift/templates/etc/_container-reconciler.conf
majewsky: done

openstack/swift/templates/etc/_account-server.conf
majewsky: done

openstack/swift/templates/account-caretaker-collect-daemonset.yaml
majewsky: see comment below

openstack/swift/templates/account-caretaker-mergify-deployment.yaml
majewsky: @ruvr Can you comment on what's the most sensible approach here?

urfuwo:
@majewsky we going to merge this before christmas?

PR #13, opened 2016-12-15
Successfully deploys Grafana - to staging at least 

dhague:
- using existing PVC for datadir.  Postgres is coming next...


openstack/grafana/templates/_deployment.yaml
```diff
@@ -0,0 +1,108 @@
+apiVersion: extensions/v1beta1
```
dhague: Note: this file begins with _ so is not deployed, Please ignore contents for now.

openstack/grafana/templates/_pvc.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.persistence.enabled }}
```
dhague: Note: this file begins with _ so is not deployed, Please ignore contents for now.

dhague:
Good point, although this PR does not use postgres (there are some bits left over from when I was trying it). I will take that on board for the next stage though, thanks.

PR #14, opened 2016-12-21
Create backup option for elektra - postgres *review need*

Dexus:
See Topic.

@BugRoger can you lease review this and the secret repo?

Thanks, Josef

### DIFF
--- a/openstack/elektra/charts/postgres/templates/backup-configmap.yaml
+++ b/openstack/elektra/charts/postgres/templates/backup-configmap.yaml
@@ -1,0 +1,14 @@
+{{- if .Values.backup.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: postgres-backup-{{.Values.name}}
+  labels:
+    system: openstack
+    type: configuration
+    component: database
+
+data:
+  postgres.backup.full: {{ .Values.backup.interval_full }}
+  postgres.backup.incr: {{ .Values.backup.interval_incr }}
+{{- end }}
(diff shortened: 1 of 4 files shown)


openstack/elektra/charts/postgres/templates/backup-configmap.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.backup.enabled }}
```
databus23: I don't understand the purpose of this configmap? I don't see where it is being used.

Dexus: Ahh, you are right i forgot to add the backup config directory for the backup container.

databus23: Ok, and whats the reason for these two settings being in a configmap instead of environment variables like the rest?

Dexus: yes, i add it now

databus23: Ok, but why specifically? To me it seems arbitrary that 2 of 11 config options are put in in a seperate file. This requires an additional k8s spec (configmap) and the deployment is more complex (because of the mounting). I think with helm there are way fewer occasions where a configmap is really needed, as the deployment is a template. The only reason I can see for adding a configmap is if a software requires a configfile and can't be configured via environment variables.

Dexus: There is no special reason. If you say it should be an ENV, we change it just in the backup container and delete the configmap.

PR #19, opened 2017-01-03
allow prometheus.yaml to be extended by the regional values.yaml

majewsky:
I need this to make Prometheus scrape our legacy Swift nodes.

### DIFF
--- a/system/kube-monitoring/charts/prometheus-collector/templates/_prometheus.yaml.tpl
+++ b/system/kube-monitoring/charts/prometheus-collector/templates/_prometheus.yaml.tpl
@@ -1,3 +1,4 @@
+{{- define "prometheus_yaml" -}}
 rule_files:
   - ./*.rules
 
@@ -236,3 +237,7 @@ scrape_configs:
 - job_name: 'prometheus-collector'
   static_configs:
     - targets: ['localhost:9090']
+
+{{ .Values.additional_scrape_configs }}
+
+{{- end -}}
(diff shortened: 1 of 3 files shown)


majewsky:
@reimannf FYI

BugRoger:
Hm. I'd prefer if we could use the service discovery to define the scrape targets. I suppose we are not talking about Kubernetes Nodes. This are just external additional static IPs that need to be scraped?

How about a headless service without selector? This won't be assigned a clusterIP and is just a configuration placeholder. Since there is no selector, the endpoints can be configured manually too.

```
apiVersion: v1
kind: Service
metadata:
  name: swift-legacy-nodes
  annotations:
    prometheus.io/scrape: 'true'
    prometheus.io/port: '8080'
spec:
  clusterIP: none
```

Endpoints:
```
apiVersion: v1
kind: Endpoints
metadata:
  name: swift-legacy-nodes
subsets:
  - addresses:
    - ip: 1.2.3.4
    - ip: 1.2.3.5
    - ip: 1.2.3.6
    ports:
      - name: metrics
        port: 8080
        protocol: TCP
```

As a bonus you will end up with DNS resolution for this service.

majewsky:
#TheMoreYouKnow. I didn't know that was even possible. Will try this tomorrow in staging and get back to you about this.

majewsky:
@BugRoger's proposal works fine, so I'll close this.

PR #25, opened 2017-01-09
add some alert definitions for node health (CPU usage, TCP connections)

majewsky:
These alert definitions come from the legacy Swift, but since they apply to the control plane rather than the Swift service, I feel more comfortable putting them in here than into the Swift chart etc.

### DIFF
--- a/global/prometheus-alertmanager/templates/config.yaml
+++ b/global/prometheus-alertmanager/templates/config.yaml
@@ -31,7 +31,7 @@ data:
       - receiver: kubernetes-info
         continue: true
         match_re:
-          service: k8s|etcd|prometheus|dns
+          service: k8s|etcd|prometheus|dns|node
         routes:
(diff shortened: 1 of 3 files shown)


BugRoger:
Cool. Thanks! LGTM

Did you try and deploy it? There's no automatic pipeline yet.

```
helm upgrade kube-monitoring system/kube-monitoring --values ../secrets/staging/values/system/kube-monitoring.yaml --install --namespace kube-monitoring
```

You need to also add the `node` service to the routing tree here:
https://github.com/sapcc/helm-charts/blob/master/global/prometheus-alertmanager/templates/config.yaml#L34-L50

Seems unneeded atm but I want to keep the ability to change the routing based on the `service` label.

majewsky:
Okay, I will test-deploy this tomorrow and merge if everything looks good.

majewsky:
Rolled out in staging and seems to work as intended.

PR #27, opened 2017-01-12
Mirror repos to swift via Jobs/ScheduledJobs

reimannf:
Chart for defining multi mirroring Jobs or ScheduledJobs to Swift.

I underlined the `templates/_job.yaml` and `templates/_scheduledjob.yaml` to skip rendering for now, because the Job is not what we finally want and the scheduled job is not yet available on our clusters.

There is still one thing open with the `swift-http-import`. I created a PR for readding the standards CAs: https://github.com/sapcc/swift-http-import/pull/1. For the staging cluster we either need to create an m3 container image to get SAP NetCA in the alpine or we just add an insecure option to `swift-http-import`.

Test run in Sydney:
https://repo.ap-au-1.cloud.sap/ubuntu/dists/xenial/

//cc @majewsky I guess you want to double check


openstack/swift/charts/swift_http_importer/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+description: swift-http-import
+name: swift_http_import
```
majewsky: chart name in here (`swift_http_import`) is inconsistent with the directory name (`swift_http_importer`) -- I don't know if this causes any trouble, but it's not desirable at least.

openstack/swift/charts/swift_http_importer/templates/configmap.yaml
```diff
@@ -0,0 +1,29 @@
...
+      region_name: {{$.Values.global.region}}
+
+    jobs:
+      {{- range $index, $job := $config.jobs}}
+      - from: {{$job.source}}
+        to:   {{$job.target}}
+      {{- end}}
+
+---
+{{end}}
```
majewsky: You can put everything in one configmap by disambiguating on the name, i.e. `{{$repo}}.yaml` instead of `config.yaml`.

reimannf:
@majewsky Finally refactored to own chart, including your suggestion with the combined configMap

majewsky:
As discussed in Slack, I'd like to have `_job.yaml` enabled since that's what we're currently rolling out (until we upgrade Kubernetes and `ScheduledJob` starts to work). Other than that, lgtm.

majewsky:
Awesome. Please :ship:

PR #32, opened 2017-01-19
Add arc backup

Dexus:
@databus23 can you please review and merge?


openstack/arc/charts/postgresql/templates/deployment.yaml
```diff
@@ -52,8 +52,39 @@ spec:
         volumeMounts:
         - name: data
           mountPath: /postgresql
+{- if .Values.backup.enabled }}
```
databus23: missing {

Dexus:
updated

PR #34, opened 2017-01-19
Add sentry backup

Dexus:
@databus23 can you please review and merge?


system/sentry/charts/postgresql/values.yaml
```diff
@@ -42,3 +42,20 @@ resources:
   requests:
     memory: 256Mi
     cpu: 100m
+
+## Bakcup for postgres
```
databus23: typo

Dexus:
updated

PR #35, opened 2017-01-23
unify swift-drive-autopilot config across clusters

majewsky:
This simplification is made possible by two changes:

1. The autopilot will not complain about an unmatching glob on
`/dev/sd[a-z][a-z]` when there are fewer than 26 drives.

2. The autopilot guarantees (both by documentation and by test coverage)
that swift-ids will always be assigned in order, so we can just put
`swift-01` through `swift-99` into the pool, and it will only use the
lowest-numbered IDs (e.g. `swift-01` through `swift-12` if there are 12
eligible drives).

@d044166 You okay with this? If so, I will merge. (Please don't merge yourself, I need to bump the image version on the autopilot in all regions.)

### DIFF
--- a/openstack/swift/charts/swift_drive_autopilot/templates/configmap.yaml
+++ b/openstack/swift/charts/swift_drive_autopilot/templates/configmap.yaml
@@ -13,7 +13,8 @@ data:
     chroot: /coreos
 
     drives:
-      - {{.Values.drive_glob}}
+      - /dev/sd[a-z]
+      - /dev/sd[a-z][a-z]
 
     # the Swift user and group is UID/GID 1000 in the Kolla containers
     chown:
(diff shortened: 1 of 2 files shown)


reimannf:
@majewsky can you provide the log output from an autopliot with that change? can a operator easily recognise that this will be ok, if there is no device on on sdaa for example?

majewsky:
Log output will be the same as before. There used to be a warning that the second glob (`/dev/sd[a-z][a-z]`) did not match anything, but [I've removed that](https://github.com/sapcc/swift-drive-autopilot/commit/1731956996272aa61a4831dd568887fe0d581b19).

As for the second question, an operator should (upon installation of a new node) only have to check whether `/srv/node` contains the desired number of mountpoints. Only if that expectation fails, will we have to look in detail at what drives were detected, which is easy because there is an initial log message like `event: new drive added: /dev/sdXX` for every drive that's managed by the autopilot.

majewsky:
@reimannf Okay to merge this, and add spare disk support in a later PR?

reimannf:
yes approved it already, but did not merge because you stated I should not.

majewsky:
I did? Huh. Okay, I can't remember why this cannot go in, so I will just see what breaks. :)

PR #44, opened 2017-02-08
add playbook to high-TCP-connections node alert

majewsky:
This commit turned out larger than expected because I want to give the
URL to the operations documentation as a variable, but the *.alerts
files were previously not templated.


system/kube-monitoring/charts/prometheus-frontend/templates/_node.alerts.tpl
```diff
@@ -0,0 +1,65 @@
...
+    service = "node",
+    severity = "warning",
+    context = "availability",
+    dashboard = "kubernetes-node?var-server={{`{{$labels.instance}}`}}"
+  }
+  ANNOTATIONS {
+    summary = "High number of open TCP connections",
+    description = "The node {{`{{ $labels.instance }}`}} has more than 20000 active TCP connections. The maximally possible amount is 32768 connections.",
+    {{ if .Values.ops_docu_url -}}
+    playbook = "{{.Values.ops_docu_url}}/docs/support/playbook/k8s_high_tcp_connections.html",
```
auhlig: Does this work? I think it has to be a label not an annotation according to https://github.com/sapcc/helm-charts/blob/master/global/prometheus-alertmanager/slack.tmpl#L25, or?

system/kube-monitoring/charts/prometheus-frontend/templates/_node.alerts.tpl
```diff
@@ -0,0 +1,65 @@
...
+    service = "node",
+    severity = "critical",
+    context = "availability",
+    dashboard = "kubernetes-node?var-server={{`{{$labels.instance}}`}}"
+  }
+  ANNOTATIONS {
+    summary = "High number of open TCP connections",
+    description = "The node {{`{{ $labels.instance }}`}} will likely reach 32768 active TCP connections within the next hour. If that happens, it cannot accept any new connections.",
+    {{ if .Values.ops_docu_url -}}
+    playbook = "{{.Values.ops_docu_url}}/docs/support/playbook/k8s_high_tcp_connections.html",
```
auhlig: same as above

system/kube-monitoring/charts/prometheus-frontend/templates/_node.alerts.tpl
majewsky: Fixed.

auhlig:
LGTM

PR #45, opened 2017-02-08
move swift rule to sep. file in prom frontend

auhlig:
move rule avg:swift_cluster_storage_used_percent_gauge{} to separate swift.rules in prom. frontend
Looks good @majewsky?

/cc @reimannf More swift rules could go here `system/kube-monitoring/charts/prometheus-frontend/swift.rules`

### DIFF
--- a/system/kube-monitoring/charts/prometheus-collector/aggregation.rules
+++ b/system/kube-monitoring/charts/prometheus-collector/aggregation.rules
@@ -1,3 +1,4 @@
 aggregated:container_cpu_system_seconds_total{} = sum(container_cpu_system_seconds_total) by (id) keep_common
 aggregated:container_cpu_usage_seconds_total{} = sum(container_cpu_usage_seconds_total) by (id) keep_common
 aggregated:container_cpu_user_seconds_total{} = sum(container_cpu_user_seconds_total) by (id) keep_common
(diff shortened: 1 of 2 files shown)


majewsky:
I don't think this is necessary. It's not like these are different concerns. The rules for Swift are still aggregation rules.

auhlig:
If it stays in the collector, then the metric won't be available in the frontend. 
I could rename the file to aggregation.rules

majewsky:
Ah, one is in the collector and the other is in the frontend.

In that case, I reject this proposal for being unecessarily obscure. :) We have already established a pattern with the `aggregated:` metrics, which are aggregated in the collector and pulled into the frontend [here](https://github.com/sapcc/helm-charts/blob/fe88b65fab3d6648f3209886e61a554598fe0ab2/system/kube-monitoring/charts/prometheus-frontend/templates/config.yaml#L35). If we follow that pattern, there are two options:

1. add another match on `__name=~"^avg:.+"` in there
2. rename the metric to `aggregated:swift_cluster_storage_used_percent_gauge` (which is what I would prefer)

auhlig:
Convincing. Then option 2 it will be.

auhlig:
To avoid confusion with existing metrics, the new one is called `aggregated:swift_cluster_storage_used_percent_gauge_average`.
Good to merge @majewsky ?

PR #46, opened 2017-02-08
Prometheus v1.5.1

auhlig:
Use [latest Prometheus](https://github.com/prometheus/prometheus/releases/tag/v1.5.1). It already runs happily in staging. 
Ok @majewsky?


majewsky:
If it looks fine in staging, then ok to merge.

PR #48, opened 2017-02-10
Scheduled jobs

reimannf:


openstack/content-repo/templates/job.yaml
```diff
@@ -1,4 +1,47 @@
 {{- range $repo, $config := .Values.repos }}
+{{- if $config.schedule }}
```
majewsky: There is a lot of code duplication between the then-branch and else-branch that can be easily reduced (the pod spec is the same in both branches).

reimannf:
Tried it. but the needed variables are not easy to forward in a `{{define}}`...

PR #49, opened 2017-02-13
Update the backup options after arc, lyra and elektra updated to helmcharts.

Dexus:
@databus23


openstack/elektra/values.yaml
```diff
@@ -38,4 +38,19 @@ postgres:
     enabled: true
   name: elektra
 #   password:
+  backup:
+    enabled: false
+    repository: sapcc/backup-tools
+    image_version: v0.4.1
```
databus23: Should this be v0.4.6?

common/postgresql/templates/deployment.yaml
```diff
@@ -53,8 +53,39 @@ spec:
...
+        env:
+          - name: MY_POD_NAME
+            value: postgres-backup-{{.Values.name}}
+          - name: MY_POD_NAMESPACE
+            value: {{.Release.Namespace}}
+          - name: OS_AUTH_URL
+            value: {{.Values.backup.os_auth_url}}
+          - name: OS_AUTH_VERSION
+            value: {{.Values.backup.os_auth_version | quote}}
+          - name: OS_IDENTITY_API_VERSION
```
databus23: What is the difference between `OS_AUTH_VERSION` and `OS_IDENTITY_API_VERSION`? Do we really need both?

common/postgresql/values.yaml
```diff
@@ -46,3 +46,19 @@ resources:
   requests:
     memory: 256Mi
     cpu: 100m
+
+backup:
+  enabled: false
+  repository: sapcc/backup-tools
+  image_version: v0.4.6
+  interval_full: 3 days
+  interval_incr: 3 days
```
databus23: Is it correct that these values are both set to 3 days?
Does this mean we only do an incremental backup every 3 days?

Dexus:
- Backup updated to 3 hours
- `OS_AUTH_VERSION` and `OS_IDENTITY_API_VERSION` are similar as of review code of the swift-client. So removed `OS_IDENTITY_API_VERSION`.
- Update version of backup-tools.

PR #59, opened 2017-02-20
Added regex config for mosquitto exporter

ArtieReus:


majewsky:
lgtm, please roll out as discussed in Slack

PR #64, opened 2017-03-03
update ingress controller to 0.9.0-beta.2 and add prometheus scrape annotations to service

auhlig:
***WIP - DO NOT MERGE***

Does it make sense to try [0.9.0-beta.1](https://github.com/kubernetes/ingress/blob/master/controllers/nginx/Changelog.md#09-beta1) of the ingress controller as this version includes https://github.com/kubernetes/ingress/pull/36,
@databus23?
Would test in staging after lunch if you agree.

### DIFF
--- a/system/kube-system/charts/ingress/templates/ingress-controller-daemonset.yaml
+++ b/system/kube-system/charts/ingress/templates/ingress-controller-daemonset.yaml
@@ -44,4 +44,4 @@ spec:
         args:
         - /nginx-ingress-controller
         - --default-backend-service={{.Release.Namespace}}/ingress-default-backend
-        - --nginx-configmap={{.Release.Namespace}}/ingress-controller-configmap
+        - --configmap={{.Release.Namespace}}/ingress-controller-configmap
--- a/system/kube-system/charts/ingress/templates/ingress-controller-daemonset.yaml
+++ b/system/kube-system/charts/ingress/templates/ingress-controller-daemonset.yaml
@@ -11,7 +11,7 @@ spec:
         app: ingress-controller
       annotations:
             prometheus.io/scrape: "true"
-            prometheus.io/port: "9091"
+            prometheus.io/port: "9101"
     spec:
       nodeSelector:
         zone: farm
@@ -47,7 +47,7 @@ spec:
           - name: https
             containerPort: 443
           - name: request-metrics
-            containerPort: 9091
+            containerPort: 9101
           - name: metrics
             containerPort: 10254
         args:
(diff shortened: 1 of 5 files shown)


system/kube-system/charts/ingress/templates/ingress-controller-daemonset.yaml
```diff
@@ -12,7 +12,7 @@ spec:
     spec:
       terminationGracePeriodSeconds: 60
       containers:
-      - image: gcr.io/google_containers/nginx-ingress-controller:0.8.3
+      - image: gcr.io/google_containers/nginx-ingress-controller:0.9.0-beta.1
```
databus23: Why not 0.9.0-beta.2?

system/kube-system/charts/ingress/templates/ingress-service.yaml
```diff
@@ -2,8 +2,10 @@ kind: Service
 apiVersion: v1
 
 metadata:
-  name: ingress-controller 
-
+  name: ingress-controller
+  annotations:
+      prometheus.io/scrape: "true"
+      prometheus.io/port: "9101"
```
MaikuMori: Why this port? Looking at the sources `/metrics` should be on the same port as `/healthz` which is `10254`.

MaikuMori:
Beta.1 has serious bug where it fails to work with services which use port name instead of port as int.

You also need to adopt other changes before merge such as `--nginx-configmap` is now `--configmap`.

system/kube-system/charts/ingress/templates/ingress-service.yaml
databus23: @MaikuMori: We are experimenting with a custom `nginx.tpl` that contains instrumentation using https://github.com/knyar/nginx-lua-prometheus providing requests metrics.

This is not in a public repository yet so this is confusing. Apologies for that.

@auhlig Maybe we should switch the service annotation to the metrics port of the nginx-controller and add another annotation to the pod template for our custom metrics.

system/kube-system/charts/ingress/templates/ingress-controller-daemonset.yaml
```diff
@@ -41,7 +44,9 @@ spec:
             containerPort: 80
           - name: https
             containerPort: 443
+          - name: metrics
+            containerPort: 10254
```
databus23: I think we are missing the port for the request metrics here:

system/kube-system/charts/ingress/templates/ingress-service.yaml
```diff
@@ -14,6 +16,8 @@ spec:
       targetPort: http
     - name: https 
       port: 443 
-      targetPort: https 
+      targetPort: https
+    - name: metrics
```
databus23: I think this is not needed here. prometheus scrapes the endpoints directly

databus23:
Applied manually in staging. LGTM

PR #70, opened 2017-03-15
set dabase dimension name as all other pg metrics. Added unit bytes to the metric name.

ArtieReus:
@databus23 check please the changes and merge asap :)


ArtieReus:
We need the same name dimensions 'datname' when creating the prometheus -> monasca mapping

databus23:
LGTM

PR #72, opened 2017-03-16
qoute backup values

Dexus:


Dexus:
@reimannf

openstack/backup-replication/values.yaml
```diff
@@ -10,7 +10,7 @@ image_version: v0.1.9
 # backup_replication_cronjob: "5 */8 * * *"
 
 # this value are seconds 864000 are 10 days
-backup_expire_after: 864000
+backup_expire_after: "864000"
```
reimannf: this is not needed any longer, or?

Dexus: no not really. will fix it now

PR #75, opened 2017-03-23
ingress controller: enable virtual host traffic status module

auhlig:
No longer region-specific. OK @databus23?


databus23:
LGTM

PR #79, opened 2017-03-27
Backup minify config settings

reimannf:


common/postgresql/values.yaml
```diff
@@ -60,13 +60,12 @@ resources:
...
-  interval_full: "3 hours"
+  image_version: v0.5.6
+  interval_full: 1 hours
   os_auth_url: DEFINED-IN-REGION-SECRETS
-  os_auth_version: "3"
-  os_username: DEFINED-IN-REGION-SECRETS
-  os_user_domain: DEFINED-IN-REGION-SECRETS
-  os_project_name: DEFINED-IN-REGION-SECRETS
-  os_project_domain: DEFINED-IN-REGION-SECRETS
   os_region_name: DEFINED-IN-REGION-SECRETS
```
majewsky: `{{.Values.backup.os_region_name}}` can be replaced by `{{.Values.global.region}}`. That value should be provided by every top-level chart's `values.yaml` file. (If not, we need to add it.)

reimannf: not sure. the global part is fairly unclear so far. e.g arc does not declare the region in global. I would keep it in that fashion until we have a real global agreement

majewsky: I see. Agreed.

PR #84, opened 2017-03-29
Tuning kube monitoring

auhlig:
Tested  in staging.


system/kube-monitoring/charts/prometheus-collector/prometheus.yaml
```diff
@@ -232,6 +232,11 @@ scrape_configs:
     target_label: __address__
     regex: ([^:]+)(:\d+)?
     replacement: ${1}:9100
+  - source_labels: [mountpoint]
```
auhlig: Shouldn't this work @BugRoger?

system/kube-monitoring/charts/prometheus-collector/aggregation.rules
```diff
@@ -4,3 +4,9 @@ aggregated:container_cpu_user_seconds_total{} = sum(container_cpu_user_seconds_t
 aggregated:swift_cluster_storage_used_percent_gauge_average{} = avg(swift_cluster_storage_used_percent_gauge) by (cluster) keep_common
 aggregated:prometheus_local_storage_ingested_samples_total_rate{} = rate(prometheus_local_storage_ingested_samples_total[5m])
 aggregated:pg_database_size_bytes_gauge_average{} = avg(pg_database_size_bytes) by (cluster,kubernetes_name,datname)
+aggregated:node_filesystem_avail{} = label_replace(node_filesystem_avail, "mountpoint", "$1", "mountpoint", "/host(/.*)")
```
auhlig: those rules replace the `/host`prefix in the label value. Altough it would be better to use the relabel_config in the prometheus config, I commented below.

system/kube-monitoring/charts/prometheus-frontend/templates/_kubernetes.alerts.tpl
```diff
@@ -321,3 +317,29 @@ ALERT KubernetesTooManyOpenFiles
     summary = "Too many open file descriptors",
     description = "{{`{{$labels.job}}`}} on {{`{{$labels.instance}}`}} is using {{`{{$value}}`}}% of the available file/socket descriptors.",
   }
+
+ALERT HighNumberOfGoRoutines
+  IF irate(go_goroutines{job="kube-system/kubelet"}[5m]) > 20
```
auhlig: Makes sense @BugRoger?

auhlig:
Before merge: Would like to remove the aggregation rules and use the relabel_config before merging, but not sure why the latter isn't working.

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -49,7 +49,7 @@ data:
         match_re:
           service: k8s|etcd|prometheus|dns|node
           cluster: {{ . }}
-          severity: critical
+          severity: warning,critical
```
BugRoger: Remove please. Too much spam in region channels with warnings being posted there as well.

auhlig:
**Open**: The relabeling as commented above doesn't work.

system/kube-monitoring/charts/prometheus-collector/prometheus.yaml
BugRoger: This might work. Not 100% sure. You either need this rule or the aggregated rules.

system/kube-monitoring/charts/prometheus-collector/aggregation.rules
BugRoger: If it works during initial ingestion, you can remove this.

system/kube-monitoring/charts/prometheus-frontend/templates/_kubernetes.alerts.tpl
BugRoger: Generally it's recommened to use `rate` for alerting. 

> irate should only be used when graphing volatile, fast-moving counters. Use rate for alerts and slow-moving counters, as brief changes in the rate can reset the FOR clause and graphs consisting entirely of rare spikes are hard to read.

Looks like a good heuristic. Though I could image quite a few false positives. What would you think about something like:

```
go_goroutines{job="kube-system/kubelet"} > avg_over_time(go_goroutines{job="kube-system/kubelet"}[12h] offset 12h) * 2
```

This would alert if there is suddenly twice as many goroutines than over the average of the last 12 hours ago.

global/prometheus-alertmanager/templates/config.yaml
auhlig: Done

system/kube-monitoring/charts/prometheus-frontend/templates/_kubernetes.alerts.tpl
auhlig: Thanks. Changed it.

system/kube-monitoring/charts/prometheus-collector/prometheus.yaml
auhlig: Will test. I might need to come back to this again.

auhlig:
Incorporated comments. Merging after discussion via IM.

PR #85, opened 2017-03-30
Renaming service to service_type

rajivmucheli:
Adding service_type : hypervisor_type, this will helps us distinguish between KVM, QEMU, VMWare.


thgrs:
please discuss the initial service -> service_type change with @rajivmucheli - the other changes were only tries to get out of the problem, which most probably came from the dimension renaming, but i think we need to fix this on influxdb side as discussed

PR #86, opened 2017-03-30
Node problem detector

auhlig:
Initial version. Currently playing with single deployment of NPD in staging. 
Will later deploy to staging via helm, if there are not objections from your side @BugRoger?


system/kube-system/charts/node-problem-detector/templates/daemonset.yaml
```diff
@@ -0,0 +1,54 @@
...
+    metadata:
+      labels:
+        app: node-problem-detector
+    spec:
+      containers:
+      - name: node-problem-detector
+        command:
+        - /node-problem-detector
+        - --logtostderr
+        - --system-log-monitors=/config/kernel-monitor.json
```
auhlig: This is only for journald. Could also add support for syslog, if required using
```
["/bin/sh", "-c", "/node-problem-detector --logtostderr --system-log-monitors=/config/kernel-monitor.json,/config/kernel-monitor-filelog.json >>/var/log/node-problem-detector.log 2>&1"]

```

auhlig:
Getting `failed to update node conditions: Operation cannot be fulfilled on nodes "minion0...": <nil>`, which leads to an [open issue](https://github.com/kubernetes/node-problem-detector/issues/65).

BugRoger:
No objections! Looks good.Well, at least the error is not critical.

> This is a known race, and should not cause any problem, as long as NPD eventually gets a chance to update.

Does it eventually manage to add one of the new conditions?

BugRoger:
Is the 0.3.0 release something you have built?

auhlig:
I grabbed the latest [release](https://github.com/kubernetes/node-problem-detector/releases/tag/v0.3.0).

auhlig:
The NPD looks good so far:
```
I0330 14:20:47.894863       1 log_monitor.go:72] Finish parsing log monitor config file: {WatcherConfig:{Plugin:journald PluginConfig:map[source:kernel] LogPath:/var/log/journal Lookback:5m} BufferSize:10 Source:kernel-monitor DefaultConditions:[{Type:KernelDeadlock Status:false Transition:0001-01-01 00:00:00 +0000 UTC Reason:KernelHasNoDeadlock Message:kernel has no deadlock}] Rules:[{Type:temporary Condition: Reason:OOMKilling Pattern:Kill process \d+ (.+) score \d+ or sacrifice child\nKilled process \d+ (.+) total-vm:\d+kB, anon-rss:\d+kB, file-rss:\d+kB} {Type:temporary Condition: Reason:TaskHung Pattern:task \S+:\w+ blocked for more than \w+ seconds\.} {Type:temporary Condition: Reason:UnregisterNetDevice Pattern:unregister_netdevice: waiting for \w+ to become free. Usage count = \d+} {Type:temporary Condition: Reason:KernelOops Pattern:BUG: unable to handle kernel NULL pointer dereference at .*} {Type:temporary Condition: Reason:KernelOops Pattern:divide error: 0000 \[#\d+\] SMP} {Type:permanent Condition:KernelDeadlock Reason:AUFSUmountHung Pattern:task umount\.aufs:\w+ blocked for more than \w+ seconds\.} {Type:permanent Condition:KernelDeadlock Reason:DockerHung Pattern:task docker:\w+ blocked for more than \w+ seconds\.}]}
I0330 14:20:47.894996       1 log_watchers.go:40] Use log watcher of plugin "journald"
I0330 14:20:47.896043       1 log_monitor.go:81] Start log monitor
I0330 14:20:47.898964       1 log_watcher.go:69] Start watching journald
I0330 14:20:47.898994       1 problem_detector.go:74] Problem detector started
I0330 14:20:47.899132       1 log_monitor.go:173] Initialize condition generated: [{Type:KernelDeadlock Status:false Transition:2017-03-30 14:20:47.899110327 +0000 UTC Reason:KernelHasNoDeadlock Message:kernel has no deadlock}]
```
This is followed occasionally by the sync error discussed above.

BugRoger:
This lists all nodes with a KernelDeadlock and the reason:
`kubectl get nodes -o template --template='{{range $node := .items}}{{range .status.conditions}}{{if and (eq .type "KernelDeadlock") (eq .status "True")}}{{$node.metadata.name}}: {{.reason | println}}{{end}}{{end}}{{end}}'`

We just had an unregistered netdevice and the node actually become `NotReady` from it. The `KernelDeadlock` condition was false though. It might have written an event but I forgot to check before rebooting. We should change `unregistered_netdevice` back to type `permanent` in the configuration file. Yes, it's not a real kernel hang but we can't monitor for the node condition if it just writes an event in the log.

auhlig:
Makes sense. Changed unregistered_netdevice to permanent.

We could test, by injecting the error in the log `[<timestamp>] unregister_netdevice: waiting for lo to become free. Usage count = 1`. The NPD should notice and change the condition accordingly.

BugRoger:
lgtm

BugRoger:
> We could test, by injecting the error in the log [<timestamp>] unregister_netdevice: waiting for lo to become free. Usage count = 1. The NPD should notice and change the condition accordingly.

`echo "aklskladkl" | systemd-cat`

PR #87, opened 2017-03-30
Update monasca-seed.yaml

thgrs:


ruvr:
Since monasca_agent is not a user of the ccadmin domain, I'ld prefer:

domains:
- name: ccadmin
projects:
- master
roles:
- user: monasca_agent@Default
 role: monasca-agent

PR #92, opened 2017-04-07
[swift] seed also during `helm install`

reimannf:


thgrs:
did you discuss this with @ruvr ? - i think he intentionally did it this way to have a working initial bootstrapping of the system, i.e. making sure that the keystone structures are there before the seeding starts ...

ruvr:
yes, that is ok. during region bootstrap we'll just have to take care that the openstack-operator is deployed before everything else.

PR #97, opened 2017-04-19
bump prometheus to v1.6.1

auhlig:
**WIP - DO NOT MERGE**

Prometheus v1.6.x deprecated the flags `-storage.local.memory-chunks` & `-storage.local.max-chunks-to-persist` in favor of `-storage.local.target-heap-size`.
https://prometheus.eu-de-1.cloud.sap/graph?g0.range_input=1h&g0.expr=container_memory_usage_bytes%7Bpod_name%3D~%22prometheus-.*%22%7D%2F1024%2F1024%2F1024&g0.tab=1 could give us an indication what could be reasonable values here. 
What do you think @BugRoger?

### DIFF
--- a/global/prometheus/values.yaml
+++ b/global/prometheus/values.yaml
@@ -1,4 +1,3 @@
-image: prom/prometheus:v1.5.2
+image: prom/prometheus:v1.6.1
 retention: 2160h0m0s # 90 days
-memory_chunks: "4194304"
-max_chunks_to_persist: "2097152"
+target_heap_size: "53687091200"
--- a/global/prometheus/templates/petset.yaml
+++ b/global/prometheus/templates/petset.yaml
@@ -20,8 +20,7 @@ spec:
           args:
             - -config.file=/etc/prometheus/prometheus.yaml
             - -storage.local.path=/prometheus
-            - -storage.local.memory-chunks={{.Values.memory_chunks}}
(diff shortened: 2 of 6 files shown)


global/prometheus/values.yaml
```diff
@@ -1,4 +1,3 @@
-image: prom/prometheus:v1.5.2
+image: prom/prometheus:v1.6.1
 retention: 2160h0m0s # 90 days
-memory_chunks: "4194304"
-max_chunks_to_persist: "2097152"
+target_heap_size: "53687091200"
```
auhlig: I guess this should be bigger?

BugRoger:
The previous parameters have been tuned so that 1 week of data fits comfortably into memory. I think it ended up to be something around 4GB... Can't tell without trying. 

Do we need to wipe the existing data? Let's just try and see, the new data model is supposed to be quite a bit more efficient too.

auhlig:
It shouldn't wipe existing data. 

According to the query from my last comment, prometheus-collector uses 2.4 Gib, frontend ~17(!) Gib, and global just ~7Gib. Can this be correct? Confusing -even more in staging.

BugRoger:
Sounds like a lot. Worth investigating...

auhlig:
Will give it a go in staging with above (quite high) values. Will reduce afterwards if possible.

auhlig:
Runs in staging. Merging, since this will be helpful. Will likely reduce `heap_size` in the future.

PR #98, opened 2017-04-24
[swift] statsd_exporter daemonset and HostNetwork

reimannf:
* statsd_exporter daemonset
* all daemonsets use HostNetwork
* template daemonset_annotation for tolerations
* template statsd_annotation for prometheus scraping
* use prometheus annotation and std container also for proxy
* Remove enable_statsd flag

### DIFF
--- a/openstack/swift/templates/_utils.tpl
+++ b/openstack/swift/templates/_utils.tpl
@@ -19,10 +19,12 @@ We truncate at 24 chars because some Kubernetes name fields are limited to this 
 {{- /**********************************************************************************/ -}}
 {{- define "swift_daemonset_annotations" }}
 scheduler.alpha.kubernetes.io/tolerations: '[{"key":"species","value":"swift-storage"}]'
-{{- if .Values.enable_statsd }}
+{{- end -}}
+
+{{- /**********************************************************************************/ -}}
+{{- define "swift_prometheus_annotations" }}
 prometheus.io/scrape: "true"
 prometheus.io/port: "9102"
-{{- end }}
 {{- end -}}
 
 {{- /**********************************************************************************/ -}}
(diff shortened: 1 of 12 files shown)


BugRoger:
Why does the `statsd-exporter` need to run in `hostNetwork` mode?

majewsky:
Because basically everyone else needs to do that (the servers need to avoid the docker-proxy, and the replicators look at `ip a` to establish their node's identity and are thus confused by pod-specific IPs), and they all need to report statsd metrics somewhere.

Now that could also be achieved by having the exporter expose its statsd port as a hostPort, but we found it valuable to have it run in host-network mode as well because we can see the host IP (rather than a pod IP) in the `instance` label on the corresponding Prometheus metrics.

BugRoger:
> Because basically everyone else needs to do that
With that logic you could just run everything `privileged` and in `hostNetwork`.

I'm just being annoying about this, because it's not a sustainable practice. If everyone just schedules everything in `hostNetwork` just because, we are back at managing ports and their conflicts.

> instance label on the corresponding Prometheus metrics.
Or you get the node in the node label ;)

majewsky:
Which node label? https://prometheus.staging.cloud.sap/graph?g0.range_input=1h&g0.expr=swift_object_auditor_timer&g0.tab=1

BugRoger:
image
Ok, on this specific metric the `instance` is the actual pod and the `node` is not broken out. Easy to add as a label if you want it.

reimannf:
_exceptional_ approval for `hostNetwork` from @BugRoger. We do not use it for getting node ips to prometheus. We do it to avoid docker-proxy when `HostPorts` are specified. The nature of swift-rings requires that (replication)-ports to be static for account, container and object server and we have dedicated nodes with taints for swift. In general that should not be a pattern bleeding into other applications.

PR #101, opened 2017-05-02
Repo cleanup

reimannf:
New options for swift-http-import:

* except
* only
* immutable

Some naming clean ups.

### DIFF
--- a/openstack/content-repo/templates/configmap.yaml
+++ b/openstack/content-repo/templates/configmap.yaml
@@ -36,17 +36,20 @@ data:
     {{end}}
     jobs:
       {{- range $index, $job := $config.jobs}}
-      - from: {{$job.from}}
-        to:   {{$job.to}}
+      - from:      {{$job.from}}
+        to:        {{$job.to}}
         {{- if $job.cert}}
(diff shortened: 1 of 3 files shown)


openstack/content-repo/templates/configmap.yaml
```diff
@@ -36,14 +36,25 @@ data:
...
+        except:    {{$job.except}}
+        {{- end}}
+        {{- if $job.only}}
+        only:      {{$job.only}}
+        {{- end}}
+        {{- if $job.immutable}}
+        immutable: {{$job.immutable}}
+        {{- else if $config.immutable}}
+        immutable: {{$config.immutable}}
         {{- end}}
```
majewsky: Since you're just passing through the `$config.jobs` into the generated YAML, you can use the `toYaml` function to greatly simplify this. See https://github.com/sapcc/helm-charts/blob/2c0bb02b8c710d2ad3a50e07a5ec0a53d77cb45e/openstack/limes/templates/configmap.yaml#L23-L24

majewsky:
Merge when you're ready.

PR #103, opened 2017-05-04
[swift] adding monasca compliant async pending rate

reimannf:


openstack/monasca/templates/etc/monasca-agent/conf.d/_prometheus.yaml.tpl
```diff
@@ -174,3 +174,5 @@ instances:
                dimensions:
                    cluster: cluster
                    service: service
+           swift:
+               gauges: [ 'swift_(async_pendings)_rate' ]
```
majewsky: You need to add the `kubernetes_namespace` to the `match_labels` field near line 126.

reimannf: done

PR #108, opened 2017-05-17
monitor alertmanagers

auhlig:
Do we want to get some alertmanagers metrics @BugRoger?
Note: I might need to add an annotation to the AM.


auhlig:
Better via scrape annotations. This would also allow automated discovery of alertmanagers.

PR #110, opened 2017-05-18
Maia prometheus

auhlig:
Prometheus for maia in flat structure. /cc @Kuckkuck @thgrs 
I reorganized the values a bit. Now it's `global`,`prometheus`,`vcenter_exporter`.
TODO:
- [x]  use prometheus service discovery rather than config. static targets (will do tomorrow)
- [x] certificates for maia prometheus


urfuwo:
+1

thgrs:
i have added dummies for the missing prometheus.ca and -.token values to staging/values/maia.yaml in the maia branch there. will fill them with @auhlig when he is back in the office. this way at least the helm chart from this pull request would run through.

auhlig:
Finally works with ca.cert, bearer_token for both service discovery and scraping the targets (pod,endpoints,svc) in maia namespace.
So this and the secrets PR is good to merge from my side. Objections @thgrs @Kuckkuck ?

PR #124, opened 2017-06-06
Hermes: Make logstash resilient to RabbitMQ connection errors

dhague:
@Kuckkuck are you OK with this? Hopefully means we don't have to keep restarting the pod when there's a RabbitMQ glitch...

### DIFF
--- a/openstack/hermes/templates/etc/_logstash.conf.tpl
+++ b/openstack/hermes/templates/etc/_logstash.conf.tpl
@@ -7,7 +7,9 @@ rabbitmq {
     queue => "{{.Values.hermes_rabbitmq_queue_name}}"
     subscription_retry_interval_seconds => 60
     id => "logstash_hermes"
-    automatic_recovery => false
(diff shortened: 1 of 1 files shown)


Kuckkuck:
I know this will work if it is a connection error, but on kubernetes the rabbitmq is sending a clean shutdown, where this switch won't work.

dhague:
But it's a start, right? The current logstash pod was only up for 24h30m before getting this:
`RabbitMQ connection was closed! {:url=>"amqp://openstack:XXXXXX@rabbitmq.monsoon3.svc.kubernetes.staging.cloud.sap:5672/", :automatic_recovery=>false, :cause=>com.rabbitmq.client.ShutdownSignalException: connection error}`
So maybe we improve things a little bit, and then we look at the next stoppage to see how we can improve them further?

dhague:
Hm, just saw your message at https://discuss.elastic.co/t/no-reconnect-after-shutdown-running-logstash-rabbitmq-input-plugin-in-docker-kubernetes/86139 - no reply yet though. Dammit. :)

PR #129, opened 2017-06-09
[repo] seed a repo user and use it

reimannf:


reimannf:
@ruvr is the user name `repo` ok or do we have a kind of naming scheme?

ruvr:
fine with me

PR #130, opened 2017-06-09
[elektra] make dashboard user real cloud admin

Carthaca:
@ruvr syntax question: do I really have to always repeat the project?


ruvr:
yes. it is a list of role-assignment tuples

Carthaca:
ok, thx

PR #134, opened 2017-06-15
Create openstack-swift.alerts

BugRoger:
How something like this? I think, a prediction is better than absolute alerts.

It will start warning us 30 days before storage reaches 80% capacity. It extrapolates based on the increase of the last 7 days.

The current alert in eu-de-1 for above 50% will never go away. The trend here is +1% per week. Until it reaches WARNING level we will get the INFO every 12h for the next 10 weeks 😄

### DIFF
--- a/system/kube-monitoring/charts/prometheus-frontend/openstack-swift.alerts
+++ b/system/kube-monitoring/charts/prometheus-frontend/openstack-swift.alerts
@@ -133,62 +133,17 @@ ALERT OpenstackSwiftRateLimit5000
     description = "Check kibana for the causing account and container",
   }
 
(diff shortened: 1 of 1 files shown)


reimannf:
I think I will just merge that for now. I would feel better with s.th. like `IF max(predict_linear(swift_cluster_storage_used_percent_gauge[1w], 60*60*24*30*2)) by (region) > 0.66` but that we reached already. 
@RohithChandran we need to have a closer look to the capacity in all regions, because the alert will be too late. maybe we get email alerts at some point and define a stricter alert for the swift dl.
Many thanks @BugRoger

PR #141, opened 2017-07-11
federate snmp exporter metrics from maia prometheus

thgrs:
@BugRoger @Kuckkuck - can you please double check and pull if it looks ok to you?

i'm federating the snmp_asr_* metrics from the maia prometheus, as the snmp exporter is a bit special in that it requires some arguments in the exporter url (the snmp machine to get the data from and the corresponding config name) and does some special rewriting (so that the snmp host appears as the instance of the resulting metrics and not the snmp exporter itself) - all this is default snmp-exporter mode. in case we would want to scrape it from the normal prometheus collector, we would need the required info about the snmp parameters, which lives in the maia secret. thus for now it is easier to scrape the snmp-exporter via the maia prometheus and federate from there to the prometheus frontend.

### DIFF
--- a/system/kube-monitoring/charts/prometheus-frontend/templates/config.yaml
+++ b/system/kube-monitoring/charts/prometheus-frontend/templates/config.yaml
@@ -111,6 +111,33 @@ data:
         - targets:
           - 'prometheus-collector.{{ .Release.Namespace }}:9090'
 
+    - job_name: 'prometheus-maia-federation'
+      scrape_interval: 30s
+      scrape_timeout: 25s
+
+      honor_labels: true
+      metrics_path: '/federate'
+
+      params:
+        'match[]':
+          - '{__name__=~"^snmp_asr_.+"}'
+
+      relabel_configs:
+        - action: replace
+          target_label: region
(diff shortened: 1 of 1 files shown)


BugRoger:
Can we clarify where this data should be held in the end? Are we intending to expose SNMP metrics to customers? I would base our decision who is scraping on that. Maybe even scrape twice?

How much data are we talking about here? It looks like a lot of time series. Would be good to just federate what we actually need.

jobrs:
We have some requests for ELB metrics from customers (HCM, PMP). I have to check their requirements again (there is some Excel floating around).

@thgrs Do you have an idea of the meaning of the metrics and which of them could be mapped to (affected) tenants with limited effort? For our internal customers, we do not need a huge number of metrics. We can start with a few ones that need to be well understood and relevant for tenant performance/avail.

jobrs:
@thgrs can this one be closed?

thgrs:
yes - i think this can be closed: we can (and do) scrape the snmp exporter from both the regular and the maia prometheus as needed meanwhile

PR #142, opened 2017-07-13
seed temp_url key for automation-artifacts swift container

reimannf:
@databus23 is that something for you? @ruvr`s seeder supports now swift container meta data.


databus23:
OMG! Looks sowas von good to me.

PR #143, opened 2017-07-20
ing ctrl deployment instead of ds

auhlig:
As discussed @databus23.
If this looks in general good to you, I would test in staging, before merging.

### DIFF
--- a/system/kube-system/charts/ingress/templates/ingress-controller-deployment.yaml
+++ b/system/kube-system/charts/ingress/templates/ingress-controller-deployment.yaml
@@ -1,17 +1,21 @@
 apiVersion: extensions/v1beta1
-kind: DaemonSet 
-
+kind: Deployment
 metadata:
   name: ingress-controller
-
+  namespace: kube-system
+  labels:
+    app: ingress-controller
 spec:
+  selector:
+    matchLabels:
+      app: ingress-controller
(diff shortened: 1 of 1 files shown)


system/kube-system/charts/ingress/templates/ingress-controller-deployment.yaml
```diff
@@ -1,35 +1,47 @@
...
-kind: DaemonSet 
-
+kind: Deployment
 metadata:
   name: ingress-controller
-
+  namespace: kube-system
+  labels:
+    app: ingress-controller
 spec:
```
databus23: This is missing `replicas: 3` or something.
Also we need to ensure that only one pod is scheduled per master.
Specifying a `hostPort` should achieve this. E.g.: https://github.com/sapcc/helm-charts/blob/47b38505d2974cecb6d861286af3d56e201000d6/openstack/elk/templates/fluent-daemonset.yaml#L61-L62

auhlig: Grr.. Yes. We already talked about this. Better now with bd1bb1d0438e5ca5762c19063e4be40611c737ac ?

auhlig:
Will merge and deploy on monday.

PR #144, opened 2017-07-27
prometheus frontend as deployment

auhlig:
As simple as this? Ok to test in staging?
Please **do not merge** yet.


BugRoger:
Looks good from my side. Good catch @majewsky.

You can try in staging. If we lose that 30GB of data, it would probably be a good thing even... :D

majewsky:
It definitely needs to be tested in staging, but Arno should coordinate at least with @reimannf while the Swift load tests are in progress.

auhlig:
Sure. I've sent a mail.
Rolling this out to staging is scheduled for tomorrow morning

auhlig:
works fine. also able to reuse existing pvc.

PR #145, opened 2017-08-01
add maia's monitoring_admin role to test_admin

jobrs:


jobrs:
... so we can test basic availability of service and exporters

ruvr:
This should go into the maia service seed. Otherwise the domain seed will fail, since the monitoring* roles haven't been defined (at initial deployment of the seed).

Please also see https://github.com/sapcc/helm-charts/tree/master/openstack/domain-seeds for explanation.

jobrs:
Indeed, that's no good. After reading the domain-seeds readme I am still not sure whether I can add roles to an existing user that way. When I redeclare the user test_admin, is it sufficient to mention only added roles (additive behaviour) or do I have to specify the full set of roles (replacing behaviour)?

ruvr:
You just need to make sure that the user has been created by another seed (in this case the domain-cc3test-seed) by adding a 'requires monsoon3/domain-cc3test-seed' in your maia seed.
Then you can just add (maia) role-assignments to the user in the seed (since the user has been created by the seed(s) listed in requires).
You obviously only add stuff that maia needs (or provides in this case). Everything else should not be the concern of the maia seed, but some other service seed..

jobrs:
Okay, done. What about the keystone-seed? Is it available meanwhile or should it remain left out as before?

jobrs:
👍

PR #146, opened 2017-08-02
seed new CC-Test Regression Project

artherd42:
dedicated project for the CC regression tests


ruvr:
This will not work due to referrals to roles that don't exist during a initial deployment of the seed. 
It already broke with the previous change. The seed is only allowed to use admin/member roles (that are provided by the keystone bootstrap). I just did not find the time to repair (revert) it. It will block the buildout of a new region because of this.
Everything else needs to go into another seeds that requires the sone and everything else referred to (service seeds that provide the roles)
See also the readme.md in https://github.com/sapcc/helm-charts/tree/master/openstack/domain-seeds.

artherd42:
@ruvr Do you want me to move the cc3test domain seed to the blackbox-test chart?

PR #148, opened 2017-08-08
Add Database Backup metrics

nobse:


majewsky:
Adding the principal contributors to the `common/postgres` chart as reviewers. The intent is to merge these changes into there and into all vendored instances of it.

reimannf:
I would now jump over all regions and pull the new backup image already

PR #149, opened 2017-08-09
Rate limiting option for nginx before hitting swift

reimannf:
Not sure whether this will ever be activated. That will not protect the masters, but at least we can protect the swift backend.


majewsky:
Did you check that our nginx contains the `ngx_http_limit_conn` module? (To check, pull the image and run `nginx --version` or something like that.)

reimannf:
As far as I understood modules `limit_conn` and `limit_req` are enabled by default. I tested both options temporary in staging with value of 1. It worked.

PR #155, opened 2017-08-17
DO-NOT-MERGE-YET: preparations to change the grafana postgres from a petset to a deploy…

thgrs:
…ment

### DIFF
--- a/openstack/grafana/values.yaml
+++ b/openstack/grafana/values.yaml
@@ -41,6 +41,9 @@ postgres:
     os_password: DEFINED-IN-REGION-SECRETS
 persistence:
   enabled: false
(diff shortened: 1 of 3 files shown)


openstack/grafana/values.yaml
```diff
@@ -41,6 +41,9 @@ postgres:
     os_password: DEFINED-IN-REGION-SECRETS
 persistence:
   enabled: false
+  name: grafana-prometheus-storage
```
auhlig: name should be `storage-grafana-postgres-0` if you want to use the existing one.

thgrs:
thx @auhlig - adjusted

PR #157, opened 2017-08-18
Bump default backup image for limes to v0.5.11

nobse:


reimannf:
Will merge after backup image is on hub.docker and cached in all regions

nobse:
We need another bump to v0.5.12 anyway...

PR #162, opened 2017-08-22
Creating OpenstackCinderVolumeStuck Alert

rajivmucheli:
@auhlig 

Please review the Cinder Alert definition and help in configuring the cinder alerts to openstack-cinder channel to slack.

### DIFF
--- a/system/kube-monitoring/charts/prometheus-frontend/openstack-cinder.alerts
+++ b/system/kube-monitoring/charts/prometheus-frontend/openstack-cinder.alerts
@@ -1,7 +1,7 @@
 ### Cinder Alerts ###
 
 ALERT OpenstackCinderVolumeStuck
-  IF (sum(openstack_stuck_volumes_count_gauge) by (host,status) *60) > 0
+  IF (sum(openstack_stuck_volumes_count_gauge) by (host,status)) > 0
   FOR 5m
   LABELS {
     service = "cinder",
--- a/system/kube-monitoring/charts/prometheus-frontend/openstack-cinder.alerts
+++ b/system/kube-monitoring/charts/prometheus-frontend/openstack-cinder.alerts
@@ -1,0 +1,15 @@
+### Cinder Alerts ###
+
+ALERT OpenstackCinderVolumeStuck
+  IF (sum(openstack_stuck_volumes_count_gauge)) by (region,host) *60) > 0
+  FOR 5m
+  LABELS {
+    service = "cinder",
+    severity = "info",
+    tier = "openstack",
+    dashboard = "cinder"
+  }
+  ANNOTATIONS {
+    summary = "Cinder Volumes in stuck state",
+    description = "Sum of Openstack Cinder Volume is more than 1",
+  }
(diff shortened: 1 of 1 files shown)


system/kube-monitoring/charts/prometheus-frontend/openstack-cinder.alerts
```diff
@@ -0,0 +1,15 @@
+### Cinder Alerts ###
+
+ALERT OpenstackCinderVolumeStuck
+  IF (sum(openstack_stuck_volumes_count_gauge)) by (region,host) *60) > 0
```
auhlig: Hi @rajivmucheli, there are 2 brackets behind `..gauge))`. Remove one. Please check the expression in prometheus beforehand. Also`region` in the by clause is optional. 
I couldn't find the metric you're trying to alert on. Is it supposed to be there yet?

auhlig:
From syntax perspective this looks good now. But since the metric doesn't exist currently it's hard to say if the alert condition makes sense. Maybe @fwiesel can tell.

fwiesel:
The metric will only appear, when there is actually a stuck value.

auhlig:
Yes. I didn't get it. The `FOR 5m` would indicate that the condition needs to apply for exactly that time.

rajivmucheli:
@auhlig i presumed *60 had to be part of the syntax. I have removed *60 now, if the syntax looks good, can i or you merge this PR ? Could you plz add the changes the on how to redirect cinder alerts to openstack-cinder slack channel ?

jobrs:
@rajivmucheli can we have a project_id label on the metric? Also I wonder what the _gauge suffix is good for. I believe unlike "_counter" for counters it is not common to add a "_gauge" suffix. See https://prometheus.io/docs/practices/naming/

fwiesel:
Counters are monotonously increasing. I think, [Gauge](https://prometheus.io/docs/concepts/metric_types/#gauge) is correct. It will hopefully decrease again.

jobrs:
absolutely, but you do not have to add the "_gauge" suffix. That's somewhat default.

rajivmucheli:
@auhlig @jobrs good to Merge now ?

jobrs:
Ok, I see you guys are talking only about the alert. 

I also wanted to create some awareness for the metric naming and the `project_id` label for those metrics which can be connected to a project.

So probably my comments are more relevant for @fwiesel. Fabian, would you be able to provide a project_id label for this metric? In which repo can I find the corresponding exporter?

fwiesel:
The metric comes from here: https://github.com/sapcc/openstack-helm/blob/master/cinder/values.yaml#L63-L80

The `_gauge` suffix is also added by the statsd exporter.

jobrs:
Not necessarily: 

> If -statsd.add-suffix is set, the exporter appends the metric type (_gauge,
_counter, _timer) to the resulting metrics. This is enabled by default for
backward compatibility but discouraged to use. Instead, it is better to
explicitly define the full metric name in your mapping and run the exporter
with -statsd.add-suffix=false

But since the suffix is there now and it does not do any harm probably we should keep it.

The other thing is more important: Can we have a project_id label on this metric. I am not aware of the DB schema, that's why I am asking

fwiesel:
Sure, I've added it. Ideally, that metric is always empty and we hopefully do not have to expose it to our customers.

jobrs:
agreed

PR #163, opened 2017-08-22
Maia instrumentation

jobrs:
* update to maia image with scrape endpoint
* enable discovery of maia and prometheus components
* enable federation of maia_.* metrics to prometheus-frontend


auhlig:
@jobrs If you're creating alerts for maia: routing as well as a openstack-maia channel in slack is already in place.

jobrs:
@auhlig good to know, thanks! do you have a template for a routing alert I could reuse?

PR #166, opened 2017-08-23
Bump common PostgreSQL backup image to v0.5.12

nobse:


reimannf:
I think you miss the service for exposing the metrics

reimannf:
If you prefere, we can do it like in openstack-helm. Don't bump the version here, but in the region values?

ruvr:
(manually) merged with additional guard condition

PR #172, opened 2017-08-24
Move replication pod to backup namespace

nobse:


nobse:
New backup replication image added

PR #175, opened 2017-08-25
Bump Grafana PostgreSQL backup image to v0.5.12

nobse:


openstack/grafana/templates/postgres-service.yaml
```diff
@@ -13,3 +13,28 @@ spec:
...
+    prometheus.io/scrape: "true"
+    prometheus.io/port: "9188"
+spec:
+  clusterIP: None
+  ports:
+    - name: backup-metrics
+      port: 9188
+      protocol: TCP
+  selector:
+    app: grafana-postgres
```
reimannf: I think the label selector is wrong. https://github.com/sapcc/helm-charts/blob/eaf6c35a79d22f13b246358df958e9e5c06a44a7/openstack/grafana/templates/postgres-deployment.yaml#L18-L19

thgrs:
looks good to me as far as i can see (incl. falks comment) - i would suggest to test this on grafana-dev in staging first and if it works well there we can run through the regions. the risk for this is quite low for most grafanas in the regions, as their content can easily get refilled from the grafana-content repo - we would only loose session infos (users have to log in again) and starred dashboard info in the worst case.

the only grafana, where we really have non-reproducable content inside is grafana-author - so i would suggest ot do this one last after the procedure has been verified for all the other grafanas.

reimannf:
thanks @thgrs  for the feedback. I changed the selector. Can you help with deploying grafana-dev only? It is link to grafana and I have no glue how to deploy grafana-dev withput pipeline???

PR #176, opened 2017-08-25
import VCenter metrics again

jobrs:
... after applying a filter to exclude data for temporary VMs (from blackbox/regression-tests)


jobrs:
@BugRoger you don't want this commit?

BugRoger:
Obsoleted by #184

PR #178, opened 2017-08-29
Add checksum for backup-replication

Dexus:
Hope that is the correct way.

### DIFF
--- a/openstack/backup-replication/templates/deployment.yaml
+++ b/openstack/backup-replication/templates/deployment.yaml
@@ -13,6 +13,8 @@ spec:
     metadata:
       labels:
         name: backup-replication
+      annotations:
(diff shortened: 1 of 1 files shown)


openstack/backup-replication/templates/service.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: backup-replication
   namespace: backup
   annotations:
+    checksum: {{ include "backup-replication/templates/replica-configmap.yaml" $ | sha256sum }}
```
reimannf: We don't need it on the service. Only for the deployment to have the spec changed.

Dexus: done

PR #180, opened 2017-08-30
Adding OpenstackCinderVolumeDetach & OpenstackCinderVolumeAttach Metrics

rajivmucheli:
@auhlig @fwiesel 

Please validate the alert and approve accordingly.

Regards,
Rajiv


auhlig:
Won't work. The `by` clause works only in combination with an aggregation @rajivmucheli

fwiesel:
Ah, indeed. It should be `max(openstack_stuck_volumes_max_duration_gauge{status=<status>}) by (host)`

PR #184, opened 2017-08-31
Disentangle kubemon maia

jobrs:
* stop scraping Maia exporters from kube-monitoring prometheus-collector. No Maia metrics will be stored in kube-monitoring. Instead Grafana can collect data from both kube-monitoring and maia-prometheus if needed. New datasource has been added to Grafana content as well.
* minimize cross-federation from kube-monitoring prom-coll to maia in favor of parallel scraping of exporters
* facilitate the above by using discovery instead of static_configs
   - maia.io/scrape marks services to be scraped by Maia's Prometheus
   - maia.io/scrape_param_<xyz> marks service which need URL-parameter <xyz>

@BugRoger: Before you merge, maia needs to get access to the Kubernetes API like you did today for staging.

### DIFF
--- a/common/pgmetrics/templates/deployment.yaml
+++ b/common/pgmetrics/templates/deployment.yaml
@@ -29,6 +29,7 @@ spec:
       annotations:
         chart-version: {{.Chart.Version}}
         configmap-etc-hash: {{ include (print $.Template.BasePath "/configmap-etc.yaml") . | sha256sum }}
+        # Would be nice to have metrics in Maia as well: maia.io/scrape: "true"
         prometheus.io/scrape: "true"
         prometheus.io/port: {{ default 9187 .Values.port | quote }}
     spec:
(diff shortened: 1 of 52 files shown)


BugRoger:
lgtm

thgrs:
sounds like a good idea - fine with me

thgrs:
i would be interested to know how this is done: "Before you merge, maia needs to get access to the Kubernetes API like you did today for staging." - i looked into this with @auhlig a while ago, but it looked quite complicated with the current kubernetes version we are using.

auhlig:
I think @BugRoger gave the service-account in the maia ns sufficient permissions, right?

BugRoger:
I renamed the authorizations for `monasca` to `maia` in staging. The same needs to be done in regions who had Monasca deployed before. In new regions we need to add this line. https://github.wdf.sap.corp/monsoon/monsoon3/commit/36d39439bea37906ea398f94f15cc4a92becc6b1

PR #189, opened 2017-09-08
Prom collector alerting

auhlig:
Currently the `up`-metric from the collector are federated through the frontend to the global prometheus. The alerting rule is triggered by this metric.
If the frontend is down, but the collector is still up, this will cause an alert, because the federation chain is broken.
Fix: Scrape (only) the up metric from the collector itself and alert on this.
Makes sense @BugRoger?


BugRoger:
lgtm

PR #194, opened 2017-09-20
Add noop.ingress to enable vice president for swift

auhlig:
/cc @majewsky @reimannf


openstack/swift/templates/ingress.yaml
```diff
@@ -0,0 +1,31 @@
...
+  rules:
+    - host: swift-{{$cluster.name}}.{{.Values.global.region}}.{{.Values.global.domain}}
+      http:
+        paths:
+        - path: /
+          backend:
+            serviceName: swift-{{$cluster.name}}
+            servicePort: 9090
+
+{{- end }}
```
majewsky: Needs a YAML document separator, `---`, at the end because multiple YAMLs are generated by this file. (Compare other templates with an outer loop.)

reimannf:
This now supports also SANs
TODO: Send `SIGHUP` to nginx whenever a change is detected in mounted secret in `/tls-secret`

reimannf:
The implementation of the cert watching still feels like a `Frickellösung` to me. @majewsky Comments?

auhlig:
Can we try this in staging @reimannf? I would need to revoke the cert before.

reimannf:
@auhlig yes we can. I prepared a corresponding PR for the secrets repo. Nevertheless we should coordinate. Is the revocation and the vice-president enrolment process now be considered as down-time free change?

auhlig:
Cool. Do we want a ~15 min meeting for that beginning of next week?
Regarding the downtime: The VP shouldn't be the problem. But we should test how long it takes until the mount is updated and the nginx is reloaded and back with the new certs.

majewsky:
I'm going to merge this now because we should roll out the cert watch to all regions before switching to the VP, to ensure that the proxies smoothly switch over to the new certs.

PR #198, opened 2017-09-27
Changing severity status of ALERT OpenstackDatapathDown to info

rajivmucheli:
@Abhishekkumar17 @auhlig 

As requested by Markus, changing the severity status of ALERT OpenstackDatapathDown to info. Would there be any need of adding blackbox in (as it wasnt there in critical initially): 

https://github.com/sapcc/helm-charts/blob/master/global/prometheus-alertmanager/templates/config.yaml#L89


auhlig:
There's no blackbox channel, so no need to add it there. Blackbox alerts are routed by to the respective channel using their `tier` and `service` label.
Will check regarding the severity.

PR #201, opened 2017-10-04
Inauguration round 1

auhlig:
Prepare transition to the vice president by annotating the ingress and, if the secret contains only the tls key,cert, by making it optional. 
Adding the annotation and removing the cert/key from the values in cc/secrets will actually switch the service to the vice-president. I will take care if none of the reviewers objects to the below changes.

### DIFF
--- a/openstack/arc/templates/api-ingress.yaml
+++ b/openstack/arc/templates/api-ingress.yaml
@@ -1,4 +1,5 @@
 {{- if .Values.api.ingress.enabled }}
+{{- if .Values.api.ingress.tlsCertificate }}
 kind: Secret
 apiVersion: v1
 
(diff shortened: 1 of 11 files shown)


auhlig:
It's the kubernetes operator that talks to the Symantec VICE API and manages tls certificates for you. No more manual certificate enrollments or renewals: https://github.com/sapcc/kubernetes-operators/tree/master/vice-president

abattye:
The elektra change has a typo in the word 'annotation'

auhlig:
Thanks @abattye @edda

auhlig:
Thanks to all involved. I'll take care from here.

auhlig:
rebased

auhlig:
@edda `helm diff elektra openstack/elektra/ --values $secrets/eu-de-1/values/elektra.yaml` shows that the image would change in elektra:
```
-           image: hub.global.cloud.sap/monsoon/monsoon-dashboard:20171017080849
+           image: hub.global.cloud.sap/monsoon/monsoon-dashboard:latest
```
Did I do something wrong?

auhlig:
Same for hermes
```
-           image: hub.global.cloud.sap/monsoon/hermes:20171016150707
+           image: hub.global.cloud.sap/monsoon/hermes:201707141751
```

edda:
@auhlig I think that's fine. But maybe double check with @databus23

majewsky:
The diff for Elektra is not surprising. You're just missing the `--set image.tag=...` that the pipeline appends. Don't know about Hermes, but I'm going to guess that it's the same there.

auhlig:
Right. I'll check the pipelines for `--set image.tag=...`

PR #206, opened 2017-10-17
Inauguration round 2: barbican

auhlig:
let barbican take advantage of the vice president certificate operator. create secret only if cert,key are found in values. set annotation. Can we try this in staging @ruvr ?

Or did I confuse this with https://github.com/sapcc/openstack-helm/tree/master/barbican ?


auhlig:

PR #207, opened 2017-10-17
Inauguration round 3: glance

auhlig:
Glance transition to vice president certificate operator. create secret only if cert,key are found in values. set annotation. Should be okay. Can we try this in staging @bozinsky?


auhlig:

PR #209, opened 2017-10-17
Inauguration round 5: keystone

auhlig:
do not create secrets via helm if removed from values. set vice-presidential annotation. Do we want to try this in staging @ruvr?


ruvr:
Implemented.

PR #213, opened 2017-10-18
Adding openstack_compute_instances metric to prometheus-global

rajivmucheli:
Hi @auhlig 

Could you please validate and approve.


majewsky:
We need to revert this. You aggregated the metrics, but then pushed the unaggregated metrics. That's too much data for the global Prometheus.

PR #218, opened 2017-10-20
Adding openstack_compute_instances_total metric

rajivmucheli:
Hi @auhlig 

Hope this fits better ?


system/kube-monitoring/charts/prometheus-frontend/aggregation.rules
```diff
@@ -7,3 +7,4 @@ swift_cluster_storage_used_percent_average{} = avg(swift_cluster_prostorage_used
 limes_consolidated_cluster_capacity{} = max(label_join(limes_cluster_capacity, "full_resource", "/", "service", "resource")) by (full_resource,os_cluster)
 limes_consolidated_domain_quota{} = max(label_join(limes_domain_quota, "full_resource", "/", "service", "resource")) by (full_resource,os_cluster,domain)
 limes_consolidated_domain_usage{} = sum(max(label_join(limes_project_usage, "full_resource", "/", "service", "resource")) by (full_resource,os_cluster,domain,project)) by (full_resource,os_cluster,domain)
+openstack_compute_instances_total{} = sum(openstack_compute_instances_gauge) by (flavor_name,vm_state,region,host)
```
auhlig: You don't need to do the `region`. It's added automatically. Besides LGTM

majewsky:
According to the ticket, you need a dashboard with:

* number of instances per region total (all states without the state deleted)
* number of instances per region per flavor (all states without deleted)

If that's all you need, why do you need to send the different VM states (and the counts from different hosts) as separate timeseries? Given these requirements, I'd do something like this:

```prometheus
openstack_compute_instances_total{} =
  sum(openstack_compute_instances_gauge{vm_state!="deleted"}) by (flavor_name)
```

The point is to keep the number of timeseries as small as possible because prometheus-global has a really long retention time, and any unnecessary extra resolution costs storage and memory and CPU.

PR #219, opened 2017-10-24
add new domain "bs" to Limes/Swift seeds

majewsky:
@ruvr Two questions:

1. I copy-pasted the parts relating to the "bn" domain and did `s/bn/bs/`. A second pair of eyes is much appreciated.
2. The `BS_DOMAIN_RESOURCE_ADMINS` part probably also needs to be set up in CAM. Can you take care, please?


ruvr:
LGTM. CAM preparation work has been done already.

PR #224, opened 2017-11-03
Repo sync - Switch from Jobs to deployments

reimannf:
As it turns out, that simulating CronJobs via Jobs (regulary deleting and
recreating the Job) is error-prone
Switching to deployments which run a script forever and sleep after execution until midnight will achieve the same behaviour.


openstack/content-repo/bin/sleep-until-midnight
```diff
@@ -0,0 +1,28 @@
...
+
+  echo "Sleeping for $sleep_seconds seconds..."
+  sleep $sleep_seconds
+}
+
+# Run forever
+while :; do
+  # Run the command immediately
+  # Run until completed succesful
+  until "$@"; do :; done
```
majewsky: I'd like to have some backoff here, just in case that the command fails fast (e.g. with a config file error).

reimannf:
Good point: Adding Exponential Backoff and Max Attempts of 10. Gives us roughly 30 mins before the pod restarts. Misconfigurations should than also be reported via alert for too many pod restarts.

openstack/content-repo/bin/sleep-until-midnight
```diff
@@ -9,19 +9,33 @@ function sleepuntil ()
 
   sleep_seconds=$(( $target_epoch - $current_epoch ))
   #If time is already reached, wait for the next day
-  if [ ${sleep_seconds} -lt 0 ]; then
-    sleep_seconds=$((86400 + sleep_seconds))
+  if [ $sleep_seconds -lt 0 ]; then
+    sleep_seconds=$(( 86400 + $sleep_seconds ))
```
majewsky: No `$` needed here before `sleep_seconds`. I think it will work, too, but it's not recommended. (Same applies the other `$((...))` instances below.)

majewsky:
No idea why Github thinks that the code I'm commenting on is outdated.

PR #225, opened 2017-11-03
Repo alert trigger based on timestamps

reimannf:


reimannf:
AFAIK a non existing metric does not trigger an alert. But will not merge before https://github.com/sapcc/helm-charts/pull/224 anyhow.

PR #226, opened 2017-11-03
Kube17 ds tolerations

SchwarzM:
daemonsets need tolerations in kube 1.7 else they will not deploy on tainted nodes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .idea
 *.tgz
+*.swp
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 .idea
 *.tgz
+*.swp
```
majewsky: Stuff like this can be put into `~/.gitignore` so that you don't need to copy it into every single repo.

PR #229, opened 2017-11-06
remove Swift dependency from Elektra seed

majewsky:
As far as I know, this was only used by the legacy resource management implementation and to create Swift accounts. Since Limes now handles both, I don't think anything else in Elektra needs to be a swiftreseller anymore.

@edda Are the above assumptions correct?

@abattye FYI (since your inquiry triggered the line of thought leading to this PR)

### DIFF
--- a/openstack/elektra/templates/seed.yaml
+++ b/openstack/elektra/templates/seed.yaml
@@ -9,7 +9,6 @@ metadata:
     heritage: "{{ .Release.Service }}"
 spec:
   requires:
-  - swift/swift-seed
   - monsoon3/domain-default-seed
   - monsoon3/domain-bn-seed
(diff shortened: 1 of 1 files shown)


edda:
@hgw77 Do you know whether swiftreseller is needed for anything else? I can't think of anything but I haven't had much contact with the Swift stuff myself...

hgw77:
From my side I go with stefans description that the swift reseller is no more used. At the moment there is a create account function in the object_storage plugin but no more used since limes is doing that for us. In the upcoming object_storage plugin based on misty that functionality is completely removed.

majewsky:
Let's do it then. I will remove the role assignments from all regions except eu-de-1, and we'll see if we notice any breakage.

PR #231, opened 2017-11-09
Prometheus 2

auhlig:
everything converted, tested and running. Switch will be manual.


majewsky:
As far as I can see, this is only the frontend and collector Promethei. We should check if anyone is impacted by losing the one week worth of data in the frontend instances (I'm not). Do you plan to do a data-retaining migration ([like in this guide](https://www.robustperception.io/accessing-data-from-prometheus-1-x-in-prometheus-2-0/)) for the global instance?

system/kube-monitoring/charts/prometheus-frontend/templates/deployment.yaml
```diff
@@ -19,6 +19,9 @@ spec:
         type: frontend
     spec:
       terminationGracePeriodSeconds: 600
+      securityContext:
+        fsGroup: 0
+        runAsUser: 0
```
auhlig: fyi: this works in k8s v1.7 @SchwarzM

auhlig:
We're going to have a hard cut. I'll send a mail when it's time @majewsky

PR #234, opened 2017-11-20
statsd-exporter v0.5

reimannf:
statsd-exporter v0.5 removed the automatic suffixing of metrics based on
its kind: `_timer`, `_gauge`, `_counter`

Unfortunatelly that setting was switched on per defualz in 0.3 and all our
framework around that metrics are built on top: alerts and dashboards.

v0.5 Removes completly the support for that, so before updating the
exporter, we need to clean up. That should allow a better transition as we
cannot assume that we can update all regions at once.


system/kube-monitoring/charts/prometheus-collector/aggregation.rules
```diff
@@ -1,11 +1,11 @@
 aggregated:container_cpu_system_seconds_total{} = sum(container_cpu_system_seconds_total) by (id) keep_common
 aggregated:container_cpu_usage_seconds_total{} = sum(container_cpu_usage_seconds_total) by (id) keep_common
 aggregated:container_cpu_user_seconds_total{} = sum(container_cpu_user_seconds_total) by (id) keep_common
-aggregated:swift_cluster_storage_used_percent_gauge_average{} = avg(swift_cluster_storage_used_percent_gauge) by (kubernetes_namespace, system, component)
+aggregated:swift_cluster_storage_used_percent_gauge_average{} = avg(swift_cluster_storage_used_percent_gauge) by (kubernetes_namespace, system, component) OR avg(swift_cluster_storage_used_percent) by (kubernetes_namespace, system, component)
```
reimannf: @auhlig: what do you think about that `OR` in the aggr rule?

auhlig: Sorry for missing that PR @reimannf. Looks good.

majewsky:
Wouldn't it be less confusing to just add the suffixes to the metric names in the 0.5 config? That way, we don't have to fiddle around with existing Grafana dashboards etc. all over the place.

reimannf:
well. everything is already prepared (incl. dashboards) and I would like to get rid of the strange suffixing

PR #245, opened 2017-12-06
Swift pike

shparker1977:
nodeSelector is now a variable
default variable value "swift-storage" is provided in values.yaml


openstack/swift/templates/rsyncd-daemonset.yaml
```diff
@@ -35,7 +34,7 @@ spec:
             path: /srv/node
         - name: swift-drive-state
           hostPath:
-            path: /run/swift-storage/state
```
reimannf: We don't want to replace that

PR #249, opened 2017-12-19
swift: split off swift-utils chart to resolve deployment ordering issues

majewsky:
When the "swift" chart is installed, the openstack-operator picks up the seeds in there and tries to create the Swift accounts defined in the seeds, but this fails because Swift is not available yet. After Swift has started up, one would need to restart the openstack-operator, so that it picks up the seeds again and creates the Swift accounts.

Therefore, everything that needs a Swift account is now in the "swift-utils" chart, which is only installed when Swift has become operational.


openstack/swift-utils/templates/account-caretaker-collect-daemonset.yaml
```diff
@@ -31,7 +35,7 @@ spec:
             name: swift-account-caretaker
         - name: swift-bin
           configMap:
-            name: swift-bin
+            name: swift-bin # NOTE: This comes from the "swift" chart.
```
reimannf: That makes me nervous. Is there any particular reason we just refer to the swift config map and don't get the needed parts in an own cm?

reimannf:
I tried to resolve the issue with swift-bin. The initial approach with just referencing the config-map from the swift release, which might be dangerous. The second approach to copy over the logic has high chances to diverge in the future.
The crucial parts like the start-up logic and the unmount helper should be maintained in one central place. So I created the `swift-common` chart, which can be required by `swift` and `swift-utils`.
I transformed `swift-utils` already but want to discuss this with @majewsky before going further.
There are some implications with that:

Pro:
* bin scripts are central
* common templates from `_utils.tpl` can go to `swift-common` and don't need to be copied over
* common values can go to `swift-common`

Con:
* refactoring of the swift chart needs to be done
* pipeline needs adoption (recognize `requirements.yaml` and do a `helm dep up` before the upgade
* No chance to render a checksum for the bin configmap in the parent charts, so I choosed a version in the value file https://github.com/sapcc/helm-charts/pull/249/files#diff-93b51de20254796861b0369919b5abd4R8 https://github.com/sapcc/helm-charts/pull/249/files#diff-039f84322060721d2bf867302da1341eR20 to have a minimum control that a deployment needs an update.

I think it is still the way to go. Before doing more refactorings, lets discuss

majewsky:
Approved (although I cannot give the green checkmark because I originally opened the PR).

PR #252, opened 2018-01-04
fix rabbit storage leak+dataloss

jobrs:
RabbitMQ stores its data files in a directory named like the hostname.

Since in Kubernetes the hostname is set to the pod-name by default, every re-deployment will make RabbitMQ loose data and leave behind an orphaned data directory on disk.

Setting the hostname to a fixed value solves the problem. Rolling-Updates are of course not possible then.

Fix is already implemented for `rabbitmq_notifications` chart. We should apply it to `openstack-helm/rabbitmq` as well and possibly merge it with `rabbitmq_notifications` on that occasion.

### DIFF
--- a/common/rabbitmq/templates/bin/_rabbitmq-start.tpl
+++ b/common/rabbitmq/templates/bin/_rabbitmq-start.tpl
@@ -30,7 +30,11 @@ function bootstrap {
 
 
 function start_application {
-   exec rabbitmq-server
+  echo "Starting RabbitMQ with lock /var/lib/rabbitmq/rabbitmq-server.lock"
+  LOCKFILE=/var/lib/rabbitmq/rabbitmq-server.lock
+  exec 9>${LOCKFILE}
+  /usr/bin/flock -n 9
+  exec gosu rabbitmq rabbitmq-server
 }
 
 bootstrap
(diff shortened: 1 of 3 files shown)


fwiesel:
I think, we might have a problem with that.
There is a chance, that a second pod comes up, while the first one is running,
and there is nothing in the [startup scripts](https://github.com/sapcc/helm-charts/blob/master/common/rabbitmq/templates/bin/_rabbitmq-start.tpl) to stop the second process to write on the same files. The Recreate deployment should do in such a way, but it does not ensure it.

I would suggest to add some locking like in our [postgres startup](https://github.wdf.sap.corp/monsoon-docker/monsoon-postgres/blob/aadd95745aa8384f30d596f37f8358bfbf6154ee/env.sh#L12-L14).

thgrs:
@jobrs please have a look at my old scripts, they contained locking as well: https://github.com/sapcc/helm-charts/blob/56eeb30bcaea18820ad398e93550d6d1912e7f33/openstack/ceilometer/templates/bin/_rabbitmq-start.tpl

jobrs:
thanks @fwiesel for the hint on actual `Recreate` behaviour
@thgrs I took over your code unchanged, thanks

jobrs:
obsolete

PR #256, opened 2018-01-20
ingress controller 0.10.2

auhlig:
upgrade the ingress ctrl next week. will test in staging beforehand.


auhlig:
Done in all regions.

PR #259, opened 2018-01-31
ingress probing

auhlig:


system/kube-monitoring/charts/prometheus-collector/templates/_prometheus.yaml.tpl
```diff
@@ -256,6 +256,7 @@ scrape_configs:
 {{ range $region := .Values.global.regions }}
 - job_name: 'blackbox-ingress-{{ $region }}'
   metrics_path: /probe
+  region_probed_from: {{ $region }}
```
auhlig: won't work. unlike `metrics_path` this is not a keyword given by prometheus. we want to add it in the relabel section

auhlig:
Tested in staging, admin cluster. Sqash'n'Merge is planned for tomorrow.

PR #260, opened 2018-02-01
WIP - do not merge - Audit designate

notque:
WIP for audit-designate mapping, and rabbitmq-notifications change.


openstack/designate/requirements.yaml
```diff
@@ -2,9 +2,14 @@ dependencies:
   - name: mariadb 
     repository: file://../../common/mariadb
     version: 0.1.1
-  - name: memcached
-    repository: file://../../common/memcached
-    version: 0.0.1
   - name: rabbitmq
+    alias: rabbitmq_notifications
```
jobrs: this renames the existing (RPC-)rabbitmq. not good. We need an additional rabbitmq for notifications.

openstack/designate/templates/api-deployment.yaml
```diff
@@ -42,7 +42,7 @@ spec:
             - name: DEPENDENCY_JOBS
               value: "designate-migration"
             - name: DEPENDENCY_SERVICE
-              value: "designate-mariadb,{{ .Release.Name }}-rabbitmq"
+              value: "designate-mariadb,rabbitmq"
```
jobrs: why this change? I did not introduce that dependency in Nova, Neutron, Cinder because the API should not be blocked by audit-specific components. If the notifications-rabbit is not up in time, the middleware will buffer.

openstack/designate/templates/central-deployment.yaml
```diff
@@ -44,7 +44,7 @@ spec:
             - name: DEPENDENCY_JOBS
               value: "designate-migration"
             - name: DEPENDENCY_SERVICE
-              value: "designate-mariadb,{{ .Release.Name }}-rabbitmq"
```
jobrs: same as above - plus that central does not have the middleware

openstack/designate/templates/etc/_designate_audit_map.yaml
```diff
@@ -0,0 +1,23 @@
...
+    singleton: true
+  # Default mappings
+  tlds:
+  tsigkeys:
+  blacklists:
+  quotas:
+    singleton: true
+
+  service_statuses:
+    service_name: 'name'
```
jobrs: why service_name? the service_name is always the same. Only the resource typeURI varies.

jobrs: do we have a resource "service_statuses" with multiple instances? the typeURI of the elements would be 'service_statuse' unless you override it.

openstack/designate/values.yaml
```diff
@@ -92,17 +92,6 @@ memcached:
   metrics:
     enabled: true
 
-rabbitmq:
```
jobrs: the rabbitmq is not ours. Please do not touch

openstack/designate/values.yaml
```diff
@@ -187,3 +176,22 @@ logging:
...
+  enabled: true
+  # do by default not attach the request payloads of create/update calls to the event
+  record_payloads: false
+  metrics_enabled: true
+  # how many messages to buffer before dumping to log (when rabbit is down or too slow)
+  mem_queue_size: 1000
+
+rabbitmq_notifications:
+  name: designate
+  # overwrite auth in region secrets
```
jobrs: the default user-names are defined by the rabbitmq chart's value file already. Let's keep them unchanged to avoid confusion and remove this section here.

openstack/designate/templates/etc-configmap.yaml
```diff
@@ -25,3 +25,8 @@ data:
 {{ include (print .Template.BasePath "/etc/_designate-policy.json.tpl") . | indent 4 }}
   logging.conf: |
 {{ include "loggerIni" .Values.logging | indent 4 }}
+
+{{- if .Values.audit.enabled }}
+    designate_audit_map.yaml: |
+{{ include (print .Template.BasePath "/etc/_designate_audit_map.yaml") . | indent 4 }}
```
jobrs: how is this file copied to /etc/designate?

openstack/designate/templates/etc/_api-paste.ini.tpl
```diff
@@ -5,6 +5,10 @@ use = egg:Paste#urlmap
 /v2: osapi_dns_v2
 /admin: osapi_dns_admin
 
+{{- define "audit_pipe" -}}
+{{- if .Values.audit.enabled }} audit{{- end -}}
```
jobrs: where is this added to the pipeline?

jobrs:
also missing: changes to the designate config. Just updated the README to be more specific: https://github.com/sapcc/openstack-audit-middleware/blob/master/README.md#oslo-messaging

openstack/designate/requirements.yaml
notque: fixed requirements with your help.

openstack/designate/templates/api-deployment.yaml
notque: It already had the requirement, I renamed it when I was renaming the rabbit. Removed the change.

openstack/designate/templates/central-deployment.yaml
notque: removed the change.

openstack/designate/templates/etc/_designate_audit_map.yaml
notque: /v2/service_status is a singleton that returns the health state of services. perhaps it shouldn't be included?

openstack/designate/values.yaml
notque: reverting

openstack/designate/values.yaml
notque: done

notque:
designate.conf is setup, uses util include, and util added to requirements. we're fine there.

openstack/designate/templates/etc-configmap.yaml
notque: added copy now, without the template audit-enabled call, as the file is not a template right now.

notque:
Just an update, it's working in staging, and I've sent the branch/pr to David to get a review. Waiting until that's sorted.

PR #262, opened 2018-02-07
Enables SSL Client Certificate Authentication for Grafana

BugRoger:
This PR enables authentication to Grafana via SSL client certificates. It subs out
the basic auth with proxy auth. Grafana now expects the user to be
passed in via the `X-Remote-User` header. This can be toggled with the
`tls_client_auth.enabled` flag in the Helm values. It defaults to off
and keeps things as they are.

The ingress is configured to verify SSL certificates using the
`nginx.ingress.kubernetes.io/auth-tls-*` annotations. This can be turned
on/off per ingress spec.

Additionally, this PR injects a bit of logic into the ingress config, to
copy the relevant part (the user-id) into the `X-Remote-User` header.

### DIFF
--- a/openstack/grafana/templates/_grafana.ini.tpl
+++ b/openstack/grafana/templates/_grafana.ini.tpl
@@ -150,7 +150,7 @@ verify_email_enabled = false
 # Background text for the user field on the login page
 login_hint = UserID[@domain]
 
-{{- if .Values.grafana.tls_client_auth.enabled}}
+{{- if .Values.grafana.auth.tls_client_auth.enabled}}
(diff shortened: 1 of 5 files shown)


BugRoger:
@thgrs Please review. I didn't really test this on live yet. It shouldn't break things but who knows... 
@auhlig Needs your eyes too.

auhlig:
Cool. I'll try in staging

auhlig:
Tested successfully with grafana-dev in staging. All disabled by default. The PR in cc/secrets will enable it. Are you okay with this @thgrs?

PR #263, opened 2018-02-08
[ingress] Enable "local" external traffic policy

fwiesel:
This should give use the real external ips


auhlig:
What about https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/configmap.md#use-proxy-protocol ?

fwiesel:
Apparently it doesn't work

PR #265, opened 2018-02-09
Editing Metrics to support Pike release

rajivmucheli:
Hi @Carthaca 

Please validate the changes:

share_type_id column is removed from shares table
prefix share_snapshots to size to overcome error in openstack_manila_snapshot metric

### DIFF
--- a/openstack/manila/values.yaml
+++ b/openstack/manila/values.yaml
@@ -79,8 +79,11 @@ pgmetrics:
             usage: "GAUGE"
             description: "Number of share networks"
     openstack_manila_shares:
-      query: "SELECT shares.project_id, shares.id, share_instances.status, COUNT(*) AS count_gauge, SUM(size) size_gauge FROM shares INNER JOIN share_instances ON shares.id=share_instances.share_id GROUP BY shares.id, shares.project_id, share_instances.status"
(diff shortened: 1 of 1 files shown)


openstack/manila/values.yaml
```diff
@@ -79,14 +79,11 @@ pgmetrics:
             usage: "GAUGE"
             description: "Number of share networks"
     openstack_manila_shares:
-      query: "SELECT coalesce(shares.share_type_id, 'N/A') AS share_type_id, shares.project_id, shares.id, share_instances.status, COUNT(*) AS count_gauge, SUM(size) size_gauge FROM shares INNER JOIN share_instances ON shares.id=share_instances.share_id GROUP BY shares.id, shares.project_id, shares.share_type_id, share_instances.status"
+      query: "SELECT shares.project_id, shares.id, share_instances.status, COUNT(*) AS count_gauge, SUM(size) size_gauge FROM shares INNER JOIN share_instances ON shares.id=share_instances.share_id GROUP BY shares.id, shares.project_id, share_instances.status"
```
Carthaca: `share_type_id` was not removed completely, but it was moved to the share_instances table.

Can you adjust to that please, so that we can keep the `share_type_id` ?

Carthaca:
thx, will be merged once we are on pike everywhere

PR #266, opened 2018-02-12
fix sso with new ing ctrl

auhlig:
obviously we have to turn on `ingress.kubernetes.io/auth-tls-pass-certificate-to-upstream` using the new ingress controller to enable sso again. off by default. enable only via regional values. i'll take care.
objections @edda?
/cc @ruvr


edda:
wouldn't it make more sense to enable by default and only disable if necessary for some reason? I would imagine that we normally enable it everywhere....

auhlig:
ah yes. i checked that this annotation wasn't used before. enabled by default now

PR #267, opened 2018-02-21
[swift] changed image_version handling

reimannf:
The image_version value is now be expected following that format:
`<openstack_release_name>-<image_version>`
Based on the OpenStack release name the intended pipeline and needed
filters are defined in `templates/etc/_proxy-server.conf.tpl`

### DIFF
--- a/openstack/swift/templates/_utils.tpl
+++ b/openstack/swift/templates/_utils.tpl
@@ -121,9 +121,9 @@ When passed via `helm upgrade --set`, the image_version is misinterpreted as a f
   {{- end -}}
 {{- end }}
 
-{{- define "swift_release_first_char" -}}
-{{- if ne .Values.image_version "DEFINED_BY_PIPELINE" -}}
-    {{ $v := .image_version | split "-"}}{{ $v._0 | trunc 1 | upper }}
+{{- define "swift_release" -}}
+{{- if ne .image_version "DEFINED_BY_PIPELINE" -}}
+    {{ $v := .image_version | split "-"}}{{ $v._0 | lower }}
   {{- end -}}
 {{- end }}
 
--- a/openstack/swift/templates/_utils.tpl
+++ b/openstack/swift/templates/_utils.tpl
@@ -114,7 +114,7 @@ checksum/object.ring: {{ include "swift/templates/object-ring.yaml" . | sha256su
 {{- /**********************************************************************************/ -}}
 When passed via `helm upgrade --set`, the image_version is misinterpreted as a float64. So special care is needed to render it correctly.
 {{- define "swift_image" -}}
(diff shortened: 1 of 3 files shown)


openstack/swift/templates/_utils.tpl
```diff
@@ -114,14 +114,16 @@ checksum/object.ring: {{ include "swift/templates/object-ring.yaml" . | sha256su
 {{- /**********************************************************************************/ -}}
 When passed via `helm upgrade --set`, the image_version is misinterpreted as a float64. So special care is needed to render it correctly.
 {{- define "swift_image" -}}
-  {{- if typeIs "string" .Values.image_version -}}
-    {{ required "This release should be installed by the deployment pipeline!" "" }}
+  {{- if and (typeIs "string" .Values.image_version) (ne .Values.image_version "DEFINED_BY_PIPELINE") -}}
```
majewsky: We don't need the typeIs anymore in this case.

reimannf: Done

openstack/swift/templates/etc/_proxy-server.conf.tpl
```diff
@@ -23,10 +24,15 @@ log_level = INFO
...
+{{- if le $swift_release "M" }}
+# Mitaka pipeline
 pipeline = catch_errors gatekeeper healthcheck proxy-logging cache cname_lookup domain_remap bulk tempurl ratelimit authtoken keystone sysmeta-domain-override staticweb container-quotas account-quotas slo dlo versioned_writes proxy-logging proxy-server
-{{- else }}
+{{- else if eq $swift_release "P" }}
+# Pike pipeline
 pipeline = catch_errors gatekeeper healthcheck proxy-logging cache cname_lookup domain_remap bulk tempurl ratelimit authtoken keystone sysmeta-domain-override staticweb copy container-quotas account-quotas slo dlo versioned_writes proxy-logging proxy-server
+{{- else if ge $swift_release "Q" }}
+# Queens or > pipeline
+pipeline = catch_errors gatekeeper healthcheck proxy-logging cache cname_lookup domain_remap bulk tempurl ratelimit authtoken keystone sysmeta-domain-override staticweb copy container-quotas account-quotas slo dlo versioned_writes symlink proxy-logging proxy-server
```
majewsky: I double-checked the specification for text/template (which says that it uses the comparison rules defined by Go), then the Go language spec, and it turns out you can just compare the release names directly and it will do the right thing. (String comparison is defined to be lexicographic.) See https://play.golang.org/p/lX3F2h3Mo0r

reimannf: Done

PR #269, opened 2018-03-07
Added tcp probes with prometheus blackbox-exporter

vonneudeck:
Unfortunately there is a bit SAP sauce.
Hopefully it won’t live long though.

### DIFF
--- a/system/kube-monitoring/charts/prometheus-collector/templates/_prometheus.yaml.tpl
+++ b/system/kube-monitoring/charts/prometheus-collector/templates/_prometheus.yaml.tpl
@@ -331,7 +331,7 @@ scrape_configs:
     target_label: instance
   - target_label: __address__
     replacement: prober.{{ $region }}.cloud.sap
-  - target_label: probed_from
+  - target_label: region_probed_from
     replacement: {{ $region }}
 {{ end }}
 
--- a/system/kube-monitoring/charts/prometheus-collector/templates/_prometheus.yaml.tpl
+++ b/system/kube-monitoring/charts/prometheus-collector/templates/_prometheus.yaml.tpl
@@ -312,6 +312,29 @@ scrape_configs:
     replacement: {{ $region }}
 {{ end }}
 
+{{ range $region := .Values.global.regions }}
+- job_name: 'blackbox-tcp-{{ $region }}'
+  metrics_path: /probe
+  scrape_interval: 15s
+  params:
+    module: [tcp_connect]
+  static_configs:
+    - targets:
+      {{- range $.Values.blackbox_exporter.tcp_probe_targets }}
+      - {{ . }}
+      {{- end }}
+  scheme: https
+  relabel_configs:
+  - source_labels: [__address__]
+    target_label: __param_target
+  - source_labels: [__param_target]
+    target_label: instance
+  - target_label: __address__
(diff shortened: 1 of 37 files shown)


auhlig:
I'm not too happy if we add more SAP sauce, as it makes it harder to transition to stable/prometheus chart. Ok to get this in for now, but we need to clean up this up afterwards.
Please test in staging. Also the alert is missing.

vonneudeck:
I would like to squash merge this now. Because I can’t meaningfully test the alert queries before rolling this out. The alerts will be done in a separate branch and PR afterwards.

system/kube-monitoring/charts/prometheus-frontend/README.md
```diff
@@ -33,20 +33,26 @@ Region is an implicit label that is automatically set by Prometheus. Do not over
    * `openstack`
    * `datapath`
    * `kubernikus`
+   * `e2e`
```
auhlig: Why are you introducing the `e2e` tier? Currently we're using `tier=datapath, context=availability` for these tests.

system/kube-monitoring/charts/prometheus-collector/templates/_prometheus.yaml.tpl
```diff
@@ -312,6 +312,29 @@ scrape_configs:
...
+  - source_labels: [__address__]
+    target_label: __param_target
+  - source_labels: [__param_target]
+    target_label: instance
+  - target_label: __address__
+    replacement: prober.{{ $region }}.cloud.sap
+  - target_label: region_probed_from
+    replacement: {{ $region }}
+{{ end }}
+
```
auhlig: ok for now, but for the next iteration: move to separate `extra_scrape_config`, just as we do in [kubernikus](https://github.com/sapcc/kubernikus/blob/master/charts/kubernikus-system/charts/prometheus/templates/config.yaml#L21-L23)

vonneudeck: Of course. I do not wish for this to stay there.

system/kube-monitoring/charts/prometheus-frontend/README.md
auhlig: Can't follow based on your explanation. The tier label doesn't change the what/why, that's up to the respective test to do that. But okay if this was already decided.

auhlig:
Please squash before merging

system/kube-monitoring/charts/prometheus-collector/templates/_prometheus.yaml.tpl
```diff
@@ -312,6 +312,29 @@ scrape_configs:
...
 
+{{ range $region := .Values.global.regions }}
+- job_name: 'blackbox-tcp-{{ $region }}'
+  metrics_path: /probe
+  scrape_interval: 15s
+  params:
+    module: [tcp_connect]
+  static_configs:
+    - targets:
+      {{- range $.Values.blackbox_exporter.tcp_probe_targets }}
```
auhlig: this assumes there's a `blackbox_exporter.tcp_probe_targets` value. surround the whole job with with`{{ if blackbox_exporter.tcp_probe_targets }} ... {{ end }}`,

PR #270, opened 2018-03-22
Skip tcp probe if targets are not set

vonneudeck:
The two phased check is necessary because the evaluation fails if `blackbox_exporter` does not exist.
Solves this https://github.com/sapcc/helm-charts/pull/269#discussion_r176355664


vonneudeck:
I could not find any. I already asked Michael and he also didn’t know a better one.

PR #275, opened 2018-04-02
Use as helm repository...

auhlig:
... so we don't need to copy the charts around and maintain them in different places, like we do [here](https://github.com/sapcc/kubernikus/tree/master/charts/kubernikus-system/vendor/vice-president). Tried this on the vice-president. 
Requires to create a archive file of the chart before it can be used (`helm package .`)
What do you think @BugRoger?


BugRoger:
That's great. Where does the tgz and `index.yaml` come from?

auhlig:
The tgz archive is created using `helm package <chart>` and the `index.yaml` was created manually - within a chart folder this should work with `helm repo index .`

PR #276, opened 2018-04-03
[Barbican] Moving to helm-charts

rajivmucheli:
Hi @Carthaca,

Please validate my commit for movement of Barbican from openstack-helm to helm-charts in staging region!


openstack/barbican/values.yaml
```diff
@@ -1,94 +1,77 @@
...
+    loggers:
+        root:
+            handlers: stdout, sentry
+            level: WARNING
+        barbican:
+            handlers: stdout, sentry
+            level: INFO
+        eventlet.wsgi.server:
+            handlers: stdout, sentry
+            level: INFO
```
Carthaca: please add rabbitmq key here with `persistence:
    enabled: false`

openstack/barbican/values.yaml
```diff
@@ -50,32 +50,6 @@ postgresql:
     imageTag: v0.5.12
     os_password: DEFINED-IN-REGION-CHART
 
-logging:
```
Carthaca: please keep those values, they are filled into logging.ini in etc-configmap

PR #280, opened 2018-04-10
**WIP**Use stable/nginx-ingress chart

auhlig:
**DO NOT MERGE**


system/kube-system/values.yaml
```diff
@@ -1,2 +1,120 @@
...
+      # set in regional values
+      externalIPs:
+        - DEFINED-AS-VALUE
+
+    defaultBackendService: "kube-system/ingress-default-backend"
+
+    podLabels:
+      app: nginx-ingress-controller
+
+    affinity:
```
auhlig: Hey @SchwarzM,
I'd like to replace the `hostPort` hack we're currently using to get 1 ingress-controller per master node with affinities. I remember you tried that already elsewhere. Can you take a look at this `affinity`? Thanks a lot in advance.

SchwarzM: lgtm

system/kube-system/templates/ingress-cacrt-secret.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.ingress.tls_client_auth.enabled }}
+apiVersion: v1
+kind: Secret
+type: Opaque
+
+metadata:
+  name: ingress-cacrt
+
+data:
+  ca.crt: {{ default "" .Values.ingress.tls_client_auth.cacrt | b64enc | quote }}
```
SchwarzM: nginx-ingress here as well ?

auhlig: Right. Thanks for taking a look

auhlig:
This is running in staging
TODO: commit enabling jaeger support after test

PR #282, opened 2018-04-13
http for repo.region.tld

reimannf:
That PR enables the option to have `http://repo.region.tld:8080/`

It only accepts non ssl traffic for configured hosts. swift api is not exposed via http in general.


openstack/swift/templates/etc/_nginx.conf.tpl
```diff
@@ -59,38 +59,28 @@ http {
...
-        }
+        {{ tuple $cluster $context | include "swift_nginx_ratelimit" | indent 8 }}
+        {{ include "swift_nginx_location" . | indent 8}}
     }
+
+    # Only allow non ssl for allowed sans, otherwise redirect
+    server {
+        listen 80 default_server;
+        server_name {{tuple $cluster $context | include "swift_endpoint_host"}};
+        return 301 https://$server_name$request_uri; # Redirect to https
```
majewsky: This should probably use `$host` instead of `$server_name` unless you have a good reason: https://serverfault.com/a/706439

PR #283, opened 2018-04-16
[swift, elk, kube-mon.] use native k8s rolling update for daemonsets

majewsky:
Now possible since we are on k8s 1.7 in all regions.

To roll this out:
1. Merge this PR.
2. Deploy all regions once with the old Concourse task (the one that does rolling upgrade manually).
3. When deployed, adjust the Concourse task to not do its own rolling upgrade, but use `k rollout status` instead. This needs coordination with @auhlig who is also using the Concourse task with daemonsets in kube-monitoring.


auhlig:
kube-monitoring, kube-system daemonsets done

auhlig:
@majewsky There's no `.spec.updateStrategy.rollingUpdate.minReadySeconds`

PR #288, opened 2018-04-19
[Glance] moving to helm-charts

rajivmucheli:
Hi @Carthaca 

Time to move glance to helm-charts now.

I am not sure if https://github.com/rajivmucheli/helm-charts/blob/fa00ceb3cd794db2e49d8683009ae4927228d4e6/openstack/glance/templates/seed.yaml#L29 is correct? since this is not reachable.

Please validate & comment.

Regards,
Rajiv

### DIFF
--- a/openstack/glance/templates/seed.yaml
+++ b/openstack/glance/templates/seed.yaml
@@ -23,13 +23,13 @@ spec:
     endpoints:
     - interface: admin
       region: {{ .Values.global.region | quote }}
-      url: http://{{include "glance_api_endpoint_host_admin" .}}:{{.Values.global.glance_api_port_internal}}
+      url: http://{{include "glance_api_endpoint_host_admin" .}}:{{.Values.glance_api_port_internal}}
     - interface: public
       region: {{ .Values.global.region | quote }}
-      url: https://{{include "glance_api_endpoint_host_public" .}}:{{.Values.global.glance_api_port_internal}}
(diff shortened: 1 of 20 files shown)


openstack/glance/templates/seed.yaml
```diff
@@ -23,13 +23,13 @@ spec:
     endpoints:
     - interface: admin
       region: {{ .Values.global.region | quote }}
-      url: '{{ .Values.services.api.scheme }}://{{ .Values.services.api.host }}:{{ .Values.services.api.port }}'
+      url: http://{{include "glance_api_endpoint_host_admin" .}}:{{.Values.api_port_internal}}
     - interface: public
       region: {{ .Values.global.region | quote }}
-      url: '{{ .Values.services.api.scheme }}://{{ .Values.services.api.host }}:{{ .Values.services.api.port }}'
+      url: https://{{include "glance_api_endpoint_host_public" .}}:{{.Values.api_port_internal}}
```
Carthaca: no need for the port, we take the default 443

rajivmucheli: would i need to remove {{.Values.api_port_internal}}? it be like:

url: https://{{include "glance_api_endpoint_host_public" .}}

or 

change api_port_internal value to 443 in values.yaml ?

Carthaca: removing the api_port_internal here, like done for barbican

PR #290, opened 2018-04-20
Add interconnect alerts

vonneudeck:
The idea is to have critical alerts when a lot of regions can’t connect
and a warning if one region can’t connect so that we have less alert fatigue.
The critical alert is not yet set critical because I don’t want to let
it loose over the weekend.

### DIFF
--- a/global/prometheus/interconnect.alerts
+++ b/global/prometheus/interconnect.alerts
@@ -1,12 +1,29 @@
-  ALERT InterconnectDown
+  ALERT RegionInterconnectDown
     IF count((count(probe_success != 1) by (region,region_probed_from) unless (count(probe_success == 1) by (region,region_probed_from) >= 1))) by (region)
     FOR 5m
     LABELS {
+      meta = "{{ $value }} region(s) can’t reach the region",
+      context = "interconnect",
       service = "interconnect",
-      severity = "notyetset",
+      severity = "warning",
       tier = "openstack"
     }
     ANNOTATIONS {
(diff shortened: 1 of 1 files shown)


vonneudeck:
This needs to be merged to check if it does what we want. As the critical alert is set to notyetcritical, we can see if it behaves as expected, add a dashboard and if everything goes as planned we can arm the critical alert on Monday.

global/prometheus/interconnect.alerts
```diff
@@ -1,12 +1,29 @@
-  ALERT InterconnectDown
+  ALERT RegionInterconnectDown
     IF count((count(probe_success != 1) by (region,region_probed_from) unless (count(probe_success == 1) by (region,region_probed_from) >= 1))) by (region)
     FOR 5m
     LABELS {
+      meta = "{{ $value }} region(s) can’t reach the region",
+      context = "interconnect",
       service = "interconnect",
-      severity = "notyetset",
+      severity = "warning",
```
auhlig: Is `warning` intended? Current interconnect alerts are `critical`. 
Also we would get alerts based on probers and blackbox tests if we merge this. Please turn of the blackbox tests to avoid duplicate alerts.

vonneudeck: This warning is intended.
The old blackbox test alerts will be turned off on Monday when we saw that these here worked and the dashboard is ready.

auhlig: Ok. Hopefully there will not be too much spam in the meantime.

PR #293, opened 2018-04-30
deploy secgroup-entanglement-exporter

majewsky:
This adds the [security group entanglement exporter](https://github.com/sapcc/secgroup-entanglement-exporter) to the Neutron chart. The exporter generates a metric called "security group entanglement" that can be used to alert customers with security group setups that strain the DVS agent.

- The image for the exporter is produced by the openstack-neutron pipeline which I extended for this purpose.
- I added `.Values.secgroup_entanglement_exporter.image_tag` to all values files for all regions in cc/secrets, so no further action is required; this can be merged and deployed directly if you like.

cc @fwiesel to make him aware of the existence of the exporter


notandy:
👍 for inventing the entanglement unit

PR #294, opened 2018-05-02
relocate the cc3test domain and support CAM provided test users

ruvr:
@artherd42  @majewsky maybe you can add your changes (if there are any) so we can do the flip in a consistent manner on staging ASAP?


majewsky:
Swift is done with #292.

PR #306, opened 2018-05-25
Neutron agent daemonset

notandy:
added neutron agent daemonset as region configurable option (should not break anything without region-specific additions)

### DIFF
--- a/openstack/neutron/templates/daemonset-network_agents.yaml
+++ b/openstack/neutron/templates/daemonset-network_agents.yaml
@@ -1,0 +1,175 @@
+{{ if $.Values.agent.enable_daemonset }}
+---
+kind: DaemonSet
+apiVersion: extensions/v1beta1
+metadata:
+  name: neutron-agents
+  labels:
+    system: openstack
+    type: agent
+    component: neutron
+spec:
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 1
+  template:
+    metadata:
+      labels:
+        name: neutron-agents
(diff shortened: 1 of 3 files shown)


openstack/neutron/templates/daemonset-network_agents.yaml
```diff
@@ -0,0 +1,194 @@
...
+          value: "network"
+          effect: "NoSchedule"
+      containers:
+        - name: neutron-dhcp-agent
+          image: {{.Values.global.imageRegistry}}/{{.Values.agent.image_name}}:{{.Values.agent.image_tag}}
+          imagePullPolicy: IfNotPresent
+          securityContext:
+            privileged: true
+          command:
+            - bash
```
fwiesel: Personally, I am more a fan of executing the script directly,
and setting the permissions on the container to executable.
But that is just a matter of taste.

openstack/neutron/templates/daemonset-network_agents.yaml
```diff
@@ -0,0 +1,194 @@
...
+      maxUnavailable: 1
+  template:
+    metadata:
+      labels:
+        name: neutron-agents
+        system: openstack
+        type: backend
+        component: neutron
+    spec:
+      hostNetwork: true
```
fwiesel: Could you also add more metadata on the pod? Like this here (https://github.com/sapcc/helm-charts/blob/master/openstack/neutron/templates/deployment-server.yaml#L27)

fwiesel: Possibly replacing "api" with "agent"

PR #318, opened 2018-06-15
Limes support team seed

reimannf:
- Adding resource_admin to the supprt team projects
- Inherit resource_viewer for all domains


reimannf:
Not needed so far

PR #319, opened 2018-06-18
[manila] added support team role assignments

Carthaca:
inspired by https://github.com/sapcc/helm-charts/commit/613ea6684d94e3e5b716d5aafc1313528efeeea8

@ruvr okay?


ruvr:
Looks OK. Potentially the *_API_SUPPORT should have cloud_sharedfilesystem_admin on the domains. But I'm not sure how manilla implements the sharedfilesystem_admin permission checks .

Carthaca:
`cloud_sharedfilesystem_admin` allows everything for all projects, so I would like to keep it limited as it is now

PR #321, opened 2018-06-20
[openstack][arc] bump backup image version

reimannf:
backup-tools:v0.5.14 has the following improvements for the backup-restore utility:
* backup-restore does not connect itself to the to be restored database anymore, which
  caused the `DROP DATABASE IF EXISTS` statement to fail
* backup-restore prints out the executed statements from the dump file,
  allowing to track down possible issues
* backup-restore keeps the restored dumpfile in `/tmp/newbackup<random>` for possible
  manual reexecution

Please merge and deploy on your own. Consider the postgres pod restart caused by that.


databus23:
Thanks!

PR #330, opened 2018-06-22
pgbouncer: enables optional prometheus scrape annotations

stefanhipfel:


common/postgresql/values.yaml
```diff
@@ -22,6 +23,7 @@ global:
       listen_port: 6543
       listen_backlog: 256
       auth_type: "md5"
+      server_tls_sslmode: "prefer",
```
fwiesel: There is a stray ','
Not sure, if it will cause problems

PR #331, opened 2018-06-25
Hermes support roles

notque:
Added information for hermes support roles. Once merged I will deploy to staging, and look at the seeder pod logs to see if there are any issues.

@ruvr Would you mind looking this over? Thanks.


ruvr:
Please check the results & deploy to the regions

PR #332, opened 2018-06-27
add fsn domain to all other relevant seeds

majewsky:


thgrs:
maia part looks good - thx for spotting the missing dependencies as well

PR #334, opened 2018-06-28
pgbouncer: uses opt var to check if prometheus is enabled

stefanhipfel:


stefanhipfel:
To enable this: https://github.wdf.sap.corp/cc/secrets/pull/230/files

PR #338, opened 2018-07-03
Update kubernetes-kubelet.alerts

B1Sandmann:
Added links to more playbooks.


auhlig:
You might also want to add it [below this line](https://github.com/B1Sandmann/helm-charts/blob/481942d937b3d5a38076383e8afa1551e1b5546d/system/kube-monitoring/charts/prometheus-frontend/kubernetes-kubelet.alerts#L12)

B1Sandmann:
Like this? I left out the "Many" Alerts because I suspected you'll get the ones for the single nodes as well but you are right more is better.

auhlig:
If multiple kubelets are down the `..ManyDown..` alert will be send - instead of an alert per kubelet to avoid spam

PR #339, opened 2018-07-04
Update prometheus.alerts

B1Sandmann:
Tried to make it more clear that a down of the Federation could be because of a partial down of the region. If someone has a better wording go ahead.


auhlig:
Thanks!

PR #349, opened 2018-07-19
[keystone] Switch to common cache template

fwiesel:
I've copied the section to the "ini_sections.cache" template (deriving the config_prefix from the chart-name) in order to avoid duplication.


ruvr:
This is part of the configuration already and I rather have not cross chart dependencies that are maintained beyond my control and could have a major impact on keystone availability.

PR #350, opened 2018-07-19
[kube-monitoring] adds ipmi service discovery helm charts

stefanhipfel:


stefanhipfel:
staging values:
https://github.wdf.sap.corp/cc/secrets/pull/245

stefanhipfel:
Added extra configmap for the ipmi-sd, which will be mounted to prom as volume when ipmi is enabled!

auhlig:
:+1:

PR #378, opened 2018-08-03
[kube-monitoring] adds pg_cluster prom alerts

stefanhipfel:


stefanhipfel:
@auhlig do you think that is ok? this way there would be an extra alert for each pg cluster..or better just one alert if any of those pg cluster backups fail? thx

auhlig:
You can cover it with one generic alert. 
Maybe you could also introduce a `service` label (for the openstack service) to the `pg_cluster_wale_backup_success` metric? Then we could route the alert to the respective `openstack-<service>` channel in slack.
The alert expression could look like `count (pg_cluster_wale_backup_success == 0) by (service)`

stefanhipfel:
we can use team label, which should have the service name!
count (pg_cluster_wale_backup_success == 0) by (team)

auhlig:
I'd rather stick to the `service` label as defined [in our alerting guidlines](https://github.com/sapcc/helm-charts/tree/master/system/kube-monitoring/charts/prometheus-frontend#labels) than introduce a new one with the same purpose. Makes sense?

stefanhipfel:
sure, I just looked at the backup service to get an idea. But are we talking about the alert label or the metrics label?
Can't we just do that?
...
labels:
      service: {{$labels.team}}

auhlig:
Ah. I assumed it's the pg exporter and could be configured there. Then `service: {{$labels.team}}` is fine

PR #379, opened 2018-08-03
add openstack-watcher-middleware to ironic. disabled by default

auhlig:
**NOTE:** requires https://github.com/sapcc/ironic/pull/3 before the watcher can be enabled. /cc @fwiesel


auhlig:
rebased

PR #380, opened 2018-08-07
[kube-monitoring] adds hw ipmi sensor alerts

stefanhipfel:
Not sure if those alerts belong here. Need feedback. Thanks!


vonneudeck:
You are right.
As these alerts are not related to ironic at all and it is pretty imaginable that we expand ipmi and its alerts to hardware we use without ironic I would say it is better to make a new file.
How about `baremetal-defects.alerts` or `hardware-defects.alerts`?
I also think we might need a new tier for that as it is neither `openstack` nor `kubernetes`. How about `hardware`? But we might put `openstack` in it for now and wait until the first alerts happen and what needs to be done than to decide these as well, if you don’t wish to create a new tier for this right now.

stefanhipfel:
I already had a baremetal alerts file in the beginning...seems to be the best to do right now

PR #381, opened 2018-08-07
Introducing custom resource classes for flavors

BerndKue:
https://specs.openstack.org/openstack/nova-specs/specs/pike/implemented/custom-resource-classes-in-flavors.html


ruvr:
So far this does not seem to be supported by the api (and hence the seeder): 
https://developer.openstack.org/api-ref/compute/#create-flavor

PR #383, opened 2018-08-14
[kube-monitoring] adds alert for pgbouncers

stefanhipfel:


stefanhipfel:
Not sure. I dont want to create too many alerts. The experience only comes from downtimes or big changes in qa, not real world issues.

We can also make two alerts, one with 30s and info/warning and 5 min => critical?

fwiesel:
Let's start with that, and then we will see.

PR #403, opened 2018-08-22
Reduce routing complexity

vonneudeck:
We avoid the double filter, have much less complexity and we avoid accidental loss if we introduce new regions or clusters.
Please double check if you also think it will work as intended, I am not completely sure if I overlooked something.

### DIFF
--- a/global/prometheus-alertmanager/templates/config.yaml
+++ b/global/prometheus-alertmanager/templates/config.yaml
@@ -61,6 +61,7 @@ data:
           service: {{ $service }}
           tier: openstack
       {{- end }}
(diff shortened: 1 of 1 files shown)


auhlig:
NVM. Was already removed. Good to merge

vonneudeck:
That is a very good idea.
I have something like that in mind but wanted to first start by detangling the existing routing a little bit.

If I understood you correctly, you want to directly route each alert by its name in the router config to clarify things.
I like that there is no dependency between alert and routing file but the alert name.
I am all in for less magic and more clarity.

PR #409, opened 2018-08-29
[kube-monitoring][limes] FailedCapacityScrapes

reimannf:
Alert for FailedCapacityScrapes

### DIFF
--- a/system/kube-monitoring/charts/prometheus-frontend/openstack-limes.alerts
+++ b/system/kube-monitoring/charts/prometheus-frontend/openstack-limes.alerts
@@ -29,6 +29,21 @@ groups:
         the ccloud cluster, so limes-collect-ccloud is probably dead.
       summary: Limes is not scraping
 
+  - alert: OpenstackLimesFailedCapacityScrapes
+    expr: sum(increase(limes_failed_capacity_scrapes[5m])) BY (capacitor) > 0
+    for: 1h
+    labels:
+      context: failedcapacityscrapes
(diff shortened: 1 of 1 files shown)


system/kube-monitoring/charts/prometheus-frontend/openstack-limes.alerts
```diff
@@ -29,6 +29,21 @@ groups:
...
+    labels:
+      context: failedcapacityscrapes
+      dashboard: limes-overview
+      service: limes
+      severity: warning
+      tier: openstack
+    annotations:
+      description: Limes cannot scrape capapcity from {{ title $labels.capacitor }}
+        for more than an hour.
+        The log for limes-collect-{{ $labels.os_cluster }} contains additional info.
```
majewsky: To use the `os_cluster` label in the description, it needs to be included in the `by` clause of `sum()`.

reimannf: Are you sure. I thought it is injected by default but cannot remember the where it was. Otherwise that would also not work: https://github.com/sapcc/helm-charts/blob/b9c22fad6b5e3ab603c40cd3884bdd88dbe8a594/system/kube-monitoring/charts/prometheus-frontend/openstack-limes.alerts#L48

reimannf: done

PR #410, opened 2018-08-30
generic name 'baremetal/ironic' for ipmi job

vonneudeck:
This PR was mostly made to ensure that there are no typos.

The beforehand used name was anyway overwritten by the value 'baremetal/ironic' provided by the ipmi_sd service discovery.
Because we want to drop the label there, we have to change it here so that nothing changes for the metric’s consumers.
This change is necessary to unblock this PR:
https://github.com/sapcc/ipmi_sd/pull/7


auhlig:
@stefanhipfel @BerndKue are more qualified to review this one

vonneudeck:
Oh we already talked about this. This is purely about the typos and also I was not sure if the pipeline is still broken.

auhlig:
Not sure what you mean. The kube-monitoring pipeline looks quite green to me.
I don't have anything to add to this

PR #411, opened 2018-08-31
[ironic] cloud_baremetal_admin for CCADMIN_COMPUTE_SUPPORT

reimannf:


ruvr:
lgtm. I would even add it as a domain role assignment with inheritance on, so it works anywhere. But that is just another nice-to-have flavor..

PR #414, opened 2018-09-03
[backup-replication] use swift-http-import for backup replication

talal:


openstack/backup-replication/templates/replication-configmap.yaml
```diff
@@ -0,0 +1,43 @@
...
+  {{ $source_region }}.yaml: |
+    swift:
+      auth_url: {{ $.Values.target.auth_url }}
+      user_name: {{ $.Values.target.user_name }}
+      user_domain_name: {{ $.Values.target.user_domain_name }}
+      project_name: {{ $.Values.target.project_name }}
+      project_domain_name: {{ $.Values.target.project_domain_name }}
+      password: {{ $.Values.target.password }}
+    statsd:
+      hostname: {{ $.Values.statsd.hostname }}
```
reimannf: I would suggest to hardcode to hostname to `statsd` completly and remove it from the value file. Also the service name is derived from `$.Values.statsd.hostname`

openstack/backup-replication/values.yaml
```diff
@@ -1,39 +1,21 @@
...
-    os_auth_version: DEFINED-IN-REGION-SECRETS
-    os_username: DEFINED-IN-REGION-SECRETS
-    os_user_domain: DEFINED-IN-REGION-SECRETS
-    os_project_name: DEFINED-IN-REGION-SECRETS
-    os_project_domain: DEFINED-IN-REGION-SECRETS
-    os_region_name: DEFINED-IN-REGION-SECRETS
-    os_password: DEFINED-IN-REGION-SECRETS
+statsd:
+  exporter_image_version: 'v0.5.0'
+  hostname: DEFINED_IN_VALUES_FILE
```
reimannf: just use `statsd` in templates and remove it here

PR #418, opened 2018-09-06
rabbitmq clustering with k8s auto discovery

notandy:
rabbitmq now uses statefulset and k8s auto discovery to create a
federation if scaled above one replica
also, introduces global rabbitmq that replaces the openstack-helm charts
one

### DIFF
--- a/common/rabbitmq_cluster/.helmignore
+++ b/common/rabbitmq_cluster/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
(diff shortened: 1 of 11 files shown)


notandy:
Merging this PR will need redeployment of all services using rabbitmq since it will chang the existing rabbitmq to statefulset. The global rabbitmq is also included, but this is safe since it need to be run by a specific pipeline job to replace the current openstack-helm one.

notandy:
this also fixes #252 since it and #257 since statefulsets are never wind-up in parallel

ruvr:
Generally a good idea. I would like to test it before we merge, to check which/what of our charts would need to be updated (and how)

PR #426, opened 2018-09-11
prepare glance for sapcc/openstack-rate-limit-middleware

auhlig:
Middleware is disabled by default.
I'll add some comment to the rate limit file to clarify how this would work. See the comments [here](https://github.com/sapcc/helm-charts/pull/426/commits/f33ae9023b9ff57e0cda7fb924f00240f93a97e8#diff-f737aed0012142fe4ea8a15737d031f7)
Values there are based on staging, qa-de-1 and **not** intended for production.

Which actions should be rate limited. I currently have:
- download image data
- create image
- update image
- list images
- show image

Would be great to get some feedback @fwiesel. Thanks in advance 💚


openstack/glance/templates/etc/_ratelimit.yaml.tpl
```diff
@@ -0,0 +1,28 @@
+rates:
+  # global rate limits counted across all projects
+  global:
```
auhlig: `global` rate limits: all requests that pass this middleware. independent of scope (initator_project_id). meant to protect glance from receiving more traffic than it can handle

fwiesel:
Looks fine to me. Though I have no idea, what values would be correct for production.

Difficulty is a bit the image download. If we publish a new image, for each building-block, the image needs to be downloaded at the same time. And that is a bit outside of the control of the user.

So, in worst case, we have (No. of Building-blocks * No of new images) parallel downloads from a single user/project.

Is probably completely hypothetical, but it might be worth testing, how nova reacts to the rate-limiting of the image download

rajivmucheli:
contd in https://github.com/sapcc/helm-charts/pull/2474

PR #429, opened 2018-09-12
[kube-montioring] fine tunes pgbouncer alerts

stefanhipfel:


fwiesel:
I would suggest to drop the criticality to warning, until we have established that it is indeed critical.
Remember, "critical" means, someone gets called out the bed.

PR #432, opened 2018-09-12
prepare placement api

notandy:


notandy:
Works on staging :) should be safe to merge since it needs to be enabled in the secrets

PR #450, opened 2018-10-09
Change liveness of pg_bouncer

SchwarzM:
PG_Bouncer currently leaks threads and while the root cause is not found / fixed we need 
to prevent it from locking up nodes.


SchwarzM:
not needed anymore leak in pgbouncer has been fixed

PR #454, opened 2018-10-18
Rework alertmanager routing

auhlig:
**Do not merge**

Changes pagerduty alerts from `severity: critical` to a set of alertnames. Also allows regexes. 
Example: Only alert when many things are down
```
alerts:
  - .*ManyDown
```

As discussed - WDYT @BugRoger?

TODO before merging:
- [x] complete list of alert
- [x] adjust secrets/values

### DIFF
--- a/global/prometheus-alertmanager/templates/config.yaml
+++ b/global/prometheus-alertmanager/templates/config.yaml
@@ -93,7 +93,6 @@ data:
         - channel: "{{.CommonLabels.tier}}-{{.CommonLabels.severity}}"
           api_url: {{ default "MISSING" .Values.slack.webhook_url | quote }}
           username: "Kubernetes Control Plane"
-          channel: "#dev-null"
           title: {{"'{{template \"slack.sapcc.title\" . }}'"}}
           title_link: {{"'{{template \"slack.sapcc.titlelink\" . }}'"}}
           text: {{"'{{template \"slack.sapcc.text\" . }}'"}}
@@ -106,7 +105,6 @@ data:
         - channel: "{{.CommonLabels.tier}}-{{.CommonLabels.service}}"
           api_url: {{ default "MISSING" .Values.slack.webhook_url | quote }}
           username: "Kubernetes Control Plane"
-          channel: "#dev-null"
           title: {{"'{{template \"slack.sapcc.title\" . }}'"}}
           title_link: {{"'{{template \"slack.sapcc.titlelink\" . }}'"}}
           text: {{"'{{template \"slack.sapcc.text\" . }}'"}}
--- a/global/prometheus-alertmanager/templates/config.yaml
+++ b/global/prometheus-alertmanager/templates/config.yaml
@@ -33,60 +33,45 @@ data:
       receiver: dev-null
 
       routes:
-
-      # staging, lab alerts > /dev/null
       - receiver: dev-null
         continue: false
         match_re:
-<<<<<<< HEAD
-          region: staging|k-staging|s-staging|lab|i-eu-de-2
-
-      # routing per tier to <tier>-{info, warning, critical}
-=======
           region: .*staging|lab
 
-      # routing per tier to <tier>-{info, warning, critical} channel
->>>>>>> page based on alert name
-      {{- range $tier := .Values.tiers }}
-      - receiver: {{ $tier }}-info
-        continue: true
+      - receiver: slack_by_tier_and_severity
         match_re:
-          tier: {{ $tier }}
-        routes:
-        - receiver: '{{ $tier }}-warning'
-          continue: true
-          match:
-            severity: warning
-        - receiver: '{{ $tier }}-critical'
-          continue: true
-          match:
-            severity: critical
-      {{- end }}
+          tier: openstack|kubernetes|kubernikus|baremetal
+          severity: info|warning|critical
 
-      # routing per service to openstack-<service> channel
-      {{- range $service := .Values.openstack_services }}
-      - receiver: openstack-{{ $service }}
-        continue: true
+      - receiver: slack_by_tier_and_service
         match_re:
-          service: {{ $service }}
           tier: openstack
-      {{- end }}
+          severity: info|warning|critical
+          service: arc|backup|barbican|cinder|designate|elektra|elk|ironic|keystone|limes|lyra|maia|manila|neutron|nova|swift
 
-      # staging, qa-de-1, lab, scaleout, internet cluster alerts > /dev/null
-      - receiver: dev-null
-        continue: false
+      - receiver: pagerduty_baremetal
         match_re:
-          region: .*staging|.*qa-de-1|lab|s-.*|i-.*
+          severity: critical
+          region: ap-ae-1|ap-au-1|ap-cn-1|ap-jp-1|ap-jp-2|ap-sa-1|eu-de-1|eu-de-2|eu-nl-1|eu-ru-1|la-br-1|na-ca-1|na-us-1|na-us-3
+          tier: baremetal
 
-      # routing to pagerduty channels
-      {{- range $pd := .Values.pagerduty }}
-      - receiver: pagerduty-{{ $pd.name }}
-        continue: false
-        {{- if $pd.alerts }}
+      - receiver: pagerduty_hypervisor
         match_re:
-          alertname: {{ join "|" $pd.alerts -}}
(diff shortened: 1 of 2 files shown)


BugRoger:
Good start!

In my opinion the current routing configuration is too convoluted. I struggle to understand what's going on. And I did write parts of it myself... :rofl:

It is mainly because of the config generation with Helm templates. It seems weird that we basically have a very similar syntax to the alertmanager YAML in our Helm values. It smells like we could just put it directly into the config. If it weren't for the multiplication in the template. 

Let's take a step back. Imagine we throw it all away. 

Wouldn't this be equivalent to the current config?

```
  - receiver: slack_by_tier_and_severity
    match_re:
      tier: openstack|kubernetes|kubernikus|baremetal
      region: [^lab|staging]

  - receiver: slack_by_tier_and_service
    match_re:
      tier: openstack
      service: arc|elektra|designate|cinder
      region: [^lab|staging]

  - receiver: pagerduty
    match_re:
      severity: critical
      tier: openstack|kubernetes|kubernikus|baremetal
      region: [^lab|staging]

  - receiver: pagerduty
    match_re:
      severity: warning
      region: [^lab|staging]
      alertname: KubernetesBond.*|KubernetesNodeDown


receivers:
  - name: slack_by_tier_and_severity
    slack_configs:
      - channel: "{{CommonLabels.tier}}-{{CommonLabels.severity}}"
        ...

  - name: slack_by_tier_and_service
    slack_configs:
      - channel: "{{CommonLabels.tier}}-{{CommonLabels.service}}"
        ...

  - name: pagerduty
    pagerduty_configs:
      - service_key: "{{CommonLabels.service}"
        ...
```

This would allow us to not multiple the slack receivers. There is no Helm values at all. You can just look at this config and see what's happening. And that directly in Github. No surprise Helm/Value evaluation. To top it all of it's completely flexible.

(I don't know what the `service_key` in the `pagerduty` receiver is. But you get the idea...)

BugRoger:
> Would you suggest to have things like tiers, services, alertnames, etc. in the values or hardcoded in the alertmanager config?

Hardcoded. Literally like above.

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -36,58 +36,42 @@ data:
...
-      - receiver: dev-null
-        continue: false
+          severity: info|warning|critical
+          service: arc|backup|barbican|cinder|designate|elektra|elk|ironic|keystone|limes|lyra|maia|manila|neutron|nova|swift
+
+      - receiver: pagerduty_baremetal
         match_re:
-          region: qa-de-1|k-qa-de-1|s-qa-de-1|lab
+          severity: critical
+          region: [^.*qa-de-1]
```
BugRoger: I'm not 100% positive this works. In case you copied that from my example.

BugRoger: I'm not 100% positive this works. In case you copied that from my example.

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -33,54 +33,45 @@ data:
...
+          service: vcenter
+
+      - receiver: pagerduty
         match_re:
-          alertname: {{ join "|" $pd.alerts -}}
-        {{- end }}
-      {{- end }}
+          severity: critical
+          region: [^.*qa-de-1]
+          tier: openstack|kubernetes|kubernikus
```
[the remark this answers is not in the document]
BugRoger:
```suggestion
          tier: openstack|kubernetes
```

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -36,58 +36,42 @@ data:
...
+      - receiver: pagerduty
+        match_re:
           severity: critical
-          {{- if $pd.match }}
-          {{- range $key, $value := $pd.match }}
-          {{ $key }}: {{ $value }}
-          {{- end }}
-          {{- end }}
-      {{- end }}
+          region: [^.*qa-de-1]
```
BugRoger:
```suggestion
          region: [^.*qa-de-1|lab|.*staging|s-.*|k-.*]
```

BugRoger: Maybe it's better to not negate it? To make it very clear what is being forwarded:

```
eu-de-1|eu-de-2|eu-nl-1|...
```

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -33,54 +33,45 @@ data:
...
-      {{- end }}
+          severity: critical
+          region: [^.*qa-de-1]
+          tier: openstack|kubernetes|kubernikus
+
+      - receiver: pagerduty
+        match_re:
+          severity: warning
+          region: [^.*qa-de-1]
+          alertname: KubernetesBond.*
```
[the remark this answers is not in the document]
BugRoger: 👍

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -36,58 +36,42 @@ data:
...
-        match:
-          tier: kubernikus
-      - receiver: dev-null
-        continue: false
-        match:
-          service: interconnect
+      - receiver: pagerduty_hypervisor
+        match_re:
+          severity: critical
+          region: [^.*qa-de-1]
```
BugRoger:
```suggestion
          region: [^.*qa-de-1]
```

global/prometheus-alertmanager/templates/config.yaml
BugRoger:
```suggestion
          region: ap-ae-1|ap-au-1|ap-cn-1|ap-jp-1|ap-jp-2|ap-sa-1|eu-de-1|eu-de-2|eu-nl-1|eu-ru-1|la-br-1|na-ca-1|na-us-1|na-us-3
```

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -36,58 +36,42 @@ data:
...
+        match_re:
           severity: critical
-          {{- if $pd.match }}
-          {{- range $key, $value := $pd.match }}
-          {{ $key }}: {{ $value }}
-          {{- end }}
-          {{- end }}
-      {{- end }}
+          region: [^.*qa-de-1]
+          tier: openstack|kubernetes|kubernikus
```
BugRoger:
```suggestion
          tier: openstack|kubernetes
```

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -36,58 +36,42 @@ data:
...
+
+      - receiver: pagerduty
+        match_re:
+          severity: warning
+          region: [^.*qa-de-1]
+          alertname: KubernetesBond.*
```
BugRoger:
```suggestion
          alertname: KubernetesNodeBond.*|KubernetesNodeNotReady
```

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -36,58 +36,42 @@ data:
...
-          {{- end }}
-          {{- end }}
-      {{- end }}
+          region: [^.*qa-de-1]
+          tier: openstack|kubernetes|kubernikus
+
+      - receiver: pagerduty
+        match_re:
+          severity: warning
+          region: [^.*qa-de-1]
```
BugRoger:
```suggestion
          region: ap-ae-1|ap-au-1|ap-cn-1|ap-jp-1|ap-jp-2|ap-sa-1|eu-de-1|eu-de-2|eu-nl-1|eu-ru-1|la-br-1|na-ca-1|na-us-1|na-us-3

```

global/prometheus-alertmanager/templates/config.yaml
auhlig: How about paging on `Many` (like in `KubernetesNodeManyNotReady`; alert is already critical). If just one node e.g. a minion is NotReady we might be able to wait until the next morning.
We could also introduce a different alert for a master being NotReady as that might be more critical. No?

auhlig: Ah. Ignore my comment^. `KubernetesNodeNotReady` is a `warning` alert.

auhlig:
I'll try this right after lunch. If this does't work we'll have a nice quite afternoon alerting wise ;p

PR #455, opened 2018-10-19
Load netapp metrics to Prometheus frontend

chuan137:


auhlig:
Let's see how much this stresses prometheus in staging, qa

PR #469, opened 2018-11-19
Add inital alerts für concourse

databus23:
Hi @auhlig please advise if this makes sense or can be improved.


databus23:
Merging after discussion with @auhlig

PR #470, opened 2018-11-19
prepare the move to prometheusv2 - start with the old values and adju…

thgrs:
…sting in a later commit

### DIFF
--- a/openstack/maia/templates/prometheus-deployment.yaml
+++ b/openstack/maia/templates/prometheus-deployment.yaml
@@ -43,24 +43,35 @@ spec:
               mountPath: /etc/prometheus
 {{- if .Values.prometheusv2.enabled }}
         - name: prometheusv2
-          image: "{{ .Values.prometheusv2.image }}"
+          image: "{{ .Values.prometheusv2.image.repository }}:{{ .Values.prometheusv2.image.tag }}"
           args:
-            - -config.file=/etc/prometheus/prometheusv2.yaml
-            - -storage.local.path=/prometheus/v2
-            - -storage.local.target-heap-size={{ .Values.prometheusv2.target_heap_size }}
-            - -storage.local.retention={{ .Values.prometheusv2.retention }}
-            - -web.console.libraries=/etc/prometheus/console_libraries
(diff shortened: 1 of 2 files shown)


auhlig:
starts a 2nd prometheus in version v2.5.0 with configmap reload sidecar. needs to write to a separate folder than prom v1.x as datamodel changed. 
not sure 100GiB is sufficient for both prometheis tsdb's

openstack/maia/templates/prometheus-deployment.yaml
```diff
@@ -41,6 +41,38 @@ spec:
               mountPath: /prometheus
             - name: config
               mountPath: /etc/prometheus
+{{- if .Values.prometheusv2.enabled }}
+        - name: prometheusv2
+          image: "{{ .Values.prometheusv2.image.repository }}:{{ .Values.prometheusv2.image.tag }}"
+          args:
+            - --config.file=/etc/prometheus/prometheusv2.yaml
+            - --storage.tsdb.path=/prometheusv2
```
thgrs: @auhlig i would suggest to use /prometheus/v2 here - this way we do not conflict with the v1 files, which we will delete later

openstack/maia/templates/prometheus-deployment.yaml
```diff
@@ -41,6 +41,38 @@ spec:
               mountPath: /prometheus
             - name: config
               mountPath: /etc/prometheus
+{{- if .Values.prometheusv2.enabled }}
+        - name: prometheusv2
+          image: "{{ .Values.prometheusv2.image.repository }}:{{ .Values.prometheusv2.image.tag }}"
+          args:
+            - --config.file=/etc/prometheus/prometheusv2.yaml
```
thgrs: @auhlig i think we would have to provide this files then as well - would it have the same content as the v1 one?

thgrs: discussed with arno: same config file, so we will use prometheus.yaml

openstack/maia/templates/prometheus-deployment.yaml
thgrs: v2 subdir is ok

PR #471, opened 2018-11-19
[nova][neutron] Add vcenter-operator templates for respective services

fwiesel:
I would like to move the neutron-dvs-agent and the nova-compute-agent to each their own deployment triggered out of the respective service charts.

Case in point is, that we deploy now the services in separate charts, leading to the issue that a deployment of either nova, or neutron might cause changes to the configuration of either the nova-compute agent or the neutron dvs-agent respectively, which might require a restart of either agent.

With separate template, this can simply be done by a checksum of the template:
(E.g. https://github.com/sapcc/helm-charts/compare/vct_nova_neutron?expand=1#diff-34b226228729bdfa2b5c1985a95265dbR37)

### DIFF
--- a/openstack/neutron/templates/vct-dvs-agent-configmap.yaml
+++ b/openstack/neutron/templates/vct-dvs-agent-configmap.yaml
@@ -1,8 +1,8 @@
-{{- if .Capabilities.APIVersions.Has "vcenter-operator.stable.sap.cc/v1" }}
+{{- if and (.Capabilities.APIVersions.Has "vcenter-operator.stable.sap.cc/v1") (contains ",dvs" .Values.ml2_mechanismdrivers )}}
 apiVersion: vcenter-operator.stable.sap.cc/v1
 kind: VCenterTemplate
 metadata:
-  name: 'vcenter-cluster-neutron-ml2-vmware-configmap'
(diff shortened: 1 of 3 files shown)


notandy:
I think this is ok, since it came really hard to maintain the neutron image inside the nova pod, this way we don't need to share neutron/nova image version between two separate pipelines. Also keeping in mind that nsx-t is changing the game again.

abattye:
Maybe I misunderstand, but personally I think this introduces an anti pattern and potentially some issues in the future.  The issue is that the hostname (defined by metadata.name) has to be the same in both pods - the old nova compute pod and the new neutron DVS pod. 

The name needs to be the same in order for it to work - so the DVS agent is registered properly and processes the binding requests for the compute pod. 

Since the hostname also forms part of pod DNS record, we introduce with this change a conflict is DNS with two A records for the compute-xxxx hostname. Personally I think this is an indicator that its not the correct pattern and we are simplifying deployment at the cost of 'breaking' the running state. Sure the DNS is likely not an issue at the moment I think, but I assume there is a working assumption on the k8s side that (per normal convention) hostnames in a domain are unique.

joker-official:
Closing because we removed DVS.

PR #473, opened 2018-11-20
add role: limes_technical_user

majewsky:
Right now, quotas can be set with the `cloud_*_admin` roles. However, human users should not be able to bypass Limes, so editing quotas should be restricted to Limes' technical user. This new role will be used for that purpose.

@ruvr If this is okay with you, please merge and roll this out to all Keystones. I'll take care of the Limes part.


ruvr:
Please check my comment on https://issues.wdf.sap.corp/browse/CCM-8000: I would suggest a more expressive role name

majewsky:
@ruvr Does this look good for you now? I would like to quickly move forward with this since this is part of a larger activity that needs to be completed until Jan 2019.

PR #474, opened 2018-11-22
*** ON HOLD *** add webhook for datahub alarm state capture

jobrs:
do not merge yet, this is early stages. Probably too early to review


jobrs:
for now I will try to get the alarm info using federation since that is easier to implement

PR #478, opened 2018-11-26
[ironic][alerts] add alert of deploy failed and clean failed states

galkindmitrii:


BerndKue:
Wouldn't it make sense to also have an alert for nodes in Error state?
The rest looks also good to me.

galkindmitrii:
There is just no last_error value in ironic..

$ os baremetal node show node007-bm091 -c last_error
+------------+-------+
| Field      | Value |
+------------+-------+
| last_error | None  |
+------------+-------+
rally@342577977b44:~/.rally$ os baremetal node list | grep node007-bm091              
| cce9018a-c3fa-4b61-8f4f-8757328db13a | node007-bm091 | None                                 | power off   | clean failed       | True        |

BerndKue:
@rajivmucheli Would it then make sense to have a default value for label "last_error" that is showing "reasons unknown" or "error unknown" for example if there is no "last_error" in ironic?

rajivmucheli:
@BerndKue Sure, Nodes in error state have last_error, maybe deploy & clean failed provision states have no data in last_error cell, coalesce() helps to replace blank cells with a value. What would like to have default value as ?

coalesce(last_error,'error unknown')  ?

BerndKue:
LGTM

PR #479, opened 2018-11-26
Sap seeds

fwiesel:
Move out seeds not required to run the service to sap-seeds.


fwiesel:
Hi,

while going over the flavours, I was thinking, that most domains are also not essential for running the service, so deploying a new domain requires deploying all services.

I've heard though, that we had a central setup before, and went to a distributed one.

PR #485, opened 2018-11-28
Neutron policy: restrict quota updates to Limes user

majewsky:


majewsky:
We tested this in staging last week and it didn't restrict quota PUTs as intended. @notandy Did you have any chance to check why that is the case?

notandy:
Urgh, can you handle it? also feel free to deploy in staging, I am still quite busy to get the nova upgrade done with year

notandy:
Mkay, will look into it, but I am not sure being done earlier this month.

PR #486, opened 2018-11-28
Nova policy: restrict quota updates to Limes user

majewsky:


openstack/nova/templates/etc/_nova-policy.json.tpl
```diff
@@ -176,9 +177,9 @@
     "compute_extension:networks:view": "!",
     "compute_extension:networks_associate": "!",
     "compute_extension:os-tenant-networks": "!",
-    "compute_extension:quotas:show": "rule:context_is_viewer",
-    "compute_extension:quotas:update": "rule:context_is_admin",
-    "compute_extension:quotas:delete": "rule:context_is_admin",
+    "compute_extension:quotas:show": "rule:context_is_viewer or rule:context_is_quota_admin",
+    "compute_extension:quotas:update": "rule:context_is_quota_admin",
+    "compute_extension:quotas:delete": "rule:context_is_quota_admin",
     "compute_extension:quota_classes": "rule:context_is_admin",
```
joker-official: Why are `quota_classes` not necessary for `nova` but are necessary for `cinder`?

openstack/nova/templates/etc/_nova-policy.json.tpl
```diff
@@ -176,9 +177,9 @@
     "compute_extension:networks:view": "!",
     "compute_extension:networks_associate": "!",
     "compute_extension:os-tenant-networks": "!",
-    "compute_extension:quotas:show": "rule:context_is_viewer",
-    "compute_extension:quotas:update": "rule:context_is_admin",
-    "compute_extension:quotas:delete": "rule:context_is_admin",
+    "compute_extension:quotas:show": "rule:context_is_viewer or rule:context_is_quota_admin",
+    "compute_extension:quotas:update": "rule:context_is_quota_admin",
+    "compute_extension:quotas:delete": "rule:context_is_quota_admin",
```
joker-official: @fwiesel Does that API still exist in nova queens? I cannot find it there. Currently, this is the `legacy_v2` api. If we roll out queens in the next week, this would essentially do nothing, right?

fwiesel: I thought, they are still available, but not as an extension.
It would be quota-sets. Not sure, how it translates to the policy though.

openstack/nova/templates/etc/_nova-policy.json.tpl
majewsky: The problem is that `quota_classes` (apparently) has only one permission for both read and write. I don't want to inadvertently take read permissions away from users that need it.

majewsky:
What's the status here? This needs to be rolled out before the holidays. If you have any questions, please let me know.

majewsky:
Bump. If this is okay, please merge and rollout. Otherwise, let me know what needs to be changed.

joker-official:
Thank you for the reminder.

joker-official:
Rolled out everywhere.

PR #487, opened 2018-11-28
Cinder policy: restrict quota updates to Limes user

majewsky:

### DIFF
--- a/openstack/cinder/templates/etc/_cinder-policy.json.tpl
+++ b/openstack/cinder/templates/etc/_cinder-policy.json.tpl
@@ -1,5 +1,6 @@
 {
   "context_is_cloud_admin":  "role:cloud_volume_admin",
+  "context_is_quota_admin":  "role:resource_service",
   "context_is_admin":  "rule:context_is_cloud_admin",
   "owner": "project_id:%(project_id)s",
   "member": "role:member and rule:owner",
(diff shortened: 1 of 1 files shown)


openstack/cinder/templates/etc/_cinder-policy.json.tpl
```diff
@@ -52,11 +53,11 @@
...
-  "volume_extension:quotas:show": "rule:context_is_viewer",
-  "volume_extension:quotas:update": "rule:context_is_admin",
-  "volume_extension:quotas:delete": "rule:context_is_admin",
-  "volume_extension:quota_classes": "rule:context_is_admin",
-  "volume_extension:quota_classes:validate_setup_for_nested_quota_use": "rule:context_is_admin",
+  "volume_extension:quotas:show": "rule:context_is_viewer or rule:context_is_quota_admin",
+  "volume_extension:quotas:update": "rule:context_is_quota_admin",
+  "volume_extension:quotas:delete": "rule:context_is_quota_admin",
+  "volume_extension:quota_classes": "rule:context_is_admin or rule:context_is_quota_admin",
+  "volume_extension:quota_classes:validate_setup_for_nested_quota_use": "rule:context_is_quota_admin",
```
joker-official: This last line won't work.

joker-official: It calls `cinder.quota_utls.validate_setup_for_nested_quota_use` which calls `cinder.quota.NestedDbQuotaDriver.validate_nested_setup` which calls `cinder.db.sqlalchemy.api.quota_allocated_update` which requires an admin context, which itself only uses `rule:context_is_admin` which would fail to validate.

joker-official: So this is not actually needed by Limes?

majewsky:
What's the status here? This needs to be rolled out before the holidays. If you have any questions, please let me know.

joker-official:
Squashed and merged. Will roll out over the next couple of days.

PR #489, opened 2018-11-28
remove unnecessary swiftreseller role assignments

majewsky:
grafik

### DIFF
--- a/openstack/swift/templates/seeds.yaml
+++ b/openstack/swift/templates/seeds.yaml
@@ -15,9 +15,6 @@ metadata:
     type: seed
 spec:
   requires:
-  #- monsoon3/keystone-seed                 # uncomment once keystone is deployed via its own helm-charts chart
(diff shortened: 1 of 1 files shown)


openstack/swift/templates/seeds.yaml
```diff
@@ -60,33 +55,19 @@ spec:
     - name: swift_dispersion
       description: 'Swift Dispersion'
       password: {{ $.Values.dispersion_password }}
-    - name: ironic
-      roles:
-      - project: service
-        role: swiftreseller
```
majewsky: This looks wrong, but it's okay because the Ironic seed [assigns this role](https://github.com/sapcc/helm-charts/blob/8d779fa475dddb591dc2bbd4aa94a12498c48087/openstack/ironic/templates/seed.yaml#L218-L219) and has the correct dependency on this seed.

reimannf: agree

PR #490, opened 2018-11-28
*WIP* Enable Debezium Change-Data-Capture on Postgres and MariaDB

jobrs:
Postgres
* use image with required plugins
* enable logical write-ahead logging
* permit WAL consumers to connect and consume WAL segments

Open Issues
- [ ] create database user for Debezium
- [ ] make sure Kafka Connect (scale-out cluster) can connect to Postgres/MariaDB (e.g. verify pg_hba.conf settings)
- [ ] implement also for MariaDB

### DIFF
--- a/common/postgresql/templates/etc/_pg_hba.conf.tpl
+++ b/common/postgresql/templates/etc/_pg_hba.conf.tpl
@@ -15,4 +15,13 @@ host    all             all             ::1/128                 trust
 #host    replication     postgres        127.0.0.1/32            trust
 #host    replication     postgres        ::1/128                 trust
 
+{{- if $.Values.cdc.enabled }}
+# All replication connections from Debezium connector running on KafkaConnect
+local   replication     debezium                                trust
+host    replication     debezium        127.0.0.1/32            trust
+host    replication     debezium        ::1/128                 trust
+# permit access from same cluster network (Kubernikus)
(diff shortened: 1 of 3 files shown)


jobrs:
in-preparation ... see https://issues.wdf.sap.corp/browse/CCM-8115

jobrs:
updated: KafkaConnect moves into the control-plane. Namespace "eventhub"

stefanhipfel:
Hi, do we need to upgrade to postgres v. 9.6 and use the debezium image?

jobrs:
Hi Stefan,

the major addition they made are the two logical decoding plugins. You missed them or are they installed by other means?

If we stick with our current helm chart (derived from the "official" one as of 9.5), we should be able to use the Debezium image as a plugin replacement. Thats no longer the case with later versions of the official helm chart.

stefanhipfel:
I removed the changes in the values file, since this is the default for all our services. If they are not overwritten in the services charts, they will fall back to default. 
That is why we cannot set anything debezium specific in the defaults

jobrs:
that's fine, meanwhile only decoderbufs is needed

https://github.wdf.sap.corp/cc/realtime-inventory/commit/5eaee7b546921503008cfb03d91582ec9d007ec9

PR #499, opened 2018-12-04
[common][postgresql] updates postgres settings to allow more connections

stefanhipfel:


stefanhipfel:
- Lets put our default postgres/pgbouncer settings in one place.
- Added Andrew B. and K.s new postgres tweaks
- allow more than 8 pgbouncer -> Postgres connections

stefanhipfel:
do not merge!!!!

fwiesel:
We also might want to set `random_page_cost` to 1.1

stefanhipfel:
Good point! Do we know the Random Reads and 
Sequential Reads of the SSD we are using to calculate the ratio for
random_page_cost
seq_page_cost?

common/postgresql/values.yaml
```diff
@@ -16,7 +16,7 @@ global:
       max_client_conn: 16384
       reserve_pool_size: 100
       min_pool_size: 1
-      default_pool_size: 50 # per user and database
+      default_pool_size: 100 # per user and database
```
fwiesel: If I read it correctly, you set the pool size higher than the number of db connections (one line below)

stefanhipfel: ok, i thought this is why we have a pooler. pool connections > db connections.
actually it is: db_count+user_count*(default_pool_size + reserve_pool_size, max_db_connections)

common/postgresql/templates/etc/_postgresql.conf.tpl
```diff
@@ -116,21 +116,21 @@ max_connections = {{.Values.max_connections | default 400 }}			# (change require
 
 # - Memory -
 
-shared_buffers = 128MB			# min 128kB
+shared_buffers = {{.Values.shared_buffers | default 128MB }}			# min 128kB
```
fwiesel: Doesn't need that to be a string?
I mean `"128MB"`?

PR #500, opened 2018-12-05
Nova queens support

notandy:
This is the production-ready implementation of nova-queens, including vspc, mks console etc.

It runs alongside with mitaka regions.

It also incoroprate a queens-upgrade-dry-run job, that clones the database and run all offline-migrations in a dry run to ensure successful upgrade.


openstack/nova/templates/_console_deployment.yaml.tpl
```diff
@@ -42,7 +42,7 @@ spec:
           name: nova-etc
       containers:
       - name: nova-console-{{ $name }}
-        image: {{ .Values.global.imageRegistry}}/{{.Values.global.image_namespace}}/ubuntu-source-nova-{{ $name }}proxy:{{index .Values (print "imageVersionNova" (title $name) "proxy") | default .Values.imageVersion | required "Please set nova.imageVersion or similar" }}
```
fwiesel: Unfortunately, this breaks the pipeline, as it cannot derive the image name without making a full template evaluation:

openstack/nova/values.yaml
```diff
@@ -86,13 +90,21 @@ portMetrics: '9102'
...
 sentry_dsn: DEFINE_IN_REGION_CHART
 
+# Use loci only for neutron
+loci:
+  neutron: true
+  nova: false
+
 #TODO we need to move to global or find another way to share image versions
+imageNameNeutron: loci-neutron
+imageVersionNeutronServerM3: null
```
fwiesel: There is no `imageVersionNeutronServerM3` anymore

openstack/nova/values.yaml
```diff
@@ -86,13 +90,21 @@ portMetrics: '9102'
...
 imageVersionNeutronServer: null
 imageVersionNeutronOpenvswitchAgent: null
-imageVersionOpenvswitchVswitchd: null
-imageVersionOpenvswitchDbServer: null
+imageVersionNeutronOpenvswitchVswitchd: null
```
fwiesel: The openvswitchd is not an neutron image, it contains simply the vswitchd, hence the name.
The image from kolla is called openstack-openvswitch-vswitchd, because there is no neutron code in there.

openstack/nova/templates/etc/_nova.conf.tpl
```diff
@@ -74,13 +74,17 @@ port = {{ .Values.consoles.shellinabox.portInternal }}
 base_url = https://{{include "nova_console_endpoint_host_public" .}}:{{ .Values.global.novaConsolePortPublic }}/shellinabox
 proxyclient_url = https://{{include "ironic_console_endpoint_host_public" .}}
 
+[mks]
```
fwiesel: I would say, this probably should rendered dependent on the value set here:
https://github.com/sapcc/helm-charts/pull/500/files#diff-e5e918ce7b67ccfe478849a3bcc2dfb1R152

openstack/nova/templates/hypervisors/_ironic-deployment.yaml.tpl
```diff
@@ -24,17 +24,18 @@ spec:
   template:
     metadata:
       labels:
-{{ tuple . "nova" "compute" | include "helm-toolkit.snippets.kubernetes_metadata_labels" | indent 8 }}
+{{ tuple . "nova" $compute_hypervisor_name | include "helm-toolkit.snippets.kubernetes_metadata_labels" | indent 8 }}
         name: nova-compute-{{$hypervisor.name}}
         hypervisor: "ironic"
       annotations:
+        pod.beta.kubernetes.io/hostname: nova-compute-{{$hypervisor.name}}
```
fwiesel: Please remove this

openstack/nova/templates/hypervisors/_ironic-deployment.yaml.tpl
```diff
@@ -24,17 +24,18 @@ spec:
   template:
     metadata:
       labels:
-{{ tuple . "nova" "compute" | include "helm-toolkit.snippets.kubernetes_metadata_labels" | indent 8 }}
+{{ tuple . "nova" $compute_hypervisor_name | include "helm-toolkit.snippets.kubernetes_metadata_labels" | indent 8 }}
```
fwiesel: Could you leave it to "nova" and "compute". Otherwise it is hard to get all the compute nodes via labels

openstack/nova/templates/hypervisors/_ironic-deployment.yaml.tpl
```diff
@@ -24,17 +24,18 @@ spec:
...
       annotations:
+        pod.beta.kubernetes.io/hostname: nova-compute-{{$hypervisor.name}}
         configmap-etc-hash: {{ include (print .Template.BasePath "/etc-configmap.yaml") . | sha256sum }}
         configmap-ironic-etc-hash: {{ tuple . $hypervisor | include "ironic_configmap" | sha256sum }}
     spec:
       terminationGracePeriodSeconds: {{ $hypervisor.default.graceful_shutdown_timeout | default .Values.defaults.default.graceful_shutdown_timeout | add 5 }}
       containers:
-        - name: nova-compute
-          image: {{.Values.global.imageRegistry}}/{{.Values.global.image_namespace}}/ubuntu-source-nova-compute:{{.Values.imageVersionNovaCompute | default .Values.imageVersion | required "Please set nova.imageVersion or similar" }}
+        - name: nova-compute-{{$hypervisor.name}}
```
fwiesel: Could you leave it on `nova-compute`, otherwise you have to change two items to get the logs of a nova-compute nodes, the name of the pod, and the name of the container, instead of only one, the name of the pod.

openstack/nova/templates/update-cells-job.yaml
```diff
@@ -0,0 +1,59 @@
+{{- if hasPrefix "queens" .Values.imageVersionNova }}
```
fwiesel: I would suggests to do the opposite: `not hasPrefix  "mitaka"`, as we will probably update to rocky soonish, and still run this job.

openstack/nova/templates/placement-api-deployment.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.placement.enabled }}
+{{- if hasPrefix "queens" .Values.imageVersionNova }}
```
fwiesel: `not hasPrefix  "mitaka"`

openstack/nova/templates/etc/_nova-compute-ironic.conf.tpl
```diff
@@ -12,15 +12,35 @@ scheduler_instance_sync_interval = {{ .Values.scheduler.scheduler_instance_sync_
 
 [ironic]
 {{ $user := print (coalesce .Values.global.ironicServiceUser .Values.global.ironic_service_user "ironic") (default "" .Values.global.user_suffix) }}
+
+{{- if hasPrefix "queens" .Values.imageVersionNova }}
```
fwiesel: `not hasPrefix  "mitaka"`

openstack/nova/templates/_console_deployment.yaml.tpl
notandy: You could match the tuple . "XYZ" instead?

openstack/nova/values.yaml
notandy: openvswitchd isn't used from kolla anymore

openstack/nova/templates/etc/_nova.conf.tpl
notandy: I don't understand.

openstack/nova/templates/hypervisors/_ironic-deployment.yaml.tpl
notandy: This has been introduced by @galkindmitrii, please resolve this

openstack/nova/templates/hypervisors/_ironic-deployment.yaml.tpl
notandy: This has been introduced by @galkindmitrii, please resolve this

openstack/nova/templates/hypervisors/_ironic-deployment.yaml.tpl
notandy: This has been introduced by @galkindmitrii, please resolve this

openstack/nova/templates/update-cells-job.yaml
notandy: I will remove all prefixes / if statements after upgraded all regions. So it make no difference really

openstack/nova/templates/placement-api-deployment.yaml
notandy: se previous comment

openstack/nova/templates/etc/_nova-compute-ironic.conf.tpl
notandy: se previous comment

openstack/nova/templates/hypervisors/_ironic-deployment.yaml.tpl
fwiesel: It looks to me more like your rebasing re-introduced an old change, which was superseeded by a later change from me

openstack/nova/values.yaml
notandy: ok

openstack/nova/templates/etc/_nova.conf.tpl
fwiesel: In the values.yaml, you write under `consoles.mks` that setting the values will enable the the mks console.
I think, it would be sensible, that the conf file would follow that. Meaning, only to add the section, if the settings are there.

notandy: ok

openstack/nova/templates/hypervisors/_ironic-deployment.yaml.tpl
notandy: I removed dmitrys commits, is that ok @galkindmitrii ?

openstack/nova/templates/etc/_nova-compute-ironic.conf.tpl
```diff
@@ -12,15 +12,35 @@ scheduler_instance_sync_interval = {{ .Values.scheduler.scheduler_instance_sync_
...
 admin_url = {{.Values.global.keystone_api_endpoint_protocol_internal | default "http"}}://{{include "keystone_api_endpoint_host_internal" .}}:{{ .Values.global.keystone_api_port_internal | default 5000}}/v3
 admin_username = {{ $user }}
 admin_user_domain_name = {{ default "Default" .Values.global.keystone_service_domain }}
 admin_password = {{ coalesce .Values.global.ironicServicePassword .Values.global.ironic_service_password  (tuple . $user | include "identity.password_for_user")  | replace "$" "$$" }}
 admin_tenant_name = {{ default "service" .Values.global.keystone_service_project }}
 admin_project_domain_name = {{ default "Default" .Values.global.keystone_service_domain }}
 api_endpoint = {{ default "http" .Values.global.ironic_api_endpoint_protocol_admin }}://{{ include "ironic_api_endpoint_host_internal" .}}:{{ .Values.global.ironic_api_port_internal | default "6385" }}/v1
+{{- end }}
+
+{{- if contains "testing" $hypervisor.name }}
```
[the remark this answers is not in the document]
notandy: ok, I removed the ironic patches from my queens branch

openstack/nova/templates/_console_deployment.yaml.tpl
notandy: btw, what exactly breaks at the pipeline? I didn't expected that

openstack/nova/values.yaml
fwiesel: Your template suggests, that it is an option to use either kolla or loci images for the openvswitch parts.

If you want to only support the loci images there, remove the option.
If you want to support the change, you would need to fix the image-version attributes.

notandy: you're right. there is no kolla support for neutron-related containers. I will remove it

openstack/nova/templates/_console_deployment.yaml.tpl
fwiesel: Could you separate this change out? There is no need to do this in one go.

notandy: just for console images or for all?

fwiesel: Hmm, let me see... maybe I can fix it in the template after all.

openstack/nova/values.yaml
notandy: I removed "kolla" support for neutron related containers since they would probably won't work anyway

openstack/nova/templates/_console_deployment.yaml.tpl
fwiesel: Okay, I gave it a try:
While I can match on tuple, which image is actually behind the image cannot be derived without template evaluation and the knowledge of the actual value.

I would say, leave all the "container_image.*" templates out, and we can revisit how to make the deployment image agnostic another time.

notandy: this will throwing us back in weeks

fwiesel: What will take weeks here?

notandy: cause I need to test it with queens and mitaka, this I cannot do in this week

notandy: please point me to the broken pipeline. I don't understand the ruby template problem

notandy: please recheck

openstack/nova/templates/_helpers.tpl
```diff
@@ -0,0 +1,51 @@
+{{- define "rabbitmq_user" -}}
+  {{.Values.rabbitmq.users.default.user}}
+{{- end -}}
+
+{{- define "rabbitmq_password" -}}
+  {{ .Values.rabbitmq.users.default.password | default (tuple . .Values.rabbitmq.users.default.user | include "rabbitmq.password_for_user")}}
```
joker-official: Can we use the above-defined `rabbitmq_user` here? Just in case that will not stay `.Values.rabbitmq.users.default.user` forever.

openstack/nova/templates/_helpers.tpl
```diff
@@ -0,0 +1,51 @@
...
+{{- end -}}
+
+{{- define "cell0_db_path" -}}
+{{.Values.cell0dbUser}}:{{.Values.cell0dbPassword | default .Values.apidbPassword | urlquery}}@{{.Chart.Name}}-postgresql.{{include "svc_fqdn" .}}:5432/{{.Values.cell0dbName}}
+{{- end -}}
+
+{{- define "rabbit_url" -}}
+  {{- $rabbitmq_host := printf "nova-rabbitmq.%s" ( include "svc_fqdn" . ) -}}
+  rabbit://:{{ include "rabbitmq_password" . | urlquery}}@{{ $rabbitmq_host }}:5672/
```
joker-official: Is the user missing here or is there no user needed?

fwiesel: `rabbit_url` is not used. Best to remove it

openstack/nova/templates/_helpers.tpl
```diff
@@ -0,0 +1,51 @@
...
+{{- end -}}
+
+{{- define "container_image_nova" -}}
+  {{- $name := index . 1 -}}
+  {{- with index . 0 -}}
+    {{- $version_name := printf "imageVersionNova%s" ($name | lower | replace "-" " " | title | nospace) -}}
+    {{- $image_name := ( .Values.loci.nova | ternary .Values.imageNameNova (printf "ubuntu-source-nova-%s" ($name | lower)) ) -}}
```
joker-official: Here's a loci reference. Didn't loci-support get reverted? Was it reverted only in parts?

joker-official: Oh. I think this is the one that was included for "loci support" and is thus currently unused.

openstack/nova/templates/bin/_db-migrate.tpl
```diff
@@ -0,0 +1,43 @@
...
+EOT
+fi
+nova-manage cell_v2 map_cell0 --database_connection postgresql+psycopg2://{{ include "cell0_db_path" . }} || true
+
+virtualenv /tmp/nova-pike
+#curl -L https://github.com/openstack/nova/archive/16.1.6.tar.gz | tar -xzC /tmp/nova-pike
+git clone --depth 1 -b stable/pike https://github.com/openstack/nova.git /tmp/nova-pike/nova
+curl -L --output /tmp/nova-pike/upper-constraints.txt https://raw.githubusercontent.com/openstack/requirements/stable/pike/upper-constraints.txt
+source /tmp/nova-pike/bin/activate
+pip install -c /tmp/nova-pike/upper-constraints.txt -r /tmp/nova-pike/nova/requirements.txt /tmp/nova-pike/nova raven psycopg2
```
joker-official: Shouldn't raven be installed from `https://github.com/sapcc/raven-python`?

openstack/nova/templates/_helpers.tpl
notandy: True, will remove it

openstack/nova/templates/_helpers.tpl
notandy: not used anymore, I remove it

openstack/nova/templates/_helpers.tpl
notandy: exactly, since I leave it since I want to reimplement it as soon the pipelines are adapted

openstack/nova/templates/bin/_db-migrate.tpl
notandy: nope, this is only a one-time virtualenv for database migration, no need for raven since it wouldn't be configured anyway

joker-official: Then why is `raven` explicitly in there anyway?

notandy: > ... to mitigate error messages

joker-official: oh. just read the email, that didn't get you update. sorry.

notandy: np

openstack/nova/templates/bin/_queens-dry-run.tpl
```diff
@@ -0,0 +1,77 @@
...
+
+
+mkdir -p /etc/nova
+cat <<EOT >> /etc/nova/nova.conf
+[DEFAULT]
+
+[database]
+connection = postgresql+psycopg2://nova:test@localhost/nova
+[api_database]
+connection = postgresql+psycopg2://nova_api:test@localhost/nova_api
```
joker-official: Does postgres not care for upper/lower-case in usernames? Because the user is created as `nova_API`.

notandy: oh you're right, typo. but did seemed to work

joker-official:
https://github.com/sapcc/helm-charts/pull/500/files#diff-bfd597b5dc494e6bced00b3ae38c3622R94 <- were'nt we going to remove that line, so it would take v3 automatically? I remember having some errors with v2 in staging in recent times.

openstack/nova/templates/hypervisors/_kvm-deployment.yaml.tpl
```diff
@@ -92,8 +99,12 @@ spec:
               readOnly: true
             - mountPath: /nova-patches
               name: nova-patches
+            - mountPath: /etc/nova/rootwrap.conf
```
joker-official: Doesn't nova need any `/etc/rootwrap.d/` stuff?

notandy: kvm nodes need it because they execute alot of local cli commands of libvirt/etc.

joker-official: But `rootwrap.conf` is not enough most of the time. The real config is in files in  `/etc/rootwrap.d/`. We just recently added one for cinder.

notandy:
> https://github.com/sapcc/helm-charts/pull/500/files#diff-bfd597b5dc494e6bced00b3ae38c3622R94 <- were'nt we going to remove that line, so it would take v3 automatically? I remember having some errors with v2 in staging in recent times.

Sorry, what do you mean?

openstack/nova/templates/hypervisors/_kvm-deployment.yaml.tpl
fwiesel: Indeed, nova also seems to have introduced new rootwrap.d files:
https://github.com/sapcc/nova/tree/stable/queens-m3/etc/nova/rootwrap.d

joker-official:
> > https://github.com/sapcc/helm-charts/pull/500/files#diff-bfd597b5dc494e6bced00b3ae38c3622R94 <- were'nt we going to remove that line, so it would take v3 automatically? I remember having some errors with v2 in staging in recent times.
> 
> Sorry, what do you mean?

It didn't take the line into the link. I was referring to `catalog_info` in the nova .conf

openstack/nova/templates/hypervisors/_kvm-deployment.yaml.tpl
notandy: nova will use the preinstalled rootwrap.d files

joker-official: They get installed automatically? They were'nt for cinder.

notandy: https://github.com/sapcc/helm-charts/blob/master/openstack/nova/templates/etc/_rootwrap.conf.tpl#L2

notandy: yes nova setup.py installs them

notandy:
Ok, I'l integrate a queens switch then since we don't know how mitaka behaves without catalog_info

notandy:
merging tmrw if no objections left, thanks!

PR #505, opened 2018-12-06
added three new flavors with 1:4 ratio

frr2018:
Added the following flavors with 1:4 ratio:

m5.16xlarge
vCPU: 64
RAM: 256

m5.32xlarge
vCPU: 128
RAM: 512

m5.64xlarge
vCPU: 256
RAM 1024

Bye,

Frank


fwiesel:
Could you rebase your changes? It looks like you are pulling in some changes, which are already in master

PR #506, opened 2018-12-08
[nova] adds dbUser to globals. Fixes pgmetrics auth error!

stefanhipfel:
…efault dbuser

### DIFF
--- a/openstack/nova/values.yaml
+++ b/openstack/nova/values.yaml
@@ -12,6 +12,8 @@ global:
   placementApiPortInternal: '8778'
   placementApiPortPublic: '443'
   master_password: null
+  
+  dbUser: nova
 
   nova_service_user: nova
   placement_service_user: placement


stefanhipfel:
the dbuser is not setup in globals, therefor pgmetrics will use the default user (postgres):
https://github.com/sapcc/helm-charts/blob/master/common/pgmetrics/templates/deployment.yaml#L41

We also have pgmetrics  deployment in the postgresql charts. I think we should remove it!

notandy:
Hi,

I would also like to see the metrics using the .Values.db_user, or even better using the local postgres domainsocket as Postgres user without authentification, this would provide metrics even if pgbouncer are unable to connect (e.g. due to pool exhaustion).

fwiesel:
We moved away from a local socket, as that requires the container to be in the same pod, and that would mean restarting the pod on a metric change.

Using a domain-scheduling would require some tricks with scheduling hostpaths, or localstorage.

notandy:
ah ok, make sense (when does kubernetes support restart of single containers!?!).
can we use the postgres user for metrics?

fwiesel:
Now that I think about it, we could co-locate the containers, but we would need a shared pid-space.
Then we could add another side-car container, which would trigger a restart of the metric container (e.g. kill the corresponding process), if the metric config changes.

Alternatively, we could patch the exporter to monitor the file for changes, which would eliminate the need for external changes.

stefanhipfel:
This was meant as a quick fix. Merge it if you want it, as I am away on a short trip.

PR #507, opened 2018-12-10
[neutron] Add rabbitmq to neutron-chart

fwiesel:
In order to migrate away from the openstack-helm rabbitmq,
first create a rabbitmq to be used.
As a next step, the service entry of the central rabbitmq will point
to this rabbit.
Then we can configure everything to use the new rabbitmq directly.
Finally, we can tear down the central rabbitmq.

### DIFF
--- a/openstack/neutron/requirements.lock
+++ b/openstack/neutron/requirements.lock
@@ -5,6 +5,9 @@ dependencies:
 - name: pgmetrics
   repository: file://../../common/pgmetrics
   version: 0.1.0
+- name: rabbitmq
+  repository: file://../../common/rabbitmq
+  version: 0.1.0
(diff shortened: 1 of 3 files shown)


fwiesel:
We also need to set the passwords accordingly, so that the neutron-rabbitmq will allow access with the same credentials as the central rabbitmq.

fwiesel:
See sapcc/openstack-helm@253d49c1362a6216a84a9cb28a49e4e2f20ced51 for the openstack-helm compatibility.

fwiesel:
I've already copied the necessary credentials, so that if this is merged and deployed, the neutron-rabbitmq will be accessible with the same credentials as the one from opentack-helm

PR #519, opened 2018-12-19
[nova] add support for sentry DSN from secret

ruvr:
@fwiesel This could save you maintaining 4 fresh nova sentry DSN's..
just drop the former sentry_dsn from values.yaml in the regions where sentry has been re-installed..

### DIFF
--- a/openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml
+++ b/openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml
@@ -51,7 +51,7 @@ template: |
             value: "nova-compute"
           - name: NAMESPACE
             value: {{ .Release.Namespace }}
-            {{- if and .Values.sentry.enabled (.Capabilities.APIVersions.Has "sentry.sap.cc/v1") }}
+            {{- if .Values.sentry.enabled }}
             - name: SENTRY_DSN
               valueFrom:
                 secretKeyRef:
--- a/openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml
+++ b/openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml
@@ -51,17 +51,13 @@ template: |
             value: "nova-compute"
           - name: NAMESPACE
             value: {{ .Release.Namespace }}
-          {{- if .Values.sentry.enabled }}
-          - name: SENTRY_DSN
-          {{- if .Values.sentry.dsn }}
-            value: {{ .Values.sentry.dsn | quote}}
-          {{ else }}
-            valueFrom:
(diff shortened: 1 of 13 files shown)


fwiesel:
Thanks, that is great.

@joker-official you wanted to change it. Is that one okay for you?

openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml
```diff
@@ -103,8 +108,17 @@ template: |
           command: ["dumb-init"]
           args: ["neutron-dvs-agent", "--config-file", "/etc/neutron/neutron.conf", "--config-file", "/etc/neutron/plugins/ml2/ml2-conf.ini", "--config-file", "/etc/neutron/plugins/ml2/ml2-vmware.ini"]
           env:
-          - name: SENTRY_DSN
-            value: {{ .Values.neutron.sentry_dsn | quote }}
+            {{- if .Values.sentry.enabled }}
```
joker-official: @fwiesel If you check for capabilities (which nobody else does btw.), you'd have to check here, too.

fwiesel:
Something like this?
I've added also a check on the existence of the sentry-operator.

joker-official:
Is there a way to update the `.Values.sentry.enabled` with the capability-check? Looks quite ugly when checked everywhere.

fwiesel:
To my knowledge, you can't compute values. The only way you can compute something is in a template.

If you want to go a minimalistic way, I would say there are two ways to go about it:

- Remove the check, and people should know that there is an order of deploying things (e.g. first the operator, then the app), and if they don't they get an error on deploying.
- Remove the option to enable/disable sentry for apps, and only check for the existence of the operator.

openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml
fwiesel: Ah, I was planning to remove that part (See #471)

fwiesel: Exactly for things like this, I think it is sensible to move the dvs agent out of the nova chart.
We need to copy values around and keep them in sync.

fwiesel:
Personally, I am not a fan of having two places to configure, and you have to match them that it works. But here you.

joker-official:
I get that, but my rationale is: If you look into the config and it's set to enabled, then it should be enabled and not only enabled if the sentry operator was deployed before or not. The place you'll look at are the values. You're not gonna read the chart for that. The deployment should rather fail and the user should sort it out.

joker-official:
superseded by https://github.com/sapcc/helm-charts/pull/2173

PR #524, opened 2019-01-08
alerting: add cloudops link; region should always be the openstack region

auhlig:
- add cloudops link
- region should always be the openstack region
- introduce `cluster_type` to distinguish between bm, kubernikus, scaleout, .. clusters. no paging for kubernikus, scaleout clusters

### DIFF
--- a/global/prometheus-alertmanager/slack.tmpl
+++ b/global/prometheus-alertmanager/slack.tmpl
@@ -6,7 +6,7 @@
 {{ define "slack.sapcc.text" }}
 {{ if eq .Status "firing" }}*[{{ .CommonLabels.severity | toUpper }}{{ if gt (len .Alerts.Firing) 1 }} - {{ .Alerts.Firing | len }}{{ end }}]* {{ end -}}
 {{ if eq .Status "resolved" }}*[RESOLVED{{ if gt (len .Alerts.Resolved) 1 }} - {{ .Alerts.Resolved | len }}{{ end }}]* {{ end -}}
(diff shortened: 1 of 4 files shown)


global/prometheus-alertmanager/slack.tmpl
```diff
@@ -25,6 +25,7 @@
 {{ if .CommonLabels.sentry }}*<https://sentry.{{ .CommonLabels.region }}.cloud.sap/monsoon/{{ .CommonLabels.sentry }}|Sentry>* {{ end -}}
 {{ if .CommonLabels.playbook }}*<https://operations.global.cloud.sap/{{ .CommonLabels.playbook }}|Playbook>* {{ end -}}
 {{ if .CommonLabels.kibana }}*<https://logs.{{ .CommonLabels.region }}.cloud.sap/{{ .CommonLabels.kibana }}|Kibana>* {{ end -}}
+{{ if .CommonLabels.cloudops }}*https://dashboard.{{ .CommonLabels.region }}.cloud.sap/ccadmin/cloud_admin/cloudops#/universal-search/{{ .CommonLabels.cloudops }}|CloudOps>* {{ end -}}
```
databus23: I think this is the wrong location. The link needs to be per alert similar as the link to prometheus. The `cloudops` can ( and will ) contain different content for each alert.

auhlig: true. moved cloudops, sentry link up

PR #527, opened 2019-01-14
[neutron] OS seeds for Neutron Tempest test.

unmerkbar:


notandy:
hi, just a question: do we need the accounts in all regions? if not you should probably add a switch for tempest.

PR #531, opened 2019-01-16
[swift] separate memcache for token validation

reimannf:


reimannf:
seems that a dedicated memcached for tokens is worth the effort

PR #533, opened 2019-01-17
enforce usage of the internal keystone endpoint (needs https://github…

ruvr:
….com/sapcc/keystonemiddleware to be functional) and enable or extend token caching times.

In order to bring down the keystone token validation request rates, I'ld like to extend token caching times client side and also make sure that service->keystone traffic takes a direct (internal) route.
This will offload ingress and also paves the way for easier source-ip based rate throttling for keystone, since all internal traffic will originate on the internal endpoint, leaving the public endpoint to be throttled without risking throttling our internal traffic.

Please note that this will only start working once all services use https://github.com/sapcc/keystonemiddleware, which does away with the hardcoded use of the 'admin' endpoint (as in upstream code) and the lacking possibility to hand down the desired endpoint to the keystone-client layer.

### DIFF
--- a/openstack/ironic/templates/etc/_ironic.conf.tpl
+++ b/openstack/ironic/templates/etc/_ironic.conf.tpl
@@ -41,7 +41,7 @@ auth_section = service_catalog
 
 [service_catalog]
 auth_section = service_catalog
-# ?? valid_interfaces = public {{- /* Public, so that the ironic-python-agent doesn't get a private url */}}
+valid_interfaces = public {{- /* Public, so that the ironic-python-agent doesn't get a private url */}}
 auth_type = v3password
 auth_interface = internal
 auth_version = v3
--- a/openstack/ironic/templates/etc/_ironic.conf.tpl
+++ b/openstack/ironic/templates/etc/_ironic.conf.tpl
@@ -33,20 +33,17 @@ public_endpoint = https://{{ include "ironic_api_endpoint_host_public" .}}
 
 [keystone]
 auth_section = service_catalog
-region = {{ .Values.global.region }}
 
 [keystone_authtoken]
 auth_section = service_catalog
-memcached_servers = {{ .Chart.Name }}-memcached.{{ include "svc_fqdn" . }}:{{ .Values.memcached.memcached.port | default 11211 }}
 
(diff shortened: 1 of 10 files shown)


openstack/ironic/templates/etc/_ironic.conf.tpl
```diff
@@ -33,20 +33,17 @@ public_endpoint = https://{{ include "ironic_api_endpoint_host_public" .}}
...
-memcached_servers = {{ .Chart.Name }}-memcached.{{ include "svc_fqdn" . }}:{{ .Values.memcached.memcached.port | default 11211 }}
 
 {{- include "ini_sections.audit_middleware_notifications" . }}
 
 [service_catalog]
 auth_section = service_catalog
-valid_interfaces = public {{- /* Public, so that the ironic-python-agent doesn't get a private url */}}
-region_name = {{ .Values.global.region }}
-# auth_section
+# ?? valid_interfaces = public {{- /* Public, so that the ironic-python-agent doesn't get a private url */}}
```
fwiesel: As it says in the comment.
If you set it to internal, ironic will pass the internal url to the ironic-python-agent.
And the agent then fails, as it is not running in the kubernetes network.

openstack/ironic/templates/etc/_ironic_inspector.conf.tpl
```diff
@@ -43,9 +36,10 @@ connection = {{ tuple . "ironic_inspector" "ironic_inspector" .Values.inspectord
 {{- include "ini_sections.audit_middleware_notifications" . }}
 
 [keystone_authtoken]
-www_authenticate_uri = {{.Values.global.keystone_api_endpoint_protocol_internal | default "http"}}://{{include "keystone_api_endpoint_host_internal" .}}:{{ .Values.global.keystone_api_port_internal | default 5000}}
+www_authenticate_uri = https://{{include "keystone_api_endpoint_host_public" .}}/v3
 auth_url = {{.Values.global.keystone_api_endpoint_protocol_internal | default "http"}}://{{include "keystone_api_endpoint_host_internal" .}}:{{ .Values.global.keystone_api_port_internal | default 5000}}/v3
 auth_type = v3password
+auth_interface = internal
```
fwiesel: Likely it will here be the same problem as above.

PR #534, opened 2019-01-18
Updated snmp generator and metrics for F5

m-kratochvil:
This update covers memory monitoring


m-kratochvil:
@thgrs

Kuckkuck:
If the generator is doing the config, than it doesn't make sense to update the generated config in the helm charts. Just chance the f5-generator.yaml and run the pipeline to generate a new config for f5.

m-kratochvil:
I fixed the metrics file, now it holds only the OIDs as it's supposed to.
@Kuckkuck @thgrs

m-kratochvil:
@thgrs

PR #541, opened 2019-01-22
[common][postgresql] use toYaml to write resources prop

stefanhipfel:


SchwarzM:
LGTM maybe someone else should also look it over

notque:
Should there be a default in values for it as well? Can resources be blank without it?

stefanhipfel:
the issue is, we want to only set requests, not limits. Our default values make that almost impossible...we would need to use null:
https://github.wdf.sap.corp/cc/secrets/pull/353

notque:
Understand the issue, and it's the right approach. But having default values for this new setup seems reasonable for requests, not limits. I'm only arguing for a basic default in values for this pr, with new values files for each system as we come up with the right requests for them. Otherwise it LGTM

common/postgresql/templates/deployment.yaml
```diff
@@ -85,15 +85,8 @@ spec:
...
-        {{- if .Values.resources.enabled }}
         resources:
-          limits:
-            cpu: {{ .Values.resources.limits.cpu | quote }}
-            memory: {{ .Values.resources.limits.memory | quote }}
-          requests:
-            cpu: {{ .Values.resources.requests.cpu | quote }}
-            memory: {{ .Values.resources.requests.memory | quote }}
-        {{- end }}
+{{ toYaml .Values.resources | indent 10 }}
```
BugRoger:
```suggestion
{{ toYaml (required "missing .resources" .Values.resources) | indent 10 }}
```

BugRoger: This will give a nicer error message when the values are missing.

common/postgresql/values.yaml
```diff
@@ -104,18 +104,6 @@ upgrades:
   revisionHistory: 3
   podReplacementStrategy: Recreate
```
BugRoger:
```suggestion
## Configure resource requests and limits	
## ref: http://kubernetes.io/docs/user-guide/compute-resources/	
##	
resources: {}
#  limits:	
#    memory: 256Mi	
#    cpu: 500m	
#  requests:	
#    memory: 256Mi	
#    cpu: 500m	
```

BugRoger: Defaults to empty resources. Guards against rendering error in the `toYAML` function. Gives an example of the expected values

notque:
I think it makes sense to move it from .Values.resources to .Values.postgres.resources so that it's clear in the values file what it's for, and it doesn't collide with anyone else making resources for other services.

ruvr:
I would suggest to not change it to .Values.postgres.resources, since that becomes .Values.postgres.postgres.resources in services that include the chart. 
Pretty unreadable and error prone.

notque:
@ruvr what do you suggest then considering right now the value is as the base level for values files? Isn't that going to collide with other services? Isn't that unreadable and error prone on a values level?

ruvr:
LGTM. Can we please merge? It is holding up keystone urgent deployments...

PR #549, opened 2019-01-29
cleanup of old asr and introduction of alot of helm require statements

notandy:


openstack/neutron/values.yaml
```diff
@@ -7,7 +7,7 @@ osprofiler: {}
 
 global:
   dbUser: neutron
-  dbPassword: DEFINED-IN-REGION-CHART
+  dbPassword: null
```
databus23: Have you tested that this actually works and fails? I would leave those values completly away. It is common practice in helm charts to have required values as a comment in the values file. E.g. https://github.com/helm/charts/blob/master/stable/postgresql/values.yaml#L80

notandy:
Sure, will adapt to the common practice

notandy:
rebased on current master

PR #560, opened 2019-02-05
[openstack/keystone] rate limit

auhlig:
Hey @ruvr,
Can you take a look? - especially at [the rates](https://github.com/sapcc/helm-charts/compare/ratelimit_keystone?expand=1#diff-e33696e96d527431f5e8390e1d9bd8dfR1)

Introduces:
- rate limit middleware. disabled by default
- 2nd memcache for rate limits to avoid negative impact on keystone memcache
- initial global, local rate limit for authenticaton requests

Thanks


auhlig:
makes sense

PR #568, opened 2019-02-07
Refactor prometheus resource values

jknipper:
This PR moves prometheus resource values to helm values file. I need to overwrite these values, eg. in scaleout klusters.


auhlig:
Thanks!

PR #574, opened 2019-02-11
[neutron] openvswitch mtu fix, neutron-openvswitch-agent-start is the only file that needs changing.

sapliedtke:
I think this is only one half of the solution, Ignition needs to make sure the network nodes have the correct MTU sizes as well

### DIFF
--- a/openstack/neutron/templates/bin/_neutron-dhcp-agent-start.tpl
+++ b/openstack/neutron/templates/bin/_neutron-dhcp-agent-start.tpl
@@ -41,11 +41,6 @@ function _start_application {
     done
 
 
-    # Make sure the MTU is correct
-
-    ovs-vsctl set int br-int mtu_request=9000
-
-
     # br-int interface isn't there, lets crash and try again - we have situation where this intermittently can happen
     # We can't start DHCP without it otherwise we lose teh DHCP ports. Kill all the ovs processes on the host to get a full
     # OVS restart which should make things better....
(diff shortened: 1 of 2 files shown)


notandy:
looks okay, even though the same interface is configured by two startup scripts.

sapliedtke:
different pods/container, we might even have to add more.

notandy:
yeah different pods, but same interface. The pods are using the host network namespace, so it's the same device

sapliedtke:
youre correct and it seems ovs-tools are working differently then i anticipated, back to the drawing board, we need to iterate through quite a few interfaces.

abattye:
I'm not sure this will work the script is working on value "cp_network_interface" as $interface. This is usually bond1. So  https://github.com/sapcc/helm-charts/pull/574/files#diff-74a8f112c7e5ce81dc59c3ce9d57c6e1R55 should I think read `ovs-vsctl set int br-${interface} mtu_request=9000`

sapliedtke:
yeah that PR was from when i didnt know yet how the setup works. Marian and Sebastian had a call with me on Friday where they told me what they know and we together figured out the right solution.
Expect a working PR by Monday Morning.

abattye:
Please don't hardcode bond1 - its configurable and can be accessed in the form br-${interface} - so the helm variable ${interface} should be used

sapliedtke:
ah ok, i wasnt sure about the variables, will update during my next shift, else its ok?

sapliedtke:
can someone point me to where those variables are defined? maybe another repo? im still learning these charts.

PR #580, opened 2019-02-13
adding jumpserver exporter to maia

viennaa:
@auhlig 
Does the exporter definition look good to you?


openstack/maia/templates/jumpserver-service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
auhlig: Currently the metrics will be scraped via the pod **and** the service as prometheus annotations are present on both resources. I think you can omit the service.

PR #584, opened 2019-02-14
Make ironic-exporter subchart optional

databus23:
This PR moves the ironic-exporter chart if the subchart folder and makes it a (local) dependency that can be disabled via value.

This PR is dependent on cc/secrets#508 which has to be merged first


databus23:
@auhlig If this works like I think I would suggest that yes

databus23:
@Kuckkuck Care to comment? I would like to merge this to get the exporter of of the scaleout deployment

databus23:
pinging @Kuckkuck

PR #592, opened 2019-02-15
**DO NOT MERGE** prepare global prometheus for additional regions

auhlig:
**DO NOT MERGE**

Configurable ingress host for global prometheus.
Allows `prometheus.global.tld` or something like `prometheus-global.<region>.tld`.
LGTY @BugRoger @databus23?

### DIFF
--- a/openstack/maia/templates/_snmp-exporter-f5.yaml.tpl
+++ b/openstack/maia/templates/_snmp-exporter-f5.yaml.tpl
@@ -1,6 +1,10 @@
 # WARNING: This file was auto-generated using snmp_exporter generator, manual changes will be lost.
 f5:
   walk:
+  - 1.3.6.1.4.1.3375.2.1.7.5.2.1.19
+  - 1.3.6.1.4.1.3375.2.1.7.5.2.1.27
+  - 1.3.6.1.4.1.3375.2.1.8.2.3.1.37
+  - 1.3.6.1.4.1.3375.2.1.8.2.3.1.38
   - 1.3.6.1.4.1.3375.2.2.10.2.3.1.1
   - 1.3.6.1.4.1.3375.2.2.10.2.3.1.10
   - 1.3.6.1.4.1.3375.2.2.10.2.3.1.11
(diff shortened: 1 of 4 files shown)


vonneudeck:
FYI: You can create draft PRs now to avoid accidental merging and make everything more explicit.
https://twitter.com/github/status/1096093162345508864

auhlig:
Uh interesting. Thanks @vonneudeck

openstack/maia/templates/_snmp-exporter-f5.yaml.tpl
```diff
@@ -1,6 +1,10 @@
 # WARNING: This file was auto-generated using snmp_exporter generator, manual changes will be lost.
```
auhlig: I guess you hit wrong branch @m-kratochvil :)

m-kratochvil: uh, I thought this .tpl file has to be modified in the prometheus_global branch to federate the new metrics into global Prometheus.. just like you need to modify the grafana_global branch dashboard files to have them available in global grafana.. not the case?

auhlig: It's on the master branch. You can configure the metrics that are federated from the regional Prometheis to the global one [here](https://github.com/sapcc/helm-charts/blob/master/global/prometheus/templates/_prometheus.yaml.tpl#L18) by adding the metric name. Please add as few metrics as possible.

However, we don't federate maia metrics to the global Prometheus.

m-kratochvil: Arno, thanks for the details. The metrics are already present in the file you mentioned on master, but global Prometheus still doesn't offer them. And it offers some other f5 metrics that are not in the file. How did those metrics get there, I'm confused totally.. 
In any case, I guess we can leave this commit here too, it will not hurt anything, correct?

auhlig: Checked the global Prometheus. It needed a force-reload. Should be up-to-date with what is checked in in this rep now. Ping me in slack if you're missing metrics.

auhlig:
Will be done via prometheus operator

PR #596, opened 2019-02-19
[Nova] Adding common/region-check/ to requirements

rajivmucheli:
Hi @joker-official 

Do validate this PR.

Regards,
Rajiv


joker-official:
Could you please split that commit into two? You're clearly doing two different things here.

PR #603, opened 2019-02-20
Make kube-monitoring alerts installable by cluster type

jknipper:
This PR moves alerts in kube-monitoring to subfolders to make them installable by cluster type.


auhlig:
On first glance this makes it more complicated.
Alerts are installable by `tier` right now.
Enable/Disable alerts via values here: https://github.com/sapcc/helm-charts/blob/master/system/kube-monitoring/charts/prometheus-frontend/values.yaml#L19

databus23:
@auhlig The idea for this change came from the fact that we are have kubernetes-*.alerts that only apply for the baremtal control plane. We could move those a alerts into a separate tier but this would require manual configuration in far away places (e.g. cc/secrets).

I quite like this approach that we have alerts organised like this:
* `alerts/*.alerts` <- go in every deployment of `kube-monitoring` (global)
* `alerts/$cluster_type/*.alerts` <- go in the specific cluster_type.

Until we have a bang bang refactor using an operator this would allow us to have a more visual separation which alerts go in which type of cluster

auhlig:
Example: kubernetes alerts would be in the controlplane tier as well in the scaleout tier.  

Pro:
- Allow tuning of alerts per cluster type. 

Con:
- Might be confusing if alerts with the same name are configured in slightly different ways for each cluster type. <- That was my main concern here.

PR #612, opened 2019-02-26
[nova-compute-node] disable integration bridge

notandy:
since it's not used by the DVS driver and breaks the NSX-T driver

@joker-official @Scsabiii


joker-official:
Why do you comment it instead of removing it?

Could you share how it breaks the NSX-T driver?

Otherwise looks fine, as the docs state, that we don't need it for DVS.
> This option should be configured only when using the NSX-MH Neutron plugin.

notandy:
compute node aren't able to bind the network device the the nsx-t bridge, they seem to expect a nsx-mh bridge that doesn't exist (won't affect dvs)

PR #617, opened 2019-02-28
Use yellow color for mixed notifications

databus23:
This commit changes the color of the slack notification when it contains both firing and resolved alerts from read (danger) to yellow (warning).
In a lot of cases a notification that carries a "resolved" and "firing" alert is an improvment: Some alerts have cleared while others are still firing. Marking them as yellow makes them easily distingushibale from alerts that only contain "firing" alerts which are never good.

### DIFF
--- a/global/prometheus-alertmanager/slack.tmpl
+++ b/global/prometheus-alertmanager/slack.tmpl
@@ -1,3 +1,4 @@
+{{ define "slack.sapcc.color" }}{{ if .Alerts.Firing | and .Alerts.Resolved }}warning{{ else if .Alerts.Firing }}danger{{ else }}good{{ end }}{{ end }}
 {{ define "slack.sapcc.iconemoji" }}:fire_engine:{{ end }}
 {{ define "slack.sapcc.title" }}{{ end }}
 {{ define "slack.sapcc.titlelink" }}{{ end }}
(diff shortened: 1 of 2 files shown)


global/prometheus-alertmanager/slack.tmpl
```diff
@@ -1,3 +1,4 @@
+{{ define "slack.sapcc.color" }}{{ if .Alerts.Firing | and .Alerts.Resolved }}warning{{ else if .Alerts.Firing }}danger{{ else }}good{{ end }}{{ end }}
```
auhlig: Not sure I understand the syntax 1st if part: Should'nt it be `{{ if and .Alerts.Firing .Alerts.Resolved) }}warning ...`? 
Besides LGTM

databus23: https://golang.org/pkg/text/template/#hdr-Pipelines

> A pipeline may be "chained" by separating a sequence of commands with pipeline characters '|'. In a chained pipeline, the result of each command is passed as the last argument of the following command.

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -204,7 +204,7 @@ data:
     receivers:
     - name: dev-null
       slack_configs:
-        - api_url: {{ default "MISSING" .Values.slack.webhook_url | quote }}
+        - api_url: {{ required "slack.webhook_url undefined" .Values.slack.webhook_url | quote }}
```
BugRoger: drive by fixing 🥇

global/prometheus-alertmanager/slack.tmpl
databus23: It looks a little weird but is perfectly valid and allows to read it like `condition1 and condition2` which I like better then the reverse polish notation `and condition1 condition2`

PR #634, opened 2019-03-06
Larger channel clean up and restructuring

geisslet:
# In basic the changes follow 3 principles

1. less dynamic channel naming - why: simply reduce risk of non delivered messages in alert manager queues (metric: alertmanager_notifications_failed_total)
2. Differ between Duty (alert) Channels and Notification Channels (some clean up needed to have no overlap)
3. Notification channels should have an owner which also defines filter on labels - contact if you wanna adopt something

# In addition:
- add a alertchain test routing for pager duty
- add a cc-nanny channel
- concourse routes now to an existing channel


geisslet:
I propose to check the file completely in raw view and not in the diff view

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -128,72 +150,94 @@ data:
...
+      - receiver: slack_by_os_service
         continue: true
         match_re:
+          tier: os
           severity: info|warning|critical
-          service: concourse
+          service: arc|backup|barbican|cinder|cfm|designate|elektra|elk|glance|hermes|ironic|keystone|limes|lyra|maia|manila|neutron|nova|sentry|swift
 
       # do not page for kubernikus controlplane
       - receiver: dev-null
```
auhlig: Since you specified the `cluster_type` on pagerduty receivers to not be ` kubernikus-controlplane|kubernikus-scaleout` this receiver can be removed.

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -415,10 +509,25 @@ data:
               text: {{"'{{template \"slack.sapcc.silence1Month.actionText\" . }}'"}}
               value: {{"'{{template \"slack.sapcc.silence1Month.actionValue\" . }}'"}}
 
-    - name: slack_vmware_warning
+    # slack duty api
+    - name: slack_api_info
       slack_configs:
-        - channel: '#alert-{{"{{ .CommonLabels.tier }}"}}-{{"{{ .CommonLabels.severity }}"}}'
-          api_url: {{ required "slack.vmware_warning_webhook_url undefined" .Values.slack.vmware_warning_webhook_url | quote }}
+        - channel: '#alert-api-{{"{{ .CommonLabels.severity }}"}}'
```
auhlig: Please be consistent. Either hardcoded severity or dynamic.

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -446,10 +555,10 @@ data:
               text: {{"'{{template \"slack.sapcc.silence1Month.actionText\" . }}'"}}
               value: {{"'{{template \"slack.sapcc.silence1Month.actionValue\" . }}'"}}
 
-    - name: slack_vmware_critical
+    - name: slack_api_critical
       slack_configs:
-        - channel: '#alert-{{"{{ .CommonLabels.tier }}"}}-{{"{{ .CommonLabels.severity }}"}}'
-          api_url: {{ required "slack.vmware_critical_webhook_url undefined" .Values.slack.vmware_critical_webhook_url | quote }}
+        - channel: '#alert-api-{{"{{ .CommonLabels.severity }}"}}'
```
auhlig: same as above

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -446,10 +555,10 @@ data:
...
+          api_url: {{ default "MISSING" .Values.slack.api_critical_webhook_url | quote }}
```
auhlig: Please use `{{ required "slack.api_critical_webhook_url undefined" .Values.slack.api_critical_webhook_url | quote }}`

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -415,10 +509,25 @@ data:
...
+          pretext: {{"'{{template \"slack.sapcc.pretext\" . }}'"}}
+          icon_emoji: {{"'{{template \"slack.sapcc.iconemoji\" . }}'"}}
+          callback_id: "alertmanager"
+          color: {{`'{{template "slack.sapcc.color" . }}'`}}
+          send_resolved: true
+          
+    - name: slack_api_warning
+      slack_configs:
+        - channel: '#alert-api-{{"{{ .CommonLabels.severity }}"}}'
+          api_url: {{ default "MISSING" .Values.slack.api_warning_webhook_url | quote }}
```
auhlig: Please use `{{ required "slack.slack.api_warning_webhook_url undefined" .Values.slack.api_warning_webhook_url | quote }}`

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -651,6 +786,25 @@ data:
             Playbook: {{"'{{template \"pagerduty.sapcc.playbook\" . }}'"}}
             firing: {{"'{{ template \"pagerduty.sapcc.firing\" . }}'"}}
 
+    - name: pagerduty_alertchain_test
+      pagerduty_configs:
+        - service_key: {{ default "MISSING" .Values.pagerduty.alerttest.service_key | quote }}
```
auhlig: Please use `{{ required "pagerduty.alerttest.service_key undefined" .Values.pagerduty.alerttest.service_key | quote }}`

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -128,72 +150,94 @@ data:
...
-      - receiver: slack_by_tier_and_severity
+      # ---- special notification channels ----
+      # owner: Thomas Grainchen
+      - receiver: slack_nannies
         continue: true
         match_re:
-          tier: os|kks|metal|net
-          severity: info|warning
-          region: admin|qa-de-1|ap-ae-1|ap-au-1|ap-cn-1|ap-jp-1|ap-jp-2|ap-sa-1|eu-de-1|eu-de-2|eu-nl-1|eu-ru-1|la-br-1|na-ca-1|na-us-1|na-us-3
+          context: nanny
```
auhlig: The `context` label is used to inhibit alerts. I'm not really sure if it's wise to use it for routing as well.

geisslet:
Thx for reviewing 👍  .. reply with `>>`

> The number of failed notifications is high, but doesn't grow anymore. I guess the reason for failed messages to slack was when the slack notification template was faulty and when I switched to the stargate (during alertmanager update/channel rename). Didn't observe any other reason in the logs.
>> IC, thx .. but still it's a good thing to haven't non existing channels addressed 

> Note: That you might have to regenerate the stargate incoming webhooks after renaming the channels. Not sure though.
>> I just rename non-stargate-interactive channels. So hope there is no impact on stargate - but thx for the hint

global/prometheus-alertmanager/templates/config.yaml
geisslet: but here was a benefit of doing it dynamic, bcz. the severity is limited by the route definition and is more or less constant. The more dynamic tier/service thing was to agil for my taste and created messages to non existing channels

global/prometheus-alertmanager/templates/config.yaml
geisslet: same as above :)

global/prometheus-alertmanager/templates/config.yaml
geisslet: very good point .. I adopt for all

global/prometheus-alertmanager/templates/config.yaml
geisslet: so what shall we sue the dev-null for then?

global/prometheus-alertmanager/templates/config.yaml
geisslet: me neither .. that's also why I thought it could be use-full for dev purposes to route to an own channel and content is owner driven .. Martin V. & I want do review the label soon .. ask for your input/experience/opinion (undoubtedly sure :))

global/prometheus-alertmanager/templates/config.yaml
auhlig: Receivers are processed in the order they are configured. The dev-null receiver is meant to catch everything else -  meaning at least `region: staging`
There are 2 options:
(1) Move dev-null receiver to the top and use it to filter alerts we don't want to see anywhere else, like `region: staging, cluster_type: ...`. 
(2) Leave dev-null receiver at bottom but without any `match_re` filter. Must ensure alerts that are routed before to their final receiver are not getting here `continue: false`. Bonus: This would enable us to see misconfigured alerts (missing labels) in the dev-null channel.

global/prometheus-alertmanager/templates/config.yaml
auhlig: I thinks it's confusing to hardcode the name of the receiver to `slack_api_info` but have it dynamic in the actual configuration`channel: '#alert-api-{{"{{ .CommonLabels.severity }}"}}'`. 
So, given the name, this receiver is meant to only handle api info alerts, no?

global/prometheus-alertmanager/templates/config.yaml
geisslet: option (2) is the better one in my eyes .. going for that - so dev-null has a defined purpose, too //👍

global/prometheus-alertmanager/templates/config.yaml
geisslet: i mixed it up - you're completely right //sorry

auhlig:
Can you resolve the conflicts - rebase against master @geisslet?

geisslet:
yeah, i will do before lunch //thx!

auhlig:
Where do all these unrelated commits and additions come from? Can you clean that up please (did you rebase already?). Hard to review with ~3300 lines of unrelated changes

geisslet:
@auhlig  .. good question (don't trust the git tool I'm using - did a rebase on origin/master that's it) .. I'm close to do a new branch and and copy the last version into it

geisslet:
close that branch .. rebase was done on old maser //grrrr // sorry .. there will be an alertmanager-route-clean-up-2

PR #635, opened 2019-03-07
Add additional exporters outside of kube-monitoring as it is currently refactored

SchwarzM:
I did not know the correct location please advise if this looks good to you and we reference these with requirements in the future

### DIFF
--- a/system/exporter/nfs-exporter/templates/deployment.yaml
+++ b/system/exporter/nfs-exporter/templates/deployment.yaml
@@ -1,4 +1,3 @@
-{{ if .Values.enabled }}
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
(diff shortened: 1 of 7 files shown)


auhlig:
Difficult question. Having them under `system/exporters` makes sense to me.
However, [infra-exporters](https://github.com/sapcc/helm-charts/tree/master/openstack/infra-exporters) might be the indented place for them. @BugRoger?

system/exporter/nfs-exporter/templates/deployment.yaml
```diff
@@ -0,0 +1,57 @@
+{{ if .Values.enabled }}
```
auhlig: `enabled` not mentioned in .Values

SchwarzM: removed the leftover flag it should be enabled / disabled via requirements

SchwarzM:
hmm nas nothing to do with openstack .... X_x

PR #651, opened 2019-03-12
Adds OpenstackSeed for DISCO (Designate IngresS Cname Operator) user

auhlig:
The Disco operator is currently deployed in the scaleout clusters and requires an OpenStack user that is able to manage Designate records. This user shall be created via an OpenstackSeed.
However, AFAIK there's no openstack-seeder there to do the magic.

So there's a more general question for which I see 2 options:
(1) Deploy the Openstack seeder to the scaleout clusters.
(2) Openstack seeder just in controlplane. Means: One has to split charts. Seeds only in baremetal. Rest of the deployment in scaleout.

What do you think @ruvr, @BugRoger, @databus23?

### DIFF
--- a/system/kube-system/charts/disco/templates/clusterrolebinding.yaml
+++ b/system/kube-system/charts/disco/templates/clusterrolebinding.yaml
@@ -11,6 +11,6 @@ roleRef:
   name: {{ template "name" . }}
 subjects:
   - kind: ServiceAccount
-    name: {{ .Values.rbac.serviceAccountName }}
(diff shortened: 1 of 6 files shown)


auhlig:
419326b will seed the disco user. not sure about role assignments though. does this look okay to you @ruvr?

ruvr:
We'd need a corresponding set of region/values/domain-seeds.yaml secrets repo updates with the actual password before this can be merged.

auhlig:
Decided to deploy the disco operator in the bm controlplane as well. LGTY @ruvr?

PR #652, opened 2019-03-13
[Barbican] Adding api resource values and adding required to db yaml …

rajivmucheli:
Hi @Carthaca 

Please validate if the api resource values are correct. os_password values are set in https://github.wdf.sap.corp/cc/secrets/pull/705.

Regards,
Rajiv

### DIFF
--- a/openstack/barbican/values.yaml
+++ b/openstack/barbican/values.yaml
@@ -41,7 +41,7 @@ pg_cluster:
     backup_schedule: "* * * * *"
 
 postgresql:
-  imageTag: '9.4.14'
+  imageTag: 9.6
   postgresDatabase: barbican
 #  postgresPassword: null
   dbInit: barbican-db-init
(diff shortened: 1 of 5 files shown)


openstack/barbican/values.yaml
```diff
@@ -15,6 +15,13 @@ global:
 pod:
   replicas:
     api: 2
+  resources:
+    requests:
+      memory: 24Mi
+      cpu: 10m
+    limits:
```
Carthaca: please don't specify limits, for the barbican service qos-class burstable is enough (not guaranteed)

openstack/barbican/values.yaml
```diff
@@ -41,7 +41,7 @@ pg_cluster:
     backup_schedule: "* * * * *"
 
 postgresql:
-  imageTag: '9.4.14'
+  imageTag: 9.6
```
Carthaca: Does that work, did you test the postgres upgrade?

rajivmucheli:
Staging is running on postgres version 9.6 since few weeks without hiccups. Stefan also suggested it works fine.

PR #670, opened 2019-03-18
[content-repo] add support for architectures if job.from.type is defined

talal:


talal:
This can be merged already.

FYI @majewsky

PR #671, opened 2019-03-18
Add new concourse chart

jknipper:
This PR adds the new concourse umbrella chart.


global/concourse/Chart.yaml
```diff
@@ -1,16 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for Kubernetes
```
databus23: Add a proper description here

global/concourse/README.md
```diff
@@ -1,246 +0,0 @@
-# Concourse Helm Chart
```
databus23: It would be nice to have a basic README explaining what this chart is how we did it.

databus23: Eat least we should briefly explain the following things:
* Daemonset per team with
* Logging reverse proxy

global/concourse/templates/daemonset.yaml
```diff
@@ -0,0 +1,139 @@
...
+          value: {{ $.Values.concourse.concourse.worker.workDir | quote }}
+        - name: CONCOURSE_BIND_IP
+          value: "127.0.0.1"
+        - name: CONCOURSE_BIND_PORT
+          value: "7777"
+        - name: CONCOURSE_LOG_LEVEL
+          value: "info"
+        - name: CONCOURSE_TSA_HOST
+          value: "concourse-web:2222"
+        - name: CONCOURSE_TSA_PUBLIC_KEY
```
databus23: Maybe we should have seperate keys per team?

global/concourse/templates/daemonset.yaml
```diff
@@ -0,0 +1,139 @@
...
+          value: "{{ $.Values.concourse.worker.keySecretsPath }}/host_key.pub"
+        - name: CONCOURSE_TSA_WORKER_PRIVATE_KEY
+          value: "{{ $.Values.concourse.worker.keySecretsPath }}/worker_key"
+        - name: CONCOURSE_GARDEN_BIND_PORT
+          value: "7777"
+        - name: CONCOURSE_BAGGAGECLAIM_DRIVER
+          value: "naive"
```
databus23: This is pretty bad performance wise, we need to use the standard btrfs one. I guess that's a little tricky with the Kubernetes volumes but this won't scale out all. Ideally the volume should be formatted with btrfs.

global/concourse/templates/daemonset.yaml
```diff
@@ -0,0 +1,139 @@
...
+            port: 8888
+        lifecycle:
+          preStop:
+            exec:
+              command:
+              - "kill"
+              - "-s"
+              - "SIGUSR2"
+              - "1"
+        env:
```
databus23: We should add the following setting, ideally configureable by value:
```
      CONCOURSE_AUTH_DURATION:                 72h
      CONCOURSE_DEFAULT_BUILD_LOGS_TO_RETAIN:  50
      CONCOURSE_MAX_BUILD_LOGS_TO_RETAIN:      100
```

jknipper: This is set via values in official chart web daemonset.

global/concourse/templates/daemonset.yaml
jknipper: We now have ssh keys per team for the workers.

jknipper:
@databus23 Please have a look. We now have:
* volume provisioning with btrfs on worker spin-up
* ssh keys per team/worker

PR #677, opened 2019-03-20
[nova][cinder] Resource requests

joker-official:


joker-official:
Rebased and merge on the commandline. Thank you.

PR #688, opened 2019-03-22
Prometheus server base chart (via operator)

auhlig:


auhlig:
TODO:
[] RBAC 
[] Example with k8s prom
[] Readme

auhlig:
Example with k8s collector: https://github.com/sapcc/helm-charts/tree/prometheus-collector-operated/system/kube-monitoring/charts/prometheus-collector-operated

PR #691, opened 2019-03-27
P8s monitor cpulimits

jobrs:
Federate completely-fair-scheduler metrics from prometheus-collector to frontend, so that you can visualize CPU utilization and limits in Grafana. This helps fine-tuning those limits and may later be used  to discover starvation caused by outdated limits.

### DIFF
--- a/system/kube-monitoring/charts/prometheus-collector/aggregation.rules
+++ b/system/kube-monitoring/charts/prometheus-collector/aggregation.rules
@@ -1,6 +1,7 @@
 groups:
 - name: cpu
   rules:
+  # TODO: remove with K8S 1.10+ / cAdvisor 0.29.0 (https://github.com/google/cadvisor/blob/master/CHANGELOG.md#0290-2018-02-20)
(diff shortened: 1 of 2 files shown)


system/kube-monitoring/charts/prometheus-collector/aggregation.rules
```diff
@@ -1,6 +1,7 @@
 groups:
 - name: cpu
   rules:
+  # TODO: remove with K8S 1.10+ / cAdvisor 0.29.0 (https://github.com/google/cadvisor/blob/master/CHANGELOG.md#0290-2018-02-20)
```
auhlig: Note that kube-monitoring is also deployed to the scalout clusters running on k8s 1.10.
Shouldn't do any harm, right. Just no more per-cpu metrics in cAdvisor 0.29.0+

jobrs: sure, this is just a reminder to clean up once everything is on 1.10 or later

PR #692, opened 2019-03-27
Use Postges with Flock Protection

BugRoger:
Liberally adapted from Keystone, where this was fixed here:
https://github.com/sapcc/helm-charts/commit/b7b4181444929bd82c4a4de49f0a1d5bf9a2c2ba

### DIFF
--- a/openstack/neutron/values.yaml
+++ b/openstack/neutron/values.yaml
@@ -275,6 +275,9 @@ pgmetrics:
           description: "Seconds since the last heartbeat"
 
 postgresql:
+  custom_repository: true
+  imageTag: "9.4.14"
+  
   postgresDatabase: neutron
   # postgresPassword:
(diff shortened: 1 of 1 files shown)


openstack/neutron/values.yaml
```diff
@@ -275,6 +275,9 @@ pgmetrics:
           description: "Seconds since the last heartbeat"
 
 postgresql:
+  custom_repository: true
+  imageTag: "9.4.14"
```
databus23:
```suggestion
  imageTag: "9.4.21"
```

databus23:
LGTM, this will break back the flock protection. But we still have some foo in out internal chart. The latest signal changes (1 year) from @fwiesel don't make any sense to me. Figuring this out shouldn't stop us from merging this

PR #702, opened 2019-04-01
[kube-monitoring][metal-ironic-alerts] new info alerts

BerndKue:
We need two info alerts for different maintenance status of the nodes.


BerndKue:
The file does not exist anymore. Not sure what I had in mind back there to split the info alerts. Must have been an issue defining the right query to catch them all. But it seems to be obsolete by now.
Closing the PR.

PR #706, opened 2019-04-04
add resource utilization and saturation metrics

jobrs:
... to prometheus collector aggregation rules, so that Grafana dashbaords are faster and permit proactive alerting

* utilization: actual usage compared to requested/scheduled usage
* saturation: exhaustion of capacity, with 100% meaning cease of operations


system/kube-monitoring/charts/prometheus-collector/aggregation.rules
```diff
@@ -38,3 +38,14 @@ groups:
...
     expr: (100 - 100 * elasticsearch_filesystem_data_free_bytes / elasticsearch_filesystem_data_size_bytes)
+
+- name: kubernetes_resources
+  rules:
+  - record: aggregated:container_memory_saturation_ratio
+    expr: sum(container_memory_working_set_bytes) by (namespace, pod_name, container_name) / sum(label_join(label_join(kube_pod_container_resource_limits_memory_bytes, "container_name", "", "container"), "pod_name", "", "pod")) by (namespace, pod_name, container_name)
+  - record: aggregated:container_memory_utilization_ratio
+    expr: sum(container_memory_working_set_bytes) by (namespace, pod_name, container_name) / sum(label_join(label_join(kube_pod_container_resource_requests_memory_bytes, "container_name", "", "container"), "pod_name", "", "pod")) by (namespace, pod_name, container_name)
+  - record: aggregated:container_cpu_saturation_ratio
+    expr: sum(irate(container_cpu_cfs_throttled_seconds_total[5m])) by (namespace, pod_name, container_name) / (sum(irate(aggregated:container_cpu_usage_seconds_total[5m])) by (namespace, pod_name, container_name) + sum(irate(container_cpu_cfs_throttled_seconds_total[5m])) by (namespace, pod_name, container_name))
```
BugRoger: > When a container exceeds its CPU limits, the Linux runtime will “throttle” the container and record the amount of time it was throttled in the series `container_cpu_cfs_throttled_seconds_total`

Sounds like this will only be >0 when the container is actually throttled. Is this intended?

jobrs: yes, so this value will be 0% unless throttling takes place.

One could better call it _over_-saturation

jobrs:
thanks for your reviews 

successful deployment on staging, no noticable impact on p8s load, grafana dashboard could be retrofitted to use this metrics ...

still, the grafana dashboard is too slow for monsoon3 :-(

SchwarzM:
thanks for trying

PR #707, opened 2019-04-04
Two new roles for security group administration:

frr2018:
securitygroup_viewer -> can show / list security groups and their rules
securitygroup_admin -> can create / update / delete security groups and security group rules

I have modified our initial config.

### DIFF
--- a/openstack/neutron/templates/etc/_neutron-policy.json.tpl
+++ b/openstack/neutron/templates/etc/_neutron-policy.json.tpl
@@ -21,6 +21,8 @@
 
     "compute_admin": "role:compute_admin and rule:owner",
     "context_is_compute_admin": "role:cloud_compute_admin or rule:compute_admin",
+    "context_is_securitygroup_admin": "role:securitygroup_admin or rule:context_is_compute_admin or rule:context_is_network_admin",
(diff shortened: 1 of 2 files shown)


openstack/neutron/templates/etc/_neutron-policy.json.tpl
```diff
@@ -21,6 +21,8 @@
 
     "compute_admin": "role:compute_admin and rule:owner",
     "context_is_compute_admin": "role:cloud_compute_admin or rule:compute_admin",
+    "context_is_securitygroup_admin": "role:securitygroup_admin or rule:context_is_compute_admin or rule:context_is_network_admin",
```
Carthaca: I believe we need `(role:securitygroup_admin and rule:owner) or ...`

openstack/neutron/templates/etc/_neutron-policy.json.tpl
```diff
@@ -21,6 +21,8 @@
...
+    "context_is_securitygroup_viewer": "role:securitygroup_viewer or rule:context_is_viewer or rule:context_is_securitygroup_admin or rule:context_is_network_viewer",
```
Carthaca: Same as above.

And why did you add `context_is_network_viewer`?

PR #710, opened 2019-04-08
Change from component to device name && add a Create SPC Ticket Link

geisslet:
Component hast the ‘snmp_exporter’ in its name - devicename is correct.
NetworkAsrNatTableIsNearlyFull got an SPC Ticket Create Link as PoC.


geisslet:
@auhlig pls. comment on naming of the spc link - wasn't sure if it should include that it is a 'create'?

also not sure about the link in the template, especially the `?`mark.

thx!

PR #737, opened 2019-04-30
Bump concourse to version 5.1.0

jknipper:
This PR updates our chart to use concourse 5.1.0. This is mainly some renaming and disabling worker stateful set in original chart.


global/concourse/templates/worker-policy.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: policy/v1beta1
```
databus23: Why do we need to add this? I don't see it being used.
Same for the worker-role and service account

jknipper: I removed the pod policy, serviceaccount/roles might be used in future.

PR #738, opened 2019-04-30
[alertmanager] switches to global wham url

stefanhipfel:
- activates canary regions for alert bm error handling

### DIFF
--- a/global/prometheus-alertmanager/templates/config.yaml
+++ b/global/prometheus-alertmanager/templates/config.yaml
@@ -156,14 +156,14 @@ data:
           region: ap-ae-1|ap-au-1|ap-cn-1|ap-jp-1|ap-jp-2|ap-sa-1|eu-de-1|eu-de-2|eu-nl-1|eu-ru-1|la-br-1|na-ca-1|na-us-1|na-us-3
 
       - receiver: slack_metal_critical
-        continue: false
+        continue: true
         match_re:
           tier: metal
           severity: critical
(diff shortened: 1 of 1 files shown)


auhlig:
`continue: false` := final receiver if regex' match.
Critical and warning alerts will never reach wham, since they are caught here: https://github.com/sapcc/helm-charts/blob/d50f48bcf60db46c79e5cb0c791c2f4603f0532f/global/prometheus-alertmanager/templates/config.yaml#L158-L170.

stefanhipfel:
So slack_metal_critical and slack_metal_warning should enable continue. 
Do we allow to set it to true? :)

global/prometheus-alertmanager/templates/config.yaml
```diff
@@ -251,7 +251,7 @@ data:
 
       # ---- webhook handler: metal ----
       - receiver: wham_metal
-        continue: false
+        continue: true
```
auhlig: being the last receiver for metal alerts this is `continue: false`

stefanhipfel: ah ok, should it be last one? (or is that mandatory to have a last handler?)...i saw my handler more like a "middleware"..

auhlig: in this case wham is the last receiver. the [dev-null receiver](https://github.com/sapcc/helm-charts/blob/776e03ead9d67d96a6ca5b957cb4d1f3df3e5dbb/global/prometheus-alertmanager/templates/config.yaml#L260-L262) is meant to catch alerts that aren't labeled properly. for this work properly labeled alerts need to have a final receiver before

stefanhipfel: ok got it!

PR #739, opened 2019-04-30
independent limes alerts

auhlig:
Our new monitoring architecture allows independent deployment of Prometheus alerts and aggregation rules rather than having them in a crowded central pipeline.
This PR uses a `PrometheusRule` to deploy limes specific alerts and aggregations.
For now it will deploy the alerts, rules in addition to the existing ones (but without double alerting). 

Will contact you on thursday @majewsky

### DIFF
--- a/openstack/limes/templates/api-service.yaml
+++ b/openstack/limes/templates/api-service.yaml
@@ -10,6 +10,8 @@ metadata:
     # uncomment to have Maia scape as well: maia.io/scrape: "true"
     prometheus.io/scrape: "true"
     prometheus.io/port: "80"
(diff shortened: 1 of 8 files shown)


openstack/limes/templates/api-service.yaml
```diff
@@ -10,6 +10,8 @@ metadata:
     # uncomment to have Maia scape as well: maia.io/scrape: "true"
     prometheus.io/scrape: "true"
     prometheus.io/port: "80"
+    # The `openstack` Prometheus shall scrape metrics of OpenStack services.
+    prometheus.io/targets: "openstack"
```
majewsky: Please use `.Values.alerts.prometheus` here.

PR #753, opened 2019-05-08
[Glance-Alerts] Moving alerts to prom-os

rajivmucheli:
Hi @auhlig,

Please validate this PR. Would i need to add Blackbox alerts to the aggregation rules?

Regards,
Rajiv


auhlig:
Blackbox alerts and rules should also be converted but deployed with the blackbox chart.

PR #759, opened 2019-05-10
[global][wham] fixes dns zone for disco change

stefanhipfel:


global/wham/templates/ingress.yaml
```diff
@@ -13,6 +13,7 @@ metadata:
     annotations:
         vice-president: "true"
         disco: "true"
+        disco/zone-name: "ingress.{{ .Values.global.region }}.cloud.sap"
```
auhlig: ah it's actually `disco/record: ...`

PR #763, opened 2019-05-13
Restart registry cache on cert renewal

jknipper:
This PR adds a liveness probe to registry cache which triggers a restart if the certificate is updated.

### DIFF
--- a/system/registry-cache/templates/deployment.yaml
+++ b/system/registry-cache/templates/deployment.yaml
@@ -19,14 +19,15 @@ spec:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
-{{- if .Values.remoteIP }}
         command:
           - sh
           - -exc
           - |
+{{- if .Values.remoteIP }}
             echo -e "\n{{ .Values.remoteIP }} {{.Values.remoteURL | trimPrefix "https://" | trimPrefix "http://"}}" >> /etc/hosts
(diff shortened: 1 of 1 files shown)


system/registry-cache/templates/deployment.yaml
```diff
@@ -38,6 +39,14 @@ spec:
...
         ports:
         - containerPort: {{ .Values.service.internalPort }}
+        livenessProbe:
+          exec:
+            command:
+              - sh
+              - -c
+              - md5sum /conf/tls.crt | cmp /tmp/tls.crt.md5
+          initialDelaySeconds: 60
+          periodSeconds: 15
```
databus23: I think this can be relaxed to at least once per minute.

system/registry-cache/templates/deployment.yaml
```diff
@@ -38,6 +39,14 @@ spec:
 {{- end }}
         ports:
         - containerPort: {{ .Values.service.internalPort }}
+        livenessProbe:
+          exec:
+            command:
+              - sh
+              - -c
+              - md5sum /conf/tls.crt | cmp /tmp/tls.crt.md5
```
databus23: have you tested that this works? are the `cmp` and `md5sum` binary available in the docker image?

jknipper: They are both part of busybox, i tested this in qa.

jknipper:
@databus23 I removed `md5sum` and just use copy.

system/registry-cache/templates/deployment.yaml
jknipper: I changed this.

PR #764, opened 2019-05-14
Restart arc on cert renewal

jknipper:
This PR adds a liveness probe in order to restart arc when the certificate gets renewed.


openstack/arc/templates/api-deployment.yaml
```diff
@@ -42,11 +50,21 @@ spec:
             initialDelaySeconds: 5
             timeoutSeconds: 1
           livenessProbe:
+{{- if .Values.api.tls.enabled }}
+            exec:
+              command:
+                - sh
+                - -c
+                - cmp /tls/tls.crt /tmp/tls.crt.livenesspobe
```
databus23:
```suggestion
                - cmp --silent /tls/tls.crt /tmp/tls.crt.livenesspobe
```

This will make cmp exit on the first difference and not output anything

jknipper: Hmm, in busybox version only `cmp -s` is available.

openstack/arc/templates/api-deployment.yaml
```diff
@@ -42,11 +50,21 @@ spec:
             initialDelaySeconds: 5
             timeoutSeconds: 1
           livenessProbe:
+{{- if .Values.api.tls.enabled }}
+            exec:
+              command:
+                - sh
```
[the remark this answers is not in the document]
jknipper: More like this?

PR #772, opened 2019-05-15
[Manila-Alerts] Moving alerts to Prom-OS tier

rajivmucheli:
Hi @chuan137 

Regards,
Rajiv


chuan137:
I think the aggregation rules are not necessary

chuan137:
done

PR #773, opened 2019-05-15
[Cinder-Alerts] Moving alerts to Prom-OS tier

rajivmucheli:
Hi @joker-official 

Regards,
Rajiv


joker-official:
Where do the .rules come from? I've found the source for .alerts. Why did you remove the .rules here but not for nova?

rajivmucheli:
Nova had a custom rule declared here : https://github.com/sapcc/helm-charts/blob/master/system/kube-monitoring/charts/prometheus-frontend/aggregation.rules#L12-L15

joker-official:
Thanks for clarifying. Merged.

PR #774, opened 2019-05-15
[Nova-Alerts] Moving alerts to Prom-OS tier

rajivmucheli:
Hi @joker-official 

Regards,
Rajiv


auhlig:
Merging after brief discussion with @joker-official

PR #788, opened 2019-05-16
Memcache ready for new Prometheus setup

auhlig:


common/memcached/templates/alerts/_memcached.alerts.tpl
```diff
@@ -0,0 +1,14 @@
+groups:
+- name: memcached.alerts
+  rules:
+  - alert: {{ include "alerts.service" . | title }}MemcachedManyConnectionsThrottled
+    expr: (rate(memcached_connections_yielded_total{app="barbican-memcached"}[5m]) * 60) > 5
```
reimannf: guess that should be generalised

auhlig: Indeed. Thanks for spotting this.

common/memcached/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: memcached
-version: 0.0.1
+version: 1.0.0
```
reimannf: you really want to bump the version... you know everyone needs to update the requirements.yaml

auhlig: reverted. bad habits .. ;)

PR #795, opened 2019-05-20
[Barbican] Removing _member_ role

rajivmucheli:
Hi @frr2018,

member role allows the user to view and download the secrets\payloads, hence removing this role. It worked fine in QA, with appl user. Anton confirmed it works well the way it should.

Regards,
Rajiv


rajivmucheli:
Hi Frank,

Sure, Blog #27 had been created on 27th May. i finally got the go ahead from customers now.

Regards,
Rajiv

PR #797, opened 2019-05-22
[hermes] prom alerts

notque:


openstack/hermes/templates/api-service.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
     system: openstack
     component: hermes
   annotations:
+    prometheus.io/targets: "openstack"
```
auhlig: use `{{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus }}` here as well

openstack/hermes/templates/logstash-service.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     system: openstack
     service: audit
   annotations:
+    prometheus.io/targets: "openstack"
```
auhlig: use `{{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus }}` here as well

PR #804, opened 2019-06-11
Add etcd backup alert for admin cluster

jknipper:
This PR adds an alert if any etcd backup is failing in admin cluster.


jknipper:
//cc @databus23

PR #808, opened 2019-06-12
Limes: make the pod OOM alert less trigger-happy

majewsky:
The old alert just checked if `predict_linear(memory usage, 8h) > 100%`. This often fires directly after deployments when pods' memory usage quickly rises during their startup, but then flattens out.

To avoid these misfirings, the alert now has a second condition, triggering only for current `memory usage > 70%`. Because the alert expresssion got quite unwieldy, I refactored part of it out into an aggregation rule.

@auhlig I'd appreciate a second pair of eyes checking the alert expressions.

One thing that I'm not sure about is if we want to move the aggregation rule to a more generic place, and have it apply to all containers. These pod-OOM alerts are just copy-pasted all over the place anyway, so it may be helpful to have this aggregated metric available for everyone.

### DIFF
--- a/openstack/limes/aggregations/consolidation.rules
+++ b/openstack/limes/aggregations/consolidation.rules
@@ -2,8 +2,7 @@ groups:
 - name: limes
   rules:
     - record: global:limes_consolidated_cluster_capacity
-      expr: max(label_join(limes_cluster_capacity, "full_resource", "/", "service", "resource"))
-          by (full_resource,os_cluster)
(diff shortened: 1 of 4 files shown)


majewsky:
Bump. Anyone want to have a look?

auhlig:
Sorry for missing this one.
`container_memory_usage_bytes` includes the cache. better: `container_memory_working_set_bytes`.
See the query [on this page](https://blog.freshtracks.io/a-deep-dive-into-kubernetes-metrics-part-3-container-resource-metrics-361c5ee46e66?gi=f8ec43262eb7) (Search for `Memory Utilization, Saturation and Errors`).

PR #820, opened 2019-06-24
Changed behaviour for securitygroup_admin/viewer 

frr2018:
To be able to view network objects. Request by LOB which requested the securitygroup_admin/viewer.


openstack/neutron/templates/etc/_neutron-policy.json.tpl
```diff
@@ -7,13 +7,13 @@
     "admin": "role:network_admin and rule:owner",
     "context_is_network_admin": "rule:context_is_admin or rule:admin",
     "context_is_editor": "rule:context_is_network_admin or rule:member",
-    "context_is_viewer":  "rule:context_is_editor or rule:viewer",
+    "context_is_viewer":  "rule:context_is_editor or rule:viewer or rule:context_is_securitygroup_admin or rule:context_is_securitygroup_viewer",
```
joker-official: Doesn't that create a cyclic graph between `context_is_viewer` and `context_is_securitygroup_viewer`? Does that work?

openstack/neutron/templates/etc/_neutron-policy.json.tpl
```diff
@@ -7,13 +7,13 @@
...
     "network_view_all": "role:network_viewer or role:member or role:network_admin or rule:context_is_admin",
 
     "network_owner": "tenant_id:%(network:tenant_id)s",
     "network_owner_or_owner": "rule:network_owner or rule:owner",
     "network_member": "role:member and rule:network_owner_or_owner",
-    "network_viewer": "role:network_viewer and rule:network_owner_or_owner",
+    "network_viewer": "(role:network_viewer and rule:network_owner_or_owner) or rule:context_is_securitygroup_viewer or rule:context_is_securitygroup_admin",
```
joker-official: `or rule:context_is_securitygroup_admin` is not necessary because `rule:context_is_securitygroup_viewer` contains that. Same above.

frr2018: Is admin is removed.

openstack/neutron/templates/etc/_neutron-policy.json.tpl
frr2018: This should work. I think the "or" condition in context_is_securitygroup_viewer prevent issues here. I will try it out in qa and if it doesnt work i will roll it back.

frr2018:
I exactly had the discussion with the customer that then it is not clearly visible that the securitygroup_viewer/admin has also rights to the network objects and compute objects. But this customer doesnt want to use the generic roles and is also requesting an special compute admin role.

joker-official:
But it's not only about the customer, it's also about keeping stuff maintainable from our side. Policy files are hard to read as it is already -.-

frr2018:
You are right. I am fighting now with the customer that they have to request network_viewer and computer_viewer.

PR #824, opened 2019-06-25
Changed the behaviour of

frr2018:
securitygroup_admin/viewer to give them the permission
to view nova related objects and to give the securitygroup_admin
the permission to attach securitygroups to compute instances.


frr2018:
I had a discussion with Johannes and will change the direction. I will submit a new pullrequest.

PR #835, opened 2019-07-03
Add alert for admin cluster velero full backup

jknipper:
This PR adds an alert for failing full backups in admin cluster. //cc @databus23

### DIFF
--- a/system/kube-monitoring/charts/prometheus-frontend/admin-velero-backup.alerts
+++ b/system/kube-monitoring/charts/prometheus-frontend/admin-velero-backup.alerts
@@ -1,0 +1,15 @@
+groups:
+- name: velero-backup.alerts
+  rules:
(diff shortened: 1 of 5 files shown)


auhlig:
Alert LGTM. 

Since the [prometheus-frontend alert swamp](https://github.com/sapcc/helm-charts/tree/master/system/kube-monitoring/charts/prometheus-frontend) is going to be retired soon the alerts should also be deployed with the application providing the metric used for the alerts - which seems to be [system/velero](https://github.com/sapcc/helm-charts/tree/master/system/velero).
Can you also do something like [this](https://github.com/sapcc/helm-charts/pull/836/files) for velero?

jknipper:
@auhlig I added alert CRD to the velero chart. Please check if i got it right.

auhlig:
Perfect @jknipper 🥇

PR #843, opened 2019-07-03
[Concourse] Move Prometheus alerts to CRD

auhlig:
Move Concourse' [existing alerts](https://github.com/sapcc/helm-charts/blob/master/system/kube-monitoring/charts/prometheus-frontend/concourse.alerts) to the PrometheusRule custom resource. `prometheus.io/targets` annotation already present on service.
LGTY @jknipper?


auhlig:
This also showed the helm chart tests work :D

jknipper:
@auhlig Yes, very nice!

PR #850, opened 2019-07-10
[swift] use map instead of array for multiple clusters

talal:


openstack/swift/values.yaml
```diff
@@ -79,7 +79,7 @@ alerts:
 # If the swift cluster is shared across multiple openstack clusters, one can
 # start multiple proxy deployments connecting to the different keystone backends
 clusters:
-  - name: DEFINED_IN_REGION_CHART
+  DEFINED_IN_REGION_CHART: # cluster name
```
majewsky: This needs to be commented out, otherwise it attempts to deploy `swift-proxy-DEFINED_IN_REGION_CHART` etc.

With the array, it was okay to have the template in there because the array was replaced by the values.yaml. Maps OTOH are merged with the values.yaml, so this key stays in there.

PR #884, opened 2019-07-31
[Seeds] Adding Neo domain

rajivmucheli:
Hi @ruvr 

Please validate this PR.

Regards,
Rajiv


reimannf:
If this PR is meant as all in one: you need as well:
- https://github.com/sapcc/helm-charts/blob/master/openstack/swift/templates/support_seed.yaml#L1
- https://github.com/sapcc/helm-charts/blob/master/openstack/limes/templates/support_seed.yaml#L1

PR #891, opened 2019-08-06
[nova] Conditionally spawn api-metadata pods

joker-official:
If the metadata APIs are enabled via setting `enabled_apis` including
"metadata", we don't have to spawn separate nova-api-metadata pods. This
is currently true in `eu-de-1` only, where we scaled down the deployment
of nova-api-metadata. To support the `eu-de-1` use-case, we were still
spawning metadata API workers in all other regions' nova-api pods. These
workers idled, but wasted memory.  This commit disables the idling
metadata workers in nova-api pods by default.


joker-official:
This has to be accompanied by a change for `eu-de-1` to enable metadata in the nova-api pod.

PR #894, opened 2019-08-13
add missing region label to all openstack metrics

artherd42:
required because otherwise existing dashboard and alerts will not work properly

### DIFF
--- a/openstack/prometheus-openstack/templates/_openstack-prometheus.yaml.tpl
+++ b/openstack/prometheus-openstack/templates/_openstack-prometheus.yaml.tpl
@@ -32,6 +32,10 @@
     target_label: kubernetes_namespace
   - source_labels: [__meta_kubernetes_pod_name]
     target_label: kubernetes_pod_name
+  # Add region label to all metrics, don't delete this without knowing what you are doing.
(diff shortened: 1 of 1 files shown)


auhlig:
All Prometheis are configured with `external_labels` like this
```
global:
  external_labels:
    region: $region
```

Meaning, the `region` label will be added to each and every metric before it leaves the Prometheus server. See details: https://prometheus.io/docs/prometheus/latest/configuration/configuration

artherd42:
Actually none of the metrics in the prometheus-openstack has the region label, 
for example https://prometheus-openstack.eu-de-1.cloud.sap/graph?g0.range_input=1h&g0.expr=openstack_assignments_total_gauge&g0.tab=1

compared to https://prometheus.eu-de-1.cloud.sap/graph?g0.range_input=1h&g0.expr=openstack_compute_instances_gauge&g0.tab=1

auhlig:
Ah. You want the region label to be visible in Prometheus itself. Then this LGTM.

`external_labels` are added as soon as the metric leaves the Prometheus server. Thus it's not present when querying in the Prometheus UI.

PR #908, opened 2019-08-20
change playbook link to new playbook

rmueller-b1:
There is a new playbook and this alert should include the new link


auhlig:
You also need to increment the version of the ns-exporter Helm chart

PR #911, opened 2019-08-21
refactor chart; add bridge monitor

auhlig:
Adds the `BridgeFilterVLANTagged` status to the node to reflect whether bridged VLAN-tagged ARP/IP traffic was passed to ARPtables/IPtables. /cc @defo89 @SchwarzM


auhlig:
TODO:
- [ ] naming (stick with node condition `BridgeFilterVLANTagged`?)
- [ ] better description of impact
- [ ] playbook

PR #912, opened 2019-08-23
[nova] Automate initial cellv2 setup

joker-official:
This wasn't necessary before, because we just recently changed to cellv2
and had to migrate the old setup. But for new regions and the new labs,
we need this automated.

This also refactors the script a bit to re-use common functionality and
extends the for-loop to also include cell into the `grep`. This also
means, that the cell0 `database_connection` will now be updated if it
changes as well. This was already the case for cell1 and cell2.

Additionally, this commit updates the `cell0_db_path`-helper, but it was
unused before, so nothing needs changing.


notandy:
Hi, I like the rewrite for cell1. But I would like a more generic approach for having n-cells support configurable through secrets since we already use a second cell in eu-de-1, this could be an option for future DCs, e.g. cell per AZ.

joker-official:
That would not only require a rewrite of the script, but also of the helm-charts and secrets (think about all the hard-coded cell2 stuff like deployments, values ...). I think we can do this as a next step, building on the changes done here, and should create a ticket for that task.

notandy:
Sounds reasonable. I am ok with this change.

PR #934, opened 2019-09-10
slack_by_os_service without staging alerts

reimannf:
The `slack_by_os_service` includes with this commit only productive regions and `qa-de-1` to avoid spamming the service specific channels with alerts from `staging`, which is in trouble most of the time.
If one is interested in `staging` alerts, join Slack channels `#alert-dev-[info|warning|critical]`


geisslet:
lgtm

PR #937, opened 2019-09-13
added AWX button to slack template

christopherhans:
Added a line of code to render a button which could directly forwards to a corresponding AWX job


christopherhans:
@auhlig could you please have a look?

PR #955, opened 2019-09-26
Keystone mariadb migration support

ruvr:
Hi,
Here are the changes that will allow us to migrate keystone to MariaDB 10.4 in one run, orchestrated by helm.
There where some issues with the plain psql2mysql from https://github.com/sapcc/psql2mysql, that basically broke the keystone-manage db_sync command.
Initially the exception looked identical to https://bugs.launchpad.net/keystone/+bug/1774688, but it turned out to be caused by pyMSQL and SQLAlchemy version clashes between openstack/requirements (and hence the keystone image) and the psql2mysql requirements.txt version pins (which overwrote the version needed by keystone).
Resetting these versions to the ones used in Openstack rocky cleared the error (https://github.com/ruvr/psql2mysql)
I also suggest to make the MariaDB chart a bit more versatile and configurable, so it can be used for local deployments also (for dev/testing)


ruvr:
I guess we'll need to revisit the charts of the services that already have been migrated and check if anything needs to be added after the MariaDB chart enhancements

PR #959, opened 2019-09-30
add playbook for alerts : certificate expires

DRichardt:


geisslet:
Opsduco still misses the page `docs/devops/alert/certificates.html`, but should fit now.
Please merge after Docu Update.
Thanks for improving.

PR #960, opened 2019-09-30
Limes: add cluster:show_basic rule to policy

talal:


openstack/limes/files/policy.json
```diff
@@ -26,9 +26,10 @@
   "domain:lower":         "rule:domain_editor",
   "domain:discover":      "rule:cluster_editor",
 
-  "cluster:list":  "rule:cluster_viewer",
-  "cluster:show":  "rule:cluster_viewer",
-  "cluster:edit":  "rule:cluster_editor",
+  "cluster:list":       "rule:cluster_viewer",
+  "cluster:show":       "rule:cluster_viewer",
+  "cluster:show_basic": "role:admin or role:member or role:resource_admin or role:resource_viewer",
```
majewsky: For historical reasons, `member` has several synonyms (see line 10) which we should all support.

talal: Updated.

PR #981, opened 2019-10-16
[arista] tcam efp alert

BerndKue:
A playbook has to be added to this document for this alert:
https://operations.global.cloud.sap/docs/devops/alert/network/switch.html#arista

Critical Alert will be fired into alert-api-critical after 30min if sum of EFP slices is above 275%.


BerndKue:
@FabianKroenner @artherd42  Shall we take this productive?
Then someone needs to push the button for the pipeline. I'm not allowed to.

PR #991, opened 2019-10-24
WIP - Glance hermes enablement

notque:


openstack/glance/templates/etc/_glance_audit_map.yaml
```diff
@@ -0,0 +1,18 @@
...
+      import: create/import # action for workflow
+    children:
+      members: # default mapping
+      tags:
+        singleton: true
+      file:
+        singleton: true
+      stage:
+        singleton: true
+  tasks:
```
jobrs: does this need a LF? not sure

notque: Line break added.

PR #992, opened 2019-10-24
[maia] enable app. cred. support in all regions

jobrs:
use latest image with application credentials support by default i.e. in all regions (that do not specify another image version explicitly)


jobrs:
chart tests fail because of recent breaking change of prometheus_server chart. We cannot include such a major change as a sideline

PR #997, opened 2019-10-29
prometheus-openstack: federate elektra open inquires metrics

andypf:


andypf:
@auhlig please take a look

PR #998, opened 2019-10-29
-- add nodeport services and change backup secret

SchwarzM:


databus23:
@SchwarzM Looks good, could you please bump the chart version so that our pipeline will push a new version to chartmusuem?

PR #1001, opened 2019-11-04
[kube-parrot] add alerts

defo89:
`enable` switch will be in the umbrella chart

### DIFF
--- a/system/kube-proxy/templates/alerts.yaml
+++ b/system/kube-proxy/templates/alerts.yaml
@@ -1,7 +1,7 @@
 {{- $values := .Values }}
 {{- if $values.alerts.enabled }}
-{{- if $values.sidecar.parrot.enabled }}
-{{- range $path, $bytes := .Files.Glob "alerts/*.alerts" }}
+{{- if $values.sidecar.parrot }}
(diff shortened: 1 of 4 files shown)


system/kube-proxy/values.yaml
```diff
@@ -35,3 +35,10 @@ parrot:
 
 externalip: {}
 #  ignoreaddress: 
+
+# Deploy kube-proxy Prometheus alerts.
+alerts: {}
+  # enabled: true
```
auhlig: I'd enable that by default.

system/kube-proxy/templates/alerts.yaml
```diff
@@ -0,0 +1,22 @@
+{{- $values := .Values }}
+{{- if $values.alerts.enabled }}
+{{- if $values.sidecar.parrot.enabled }}
```
auhlig: i think it's `$values.sidecars.parrot`

defo89: tnx, you are right - fixed this

auhlig:
🎉

PR #1004, opened 2019-11-07
added support for nodeAffinity, to allow enforcement of co-locating p…

ruvr:
…ods within one availability zone


ruvr:
This is required to get the latency in na-us-* down, since the inter DC links there are way to slow..

PR #1006, opened 2019-11-07
Alert updates for IPMI-Exporter

BerndKue:


stefanhipfel:
lgtm
but we need to discuss how to group by service

BerndKue:
That's why I now added the job label in the queries.
bPOD metrics -> job="ipmi/ironic"
vPOD metrics -> job="vmware-esxi"
CP metrics -> job="cp/netbox"

Did I miss an alert to filter by job?

PR #1012, opened 2019-11-13
[alerting] rm silence buttons; add supernova way

geisslet:
remove the silence buttons in slack; 
add link in slack template for supernova for silencing there; 
add global region in routing for API managed service (thousandeyes http alerts of stargate and supernova need it)


geisslet:
please do not merge - just check

PR #1017, opened 2019-11-18
Adding private 4TB, 3TB and 6Tb bigVM flavor

dhoeller:


joker-official:
We need this rolled out first: https://github.com/sapcc/nova/pull/122

PR #1019, opened 2019-11-19
improve specificity of Kubernikus object-store quota alert

majewsky:
We have reports of false-positive alerts here when projects use quota autoscaling. Quota tracks usage closely in this case, so this alert may fire even though no manual intervention is required.

The fix uses that most projects have quota bursting enabled, so the backend quota (in Swift) will actually be 10% higher than the nominal quota (in Limes). Therefore there will be ample margin between backend quota and actual usage in projects with quota autoscaling.

@databus23 What do you think? Does this fix the problem for you, or do we still need CCM-12045 regardless?

### DIFF
--- a/system/kubernikus-monitoring/alerts/kluster.alerts
+++ b/system/kubernikus-monitoring/alerts/kluster.alerts
@@ -50,7 +50,7 @@ groups:
       summary: "{{ $labels.kubernetes_name }} is unavailable"
 
   - alert: KubernikusKlusterLowOnObjectStoreQuota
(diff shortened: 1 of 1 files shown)


databus23:
So in the specific case the "autoscaled" quota was around 100MB. For this case this would not help much as 10 MiB difference is negligible. In general I'm ok with changing this but I think I need to change the federation for the kubernikus prometheus so that this new metric is available as well first.

majewsky:
I have another trick up my sleeve, but it will require changing existing autoscaling configs in a batch job, which is a bit more involved than this fix here. Regardless, I'll merge this, it should always be a good idea anyway.

PR #1020, opened 2019-11-19
bump ntp-exporter to v1.1.0

majewsky:
Container port changes from 9100 to 9559 because of this.

Since at least d015d5b, we use regular annotation-based service discovery for the ntp-exporter pods (instead of node-based service discovery) and thus don't need to bind to a host port anymore.


majewsky:
Do I need to take care to bump the versions in requirements.yaml somewhere else?

auhlig:
Sure. The ntp-exporter version needs to be bumped in the requirements.yaml and lock files as well as the version of the Chart using it

PR #1022, opened 2019-11-22
Raise garden max containers for concourse workers

jknipper:
This PR changes max. garden containers per worker to 500, default is currently 250. It also changes netmask of used network to 10.254.0.0/16.


jknipper:
@databus23 Maybe we try this manually for one worker pool before rolling it out. What do you think?

PR #1024, opened 2019-11-25
[Keystone][Alert] Fix OpenstackKeystoneDatabaseSize Expr for mariadb …

rajivmucheli:
…database retrivel

Hi @ruvr,

The Expression works fine, but we need to change dashboard (shall i create a tile in database section ?) and playbook links.

Regards,
Rajiv

### DIFF
--- a/openstack/keystone/templates/prometheus-alerts.yaml
+++ b/openstack/keystone/templates/prometheus-alerts.yaml
@@ -247,7 +247,7 @@ spec:
         summary: "Datapath keystone {{`{{ $labels.check }}`}} is flapping"
 
     - alert: OpenstackKeystoneDatabaseSize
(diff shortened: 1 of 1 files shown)


ruvr:
The query does not report correct figures. Looking at the qa database on the filesystem, I collected the following schema sizes: keystone 31M, mysql 11M, performance_schema 4K.
The query reports 10.7, 10.2 and 0.2.
This needs to be fixed, otherwise the alert is useless. 
Please check and make sure that the alert actually alerts in time (storage depletion predicted).

ruvr:
I don't see a need for the figures in the keystone dashboards, just alerting would be ok.
It would be useful in the mySQl dashboard though..

rajivmucheli:
k8s upgrade will activate https://github.com/sapcc/helm-charts/commit/84e0f76734ab7d44b4fea8a421559f7ddccb54fb

rajivmucheli:
https://github.com/sapcc/helm-charts/pull/2415

PR #1030, opened 2019-11-26
Add pid into backdoor socket path

joker-official:
For this to work properly, a patch to oslo.service is necessary. We've
got our own fork of oslo.service that includes that patch already in the
upper-constraints for queens.

---
Services not having the patch will have socket path of `eventlet_backdoor-{pid}.socket` instead of `eventlet_backdoor.socket`.

### DIFF
--- a/openstack/utils/templates/snippets/_debug.tpl
+++ b/openstack/utils/templates/snippets/_debug.tpl
@@ -100,5 +100,5 @@ data:
 
 
 {{ define "utils.snippets.debug.eventlet_backdoor_ini" }}
-backdoor_socket=/var/lib/{{.}}/eventlet_backdoor.socket
+backdoor_socket=/var/lib/{{.}}/eventlet_backdoor-{pid}.socket
 {{ end }}


notandy:
are you sure that every project dependency has been updated using the new path (e.g. also for mitaka, stein, etc?)

joker-official:
I'm sure that I described which version has the upper-constraints changed already in the PR description: queens.
We've got branches for `oslo.service` for `rocky` and `stein`. They are _not_ used in the upper-constraints, yet.

Is the lbaas (which is afaik the only mitaka service left) using oslo.service and has a backdoor enabled?

Additionally, even with this change, the backdoor still gets created (as described) with a different name, if you're running an unpatched `oslo.service`. If your service only starts one process, it doesn't even matter much.

notandy:
Just asking if the curly brakets don't break anything for existing (unpatched) oslo.service, if that's ensured I would say it's good to go.

notandy:
regards lbaas: yes lbaas agent use oslo.service from mitaka and have backdoor socket configured

joker-official:
> Services not having the patch will have socket path of eventlet_backdoor-{pid}.socket instead of eventlet_backdoor.socket.

I checked this by changing cinder-api's `backdoor_socket` to the here-proposed value in `staging`. It still worked, I could connect to it, it was just named differently.

PR #1036, opened 2019-12-03
[prometheus-exporters] added esxi-exporter

christopherhans:
@viennaa

### DIFF
--- a/prometheus-exporters/esxi-exporter/templates/service.yaml
+++ b/prometheus-exporters/esxi-exporter/templates/service.yaml
@@ -9,7 +9,7 @@ metadata:
     prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
 spec:
   selector:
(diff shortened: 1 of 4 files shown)


prometheus-exporters/esxi-exporter/templates/service.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: esxi-exporter
+  namespace: infra-monitoring
+  annotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/port: {{ .Values.listen_port | quote }}
+    prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
+spec:
+  selector:
+    app: {{ include "infraMonitoring.name" . }}
```
viennaa: Looks like this could not be resolved?
https://ci.eu-de-2.cloud.sap/builds/98120

christopherhans: jup. resolved.

PR #1082, opened 2020-01-21
Switch to Swift health exporter

talal:
/cc @reimannf


openstack/swift/alerts/swift-api.alerts
```diff
@@ -32,7 +32,7 @@ groups:
       summary: swift-service-unavailable
 
   - alert: OpenstackSwiftHealthCheck
-    expr: avg(swift_health_statsd_exit_code) BY (region) > 0.2
+    expr: (avg(swift_recon_exit_code) BY (region) OR avg(swift_dispersion_exit_code) BY (region)) > 0.2
```
majewsky: The `or` might mask alarms here. `x or y` returns x if it is not 0. So if `avg(swift_recon_exit_code)` is above 0, but below 0.2, that value will be returned from the `... or ...` expression and the alert will not fire, even if `avg(svg_dispersion_exit_code)` was higher than 0.2.

Please make two separate alerts for swift-recon failing and swift-dispersion-report failing. Also, since we have the actual failed commandline in the labels for those metrics, that would be nice to have in the alert description, e.g. `swift-recon --unmounted has failed` instead of just `Swift health check failures`.

talal:
Done

openstack/swift/templates/swift-health-exporter-service.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - name: api
+      port: {{$cluster.proxy_public_port}}
+      targetPort: 443
+    {{- if and $cluster.proxy_public_http_port $cluster.sans_http }}
+    - name: api-http
+      port: {{$cluster.proxy_public_http_port}}
+      targetPort: 80
+    {{- end}}
+  externalIPs:
+    - {{$cluster.proxy_public_ip}}
```
reimannf: I am quite unsure here, if you expose the swift API the second time. Did you tried to overwrite the prom annotation in the proxy-service? There is an additional scrape annotation on pod level and so you might not need this additional service.

talal: This issue has been resolved over Slack discussion. Tests ongoing in QA.

openstack/swift/etc/statsd-exporter.yaml
```diff
@@ -142,65 +142,3 @@ mappings:
   name: swift_object_replicator
 - match: swift.object-updater.timing
   name: swift_object_updater
-
-  ##############################################################################
-  # extract storage IP from swift-recon metrics (sent by swift-health-statsd)
```
reimannf: That should stay in if there is support for health-statsd and health-export

talal: commit reverted

PR #1091, opened 2020-01-24
removing blank lines

DRichardt:


geisslet:
fail is related to fix in https://github.com/sapcc/helm-charts/pull/1090

PR #1092, opened 2020-01-24
C5255888: route barbican warning to service channel

DRichardt:


auhlig:
The severity of the alert was set to `info`.

PR #1109, opened 2020-02-03
WIP: Glance hermes enablement

notque:
Work in progress, Enable Hermes in Glance. Still going over the details to make sure we're not missing anything.


notque:
Grabbed wrong branch. Need to clean up branches.

PR #1123, opened 2020-02-10
added esxi-exporter to prometheus-infra-collector & federation

christopherhans:
CPO esxi-exporter is deployed with its own pipeline.
The exporter is using atlas and therefore needs to be added to the prometheus config.

### DIFF
--- a/system/infra-monitoring/templates/_prometheus-infra-collector.yaml.tpl
+++ b/system/infra-monitoring/templates/_prometheus-infra-collector.yaml.tpl
@@ -432,3 +432,23 @@
     - target_label: __address__
       replacement: vrops-exporter:9160
 {{- end }}
+
+#exporter is leveraging service discovery but not part of infrastructure monitoring project itself.
+{{- $values := .Values.esxi_exporter -}}
+{{- if $values.enabled }}
+- job_name: 'esxi-config'
+  scrape_interval: {{$values.scrapeInterval}}
+  scrape_timeout: {{$values.scrapeTimeout}}
+  file_sd_configs:
+      - files :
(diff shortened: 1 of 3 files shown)


system/infra-monitoring/templates/_prometheus-infra-collector.yaml.tpl
```diff
@@ -432,3 +432,23 @@
...
+  file_sd_configs:
+      - files :
+        - /etc/prometheus/configmaps/atlas-sd/netbox.json
+  metrics_path: /
+  relabel_configs:
+    - source_labels: [job]
+      regex: vcenter
+      action: keep
+    - source_labels: [server_name]
+      target_label: __param_target
```
viennaa: Only param queried would be `.../?target=vcenter-name`
Is this wanted?

system/infra-monitoring/values.yaml
```diff
@@ -176,4 +176,9 @@ vrops_exporter:
   scrapeTimeout: 295s
 
 px_exporter:
-  enabled: false
\ No newline at end of file
+  enabled: false
+
+esxi_exporter:
+  enabled: false
+  scrapeInterval: 60m
```
viennaa: is a single run taking 60m? If not, what is the expected runtime here?

system/infra-monitoring/templates/_prometheus-infra-collector.yaml.tpl
christopherhans: yes, this is wanted

system/infra-monitoring/values.yaml
christopherhans: Depends on the region, but the information doesn't needs to be updated more frequently.

PR #1163, opened 2020-02-13
[mariadb] refine backup alerts

Carthaca:
Longer range to remove false positives due to backup container restart.
If the backup age has been older than 3 hours for 1 hour, it is 4 hours old now.


Carthaca:
@rajivmucheli @BerndKue fyi

PR #1216, opened 2020-02-19
[mariadb] warning alters when # of running threads is more than 15.

chuan137:
To detect the mysqldump issues earlier. The MariaDBWaitingForLock alert will fire only when other writing queries  take place and time out. But the threshold 15, might be aggressive.


chuan137:
https://prometheus-openstack.eu-nl-1.cloud.sap/graph?g0.range_input=12h&g0.end_input=2020-02-15%2020%3A38&g0.expr=mysql_global_status_threads_running%7Bapp%3D~%22manila.*%22%7D&g0.tab=0

PR #1222, opened 2020-02-21
Fixed Playbook link for Ironic Canary

B1Sandmann:


B1Sandmann:
Tests fails because of
```
[ERROR] templates/: render error in "ironic/charts/postgresql/templates/pgbouncer-deployment.yaml": template: ironic/charts/postgresql/templates/pgbouncer-deployment.yaml:65:174: executing "ironic/charts/postgresql/templates/pgbouncer-deployment.yaml" at <.Values.global.pgbou...>: invalid value; expected string
```
But I didn't touch this.

B1Sandmann:
As #1252 is now merged I think this should be good to go? cc @BerndKue

B1Sandmann:
Bumped chart version so it can be deployed.

PR #1237, opened 2020-02-26
Swift: add health-exporter deployment

talal:


openstack/swift/templates/health-exporter-deployment.yaml
```diff
@@ -0,0 +1,73 @@
+{{- if $.Values.health_exporter }}
+{{- if not $.Values.health_statsd }}
+{{- range $cluster_id, $cluster := .Values.clusters }}
```
reimannf: Once per region is enough. We don't need that per swift-cluster as there is only identical information.

openstack/swift/templates/health-exporter-deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+      maxSurge: 2
+  selector:
+    matchLabels:
+      component: swift-health-exporter
+  template:
+    metadata:
+      labels:
+        component: swift-health-exporter
+      annotations:
+        checksum/swift-cluster.etc: {{ include "swift/templates/etc/cluster-configmap.yaml" $ | sha256sum }}
```
reimannf: This is per swift cluster. we do not need. Just add the ring checksum as well `{{- include "swift_conf_annotations" $ | indent 8 }}`

PR #1238, opened 2020-02-27
[Barbican] Set resources and replicas values

rajivmucheli:
Adding values

### DIFF
--- a/openstack/barbican/templates/api-deployment.yaml
+++ b/openstack/barbican/templates/api-deployment.yaml
@@ -12,7 +12,7 @@ metadata:
     type: api
     component: barbican
 spec:
-  replicas: {{ .Values.pod.replicas.api }}
+  replicas: {{ .Values.api.replicas }}
   revisionHistoryLimit: {{ .Values.pod.lifecycle.upgrades.deployments.revisionHistory }}
   strategy:
     type: {{ .Values.pod.lifecycle.upgrades.deployments.podReplacementStrategy }}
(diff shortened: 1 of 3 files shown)


openstack/barbican/templates/api-deployment.yaml
```diff
@@ -12,7 +12,7 @@ metadata:
     type: api
     component: barbican
 spec:
-  replicas: {{ .Values.pod.replicas.api }}
+  replicas: {{ .Values.api.replicas }}
```
Carthaca: I suggest to set a default

openstack/barbican/templates/api-deployment.yaml
```diff
@@ -57,6 +57,15 @@ spec:
                   name: sentry
                   key: barbican.DSN.python
             {{- end }}
+            {{- if .Values.api.resources.enabled }}
```
Carthaca: This will not work without a default value or some additional guards

openstack/barbican/templates/api-deployment.yaml
rajivmucheli: we cannot set to default due CPU constraints in qa-de-2, 

example: https://github.wdf.sap.corp/cc/secrets/blob/barbican_qa-de-2_values/qa-de-2/values/barbican.yaml#L44

Carthaca: That is a different perspective - you can override the value. If you don't want a default - at least mark the value as required please

openstack/barbican/templates/api-deployment.yaml
rajivmucheli: https://github.wdf.sap.corp/cc/secrets/blob/barbican_qa-de-2_values/staging/values/barbican.yaml#L35

https://github.wdf.sap.corp/cc/secrets/blob/barbican_qa-de-2_values/staging/values/barbican.yaml#L35

testing in qa-de-2 and staging at the moment

openstack/barbican/templates/api-deployment.yaml
rajivmucheli: https://github.com/sapcc/helm-charts/pull/1238/commits/efb62f288c1eb7820187c484ecd4e1891a30bb33 hope this is fine?

rabbitmq_notifications which is required for barbican-audit which is enabled in all regions expect staging

rajivmucheli:
unsure how this passed : https://github.com/sapcc/helm-charts/pull/1263

PR #1243, opened 2020-02-28
[mariadb] fallback to derived passwords

Carthaca:
- use masterpassword mechanism if passwords are not overriden
- for ease of use: add MYSQL_PWD env


galkindmitrii:
OK. FYI: there is also a hammer tool on internal github which supports reading the secret and opening cli directly to both postgres and maria.

Carthaca:
for whatever reason `ccm dbcli` does not work on my mac ¯\_(ツ)_/¯

PR #1247, opened 2020-03-02
[utils] fallback to derived password for mariadb connection

Carthaca:
and allow for derived password


galkindmitrii:
I suppose that will break things for a number of services as I don't see the values set everywhere..

Carthaca:
right, the risk is not worth the exercise

PR #1248, opened 2020-03-03
[tempest][cinder] Add cleanup script

joker-official:
We've run into quota errors with cinder recently. There were a couple of
volumes in different states, but also groups and snapshots. We now try to
delete them after every run.


hemna:
lgtm

PR #1259, opened 2020-03-05
global grafanaAsAServicePassword

artherd42:


Carthaca:
can you adjust in openstack/domain-seeds/values.yaml as well, please

PR #1278, opened 2020-03-13
[Kubernetes] fix playbook link

B1Sandmann:


auhlig:
Could you bump the chart version so this actually gets deployed @B1Sandmann?

B1Sandmann:
@auhlig sorry not so familiar with helm yet. This should do it I believe?

auhlig:
Yes. LGTM

PR #1285, opened 2020-03-16
[openstack][nova] Fix typo in bigvm antiaff. host group name

grandchild:
Probably didn't break anything, but might avoid some confusion later.


joker-official:
We can't change it without also changing all currently existing production hostgroups named by the old name. Is it important enough to you to do that?

grandchild:
No not important at all. Hence the PR.

PR #1286, opened 2020-03-18
added helm for bm-cablecheck-exporter

unnuetz:
added helm for bm-cablecheck-exporter
created new job for custom scraping of more than 1 minute as default

### DIFF
--- a/system/infra-monitoring/templates/_prometheus-infra-collector.yaml.tpl
+++ b/system/infra-monitoring/templates/_prometheus-infra-collector.yaml.tpl
@@ -546,9 +546,9 @@
 {{- end }}
 - job_name: 'cablecheck-exporter'
   params:
-    job: [cablecheck-exporter]
-  scrape_interval: {{$values.cablecheck_scrapeInterval}}
-  scrape_timeout: {{$values.cablecheck_scrapeTimeout}}
+    job: [bm-cablecheck-exporter]
+  scrape_interval: {{$values.bm-cablecheck_scrapeInterval}}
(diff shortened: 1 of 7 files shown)


system/infra-monitoring/templates/_prometheus-infra-collector.yaml.tpl
```diff
@@ -544,3 +544,11 @@
     - target_label: __address__
       replacement: esxi-exporter:9203
 {{- end }}
+- job_name: 'cablecheck-exporter'
```
viennaa: the whole part is not needed, we don't do relabling nor service discovery

system/infra-monitoring/vendor/bm-cablecheck-exporter/templates/_helpers.tpl
```diff
@@ -0,0 +1,31 @@
+{{/* vim: set filetype=mustache: */}}
```
viennaa: please completely remove this file too

system/infra-monitoring/vendor/bm-cablecheck-exporter/templates/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
```
viennaa: please add annotations similar to:
https://github.com/sapcc/helm-charts/blob/master/prometheus-exporters/vrops-exporter/templates/service.yaml#L6-L9

PR #1323, opened 2020-03-27
Remove deprecated APIs

talal:


talal:
Dummy registry value has been added.

PR #1267, opened 2020-03-10
[Glance] adding image_admin and image_viewer roles

rajivmucheli:
Hi @Carthaca,

I was able to run tempest pipeline for all services in qa, application credentials worked fine as well. Please validate.

Regards,
Rajiv

PR #1328, opened 2020-03-30
[Glance][Policy] implementing property protections

rajivmucheli:
Hi @Carthaca,

Please validate this PR, i tested each component and tempest pipeline is running fine since 3 days. This commit is in QA. If this property protection approach is acceptable, i will add the seeder part from https://github.com/sapcc/helm-charts/pull/1267.


rajivmucheli:
do we go ahead with both image_viewer and image_admin roles ?

Johannes informed pipelines need these roles to be manually added, else canary will fail.

openstack/glance/templates/etc/_glance-protection.pp.tpl
```diff
@@ -0,0 +1,12 @@
+[.*]
+create = pp:image_admin_only
+read = pp:image_viewer_only
+update = pp:image_admin_only
+delete = pp:image_admin_only
+
+
+[^upload_.*]
```
Carthaca, line 8: Can you explain what do you want to achieve here? Afaik we are not using image properties that start with `upload_`, why protect them differently?

This may make sense for special properties like `hypervisor_type`, `hw_vif_model` etc.

Carthaca:
> do we go ahead with both image_viewer and image_admin roles ?
> 
> Johannes informed pipelines need these roles to be manually added, else nova canary will fail.

This should be done in a seed, if it is needed everywhere. Or in a one-time script if for transition only.

openstack/glance/templates/etc/_glance-protection.pp.tpl
rajivmucheli: this section resolved https://ci.eu-de-2.cloud.sap/teams/core/pipelines/tempest/jobs/run-glance-tempest-plugin-in-qa-de-1/builds/72

rajivmucheli:
https://github.com/sapcc/helm-charts/blob/glance-pp/openstack/glance/templates/etc/_glance-protection.pp.tpl#L3 should also allow any user role to perform read operations, which will not impact the build pipelines and customer read impact will be 0.

Carthaca:
> https://github.com/sapcc/helm-charts/blob/glance-pp/openstack/glance/templates/etc/_glance-protection.pp.tpl#L3 should also allow any user role to perform read operations, which will not impact the build pipelines and customer read impact will be 0.

Does this work? Is `@` a valid policy in this case?

Carthaca:
I believe the API support group should have image_admin
and maybe the COMPUTE support group as well. @frr2018 what do you think?

rajivmucheli:
1. Yes, its valid and tested in qa-de-1 since a week, more info is here https://docs.openstack.org/glance/latest/admin/property-protections.html

2. Yes, Elektra shows these roles but these roles are not present in seeder.

frr2018:
> I believe the API support group should get image_admin
> and maybe the COMPUTE support group as well. @frr2018 what do you think?

Thats a OK. With that these guys can also have a look and modify images during incidents.

frr2018:
I am not really sure if this will solve https://ci.eu-de-2.cloud.sap/teams/core/pipelines/tempest/jobs/run-glance-tempest-plugin-in-qa-de-1/builds/72

because it looks like it tries to update "ramdisk_id" which looks like it is a protected property which cant be updated.

rajivmucheli:
https://github.com/sapcc/helm-charts/blob/glance-pp/openstack/glance/templates/etc/_glance-protection.pp.tpl#L3 this solves the issue, @ allows all roles to perform read operations. QA tempest is running fine since last week with this setting.

openstack/glance/templates/etc/_glance-policy.json.tpl
```diff
@@ -46,5 +47,8 @@
     "get_metadef_tags":"",
     "modify_metadef_tag":"",
     "add_metadef_tag":"",
-    "add_metadef_tags":""
+    "add_metadef_tags":"",
+
+    "pp:image_admin_only":"role:image_admin or role:cloud_image_admin or role:admin"
```
Carthaca, line 52: this can be removed, it is not used in glance-protection.pp.tpl

PR #1334, opened 2020-03-31
[Keystone] Add conditional apps/v1

rajivmucheli:
Hi @Carthaca,

Please validate.

Regards,
Rajiv

### DIFF
--- a/openstack/keystone/templates/ingress-api.yaml
+++ b/openstack/keystone/templates/ingress-api.yaml
@@ -36,7 +36,11 @@ data:
 ---
 {{- end }}
 {{- end }}
+{{- if .Capabilities.APIVersions.Has "apps/v1" }}
+apiVersion: apps/v1
+{{- else }}
 apiVersion: extensions/v1beta1
+{{- end }}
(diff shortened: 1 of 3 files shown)


openstack/keystone/templates/ingress-api.yaml
```diff
@@ -36,7 +36,11 @@ data:
 ---
 {{- end }}
 {{- end }}
+{{- if .Capabilities.APIVersions.Has "apps/v1" }}
```
Carthaca: this is of `kind: Ingress`
so I suggest to use `{{- if $.Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}`

openstack/keystone/templates/ingress-api.yaml
```diff
@@ -36,7 +36,11 @@ data:
...
+apiVersion: apps/v1
```
Carthaca: this is of `kind: Ingress`
so I suggest to use `apiVersion: networking.k8s.io/v1beta1`

PR #1369, opened 2020-04-03
[keystone][policy] create an ec2 credential within project_id scope

rajivmucheli:


Carthaca:
Seems to have been untested. I will work on this.

PR #1386, opened 2020-04-08
bump ntp-exporter to 1.1.2

majewsky:
My scheduled task for upgrading all the things came up, so ntp-exporter got the "upgrading all the things" treatment.


auhlig:
triggered my "merging all the things" routine

PR #1395, opened 2020-04-09
[rabbitmq] Fix "for" condition on alerts

joker-official:
It's useless if used as a label.


geisslet:
lgtm //sorryAndThanks

PR #1416, opened 2020-04-15
[vrops-exporter] Added InaccessibleShadowVM alert

christopherhans:
Multiple alert conditions which indicates an inaccessible shadow vm.

### DIFF
--- a/prometheus-exporters/vrops-exporter/alerts/virtual_machine.alerts
+++ b/prometheus-exporters/vrops-exporter/alerts/virtual_machine.alerts
@@ -1,0 +1,20 @@
+groups:
+- name: virtual_machine.alerts
+  rules:
+  - alert: InaccessibleShadowVM
+    expr: >
+      vrops_vm_stats{statkey="number_vcpus_total"} == 1
+      and on(virtualmachine) vrops_vm_stats{statkey="vm_memory_kilobytes"} == 131072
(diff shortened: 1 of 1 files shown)


prometheus-exporters/vrops-exporter/alerts/virtual_machine.alerts
```diff
@@ -0,0 +1,20 @@
+groups:
+- name: virtual_machine.alerts
+  rules:
+  - alert: InaccessibleShadowVM
+    expr: >
+      vrops_vm_stats{statkey="number_vcpus_total"} == 1
+      and on(virtualmachine) vrops_vm_stats{statkey="vm_memory_kilobytes"} == 131072
+      and on(virtualmachine) vrops_vm_properties{propkey="summary_ethernetCards"} == 0
+      and on(virtualmachine) vrops_vm_properties{propkey="runtime_connectionState: disconnected"} == 1
+    for: 5m
```
viennaa, line 10: do you want to fire straight? usually i am giving it 15-30mins to sit on a matched condition before firing.

christopherhans: You are right. Increased to 20 min

christopherhans:
@viennaa we should give it a try

PR #1421, opened 2020-04-16
[cinder] Add new vmware_low_iops backend

hemna:
This patch adds a new cinder backend labelled vmware_low_iops
that will service the slower performing iops volumes.  By adding
this backend to the existing cinder-volume.conf, cinder will
fork another cinder volume service in the same container that will
service provisioning requests.   Since the default volume type is
vmware, this new backend won't get any provisioning requests unless
there is a new volume type associated with it's volume_backend_name.

With this patch we need a new cinder volume type created, named
'standard_hdd' with the following extra specs:

vmware:storage_profile : cinder_low_iops                                                          
vmware:vmdk_type : thick                                                                          
volume_backend_name : vmware_low_iops                                                                                                                                                   

There needs to be a new storage profile created in vcenter named
'cinder_low_iops' that has SIOC enabled and a value of 1000 iops
specified.  We need to ensure this is populated on all systems before
this rolls out.

### DIFF
--- a/openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
+++ b/openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
@@ -20,7 +20,7 @@ template: |
   data:
     cinder-volume.conf: |
       [DEFAULT]
-      enabled_backends = vmware
+      enabled_backends = vmware,vmware_low_iops
 
       [backend_defaults]
       vmware_host_ip = {= host =}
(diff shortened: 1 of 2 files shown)


openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
```diff
@@ -41,6 +41,13 @@ template: |
       volume_backend_name = vmware
       extra_capabilities = '{"vcenter-shard": "{= host.split(".")[0] =}"}'
 
+      [vmware_low_iops]
+      volume_driver = cinder.volume.drivers.vmware.vmdk.VMwareVcVmdkDriver
+      volume_backend_name = vmware_low_iops
+      vmware_storage_profile = cinder-low-iops
```
joker-official, line 47: With this, we're in a situation where we can change the name of the `vmware_storage_profile` only for the `vmware` backend, not for the `vmware_low_iops` backend. I'd prefer this to be configurable, too. I'd also prefer if the `vmware_storage_profile` for the `vmware` backend would show up more explicitly. It's currently hidden in the `backend_defaults` section in the for-loop.

PR #1427, opened 2020-04-20
[manila] Manila share instance metric

chuan137:
Provides share_id <-> share_instance_id mapping

### DIFF
--- a/openstack/manila/values.yaml
+++ b/openstack/manila/values.yaml
@@ -160,6 +160,29 @@ mysql_metrics:
           availability_zones.id;
       values:
       - "size_gauge"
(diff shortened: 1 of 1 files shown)


Carthaca:
I would rather put this into openstack_manila_shares_count_gauge instead of introducing another metric if this works for you. I suggest to add `share_instances.id AS share_instance_id` and `WHERE share_instances.deleted='False'` there.

 I think we don't need the deleted shares at that place either, it just consumes a lot of memory in the exporter pod.

chuan137:
I have changed the metrics as you suggested @Carthaca 

I have to revert some fixes for train release because they will fail otherwise in production regions

PR #1443, opened 2020-04-26
[alertmanager] add awx routing

geisslet:
vmware alerts should be able to trigger actions in awx

pls check related secrets change, too

### DIFF
--- a/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
+++ b/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
@@ -332,8 +332,8 @@ receivers:
     - send_resolved: true
       http_config:
         basic_auth:
-          username: {{ required ".Values.awx.basicAuthUser undefined" .Values.awx.user | quote }}
-          password: {{ required ".Values.awx.basicAuthPwd undefined" .Values.awx.pwd | quote }}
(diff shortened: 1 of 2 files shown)


global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
```diff
@@ -322,6 +327,15 @@ receivers:
     - send_resolved: true
       url: {{ required ".Values.elastic.logstashURL undefined" .Values.elastic.logstashURL | quote }}
 
+  - name: awx
+    webhook_configs:
+    - send_resolved: true
+      http_config:
+        basic_auth:
+          username: {{ required ".Values.awx.basicAuthUser undefined" .Values.awx.user | quote }}
```
auhlig, line 335: `.Values.awx.basicAuthUser` vs `.Values.awx.user`

global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
```diff
@@ -322,6 +327,15 @@ receivers:
...
+          password: {{ required ".Values.awx.basicAuthPwd undefined" .Values.awx.pwd | quote }}
```
auhlig, line 336: `.Values.awx.basicAuthPwd` vs `.Values.awx.pwd`

PR #1450, opened 2020-04-29
[cinder] make enabled_backends configurable

hemna:
This patch makes the cinder-volume.conf enabled_backends
configurable.   This allows each region to optionally enable
the standard_hdd backend along with vmware backend.  This patch
makes the default value for enabled_backends=vmware.

If you want to enable the slow performing cinder backend,
then edit the secrets <region>/values/cinder.yml and add
backends:
  enabled: vmware,standard_hdd

### DIFF
--- a/openstack/cinder/values.yaml
+++ b/openstack/cinder/values.yaml
@@ -57,6 +57,7 @@ backend_defaults:
   max_over_subscription_ratio: 1.5
 
 backends:
+    enabled: vmware
     vmware:
         vmware_storage_profile: cinder-vvol
     standard_hdd:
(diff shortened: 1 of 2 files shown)


openstack/cinder/values.yaml
```diff
@@ -57,6 +57,7 @@ backend_defaults:
   max_over_subscription_ratio: 1.5
 
 backends:
+    enabled: vmware
```
grandchild, line 60: Just out of curiosity: Am I reading this correct, that the default before was `vmware,standard_hdd` and is now only `vmwware` unless overridden by a value from secrets?

grandchild: I'm simply overly-cautious and thinking about what happens if the change in secrets is (for some reason) not applied at the same time.

hemna: Yes, you are correct.   standard_hdd is a new backend that hasn't been deployed to production.   The requirements of deploying this backend has changed to only wanting to deploy it to very specific regions until we have more customers asking for it.  That's why it was removed and added as an optional configurable change here as well as secrets.

grandchild: I see. Thanks!

PR #1452, opened 2020-04-30
[blackbox] add ews env to datapath tests

Carthaca:


Carthaca:
we use ews health check instead https://dirteam.com/dave/2015/03/12/using-your-browser-to-check-exchange-2013-protocol-health/

PR #1502, opened 2020-05-12
[cablecheck-exporter] added cache mode

unnuetz:
[bm-cablecheck-exporter] V11
[vpod-cablecheck-exporter] V9
reduced polling time from 15m to 1m (collector provides now cached metrics, internally cablecheck will queried every 30m to reduct netbox/switch api load)


viennaa:
If I got this right, you are providing static metrics with this change. That's not how the exporter should be running since it is not reflecting real data. To avoid your timeline gaps you need to optimize your exporter to finish within 5minutes, better 1 minute.

PR #1505, opened 2020-05-13
link mysql too many connections playbook

DRichardt:


Carthaca:
`common/mariadb/Chart.yaml
  5:1       error    duplication of key "version" in mapping  (key-duplicates)`
Please increment the version, don't add an additional key - i.e. remove line 4 please

common/mariadb/Chart.yaml
```diff
@@ -2,3 +2,4 @@ apiVersion: v1
 description: A Helm chart for Kubernetes
 name: mariadb
 version: 0.1.5
+version: 0.1.6
```
B1Sandmann, line 5:
```suggestion
version: 0.1.6
```
I think that is what you meant

galkindmitrii:
Is that really required? For _severity: info_ alerts we don't have any playbooks mostly.

PR #1510, opened 2020-05-13
[SRE] clean up & housekeeping

geisslet:
- replace the useless sophisticated by simple availability (last hour) alerts to trigger action
- this lead to changes in the aggregation and a shift to percentage SLO values

### DIFF
--- a/openstack/sre/templates/prometheus-aggregations.yaml
+++ b/openstack/sre/templates/prometheus-aggregations.yaml
@@ -19,7 +19,7 @@ spec:
       labels:
           ingress: {{ $api | quote }}
           region: {{required ".Values.global.region missing" $values.global.region}}
-      expr: {{ $slo.warning | default $values.slo.default_warning  | quote }}
+      expr: {{ $slo.warning | default $values.slo.defaults.api_availability_percent.warning | quote }}
 {{- end }}
 
 {{- range $api, $slo := $values.slo.api_availability_percent }}
@@ -27,7 +27,7 @@ spec:
       labels:
           ingress: {{ $api | quote }}
           region: {{required ".Values.global.region missing" $values.global.region}}
-      expr: {{ $slo.critical | default $values.slo.default_critical | quote }}
+      expr: {{ $slo.critical | default $values.slo.defaults.api_availability_percent.critical | quote }}
 {{- end }}
 
 {{- range $period := $values.periods }}
--- a/openstack/sre/templates/prometheus-aggregations.yaml
+++ b/openstack/sre/templates/prometheus-aggregations.yaml
@@ -14,13 +14,20 @@ spec:
   groups:
   - name: openstack-sre.aggregations
     rules:
-{{- range $api, $slo := $values.slo.api_error_ratio_percent }}
-    - record: global:api_errors_per_request_slo:percent
+{{- range $api, $slo := $values.slo.api_availability_percent }}
+    - record: global:api_availability_warning_slo:percent
       labels:
           ingress: {{ $api | quote }}
           region: {{required ".Values.global.region missing" $values.global.region}}
-      expr: {{ $slo | quote }}
+      expr: {{ $slo.warning | default $values.slo.default_warning  | quote }}
+{{- end }}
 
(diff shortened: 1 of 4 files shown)


openstack/sre/templates/prometheus-alerts.yaml
```diff
@@ -13,48 +13,24 @@ spec:
...
-            (100 * global:api_errors_per_request_sli:ratio_rate6h > on (ingress) global:api_errors_per_request_slo:percent)
-          )
-      for: 1h
+    - alert: ApiAvailabilityCritical
+      expr: global:api_errors_per_request_sli:ratio_rate1h <= on (ingress,region) global:api_availability_critical_slo:percent
       labels:
         context: availability
         service: "{{`{{$labels.ingress}}`}}"
-        severity: info
+        severity: warning
```
joker-official, line 32: Why just warning here? It's supposed to be critical in the end, right?

openstack/sre/templates/prometheus-alerts.yaml
```diff
@@ -13,48 +13,24 @@ spec:
...
-          )
-          or
-          (
-            (100 * global:api_errors_per_request_sli:ratio_rate3d > on (ingress) global:api_errors_per_request_slo:percent)
-            and
-            (100 * global:api_errors_per_request_sli:ratio_rate6h > on (ingress) global:api_errors_per_request_slo:percent)
-          )
-      for: 1h
+    - alert: ApiAvailabilityCritical
+      expr: global:api_errors_per_request_sli:ratio_rate1h <= on (ingress,region) global:api_availability_critical_slo:percent
```
joker-official, line 28: Don't we need the same thing as above here, too? I mean the `(1 - X) * 100`?

openstack/sre/templates/prometheus-aggregations.yaml
```diff
@@ -14,13 +14,20 @@ spec:
...
     rules:
-{{- range $api, $slo := $values.slo.api_error_ratio_percent }}
-    - record: global:api_errors_per_request_slo:percent
+{{- range $api, $slo := $values.slo.api_availability_percent }}
+    - record: global:api_availability_warning_slo:percent
       labels:
           ingress: {{ $api | quote }}
           region: {{required ".Values.global.region missing" $values.global.region}}
-      expr: {{ $slo | quote }}
+      expr: {{ $slo.warning | default $values.slo.default_warning  | quote }}
```
joker-official, line 22: One space too much in front of `| quote`.

openstack/sre/values.yaml
```diff
@@ -9,40 +9,42 @@ prometheus:
...
-    maia: 0.1
-    maia-oprom: 0.1
-    manila-api: 0.1
-    neutron-server: 0.1
-    nova-api: 0.1
-    nova-console: 0.1
-    nova-placement-api: 0.1
-    octavia-api: 0.1
+  default_warning: 99.8
+  default_critical: 99.6
```
joker-official, line 20: Imho, putting the defaults here just as "default_warning" doesn't make sense if we ever want to support more than one metric, which having "api_availability_percent" as an extra dict implies. It needs to somehow align with supporting that, e.g.
```
slo:
  defaults:
    api_availability_percent:
      warning: 99.8
      critical: 99.6
```

That being said, it currently doesn't matter and could be done once we need more than one default or metric. Your choice.

openstack/sre/templates/prometheus-alerts.yaml
geisslet: true

openstack/sre/templates/prometheus-alerts.yaml
geisslet: true // sorry

openstack/sre/templates/prometheus-aggregations.yaml
geisslet: ok

openstack/sre/values.yaml
geisslet: very true

PR #1541, opened 2020-05-20
[mariadb] backupv2 img bump

stefanhipfel:


common/mariadb/values.yaml
```diff
@@ -82,7 +82,7 @@ backup_v2:
   enabled: false
   backup_dir: "./backup"
   image: monsoon/maria-back-me-up
-  image_version: 7fc55517
+  image_version: 136e451d
```
Carthaca, line 85: Is there a meaning behind the version, can this be correlated to the source https://github.com/sapcc/maria-back-me-up somehow?

stefanhipfel: yes
https://github.com/sapcc/maria-back-me-up/commit/**136e451d**2119b661143775ea17c61da8f395de7c

the version is part of the git commit id

Carthaca: you just pushed this now? ^^

stefanhipfel: i forgot to push it (lunch time)...was still local on my pc ;)

PR #1549, opened 2020-05-25
[prometheus-infra] added metrics to federation

richardtief:


global/prometheus-infra/templates/_prometheus.yaml.tpl
```diff
@@ -79,6 +79,8 @@
       - '{__name__=~"^vrops_hostsystem_stats",statkey=~"cpu_usage_average_percentage|summary_running_VMs_number"}'
       - '{__name__=~"^vrops_hostsystem_properties",propkey=~"runtime_maintenanceState: inMaintenance"}'
       - '{__name__=~"^vrops_cluster_stats",statkey=~"cluster_running_hosts|memory_usage_percentage|cpu_capacity_usage_percentage"}'
+      - '{__name__=~"^vrops_vms_stats",statkey=~"vm_memory_kilobytes"}'
+      - '{__name__=~"^vrops_vms_properties", propkey=~"runtime_powerState: Powered On"}'
```
viennaa, line 83: wrong metric name, should be `vm` instead of `vms`

PR #1568, opened 2020-05-27
[Glance] enabling swift object size and object chunk size

rajivmucheli:
Hi @reimannf 

wrt 230 Gb server sizes, i found :

1. the chunk count reduced from 1219 to 477
2. upload speeds are up by 7Mbps
3. upload time reduced by ~30mins

I would like to know, if enabling these flags would impact swift ? swift-ratelimiting document explains operations per second only.

Regards,
Rajiv

### DIFF
--- a/openstack/glance/templates/etc/_glance-api.conf.tpl
+++ b/openstack/glance/templates/etc/_glance-api.conf.tpl
@@ -61,6 +61,8 @@ swift_store_auth_insecure = True
 swift_store_create_container_on_put = True
 {{- if .Values.swift.multi_tenant }}
 swift_store_multi_tenant = True
+swift_store_large_object_size = 5120
+swift_store_large_object_chunk_size = 512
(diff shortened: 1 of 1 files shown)


reimannf:
There are 2 impacts:
- You only chunk when the Image is >= 5GiB, which was the default anyhow based on glance docs and is the upper limit for object sizes in Swift. Latest at this size you have to chunk.
- You create bigger but lesser chunks

Swift in general let you choose the segment size when uploading Large Objects. It will have no affect on rate limit, as this is per container and number of parallel PUTs. I think glance does a serial upload of chunks. So from pure Swift perspective this is as good as the previous setting.

Just my 2 cents: You are now tuning glance against the 230GiB image scenario, I am not sure this is the main usage pattern.

rajivmucheli:
Yes, these flags are enabled to onboard Neo customer, they plan to create 30 images in parallel of same size.

So there is no impact on swift, since glance does a serial upload, correct? Can I go ahead?

reimannf:
from **swift** side, no objections

PR #1580, opened 2020-05-27
[cinder] add volume type seeding

hemna:
This patch adds the 2 volume types to seed every deployment of
cinder with. This adds vmware and standard_hdd volume types and
all the extra specs for each volume type.

### DIFF
--- a/openstack/cinder/templates/type-seeds.yaml
+++ b/openstack/cinder/templates/type-seeds.yaml
@@ -1,0 +1,30 @@
+{{- if .Capabilities.APIVersions.Has "openstack.stable.sap.cc/v1" }}
+apiVersion: "openstack.stable.sap.cc/v1"
+kind: "OpenstackSeed"
+metadata:
+  name: cinder-type-seed
+  labels:
+    component: cinder
+    app: cinder
+    type: seed
+spec:
+  requires:
+  - monsoon3/cinder-seed
+
+  volume_types:
+  - name: vmware
+    is_public: true
+    description: "premium_ssd"
+    extra_specs:
+      volume_backend_name: "vmware"
+      "vmware:storage_profile": "cinder-vvol"
+
+  - name: standard_hdd
+    is_public: true
+    description: "standard_hdd"
+    extra_specs:
+      volume_backend_name: "standard_hdd"
+      "provisioning:min_vol_size": "500"
+      "vmware:storage_profile": "cinder-standard-hdd"
+      "vmware:vmdk_type": "thick"
+{{- end }}


openstack/cinder/templates/type-seeds.yaml
```diff
@@ -0,0 +1,37 @@
...
+kind: "OpenstackSeed"
+metadata:
+  name: cinder-type-seed
+  labels:
+    component: cinder
+    app: cinder
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+    type: seed
```
joker-official, line 13: I'm not sure about these. the `cinder-seed` only has a `name` as metadata. They probably will not matter.

openstack/cinder/templates/type-seeds.yaml
```diff
@@ -0,0 +1,37 @@
...
+spec:
+  requires:
+  - monsoon3/cinder-seed
```
joker-official, line 16: I wasn't sure how this works, but after looking at it, it looks fine.

openstack/cinder/templates/type-seeds.yaml
```diff
@@ -0,0 +1,37 @@
+{{- if .Release.IsUpgrade }}
```
joker-official, line 1: I don't understand, why we need this. Does this mean, we don't want to put this into kubernetes on first deployment? If yes, why?

openstack/cinder/templates/type-seeds.yaml
```diff
@@ -0,0 +1,37 @@
...
+{{- if .Capabilities.APIVersions.Has "openstack.stable.sap.cc/v1"}}
```
joker-official, line 2: This is fine, I guess. The `cinder-seed` does not have it though, so cinder can currently not be deployed without that `CustomResourceDefinition` existing.

You should add a space before the closing `}}`, though.

openstack/cinder/templates/type-seeds.yaml
```diff
@@ -0,0 +1,37 @@
...
+    type: seed
+spec:
+  requires:
+  - monsoon3/cinder-seed
+
+  volume_types:
+  - name: vmware
+    is_public: true
+    specs:
+      description: "premium_ssd"
```
joker-official, line 22: Reading the code, I don't think this will work:
https://github.com/sapcc/kubernetes-operators/blob/31feb0820940f7b7dad40b7492478307838f5551/openstack-seeder/python/openstack_seeder.py#L2198

https://github.com/sapcc/kubernetes-operators/blob/31feb0820940f7b7dad40b7492478307838f5551/openstack-seeder/pkg/seeder/apis/v1/types.go#L76

It needs to be without `specs`, I guess.

openstack/cinder/templates/type-seeds.yaml
```diff
@@ -0,0 +1,37 @@
...
+    extra_specs:
+      volume_backend_name: "vmware"
+      "vmware:storage_profile": "cinder-vvol"
+
+  - name: standard_hdd
+    is_public: true
+    specs:
+      description: "standard_hdd"
```
joker-official, line 30: Same problem here: without specs.

openstack/cinder/templates/type-seeds.yaml
Carthaca: I think it should be `extra_specs` and will work then

joker-official: No. `extra_specs` is something else and already used.

Carthaca: I see, my bad - description can be used directly without the parent key

openstack/cinder/templates/type-seeds.yaml
hemna: removed

openstack/cinder/templates/type-seeds.yaml
hemna: removed

openstack/cinder/templates/type-seeds.yaml
hemna: updated

openstack/cinder/templates/type-seeds.yaml
hemna: updated

PR #1581, opened 2020-05-28
[Glance] Add communitize_image policy and image_community_admin role

rajivmucheli:
Hi @Carthaca 

Regarding Jira task CCM-13726, please validate.

Regards,
Rajiv

### DIFF
--- a/openstack/glance/templates/etc/_glance-policy.json.tpl
+++ b/openstack/glance/templates/etc/_glance-policy.json.tpl
@@ -1,6 +1,6 @@
 {
     "context_is_cloud_admin":  "role:cloud_image_admin",
-    "context_is_admin":  "rule:context_is_cloud_admin",
+    "context_is_admin":  "rule:context_is_cloud_admin or role:admin or role:image_admin",
     "default": "rule:context_is_admin",
 
     "add_image": "",
@@ -9,6 +9,7 @@
     "get_images": "",
     "modify_image": "",
     "publicize_image": "rule:context_is_cloud_admin",
+    "communitize_image": "context_is_admin or role:image_community_admin",
     "copy_from": "",
     "download_image": "",
     "upload_image": "",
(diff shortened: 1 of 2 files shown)


openstack/glance/templates/etc/_glance-policy.json.tpl
```diff
@@ -1,6 +1,6 @@
 {
     "context_is_cloud_admin":  "role:cloud_image_admin",
-    "context_is_admin":  "rule:context_is_cloud_admin",
+    "context_is_admin":  "rule:context_is_cloud_admin or role:admin or role:image_admin",
```
Carthaca, line 3: Why are you relaxing this rule?

rajivmucheli: role:admin or role:image_admin can also set images visibility to community, these roles neednt add role:image_community_admin again.

Carthaca: If I read the policy correctly they can't do this currently, rule `default` doesn't allow that.

Do you want to allow communitize_image for roles `admin` and `image_admin`, too? Please add it directly to that rule.

openstack/glance/templates/seeds.yaml
```diff
@@ -29,6 +29,7 @@ spec:
   - name: cloud_image_admin
   - name: image_admin
   - name: image_viewer
+  - name: image_community_admin
```
Carthaca: Please remove the unused role here

openstack/glance/templates/etc/_glance-policy.json.tpl
```diff
@@ -9,6 +9,7 @@
     "get_images": "",
     "modify_image": "",
     "publicize_image": "rule:context_is_cloud_admin",
+    "communitize_image": "role:image_admin",
```
Carthaca, line 12: please add the permission for cloud_image_admin, e.g. `role:image_admin or rule:context_is_admin`

PR #1646, opened 2020-06-19
[SRE] add a rolling 7 days alert; shorten message

geisslet:


openstack/sre/templates/prometheus-alerts.yaml
```diff
@@ -13,24 +13,65 @@ spec:
...
         context: availability
         service: "{{`{{$labels.ingress}}`}}"
         severity: warning
         tier: sre
+        playbook: 'docs/devops/alert/sre-alerts.html'
       annotations:
-        summary: "An API fell below availability warning SLO in the last hour"
-        description: "{{`{{$labels.ingress}}`}} fell below availability warning SLO in the last hour."
+        summary: "API availability warning SLO hit: 1h"
+        description: "{{`{{$labels.ingress}}`}} fell below availability warning SLO in the last hour (default:  *99.8%*)."
```
joker-official, line 27: This makes me think, that we have to create one alert per service we defined, so we can show the actual value instead of just a default. This would also help in later on routing the alerts to the right channels, e.g. cc-os-nova for the nova alerts.

openstack/sre/templates/prometheus-alerts.yaml
```diff
@@ -13,24 +13,65 @@ spec:
...
+        severity: info
+        tier: os
+        playbook: 'docs/devops/alert/sre-alerts.html'
+      annotations:
+        summary: "API availability critical SLO hit: 7d"
+        description: "{{`{{$labels.ingress}}`}} ffell below weekly availability warning SLO (default:  *99.6%*)."
+
+    # meta alert
+    - alert: SRE-MetricAbesent
+      expr: absent(global:api_errors_per_request_sli:ratio_rate1h) or absent(global:api_availability_critical_slo:percent)
```
joker-official, line 68: How do we define that this alert prohibits those other alerts from above from firing? Or do the ones above not fire at all if one of the metrics is missing?

What about the `global:api_errors_per_request_sli:ratio_rate7d` we use above? Don't we have to check for its absence, too?

openstack/sre/templates/prometheus-alerts.yaml
geisslet: Mmh .. you know what that I didn't found a way to put `(ingress,region) global:api_availability_critical_slo:percent` into the text sucks -  don't like this double definition.

But I agree on your proposal - just need to add the service and it should be routed than -- maybe some relabel magic from ingress name? 
Should we do it from start on?

joker-official: I assumed we could use our chart's values and with that have both service-name and the actual value we also fed to prometheus. We can then use both to build the alert - each with their own name e.g. `NovaAvailabilitySucksAlert` from `{{service}AvailabilitySucksAlert`. Don't know how to capitalize with go-template, yet, but we have something like that for other alerts too, don't we?

openstack/sre/templates/prometheus-alerts.yaml
geisslet: Yes @ "Or do the ones above not fire at all if one of the metrics is missing"

YNeos @  "Don't we have to check for its absence, too" : 
- yes part: to be on the safe side we should check, too
- no part: normally I just choose one metric as sample for the concept (it's a ~8o% approach of risk coverage)

joker-official: ok. then let's keep it that way.

openstack/sre/templates/prometheus-alerts.yaml
```diff
@@ -14,28 +14,28 @@ spec:
...
         context: availability
         service: "{{`{{$labels.ingress}}`}}"
-        severity: warning
+        severity: info
         tier: sre
         playbook: 'docs/devops/alert/sre-alerts.html'
       annotations:
-        summary: "API availability warning SLO hit: 1h"
+        summary: "API availability warning SLO hit: 2h"
         description: "{{`{{$labels.ingress}}`}} fell below availability warning SLO in the last hour (default:  *99.8%*)."
```
joker-official, line 27: > in the last hour
needs to be last 2 hours

geisslet: true

openstack/sre/templates/prometheus-alerts.yaml
geisslet: true - I put it into the todo

PR #1688, opened 2020-06-25
cp alerts query fix

ruizufase:
There was an issue reported that some alerts on the vpod channel where from not vpod servers, it was found that the query for the alerts was pulling some server which serial number coincided with the query 'CP' and those server were routed to the wrong slack channel, this will fix the issue by adding '-' to the atlas query so it ignores serial numbers.

### DIFF
--- a/system/infra-monitoring/vendor/atlas/templates/configmap.yaml
+++ b/system/infra-monitoring/vendor/atlas/templates/configmap.yaml
@@ -270,7 +270,7 @@ data:
               target: 2
               region: "{{ .Values.global.region }}"
               status: "active"
-              q: "cp"
+              q: "-cp"
               role: "server"
               management_ip: true
               has_primary_ip: true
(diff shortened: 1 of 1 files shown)


system/infra-monitoring/vendor/atlas/templates/configmap.yaml
```diff
@@ -270,7 +270,7 @@ data:
               target: 2
               region: "{{ .Values.global.region }}"
               status: "active"
-              q: "cp"
+              q: "-cp"
```
viennaa, line 273: this one is not correct. `cp/netbox` is only used by ipmi-exporter and my guess would be, that you need this change for the bios-exporter?

ruizufase: yes, that's tight, I changed the cp/netbox as well as it is anyways wrong, and it will have the same issue pulling servers that are not CP, but if you want me to undo that I will do it.

viennaa: Good point tho and good catch ;)

PR #1712, opened 2020-06-30
[nova] Always start metadata service in own pod

joker-official:
When having multiple cells, we disabled the standalone metadata service,
because that's not supported. Instead, one has to run the metadata
service "globally" as an api service. But this only means, one should
not use the "nova-api-metadata" executable to run the metadata service.

Therefore, we now run the metadata service in its own pod with the
nova-api executable.

We need to make sure we switch neutron to using the standalone api
endpoint everywhere before completely removing metadata from the api
service.

Afterwards, we should be able to switch both api and metadata to using
uwsgi. This is not possible with nova-api also having to spawn the
metadata service on a different port.

### DIFF
--- a/openstack/nova/templates/api-deployment.yaml
+++ b/openstack/nova/templates/api-deployment.yaml
@@ -90,6 +90,7 @@ spec:
           ports:
             - name: nova-api
               containerPort: {{.Values.global.novaApiPortInternal}}
+            # TODO remove this after migrating neutron to always use the explicit metadata service
             - name: nova-metdata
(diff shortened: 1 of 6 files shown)


joker-official:
Hey @notandy , please have a look at this. We talked about it before and came to the conclusion, that it's not possible, but I think I found the way to make it work.

notandy:
Implies a bit of overhead running nova-api as a pure metadata provider in a single cell region, but I see what you want to achive (running uwsgi-nova). I am ok with that.

PR #1730, opened 2020-07-02
[apic] add two new alerts for check on accessibility

geisslet:
decided to leave the check on 200, just because it seems to be the more resilient way of check.


geisslet:
obsolete

PR #1740, opened 2020-07-06
federating power and boot settings for baremetal servers for new dashboard

ruizufase:
We had a discussion with our service provider and they will need to have these settings on a global dashboard.


ruizufase:
discussed this with stakeholders, decided to leave it as a regional dashboard.

PR #1742, opened 2020-07-08
[bm-cablechecker] implemented swift credentials

unnuetz:


viennaa:
is it intentional, that you removed all these deployment and helm-chart files?

unnuetz:
yes, these are old, and not needed anymore

PR #1743, opened 2020-07-08
split supernova-incident-contact-list chart from supernova chart

majewsky:
By putting this ConfigMap in a separate chart, we can use our standard deployment tooling, as explained in the new chart's README.

The new chart uses a different name for the ConfigMap, so as not to cause a collision when we first install it.

Please only approve, don't merge yet. I'm still preparing the requisite changes for the Concourse pipeline.

### DIFF
--- a/global/supernova/templates/contact-list-configmap.yaml
+++ b/global/supernova/templates/contact-list-configmap.yaml
@@ -1,9 +1,0 @@
-apiVersion: v1
-kind: ConfigMap
-
-metadata:
-  name: incident-contact-list
-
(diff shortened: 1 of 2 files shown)


majewsky:
I committed the new chart on master, but left the changes to the existing chart in here. That way, I can deploy the new chart already, and we can merge this once we're ready to cut over to the new ConfigMap.

With regards to the missing `ci/test-values.yaml`, you can commit it onto this branch if you like to get some immediate feedback from the CI.

PR #1744, opened 2020-07-08
[neutron] Allow CRUD on trunks and subports

sebageek:


openstack/neutron/templates/etc/_neutron-policy.json.tpl
```diff
@@ -262,12 +262,12 @@
...
     "get_auto_allocated_topology": "rule:context_is_network_admin",
 
-    "create_trunk": "rule:context_is_admin",
-    "get_trunk": "rule:context_is_admin",
-    "delete_trunk": "rule:context_is_admin",
-    "get_subports": "rule:context_is_admin",
-    "add_subports": "rule:context_is_admin",
-    "remove_subports": "rule:context_is_admin",
+    "create_trunk": "rule:context_is_network_admin",
+    "get_trunk": "rule:context_is_network_viewer",
```
Carthaca, line 266: I suggest to use `context_is_viewer`, trunks don't seem to have a direct relation to networks

openstack/neutron/templates/etc/_neutron-policy.json.tpl
```diff
@@ -262,12 +262,12 @@
...
+    "delete_trunk": "rule:context_is_network_admin",
+    "get_subports": "rule:context_is_network_viewer",
```
Carthaca, line 268: Same like above: `context_is_viewer`

openstack/neutron/templates/etc/_neutron-policy.json.tpl
sebageek: Trunks create a relationship between Ports. A Trunk has a parent port and a list of subports with segmentation information (segmentation type + segmentation id). I think without a being able to view the ports it would not be any help to see the trunks, therefore I would choose the same policy that `get_port` has.

Carthaca: Okay ..and vice versa: not seeing the trunks, but the ports, would probably also not make much sense.

sebageek:
My original observation that quota is needed is based upon that I tried to create a trunk and quota was missing. It only worked after @joker-official and I created a manual entry for it.

Investigating somewhat how this works in neutron, the resource is created in neutron/extensions/trunk.py with `register_quota=True`. This somewhen ends up here:

https://github.com/sapcc/neutron/blob/stable/queens-m3/neutron/quota/resource.py#L94-L96

The mentioned `self.flag` should be something like `quota_trunk` (see here: https://github.com/sapcc/neutron/blob/stable/queens-m3/neutron/quota/resource_registry.py#L181), which is not set. The alternative of `default_quota` is set to 0 here https://github.com/sapcc/helm-charts/blob/94ec020971bf20df229eb93440b08ae10381684b/openstack/neutron/templates/etc/_neutron.conf.tpl#L129

So unless there isn't anything set and this default is used it should be 0.

PR #1745, opened 2020-07-09
[Designate-seeds] Adding cloud_identity_viewer and cloud_resource_vie…

rajivmucheli:
…wer to *_domain_dns_ops access level

@Carthaca 

CAM profile :  CC Converged Cloud DNS Operations has *_domain_dns_ops and *_domain_users access levels, i have added cloud_identity_viewer and cloud_resource_viewer roles to designate/seeds.yaml do i need to create *_domain_dns_ops spec in domain-seeds.yaml

### DIFF
--- a/openstack/designate/templates/seeds.yaml
+++ b/openstack/designate/templates/seeds.yaml
@@ -38,8 +38,9 @@ spec:
   - name: dns_webmaster
   - name: dns_admin
   - name: dns_viewer
-  - name: cloud_identity_viewer
-  - name: cloud_resource_viewer
+  - name: role_viewer
+  - name: resource_viewer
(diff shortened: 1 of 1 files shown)


openstack/designate/templates/seeds.yaml
```diff
@@ -38,8 +38,9 @@ spec:
   - name: dns_webmaster
   - name: dns_admin
   - name: dns_viewer
-  - name: cloud_identity_viewer
-  - name: cloud_resource_viewer
+  - name: role_viewer
```
Carthaca, line 41: you don't need to add these 3 roles here, they are already introduced in other seeds - just make sure to have those covered under `spec:requires` above

PR #1757, opened 2020-07-20
D072896 am rm pd

geisslet:
removal of the old pd instance to avoid notification fails after August 1st - removal of secret keys follows separately .

+ some minor sre routing

### DIFF
--- a/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
+++ b/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
@@ -756,21 +756,6 @@ receivers:
 
   - name: pagerduty_api
     pagerduty_configs:
-      - service_key: {{ required ".Values.pagerduty.api.serviceKey undefined" .Values.pagerduty.api.serviceKey | quote }}
-        description: {{"'{{ template \"pagerduty.sapcc.description\" . }}'"}}
-        component: {{"'{{template \"pagerduty.sapcc.tier\" . }}'"}}
-        group: {{"'{{template \"pagerduty.sapcc.service\" . }}'"}}
-        details:
-          Details: {{"'{{template \"pagerduty.sapcc.details\" . }}'"}}
-          Region: {{"'{{template \"pagerduty.sapcc.region\" . }}'"}}
-          Tier: {{"'{{template \"pagerduty.sapcc.tier\" . }}'"}}
-          Service: {{"'{{template \"pagerduty.sapcc.service\" . }}'"}}
-          Context: {{"'{{template \"pagerduty.sapcc.context\" . }}'"}}
-          Prometheus: {{"'{{template \"pagerduty.sapcc.prometheus\" . }}'"}}
-          Dashboard: {{"'{{template \"pagerduty.sapcc.dashboard\" . }}'"}}
-          Sentry: {{"'{{template \"pagerduty.sapcc.sentry\" . }}'"}}
-          Playbook: {{"'{{template \"pagerduty.sapcc.playbook\" . }}'"}}
-          firing: {{"'{{ template \"pagerduty.sapcc.firing\" . }}'"}}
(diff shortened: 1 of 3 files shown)


global/prometheus-alertmanager-operated/values.yaml
```diff
@@ -28,13 +28,6 @@ slack:
   vmware: {}
 
 # Values for PagerDuty receivers.
-pagerduty:
```
auhlig, line 31: I see `.Values.pagerduty.*` still being used in various places, for instance here: https://github.com/sapcc/helm-charts/blob/2db664cd2ac9e95cfc293a52750567e634139912/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl#L759

global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
```diff
@@ -98,13 +98,13 @@ route:
     match_re:
       tier: os
       severity: info|warning|critical
-      service: arc|backup|barbican|castellum|cinder|cfm|designate|elektra|elk|glance|hermes|ironic|keppel|keystone|limes|lyra|maia|manila|neutron|nova|octavia|sentry|swift
+      service: arc|backup|barbican|castellum|cinder|cfm|designate|elektra|elk|glance|hermes|ironic|keppel|keystone|limes|lyra|maia|manila|neutron|nova($|-.*)|octavia|sentry|swift
```
auhlig, line 101: Why is that needed?

joker-official: So SRE alerts für `nova-api` and `nova-console` and `nova-placement-api` get routed to `#cc-os-nova`, too.

auhlig: I see. Please be explicit in the config making this `nova|nova-api|nova-console|nova-placement-api`.

joker-official: Why? Everything starting with nova- should go to cc-os-nova. Why do we have to search for all possible services and always update this rule if we add one?

auhlig: Because we started with regexes a while ago which proved to become messy over time. Thus I want to keep the configuration explicit.

joker-official: Then I vote for removing this and trying to relabel the "service" to nova in the alerts.

joker-official: Or how about setting the `($|-.*)` at the end, so that it's valid for all services. Would that work for you? So that `arc-api` ingress also gets into the `arc` channel etc.

auhlig: Would also work. You could still use e.g. the `context` label for logical grouping of alerts.

auhlig: Please no regex-ing here anymore. I've spent too many hours chasing alerts.

joker-official: ok then, removal it is. @geisslet could you remove that, please? We will have to put the regexes into the Alert-Definition instead.

global/prometheus-alertmanager-operated/values.yaml
geisslet: Don't get it. 
Of course `.Values.pagerduty_sap.` is still in :: just `.Values.pagerduty.` is removed (and  I can't find any anymore). 
Do I miss something?

auhlig: I though I saw 2 occurrences of the old `.Values.pagerduty.`. Nevermind they seem to be gone

PR #1758, opened 2020-07-22
Jaeger Operator

chuan137:
* deploy Jaeger Operator
* deploy an Jaeger instance with elasticsearch storage
* deploy Jaeger angent as sidecars in cinder service


joker-official:
Could you explain a little more, how this works? If we have a sidecar in cinder, what does the operator do? Who uses the CRD of the operator?

Also please try to fix the helm-charts validation.

openstack/utils/templates/_helpers.tpl
```diff
@@ -60,3 +60,25 @@ enabled = false
...
+      name: jg-binary-trft
+      protocol: UDP
+    - containerPort: 5778
+      name: config-rest
+      protocol: TCP
+    - containerPort: 14271
+      name: admin-http
+      protocol: TCP
+  args: ["--collector.host-port=openstack-jaeger-collector.jaeger-infra.svc:14267"]
+{{- end }}
```
chuan137, line 84: a helper function to define the jaeger agent side car container

chuan137:
Sure. The openstack-jaeger (system/jaeger-operator/templates/openstack-jaeger.yaml) uses the CRD, and is consumed by the operator. The Jaeger application will be deployed together with a few services (collector, agent, UI).

system/jaeger-operator/templates/openstack-jaeger.yaml
```diff
@@ -0,0 +1,22 @@
...
+    tls:
+      - hosts:
+        - tracing.{{ .Values.global.region }}.cloud.sap
+        secretName: tls-tracing-{{ .Values.global.region }}-cloud-sap
+
+  storage:
+    type: elasticsearch
+    options:
+      es:
+        server-urls: http://elasticsearch.hermes.svc:9200
```
chuan137, line 22: This file uses the CRD Jaeger, and will be consumed by operator. A jaeger application will be spawned together with serveral services (collector, agent, UI) and also ingress rules.

openstack/cinder/templates/api-deployment.yaml
```diff
@@ -130,6 +130,7 @@ spec:
               mountPath: /etc/statsd/statsd-exporter.yaml
               subPath: statsd-exporter.yaml
               readOnly: true
+{{- include "jaeger_agent_sidecar" . | indent 8 }}
```
chuan137, line 133: Injecting the Jaeger agent as side car container here.

chuan137: I should test if jaeger should be deployed first

openstack/utils/templates/_helpers.tpl
joker-official: If we have versioning constraints, does that mean we have to update every service using jaeger if we update the jaeger operator immediately?

openstack/utils/templates/_helpers.tpl
```diff
@@ -60,3 +60,25 @@ enabled = false
     {{- $options := merge .Values.osprofiler .Values.global.osprofiler }}
     {{- if $options.enabled }} osprofiler{{ end -}}
 {{- end }}
+
+{{- define "jaeger_agent_sidecar" }}
```
chuan137, line 64: This is a helper function defining the jaeger agent side car container. The version of the Jaeger-agent image must be the same as the Jaeger operator. So I add the jaeger operator as the subchart and import the version number from there.

The agent collects the profiling data, and send them to the openstack-jaeger-collector.

system/jaeger-operator/templates/openstack-jaeger.yaml
joker-official: Does that mean, the jaeger-operator is really supposed to be used on per service basis, but we don't use it that way and thus only need to use the CRD once for a region-global jaeger instance?

system/jaeger-operator/values.yaml
```diff
@@ -0,0 +1,14 @@
+# Default values for jaeger.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+#
+global:
+  region: DEFINED_IN_VALUES_FILE
```
joker-official, line 6: This is an anti-pattern. Please don't define default-values that are not usable. Rather comment it out.

system/jaeger-operator/templates/openstack-jaeger.yaml
joker-official: What I mean is: What's the point of having an operator, if it only ever spawns 1 pod + 1 ingress, because nobody else uses it? Wouldn't it make more sense to just put those 1 objects in the helm-chart manually and thus let the user see what we're doing?

Are we planning on using it further in the future and this is just a simplification for the PoC?

openstack/cinder/requirements.yaml
```diff
@@ -23,3 +23,9 @@ dependencies:
   - name: region_check
     repository: https://charts.eu-de-2.cloud.sap
     version: 0.1.1
+  - name: jaeger
+    repository: file://../../system/jaeger-operator
```
joker-official, line 27: Should we use charts.eu-de-2.cloud.sap here? Or do we all have to use the same version anyways, because otherwise jaeger doesn't work anymore?

chuan137: yes, the chart repository should be used. I will fix it once we are done with the first round test.

system/jaeger-operator/values.yaml
chuan137: I added this to fix the linting error

> executing "jaeger/templates/openstack-jaeger.yaml" at <.Values.global.region>: nil pointer evaluating interface {}.region

how to fix it otherwise?

joker-official: Where do you get that error? In a local `helm template` run? In the automatic helm-charts tests? In your attempt to deploy the chart?

system/jaeger-operator/templates/openstack-jaeger.yaml
chuan137: @joker-official jaeger-operator and jaeger are different. I think only one operator is needed, since it can watch all the namespaces. How many jaegers CRD are needed, I don't have an answer now. I would say starting with one per namespace, therefore a openstack-jaeger CRD in this PR.

I think there are benefits using operator. It's easy to setup and upgrade. We can simply change the version number and without worring the changes in the external tool. Operators are extensively used in our environment already, in my experiences, it is a plus to hide the details from the end user if they will only use it but never change it. 

And It looks simple, because is a minimal setup at the moment. It could expand to more complex deployment if we need the storage supports. Even at the moment it is not very simple. The operator introduces the roles and cluserroles, which is quite nice. The services are deployed besides the pods.

system/jaeger-operator/values.yaml
chuan137: the error is from the pull request validation. https://ci.eu-de-2.cloud.sap/builds/396813

system/jaeger-operator/templates/openstack-jaeger.yaml
joker-official: I would have assumed, we need one CRD per service using jaeger, but I don't know how jaeger looks like and how it presents the different traces.

I get that it looks easy to use an operator. But using an operator also means less control as there's additional code running with additional assumptions. If we want to know what it does and why, we have to read it. That being said, I'm not against using the operator, I just wanted to know your thoughts on this topic.

openstack/utils/templates/_helpers.tpl
joker-official: What happens if the agent doesn't have the same version number anymore? Does it crash? Will it bring down cinder-api, because one of the containers crashed?
It's fine, if you don't have answers, yet, but please figure this out.

system/jaeger-operator/values.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+jaeger:
+  version: 1.18.0
+
+ingress:
+  enabled: true
+  hosts:
+    - xxx
```
joker-official, line 14: What is that `xxx` for? Can we have a comment here, so people can understand what we're setting?

system/jaeger-operator/values.yaml
joker-official: Ok, then you need to provide an extra values file for the CI to use for validation. [Here](https://github.com/sapcc/helm-charts/blob/master/openstack/nova/ci/test-values.yaml) is an example from nova. It uses any `.yaml` file in the chart's `ci` subdirectory.

system/jaeger-operator/values.yaml
joker-official: If this is also just a default-value because the helm-charts validation CI needs it, then please remove it in favor of setting explicit values in the `ci/` sub directory of the `jaeger-operator` chart.

system/jaeger-operator/templates/openstack-jaeger.yaml
chuan137: > I would have assumed, we need one CRD per service using jaeger, but I don't know how jaeger looks like and how it presents the different traces.

this needs to find out during this PoC, if the traces from different services play well together. 
> 
> I get that it looks easy to use an operator. But using an operator also means less control as there's additional code running with additional assumptions. If we want to know what it does and why, we have to read it. That being said, I'm not against using the operator, I just wanted to know your thoughts on this topic.

yes we have to understand what the operator does. In this sense, this PR is a PoC. I will explore different options and expect adjustment in the settings. My impression is that the operator is actively developed and pretty good in fine controlling the deployment. I think there is a good chance we can use the operator in production. 

Unfortunately, as you said, it is less transparent and makes review harder. To summarize the current setting: the CRD openstack-jaeger defines the ingress rule, and for the storage the elasticsearch befind kibana is re-used. The operator spawns a all-in-one pod, which I assume could be split up. In addition, as you know it already, a jaeger agent container must be injected into each service pod.

I would like to see this simplest setting can collect data from cinder, and would like to play around the deployment strategy.

system/jaeger-operator/values.yaml
chuan137: these are the dummy values referenced in the default NOTES.txt...I have updated the NOTES, so that they are removed as well. (commit 7d10f69)

system/jaeger-operator/values.yaml
chuan137: thx, fixed by values in ci/

openstack/utils/templates/_helpers.tpl
chuan137: good point, I don't know the answer yet, will figure it out.

hemna: what happens when the agent is down?

chuan137: To summarize the questions on the sidecar.

I have not found any document on the behavior of agent. My bet is some warning or error message spit out, either from the agent when posting the message to collector, or from the collector side when parsing the message. I need to spend more time to find out how jaeger behaves. 

But what if the agent really crashes due to some reason? In this case, a custom liveness probe should be set. For cinder api, it is set to 

> Liveness:   http-get http://:cinder-api/ delay=15s timeout=5s period=10s #success=1 #failure=3

K8s would not restart the pod when the agent crashes, but only when cinder api fails. There is already another sidecar in the cinder-api pod, the statsd exporter.

PR #1780, opened 2020-08-05
[nova] Alert on task_state no vm_state

joker-official:
vm_state for an instance stuck in "deleting" is "active". Instead, we
have to alert on an instance having the task_state "deleting" for too
long.


joker-official:
Thanks for looking at it.

PR #1784, opened 2020-08-06
[cinder] Update the expression for low overcommit warning

hemna:
This patch updates the expression for tracking the amount of space
left before cinder reaches the max overcommit ratio.  The new
epxression matches when there is less than 10% overcommit ratio left
and shows the amount of space in TiB that equates to.

### DIFF
--- a/prometheus-exporters/openstack-exporter/alerts/cinder.alerts
+++ b/prometheus-exporters/openstack-exporter/alerts/cinder.alerts
@@ -3,14 +3,14 @@ groups:
   rules:
   - alert: CinderBackendShardLowOvercommitWarning
     expr: >
-      sum(cinder_max_oversubscription_ratio - cinder_overcommit_ratio) by (backend, shard) < .5
+      sum(cinder_free_until_overcommit_gib) by (region, shard, backend) / 1024 + sum(cinder_free_until_overcommit_gib / cinder_total_capacity_gib * 100 < 10) by (region, shard, backend) * 0
     for: 15m
     labels:
       severity: info
       tier: os
       service: cinder
       context: "openstack-exporter"
-      meta: "Cinder backend {{ $labels.shard }}/{{ $labels.backend }} max over commit is almost reached (< 0.5). Shard needs to be increased."
+      meta: "Cinder backend {{ $labels.shard }}/{{ $labels.backend }} has less than 10% storage left before overcommit is reached."
     annotations:
-      description: "Cinder backend {{ $labels.shard }}/{{ $labels.backend }} max over commit is almost reached (< 0.5). Shard needs to be increased."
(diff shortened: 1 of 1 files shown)


prometheus-exporters/openstack-exporter/alerts/cinder.alerts
```diff
@@ -3,14 +3,14 @@ groups:
   rules:
   - alert: CinderBackendShardLowOvercommitWarning
     expr: >
-      sum(cinder_max_oversubscription_ratio - cinder_overcommit_ratio) by (backend, shard) < .5
+      sum(cinder_free_until_overcommit_gib) by (region, shard) / 1024 + sum(cinder_free_until_overcommit_gib / cinder_total_capacity_gib * 100 < 10) by (region, shard) * 0
```
joker-official, line 6: I think this needs a `backend` in the `by (region, shard)`, too.

hemna: ok updated

prometheus-exporters/openstack-exporter/alerts/cinder.alerts
```diff
@@ -3,14 +3,14 @@ groups:
...
       tier: os
       service: cinder
       context: "openstack-exporter"
-      meta: "Cinder backend {{ $labels.shard }}/{{ $labels.backend }} max over commit is almost reached (< 0.5). Shard needs to be increased."
+      meta: "Cinder backend {{ $labels.shard }}/{{ $labels.backend }} has less than 10% storage left {{ $value }} TiB before overcommit is reached."
     annotations:
-      description: "Cinder backend {{ $labels.shard }}/{{ $labels.backend }} max over commit is almost reached (< 0.5). Shard needs to be increased."
-      summary: "Cinder backend {{ $labels.shard }}/{{ $labels.backend }} max over commit is almost reached (< 0.5). Shard needs to be increased."
+      description: "Cinder backend {{ $labels.shard }}/{{ $labels.backend }} has less than 10% storage left {{ $value }} TiB before overcommit is reached."
+      summary: "Cinder backend {{ $labels.shard }}/{{ $labels.backend }} has less than 10% storage left {{ $value }} TiB before overcommit is reached."
```
joker-official, line 16: @viennaa One of those is used for grouping alerts, iirc, so we should not include the actual value there. Otherwise, the alert might resolve and re-fire all the time the value changes a little. Do you known which one is used?

hemna: should I remove {{ $value }} ?

viennaa: Never use $value please

viennaa: > @viennaa Do you known which one is used?

Unfortunately not, would be a question for @geisslet

hemna: ok removed {{ $value }}

PR #1785, opened 2020-08-10
[keystone] use variables for resource naming; dynamic requirements

galkindmitrii:
in templates; bump chart version and requirements.

### DIFF
--- a/openstack/keystone/requirements.yaml
+++ b/openstack/keystone/requirements.yaml
@@ -2,7 +2,7 @@ dependencies:
   - condition: percona_cluster.enabled
     name: percona_cluster
     repository: https://charts.eu-de-2.cloud.sap
-    version: 1.0.5
+    version: 1.0.6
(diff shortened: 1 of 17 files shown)


openstack/keystone/requirements.yaml
```diff
@@ -2,15 +2,16 @@ dependencies:
...
+    version: 1.0.6
   - condition: mariadb.enabled
     name: mariadb
     repository: https://charts.eu-de-2.cloud.sap
     version: 0.1.16
   - name: memcached
     repository: https://charts.eu-de-2.cloud.sap
     version: 0.0.3
-  - name: mysql_metrics
+  - condition: mysql_metrics.enabled
```
Carthaca, line 13: Does this require a respective value in values.yaml?

galkindmitrii: you're absolutely right, thanks Maurice!

PR #1789, opened 2020-08-12
[Cinder][Alert] Update low overcommit to 20%

hemna:
This patch increases the low overcommit ration from 10% to 20%
to help us catch the warnings earlier.


hemna:
I spoke with David about the alerting and he recommended setting it to 20% just to give us a heads up earlier about smaller vcenter deployments, such as eu-de-2 vc-b-1, which only had 16G left.   I'm guessing we can move this back to 15% at some point.

joker-official:
`vc-b-1` was alerting since you created the alert, iirc. We will probably have to work with silences then. Otherwise, I fear we will be overwhelmed by the number of alerts "we can ignore" and will ignore/oversee important ones, too.

PR #1802, opened 2020-08-17
system/absent-metrics-operator: use Docker repo

talal:


system/absent-metrics-operator/templates/_utils.tpl
```diff
@@ -1,7 +1,3 @@
-{{- define "absent_metrics_operator_image" -}}
-  {{- if contains "DEFINED" $.Values.absent_metrics_operator.image_tag -}}
-    {{ required "This release should be installed by the deployment pipeline!" "" }}
-  {{- else -}}
-    {{$.Values.global.registry}}/absent-metrics-operator:{{$.Values.absent_metrics_operator.image_tag}}
-  {{- end -}}
+{{- define "image" -}}
+{{- required ".Values.image.repository missing" .Values.image.repository -}}:{{- required ".Values.image.tag missing" .Values.image.tag -}}
```
majewsky, line 2: Since these variables have default values set in values.yaml, I don't see the need for the `required` checks here. In fact, the whole `_utils.tpl` can just be scraped if it just substitutes `{{.Values.image.repository}}:{{.Values.image.tag}}`.

system/absent-metrics-operator/ci/test-values.yaml
```diff
@@ -3,5 +3,6 @@ global:
   region: xx-yy-1
   tld: example.org
 
-absent_metrics_operator:
-  image_tag: "12345"
+image:
+  repository: org/repo
+  tag: "12345"
```
majewsky, line 8: This can just use the default values for `.Values.image`.

PR #1805, opened 2020-08-17
[infra-monitoring]: Fixed metric for vasa filesystem alert

christopherhans:
Used wrong metric (percentage instead of percent).


christopherhans:
@viennaa

PR #1808, opened 2020-08-18
Add Prometheus Probes

defo89:


auhlig:
LGTM.
But creating/updating charts and using them needs to be splitted to give them time to be pushed to the chart registry

PR #1820, opened 2020-08-21
openstack/maia: add service label to PromRule

talal:
This will ensure that the corresponding absent alerts will _always_ get routed to the appropriate Slack channel. More info [here](https://github.com/sapcc/absent-metrics-operator/blob/master/doc/playbook.md#tier-and-service-labels).

### DIFF
--- a/openstack/maia/templates/prometheus-alerts.yaml
+++ b/openstack/maia/templates/prometheus-alerts.yaml
@@ -10,6 +10,7 @@ metadata:
   labels:
     app: maia
     tier: os
+    service: maia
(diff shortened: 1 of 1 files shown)


auhlig (approved):

auhlig:
To be clear, this only sets the label on kubernetes level allowing `k get prometheusrule -l service=maia` and does *not* modify the alerts itself.

talal:
The absent metrics operator falls back to these k8s-level labels if the `tier` and `service` can't be determined at the alert-rule level.

PR #1825, opened 2020-08-25
[keystone] deploy seeds if it is 1st install; fix seed naming; allow to

galkindmitrii:
specify both public/internal external service IPs; bump chart.

### DIFF
--- a/openstack/keystone/templates/seed.yaml
+++ b/openstack/keystone/templates/seed.yaml
@@ -1,8 +1,7 @@
-{{- if .Release.IsUpgrade }}
 apiVersion: "openstack.stable.sap.cc/v1"
 kind: "OpenstackSeed"
 metadata:
-  name: keystone-seed
+  name: {{ .Release.Name }}-seed
   labels:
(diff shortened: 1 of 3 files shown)


openstack/keystone/templates/seed.yaml
```diff
@@ -1,8 +1,7 @@
-{{- if .Release.IsUpgrade }}
```
Carthaca, line 1: In an 'ordinary' keystone installation the seeds won't work on install - but it's okay I guess, the seeder will re-try to apply the seeds.

Carthaca (approved):
lgtm

openstack/keystone/templates/seed.yaml
galkindmitrii: yes, correct. The problem is that now we have to run upgrade / 2nd deployment before the seeds even appear. Looking at other services, almost nobody seem to have this clause right now.

galkindmitrii:
Merci

PR #1842, opened 2020-09-02
Add absent metrics operator to all cluster types

talal:
/cc @BugRoger @databus23 

This PR only adds the operator. I will enable the operator on a regional basis as per the bronzer, silver, and gold tier pattern.


auhlig (approved):
Bring it on!

PR #1873, opened 2020-09-17
[cinder] Upgrade the low overcommit alert to warning

hemna:
This patch updates the cinder low overcommit ration alert from
info to warning.  Also adds the link to the playbook.


geisslet (approved):
yep!

joker-official (approved):

PR #1874, opened 2020-09-18
[keystone] added image puller ds for keppel

galkindmitrii:
similar to what keppel has for postgres.
technically only db readiness image is stored in keppel, but theoretically db restarts will be a bit faster if mariadb and metrics are pre-pulled on all nodes.


majewsky:
Looks good as far as I can see. The Keystone images should also be added, at least the bare minimum required to get auth to work (i.e. keystone-api, possibly keystone-rabbitmq, maybe the migration job?).

Carthaca (changes requested):
Hi Dmitry, thanks for this addition. I found just some copy & paste errors, fixing this should make the ci green, too, I hope

Carthaca (approved):
Thank you 👍

PR #1881, opened 2020-09-23
[Cinder] Add critical alert for overcommit <10% left

hemna:
This patch adds a new critical alert for cinder pool storage
overcommit ratio reaching 10% or less leftover.


joker-official (changes requested):
We're not allowed to implement it as critical immediately. We're supposed to start as info and upgrade to warning and critical later on with confirmation from Martin V.

hemna:
updated it to info

joker-official (approved):

PR #1882, opened 2020-09-25
[SRE] Push alerts to more service channels

joker-official:
Description in the commit messages.


geisslet:
@dhoeller agreed on the updated routing - global dashboard will be added later, bzs it needs a new label in the slack template - will be separate PR

PR #1884, opened 2020-09-25
Fix hermes playbook link

B1Sandmann:
The old link is wrong the new one is working
It also seems that `{{ $labels.service }} {{ $labels.check }}`  are empty but I don't know what should be there and what should be there instead.


B1Sandmann:
Here is an example where the playbook link is wrong and one can see that the labels are empty: https://convergedcloud.slack.com/archives/CFH0SEAGK/p1601024222025000

B1Sandmann:
fyi @notque

B1Sandmann:
The fail seems to be issues that are not related to this change but I think @notque can  judge this better.

notque (approved):
LGTM, looks like something changed in one of the moves. I'll need to go over it again.

PR #1887, opened 2020-09-29
[vrops-exporter] adapt deployment file

richardtief:


prometheus-exporters/vrops-exporter/templates/vrops-exporter-deployment.yaml
```diff
@@ -29,6 +36,12 @@ spec:
            args:
              - -m
              - /config/collector_config.yaml
+             - -t
+             - {{ required "$target.hostname missing" $target.hostname }}
```
richardtief, line 40: is "required" reasonable for the default case?

viennaa: yes, in case it is not defined

viennaa (approved):

PR #1889, opened 2020-09-29
allow zone transfers to nsi name servers

braunmatt:


system/bind/templates/config.yaml
```diff
@@ -25,6 +33,9 @@ data:
      };
      include "/etc/bind/rndc.key";
      include "/etc/bind/tsig.key";
+     {{- if .Values.nsi_key }}
+     include "/etc/bind/keys/nsi.key";
```
galkindmitrii, line 37: ConfigMap will be mounted into /etc/bind: https://github.com/sapcc/helm-charts/blob/master/system/bind/templates/deployment.yaml#L54-L55 , so path has to be changed

galkindmitrii (changes requested):

braunmatt (commented):
correct path

galkindmitrii (approved):
lgtm

PR #1890, opened 2020-09-29
[Keystone] switch vice-president to kubernetes.io/tls-acme

rajivmucheli:
Hi @Carthaca 

Please validate this PR.

Regards,
Rajiv

### DIFF
--- a/openstack/keystone/templates/ingress-api.yaml
+++ b/openstack/keystone/templates/ingress-api.yaml
@@ -16,7 +16,7 @@ data:
   ca.crt: {{ .Values.services.ingress.x509.ca | b64enc }}
 ---
 {{- end }}
-{{- if not .Values.services.ingress.vice_president }}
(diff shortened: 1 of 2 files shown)


openstack/keystone/templates/ingress-api.yaml
```diff
@@ -66,7 +66,7 @@ metadata:
     ingress.kubernetes.io/auth-tls-verify-depth: "3"
     {{- end }}
     {{- if .Values.services.ingress.vice_president }}
```
Carthaca, line 68: To reduce confusion, this key `.Values.services.ingress.vice_president` should be renamed, too

Carthaca (changes requested):

openstack/keystone/templates/ingress-api.yaml
Carthaca: same for line 73 below

Carthaca (approved):
lgtm

PR #1909, opened 2020-10-02
[nova] prepare MariaDB migration

galkindmitrii:


openstack/nova/templates/online-migration-job.yaml
```diff
@@ -28,7 +28,11 @@ spec:
         - name: DEPENDENCY_JOBS
           value: "nova-migration"
         - name: DEPENDENCY_SERVICE
-          value: "nova-postgresql"
+        {{- if eq .Values.postgresql.enabled true }}
+          value: "{{ .Release.Name }}-postgresql"
+        {{- else }}
+          value: "{{ .Release.Name }}-mariadb,{{ .Release.Name }}-api-mariadb"
```
joker-official, line 34: While we're at it, can we add cell2, if it's enabled, too?

joker-official (approved):
I looks ok to me. Can't think of an additional problem, from the top of my head. Just one remark, we might as well fix right now.

PR #1910, opened 2020-10-05
[prometheus-infra] added static targets

richardtief:


viennaa (approved):
approved until we have a better solution

PR #1930, opened 2020-10-08
[nova] run nova-manage placement sync_aggregates in online-migration

mariusleu:


joker-official (approved):

joker-official (changes requested):
This should only run when we're on rocky. So we either need to get the version somehow (currently no idea - maybe some secrets value or from an image?) or we need to check for the availability of the command.

openstack/nova/templates/bin/_db-online-migrate.tpl
```diff
@@ -1,7 +1,13 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 set -x
 set -e
 
-nova-manage --config-file /etc/nova/nova.conf db online_data_migrations
+nova_manage="nova-manage --config-file /etc/nova/nova.conf"
+available_commands_text=$(nova-manage --help | awk '/Command categories/ {getline; print $0}')
```
joker-official, line 7: Does `getline` read the next line?

openstack/nova/templates/bin/_db-online-migrate.tpl
```diff
@@ -1,7 +1,13 @@
...
 
+$nova_manage db online_data_migrations
+
+if [[ $available_commands_text == *"placement"*  ]]; then
```
joker-official, line 11: I don't feel like that's very accurate. Can we do something like
```
if echo "${available_commands_text}" | grep -q -E '[{,]placement[},]'; then
```
```
$ text='{version,bash-completion,network,cell_v2,db,cell,floating,api_db}'
$ echo "${text}" | grep -q -E '[{,]placement[},]' && echo "success"
$ text='{version,bash-completion,network,cell_v2,db,cell,floating,api_db,placement}'
$ echo "${text}" | grep -q -E '[{,]placement[},]' && echo "success"
success
```

Then we can also switch back to using `#!/bin/sh`.

openstack/nova/templates/bin/_db-online-migrate.tpl
mariusleu: Yes.

joker-official (approved):

PR #1934, opened 2020-10-08
[cinder] update cinder requirements for keppel

hemna:
This patch updates some of cinder's dependencies to versions
that use keppel instead of quay.

### DIFF
--- a/openstack/cinder/requirements.lock
+++ b/openstack/cinder/requirements.lock
@@ -4,24 +4,24 @@ dependencies:
   version: 0.1.1
 - name: mariadb
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.1.20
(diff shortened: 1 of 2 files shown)


joker-official (changes requested):
Please run a `helm template`, too. Because you also need to update the requirements.lock, which is done by that.

majewsky (approved):
The `requirements.yaml/lock` look good to me. I had a brief look at the `helm diff` for qa-de-1 and saw the expected image changes, so that looks plausible as well. The `helm diff` is quite large actually, esp. because some subcharts do the appVersion change for Deployments etc., so I recommend to take a close look at the diffs when deploying. Other than that, this is good to go from my PoV.

joker-official (approved):

PR #1937, opened 2020-10-09
CNMP: Big root disk sized flavor for HANA nodes

reimannf:
For kubernikus clusters running node pools in the CNMP context
there is the need of huge flavors for HANA nodes. The bigger root
disk is needed to pull container images carrying the hana content.

It is made explicit private.

### DIFF
--- a/openstack/sap-seeds/templates/flavor-seed.yaml
+++ b/openstack/sap-seeds/templates/flavor-seed.yaml
@@ -312,6 +312,15 @@ spec:
     extra_specs:
       "vmware:hv_enabled": "True"
       "hw_video:ram_max_mb": "16"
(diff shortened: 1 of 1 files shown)


frr2018 (commented):
The requested flavor is not numa aligned. Because most of the HVs in EU-NL-1 are 2TB RAM nodes, i would recommend the following sizes:

970GB | 84
or
1450GB | 126

This fits to the Numa Setup of these nodes. Please create it as a private flavor on your own in the project. I think it doesnt make sense to create this in the flavor seed.

Thanks and bye,

Frank

reimannf:
ok. we go with the first option and a not seeded private flavor in eu-nl-1.

PR #1944, opened 2020-10-13
Update netapp-capacity.alerts

killermoehre:
Use the right unit prefix for the value, else it's off by three magnitudes.


Carthaca (approved):
👍

PR #1949, opened 2020-10-14
[cinder] reduce the workers per pod, but increase replicas;

galkindmitrii:
also adjust max_connections & connect_timeout for mariadb and pool_size for sqlalchemy;

### DIFF
--- a/openstack/cinder/values.yaml
+++ b/openstack/cinder/values.yaml
@@ -19,7 +19,7 @@ osprofiler:
 
 pod:
   replicas:
-    api: 2
+    api: 4
     scheduler: 1
   lifecycle:
     upgrades:
@@ -33,7 +33,7 @@ pod:
     api:
       requests:
         cpu: "1500m"
-        memory: "5Gi"
+        memory: "3Gi"
     scheduler:
       requests:
         cpu: "100m"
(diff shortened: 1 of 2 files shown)


openstack/cinder/values.yaml
```diff
@@ -32,8 +32,8 @@ pod:
   resources:
     api:
       requests:
-        cpu: "1500m"
-        memory: "5Gi"
+        cpu: "1000m"
```
joker-official, line 35: Are you sure it will use less than 1,5 CPUs now? I'd think it uses the same, just less memory.

galkindmitrii: hmm, I'd actually expect it to use less cpu will fewer workers.
But those are not the limits, rather the requests for scheduler for a better placement, so it will not cap even if real usage goes to 2+ cpu.

joker-official: yeah, still. I expect it didn't use all the workers before and will not use now. So the usage should stay stable.
If anything, the number might be reduced with a bigger number of pods, because requests are spread to more pods.

galkindmitrii: Ok, back to 1500

joker-official (approved):

PR #1955, opened 2020-10-19
[designate][seed] add core_dns_registry

Carthaca:
with designate cloud read permissions, requested for Core DNS Team

coreDnsRegistryServicePassword password needs to be set


galkindmitrii (approved):
lgtm, thanks

PR #1957, opened 2020-10-20
Add asr1k config map hash to neutron deployments

sebageek:
neutron-server and the asr1k agents should be restarted on a configmap
change, hence we need to include the configmap hash.


notandy:
👍

PR #1963, opened 2020-10-21
Nova cleanup deprecations

mariusleu:
Deprecation warnings found in Rocky are cleaned up here.

### DIFF
--- a/openstack/nova/templates/etc/_nova-scheduler.conf.tpl
+++ b/openstack/nova/templates/etc/_nova-scheduler.conf.tpl
@@ -1,17 +1,17 @@
 [DEFAULT]
 # Scheduling
-scheduler_driver_task_period = {{ .Values.scheduler.driver_task_period | default 60 }}
 scheduler_host_manager = {{ if .Values.global.hypervisors_ironic }}ironic_host_manager{{ else }}host_manager{{ end }}
 scheduler_driver = {{ .Values.scheduler.driver }}
 
 [scheduler]
 discover_hosts_in_cells_interval = 60
 workers = {{ .Values.scheduler.workers | default 1 }}
+driver_task_period = {{ .Values.scheduler.driver_task_period | default 60 }}
 
 [filter_scheduler]
 available_filters = {{ .Values.scheduler.available_filters | default "nova.scheduler.filters.all_filters" }}
 default_filters = {{ .Values.scheduler.default_filters}}
-tracks_instance_changes = {{ .Values.scheduler.tracks_instance_changes }}
+tracks_instance_changes = {{ .Values.scheduler.scheduler_tracks_instance_changes }}
 bigvm_host_size_filter_uses_flavor_extra_specs = true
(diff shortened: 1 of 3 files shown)


openstack/nova/templates/etc/_nova.conf.tpl
```diff
@@ -28,19 +24,19 @@ default_availability_zone = {{.Values.global.default_availability_zone}}
...
 
-wsgi_default_pool_size = {{ .Values.wsgi_default_pool_size | default .Values.global.wsgi_default_pool_size | default 100 }}
 sync_power_state_pool_size = {{ .Values.sync_power_state_pool_size | default 500 }}
 sync_power_state_interval = {{ .Values.sync_power_state_interval | default 1200 }}
 sync_power_state_unexpected_call_stop = false
 
 prepare_empty_host_for_spawning_interval = 600
 
-dhcp_domain = openstack.{{ required ".Values.global.region is missing" .Values.global.region }}.{{ required ".Values.global.tld is missing" .Values.global.tld }}
-
```
joker-official, line 39: I think we still need that and it was undeprecated in a later release, because that's how we set the domain-name.

openstack/nova/templates/etc/_nova-scheduler.conf.tpl
```diff
@@ -1,23 +1,23 @@
...
 
 [scheduler]
 discover_hosts_in_cells_interval = 60
 workers = {{ .Values.scheduler.workers | default 1 }}
+driver_task_period = {{ .Values.scheduler.driver_task_period | default 60 }}
 
 [filter_scheduler]
+available_filters = {{ .Values.scheduler.available_filters | default "nova.scheduler.filters.all_filters" }}
+default_filters = {{ .Values.scheduler.default_filters}}
+tracks_instance_changes = {{ .Values.scheduler.tracks_instance_changes }}
```
joker-official, line 14: This was `.Values.scheduler.scheduler_tracks_instance_changes` before, but I see now values file change.

joker-official (changes requested):
We need to squash those 2 commits latest at merge.

openstack/nova/templates/etc/_nova.conf.tpl
mariusleu: OK.
Seems it gets back in Train under [api].

https://github.com/openstack/nova/blob/886b0a5d748ae1deda3a039734f831d7c0cf0476/releasenotes/notes/undeprecate-dhcp_domain-opt-77c9154c5b06e0ff.yaml#L2-L8

openstack/nova/templates/etc/_nova-scheduler.conf.tpl
mariusleu: Updated the value.

joker-official (approved):

grandchild:
@mariusleu @joker-official is anyone of you still holding out on merging this for a reason?

PR #1969, opened 2020-10-22
[Cinder]  Bump the mariadb version to 0.2.7

hemna:
Update the mariadb chart version to 0.2.7


joker-official (approved):

joker-official (changes requested):
Needs an update to requirements.lock, too.

joker-official (approved):

PR #1974, opened 2020-10-27
[cronus]: fix region variable path

kayrus:


dhalimi (commented):
It requires a change in cc/secrets/.../cronus.yaml and to add a logic

dhalimi (approved):

PR #1979, opened 2020-10-28
Treat shoot networks in a special way

killermoehre:
Shoot networks are some kind of broken. As the alternative would be to silence them every night, I propose to make them an `warning` instead. So the support only reacts to them during the business hours.


geisslet:
lgtm - double sensitivity could be helpfull, but not sure though - let's see.
will merge later

PR #1983, opened 2020-10-29
bring back the resource types

BerndKue:


openstack/ironic/templates/seed.yaml
```diff
@@ -101,7 +101,7 @@ spec:
       "capabilities:cpu_arch": "x86_64"
       "quota:separate": "true"
       "quota:instance_only": "true"
-#      "resources:CUSTOM_ZG1BCM1_MEDIUM": "1"
+      "resources:CUSTOM_ZG1BCM1_MEDIUM": "1"
```
joker-official, line 104: This is the same as above. It shouldn't be, shouldn't it? There seems to be a `1` missing.

openstack/ironic/templates/seed.yaml
```diff
@@ -134,7 +134,7 @@ spec:
       "capabilities:cpu_arch": "x86_64"
```
joker-official, line 134: above, there's still one commented. why?

joker-official (changes requested):

BerndKue:
Well spotted. Thanks for finding the two mistakes.

joker-official (approved):

PR #1986, opened 2020-10-29
[cinder] Add CinderBackendShardLowFreeSpaceWarning

hemna:
This patch adds new CinderBackendShardLowFreeSpaceWarning
alert.  This alerts when the amount of free space is less
than 25% remaning.  This should cover both thin provisioned
and thick provisioned backends.


viennaa (approved):

viennaa:
Looks good, you can merge, if you're ready.

PR #2053, opened 2020-11-06
[cronus]: add cronus cloud role and assign it to limes

kayrus:


openstack/cronus/templates/cronus/_config.yaml.tpl
```diff
@@ -32,7 +32,7 @@ cronus:
   policy:
     project_scope: project_id:%(project_id)s and (role:email_admin or role:email_user)
     domain_scope: domain_id:%(target.project.domain.id)s and (role:email_admin or role:email_user)
-    cronus:usage_viewer: rule:project_scope or rule:domain_scope
+    cronus:usage_viewer: rule:project_scope or rule:domain_scope or role:cloud_email_admin
```
majewsky, line 35: Use the `resource_service` role which is the designated role for the Limes service user. Then you don't need to change the Limes seed.

```suggestion
    cronus:usage_viewer: rule:project_scope or rule:domain_scope or role:cloud_email_admin or role:resource_service
```

kayrus: done

majewsky (approved):

PR #2056, opened 2020-11-10
[nova] Increase http_retries for cinder

joker-official:
If we fail to connect to cinder, we might get into an invalid state the
nanny has to fix. Let's retry a couple more times to try and keep these
inconsistencies to a minimum. Default is 3.

The specific inconsistency we've seen is that a volume was set to state
"detaching" by nova-api, but then nova-compute failed to actually get
the volume from cinder and thus couldn't update or roll back the state.


joker-official:
Maybe we should do the same thing for neutron?

grandchild (approved):

grandchild:
> Maybe we should do the same thing for neutron?

I don't know enough about nova-to-neutron behavior to answer that...

PR #2061, opened 2020-11-11
[nova] switch to projected volumes

mariusleu:


joker-official:
This commit needs much more info. At least the JIRA issue number, but probably even better something from fe6e986d

Also please rebase.

openstack/nova/templates/api-deployment.yaml
```diff
@@ -90,58 +90,10 @@ spec:
...
-              readOnly: true
-            {{- if .Values.watcher.enabled }}
             - name: nova-etc
-              mountPath: /etc/nova/watcher.yaml
-              subPath: watcher.yaml
-              readOnly: true
-            {{- end }}
+              mountPath: /etc/nova
+            - name: nova-bin
+              mountPath: /var/lib/kolla/venv/bin
```
joker-official, line 96: Did you confirm this works? Is everything else still there in `/var/lib/kolla/venv/bin`?

openstack/nova/templates/api-deployment.yaml
```diff
@@ -153,17 +105,57 @@ spec:
...
-          configMap:
-            name: nova-bin
+          projected:
             defaultMode: 0755
+            sources:
+              - configMap:
+                  items:
+                    - key: iptables-restore.mock
+                      path: iptables-restore
+                  name: nova-etc
```
joker-official, line 152: Should be `nova-bin`.

openstack/nova/templates/api-metadata-deployment.yaml
```diff
@@ -141,17 +99,53 @@ spec:
...
-          configMap:
-            name: nova-etc
+          projected:
+            defaultMode: 420
+            sources:
+              - configMap:
+                  items:
+                    - key: nova.conf
+                      path: nova.conf
+                    - key: nova-api.conf
```
joker-official, line 113: Should be `nova-api-metadata.conf`

openstack/nova/templates/api-metadata-deployment.yaml
```diff
@@ -141,17 +99,53 @@ spec:
...
-          configMap:
-            name: nova-bin
+          projected:
             defaultMode: 0755
+            sources:
+              - configMap:
+                  items:
+                    - key: iptables-restore.mock
+                      path: iptables-restore
+                  name: nova-etc
```
joker-official, line 142: Should be `nova-bin`.

openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml
```diff
@@ -71,39 +71,9 @@ template: |
...
-          - mountPath: /etc/nova/rootwrap.conf
-            name: nova-etc
-            subPath: rootwrap.conf
-            readOnly: true
-          - mountPath: /etc/nova/rootwrap.d/compute.filters
-            name: nova-etc
-            subPath: compute.filters
-            readOnly: true
+          - mountPath: /etc
+            name: sudoers
```
joker-official, line 76: Doing this, we mount one configmap over another again, as we mount this to `/etc` and then `nova-etc` to `/etc/nova`. We don't want that. Can we find another solution? Maybe mount the whole `/etc` because of `sudoers` and prepend all paths for Nova's config files with `nova/`?

openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml
```diff
@@ -172,27 +118,75 @@ template: |
...
+                  path: hostgroup-reservations.json
+                name: nova-compute-vmware-{= name =}
+        - name: sudoers
+          projected:
+            defaultMode: 420
+            sources:
+            - configMap:
+                items:
+                - key: sudoers
+                  path: sudoers
```
joker-official, line 155: `name:` is missing, I think. That way, it doesn't know where to project from.

openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml
```diff
@@ -172,27 +118,75 @@ template: |
...
         - name: hypervisor-config
           configMap:
             name: nova-compute-vmware-{= name =}
```
joker-official, line 158: Why do we keep this?

openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml
```diff
@@ -172,27 +118,75 @@ template: |
...
+            defaultMode: 420
+            sources:
+            - configMap:
+                items:
+                - key: neutron.conf
+                  path: neutron.conf
+                - key: api-paste.ini
+                  path: api-paste.ini
+                - key: policy.json
+                  path: neutron-policy.json
```
joker-official, line 171: Needs to be switched. See https://github.com/sapcc/helm-charts/blob/2ee3d7ef283269da28ed769fc20d8d503548ffa5/openstack/neutron/templates/vct-nsxv3-agent-deployment.yaml#L110-L111

openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml
```diff
@@ -172,27 +118,75 @@ template: |
...
+                - key: logging.conf
+                  path: logging.conf
+                - key: ml2-conf.ini
+                  path: plugins/ml2/ml2-conf.ini
+                name: neutron-etc
+            - configMap:
+                items:
+                - key: ml2-vmware.ini
+                  path: plugins/ml2/ml2-conf.ini
```
joker-official, line 180: Should be `plugins/ml2/ml2-vmware.ini`

joker-official (changes requested):

openstack/nova/templates/api-deployment.yaml
mariusleu: It seems that here we need to keep using `subPath` otherwise the `bin` directory will get overwritten. I think there's no problem with that, since this is the only place where nova-bin is being used. Then does it still make sense to project a volume here? Probably not.

openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml
mariusleu: But then, whole /etc on the container will get overwritten with what's on our projected volume.

mariusleu: The way would be to use subPath for /etc/sudoers.

fwiesel:
I've rebased the code and added some of the fixed requested. I am not 100% sure I covered all of them.

PR #2064, opened 2020-11-12
bump most kube-monitoring/kube-system deps to latest

majewsky:
I went through all the deps once again and there are already some new changes in that we might want to push through, particularly because of images moving away from Docker Hub, and of course, because of the deprecated API versions:

* **kube-monitoring**
  * ntp-exporter 2.0.0 -> 2.1.0 switches the image from Docker Hub to the mirror in Keppel.
* **kube-system**
  * ccauth 0.1.0 -> 0.1.1 switches the API version on the Ingress, but nothing else, so Helm should not touch it.
  * kube-dns 0.3.3 -> 0.3.4 removes an API version conditional for older Kubernetes. No diff on k8s 1.15.
  * kube-fip-controller 0.1.10 -> 0.1.12 switches the API version on the Deployment. To make the Deployment compliant with `apps/v1`, I added an explicit `spec.selector`. (This field used to be optional on `extensions/v1beta1`.) The value is the same as the one that was autofilled, so this could be a no-op, but Helm might still choose to touch this Deployment.
  * kube-proxy 0.6.23 -> 0.6.24 removes an API version conditional for older Kubernetes. No diff on k8s 1.15.
  * node-problem-detector 1.3.5 -> 1.3.6 switches the image from Docker Hub to the mirror in Keppel.
  * sysctl 0.0.1 -> 0.0.2 switches the API version on the DaemonSet, but nothing else, so Helm should not touch it.

### DIFF
--- a/system/kube-monitoring-admin-k3s/Chart.lock
+++ b/system/kube-monitoring-admin-k3s/Chart.lock
@@ -22,7 +22,7 @@ dependencies:
   version: 0.1.7
 - name: ntp-exporter
   repository: https://charts.eu-de-2.cloud.sap
-  version: 2.0.0
+  version: 2.1.0
 - name: oomkill-exporter
   repository: https://charts.eu-de-2.cloud.sap
   version: 0.1.9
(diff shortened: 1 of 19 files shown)


majewsky:
I'm having trouble with the Chart.lock for kube-system-metal:

```
$ helm3 dep up
Saving 20 charts
Downloading cc-rbac from repo https://charts.eu-de-2.cloud.sap
Downloading disco from repo https://charts.eu-de-2.cloud.sap
Downloading go-pmtud from repo https://charts.eu-de-2.cloud.sap
Save error occurred:  could not find : chart kube-dns not found in https://charts.eu-de-2.cloud.sap
Deleting newly downloaded charts, restoring pre-update state
Error: could not find : chart kube-dns not found in https://charts.eu-de-2.cloud.sap
```

Is the Chartmuseum upload having problems?

defo89:
@majewsky It seems that we are already running in version conflicts with `kube-monitoring-*`. But good news is that ntp-exporter should already be updated to 2.1.1 everywhere.

I have bumped `kube-system-*` charts separately. Also kube-dns was fixed - it was failing helm3 linter checks.

PR #2066, opened 2020-11-13
[alertmanager] metal : handle QA regions like all prod regions

geisslet:
QA regions shall be handled like prod regions for HW break&fix


geisslet:
@BerndKue fyi

BerndKue (approved):

auhlig (approved):
🎉

PR #2068, opened 2020-11-13
[cinder] switch to projected volumes

mariusleu:
This is part of the ticket CCM-9905

### DIFF
--- a/openstack/cinder/templates/api-deployment.yaml
+++ b/openstack/cinder/templates/api-deployment.yaml
@@ -83,44 +83,8 @@ spec:
             - name: cinder-api
               containerPort: {{.Values.cinderApiPortInternal }}
           volumeMounts:
-            - name: etccinder
+            - name: cinder-etc
               mountPath: /etc/cinder
-            - name: cinder-etc
-              mountPath: /etc/cinder/cinder.conf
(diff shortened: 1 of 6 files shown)


hemna:
projected volumes?

mariusleu:
> projected volumes?

This is a [feature in Kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-projected-volume-storage/) which is able to gather multiple files into a single volumeMount. According to the ticket I mentioned, the old way was sometimes leading to CrashLoopBackOff.

openstack/cinder/templates/api-deployment.yaml
```diff
@@ -132,14 +96,43 @@ spec:
...
-          emptyDir: {}
         - name: cinder-etc
-          configMap:
-            name: cinder-etc
+          projected:
+            defaultMode: 420
+            sources:
+              - configMap:
+                  items:
+                  - key: ciner.conf
```
hemna, line 109: ciner.conf -> cinder.conf

hemna (commented):
If this helps prevent CrashLoops, then that's great.   Just one misspelling in the first file
ciner.conf -> cinder.conf

PR #2069, opened 2020-11-13
Added vmware-vspc metrics to deployment

GlaciErrDev:


joker-official (commented):
I think there are some annotations missing for this exporter to be scraped by prometheus. Have a look [here](https://github.com/sapcc/helm-charts/blob/0598a2aca4001f4ae24bcf268d2a857f5465e22a/openstack/nova/templates/hypervisors/hypervisors-vmware-deployment-vct.yaml#L37-L40)

GlaciErrDev:
@joker-official sure, forgot about annotations.

galkindmitrii:
lgtm, I assume the image version will change when a new pipeline is there

PR #2088, opened 2020-11-23
castellum: use Secret for passwords

talal:


openstack/castellum/templates/_utils.tpl
```diff
@@ -11,8 +11,13 @@
...
   value: "nfs-shares,project-quota"
-- name: CASTELLUM_DB_URI
-  value: "postgres://postgres:{{ .Values.postgresql.postgresPassword }}@castellum-postgresql.{{ .Release.Namespace }}.svc/castellum?sslmode=disable"
+- name: CASTELLUM_DB_PASSWORD
+  valueFrom:
+    secretKeyRef:
+      name: castellum-secret
+      key: postgres_password
+- name: CASTELLUM_DB_HOSTNAME
+  value: "castellum-postgresql.{{ .Release.Namespace }}.svc"
```
majewsky, line 20: Relating to my review comment on the code PR, this section needs an explicit mention of `sslmode=disable` since this value should not be the default for connection options.

openstack/castellum/templates/secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: castellum-secret
+data:
+  service_user_password: {{ required ".Values.castellum.service_user.password" .Values.castellum.service_user.password | b64enc }}
+  postgres_password: {{ required ".Values.postgresql.postgresPassword" .Values.postgresql.postgresPassword | b64enc }}
+  rabbitmq_password: {{ required ".Values.castellum.rabbitmq.password" .Values.castellum.rabbitmq.password | b64enc }}
```
majewsky, line 8: Audit logging is optional in the `values.yaml`, so this value should not be `required` here:
```suggestion
  rabbitmq_password: {{ .Values.castellum.rabbitmq.password | b64enc }}
```

openstack/castellum/templates/secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: castellum-secret
+data:
+  service_user_password: {{ required ".Values.castellum.service_user.password" .Values.castellum.service_user.password | b64enc }}
```
majewsky, line 6: The `required` will never trigger because the default is the non-empty string `DEFINED_IN_VALUES_FILE`. Since we did without the `required` calls before, I think we'll be good without going forward as well.

majewsky (changes requested):

majewsky (approved):

PR #2101, opened 2020-11-25
[esxi-exporter] added syslog variables

christopherhans:


prometheus-exporters/esxi-exporter/values.yaml
```diff
@@ -4,9 +4,11 @@ docker_repo: DEFINED-IN-REGION
...
 esxi_password: DEFINED-IN-REGION
 esxi_connections: DEFINED-IN-REGION
+syslog_host: DEFINED-IN-REGION
+syslog_port: DEFINED-IN-REGION
 collector:
   - ConfigCollector
   - CriticalServiceCollector
 alerts:
-  enabled: false
+  enabled: true
```
viennaa, line 13: that should stay at false, alerts are enabled in the secrets values.yaml only.

viennaa (changes requested):

viennaa (approved):

PR #2102, opened 2020-11-26
Add CRC error alert definition

mithunmgore:
Added info alert for CRC errors.

Signed-off-by: Mithun Gore <mithungore@gmail.com>

### DIFF
--- a/prometheus-exporters/ucs-exporter/alerts/portlicense.alerts
+++ b/prometheus-exporters/ucs-exporter/alerts/portlicense.alerts
@@ -7,6 +7,8 @@ groups:
     for: 15m
     labels:
       severity: info
+      tier: vmware
+      service: compute
       context: "ucs-exporter"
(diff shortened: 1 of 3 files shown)


christopherhans (changes requested):
The alert is missing tier & service. This is important for the alert routing. Please add
```
tier: vmware
service: compute
```
i.e.: https://github.com/sapcc/helm-charts/blob/master/prometheus-exporters/vrops-exporter/alerts/virtualmachine.alerts#L13-L14

christopherhans (approved):
@viennaa pls merge

viennaa:
Shouldn't there be a `for` condition in there too?

christopherhans:
As the CRC alerting conditions are not 100% sure, we agreed with Martin&Biswajit to create the definition as in this PR. The Ops colleagues will actively monitor the alert in the info channel and decide if further adjustments are needed.

viennaa:
Alright, your call.

PR #2103, opened 2020-11-26
openstack/limes: use Secret for passwords

talal:


openstack/limes/templates/configmap.yaml
```diff
@@ -9,7 +9,9 @@ data:
 {{ .Files.Get "files/policy.json" | indent 4 }}
   limes.yaml: |
     database:
-      location: "postgres://postgres:{{ .Values.postgresql.postgresPassword }}@limes-postgresql.{{ .Release.Namespace }}.svc/limes?sslmode=disable"
+      password: "{ fromEnv: LIMES_DB_PASSWORD }"
```
majewsky, line 12: The quotes around this value need to be removed. This sets the password to the actual string `{ fromEnv: LIMES_DB_PASSWORD }`.

majewsky (changes requested):

majewsky (approved):

PR #2114, opened 2020-12-01
Upgrade the alerts for cinder storage

hemna:
This patch updates the alerts for several of the
cinder low storage / empty storage alerts from info
to warning/critical

TODO:
The empty storage playbook is being updated by Gergely


dhoeller (approved):
LGTM

PR #2118, opened 2020-12-02
openstack/content-repo: use Secret for passwords

talal:


openstack/content-repo/templates/secret.yaml
```diff
@@ -0,0 +1,20 @@
...
+  target_auth_password: {{ required ".Values.password" .Values.password | b64enc }}
+{{- range $repo, $config := .Values.repos }}
+{{- range $index, $job := $config.jobs}}
+{{- if $job.from.password }}
+  {{- $objPrefix := $job.from.object_prefix | replace "/" "" }}
+  {{- $uniqueID := (printf "%s_%s_%s_%s_%s" $repo $job.from.project_domain_name $job.from.project_name $job.from.container $objPrefix) }}
+  source_{{ $uniqueID }}_auth_password: {{ $job.from.password | b64enc }}
+{{- end }}
+{{- end}}
+{{- end}}
```
majewsky, line 20: This leads to quite some useless duplication, and what is more concerning: When we add a new sync job reusing an existing password, we have to change the secret and thus restart everything because of the checksum change.

I would much rather have the passwords deduplicated in `.Values.passwords` and have `.Values.repos{}.jobs.from.password` use the fromEnv syntax directly.

majewsky (changes requested):

majewsky (approved):

PR #2125, opened 2020-12-04
openstack/swift-drive-autopilot: use Secret for passwords

talal:


majewsky (approved):

majewsky:
When testing this in QA, please test the reboot of a storage node. The encryption key is only used when opening LUKS containers, which typically only happens on first startup after a reboot.

PR #2137, opened 2020-12-10
remove vice-president annotation

ArtieReus:


openstack/arc/templates/api-ingress.yaml
```diff
@@ -20,7 +20,7 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
   {{- if .Values.api.ingress.vice_president }}
```
jknipper, line 21: Maybe change parameter name to make it less confusing.

databus23 (approved):

openstack/arc/templates/api-ingress.yaml
databus23: 👍 @ArtieReus

PR #2141, opened 2020-12-14
Added new nova value to scheduler section

GlaciErrDev:
- New value:
    'reserved_ram_threshold_for_resize'


joker-official:
I don't like the commit-message. It's too generic and one has to click on the PR to even know what's it about. Could we change it to something like `Make reserved_ram_threshold_for_resize configurable` or something? We can then in the body explain that it's a new scheduler thing for a filter to reserve ram for resize and why we want that filter in the first place.

joker-official:
Wouldn't we also have to add it to the filter to `default_filters`, to make use of it? Is this another PR? Could you please link those two, if yes?

joker-official (changes requested):
Blocker for merging. David hasn't decided, yet, if we want to do this.

PR #2146, opened 2020-12-17
removed instance label

richardtief:


system/infra-monitoring/templates/_prometheus-infra-collector.yaml.tpl
```diff
@@ -644,7 +644,7 @@
         - infra-monitoring
   metrics_path: /
   relabel_configs:
-    - action: keep
+    - action: labeldrop
       source_labels: [__meta_kubernetes_service_name]
       regex: .*vrops-exporter.*
```
viennaa, line 649: shouldn't this be done after the collector label is set?

PR #2149, opened 2020-12-18
[cinder] prepare rate-limiting middleware:

galkindmitrii:
- no global limits at the moment; only (target-) project limits now.
- each individual endpoint has to be specified, no support for wildcards
- http 429 is returned when limit is hit; or 403 if blacklisted.
- json message is returned when limiting takes place
- rate-limiting settings ignored when backend (redis) is not reachable

### DIFF
--- a/openstack/cinder/templates/etc/_api-paste.ini.tpl
+++ b/openstack/cinder/templates/etc/_api-paste.ini.tpl
@@ -20,6 +20,10 @@ use = call:cinder.api:root_app_factory
 {{- if .Values.watcher.enabled }} watcher{{- end -}}
 {{- end }}
 
+{{- define "rate_limit_pipe" -}}
+{{- if .Values.api_rate_limit.enabled }} rate_limit{{- end -}}
+{{- end }}
+
 {{- if $mitaka }}
 [composite:openstack_volume_api_v1]
 use = call:cinder.api.middleware.auth:pipeline_factory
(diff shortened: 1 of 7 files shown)


joker-official:
Why doesn't it use the standard HTTP return code?

openstack/cinder/templates/etc/_ratelimit.yaml.tpl
```diff
@@ -0,0 +1,98 @@
...
+blacklist_response:
+  status: 497 Blacklisted
+  status_code: 497
+  json_body: { "message": "You have been blacklisted. Please contact administrator." }
+
+# Group multiple CADF actions to one rate limit action.
+groups:
+  write:
+    - update
+    - delete
```
joker-official, line 23: Where does create come in? Do we allow it unlimited or is it an `update`?

openstack/cinder/templates/etc/_ratelimit.yaml.tpl
```diff
@@ -0,0 +1,99 @@
...
+{{ if .Values.api_rate_limit.project_whitelist -}}
+# List of whitelisted scopes keys (domainName/projectName).
+whitelist: {{ .Values.api_rate_limit.project_whitelist }}
+{{- end }}
+{{- end }}
+
+# Override default ratelimit response.
+ratelimit_response:
+  status: 498 Rate Limited
+  status_code: 498
```
hemna, line 11: is this the returned HTTP status code?   Shouldn't it be 429 ?
429 Too Many Requests

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

galkindmitrii:
@joker-official as for http status code, I've also asked Arno and got response that 498 is already used for that purposes in Openstack.
That is an open point that we should decide on and properly communicate to the end-users.

openstack/cinder/templates/etc/_ratelimit.yaml.tpl
galkindmitrii: @auhlig, do we need to explicitly specify all methods here? - in watcher I can see also: 
update/os-complete
update/os-begin_detaching
update/os-reset_status
update/os-force_delete
and so on.

openstack/cinder/templates/etc/_api-paste.ini.tpl
```diff
@@ -31,13 +35,13 @@ keystone_nolimit = cors http_proxy_to_wsgi request_id faultwrap sentry sizelimit
 [composite:openstack_volume_api_v2]
 use = call:cinder.api.middleware.auth:pipeline_factory
 noauth = cors http_proxy_to_wsgi request_id {{- include "watcher_pipe" . }} faultwrap sentry sizelimit {{- include "osprofiler_pipe" . }} noauth apiv2
-keystone = cors http_proxy_to_wsgi request_id faultwrap sentry sizelimit {{- include "osprofiler_pipe" . }} authtoken keystonecontext {{- include "watcher_pipe" . }} {{- include "audit_pipe" . }} apiv2
+keystone = cors http_proxy_to_wsgi request_id faultwrap sentry sizelimit {{- include "osprofiler_pipe" . }} authtoken keystonecontext {{- include "watcher_pipe" . }} {{- include "audit_pipe" . }} {{- include "rate_limit_pipe" . }} apiv2
 keystone_nolimit = cors http_proxy_to_wsgi request_id faultwrap sentry sizelimit {{- include "osprofiler_pipe" . }} authtoken keystonecontext {{- include "watcher_pipe" . }} {{- include "audit_pipe" . }} apiv2
```
galkindmitrii, line 39: Hi @hemna,
- do I understand correctly that _keystone_nolimit_ pipe should not have any rate limiting?

openstack/cinder/templates/etc/_ratelimit.yaml.tpl
galkindmitrii: Please see: https://github.com/sapcc/helm-charts/pull/2149#issuecomment-748096203

openstack/cinder/templates/etc/_api-paste.ini.tpl
hemna: no idea about keystone

openstack/cinder/templates/etc/_ratelimit.yaml.tpl
auhlig: You can set it as you like. Swift uses 498, but 429 is the standard. I think it's a good idea to enforce http standard

openstack/cinder/templates/etc/_ratelimit.yaml.tpl
auhlig: Yes. 
Rate limits must be set explicitly to avoid potential confusion why/when/if limits are implicitly applied.
But I get the point of making this more flexible

openstack/cinder/templates/etc/_ratelimit.yaml.tpl
galkindmitrii: set to 429 now

openstack/cinder/templates/etc/_ratelimit.yaml.tpl
galkindmitrii: yes, wildcards would be helpful :)

hemna (commented):
lgtm

PR #2150, opened 2020-12-21
Swift: secret handling

talal:


openstack/swift/templates/etc/_proxy-server.conf.tpl
```diff
@@ -114,7 +115,7 @@ token_cache_time = {{$cluster.token_cache_time | default 600}}
 region_name = {{$cluster.region_name | default $context.global.region}}
 user_domain_name = {{$cluster.swift_service_user_domain}}
 username = {{$cluster.swift_service_user}}
-password = {{$cluster.swift_service_password}}
+password = "fromEnv: {{ $cluster_id | upper }}_SERVICE_PASSWORD"
```
majewsky, line 118: If I'm not mistaken, this should be:

```suggestion
password = "{ fromEnv: {{ $cluster_id | upper }}_SERVICE_PASSWORD }"
```

openstack/swift/templates/etc/_proxy-server.conf.tpl
```diff
@@ -208,7 +209,7 @@ username = {{ $cluster.swift_service_user }}
 user_domain_name = {{ $cluster.swift_service_user_domain }}
 project_name = {{ $cluster.swift_service_project }}
 project_domain_name = {{ $cluster.swift_service_project_domain }}
-password = {{ $cluster.swift_service_password }}
+password = "fromEnv: {{ $cluster_id | upper }}_SERVICE_PASSWORD"
```
majewsky, line 212: Same here:

```suggestion
password = "{ fromEnv: {{ $cluster_id | upper }}_SERVICE_PASSWORD }"
```

majewsky (changes requested):
I had a lot more notes initially, but most of them come down to a single theme: Most services don't actually need all passwords. The following containers definitely work without any passwords:

- nginx
- rsyncd
- all statsd exporters

For the `dispersion-password` specifically, as far as I'm aware, it's only used by `swift-dispersion-report` and thus only needs to be provided to the health-exporter containers. @reimannf Please correct me if I'm missing something here.

majewsky (approved):
Please go ahead with the rollout to QA. Remember to announce the deployment in the respective region channels in Slack; it's not out of the question that there will be some service disruption in case we forgot about some pod or some password.

PR #2160, opened 2020-12-30
clair: secret handling

talal:


majewsky:
In the meantime, there is a new Clair deployment within Keppel that is about to replace this Clair instance, so it's not worth continuing with this changeset. For what it's worth, the new Clair has proper secret handling from the get-go, largely based on your work in this changeset.

PR #2161, opened 2021-01-04
[Cinder] Allow for defining a custom cinder backend

hemna:
This patch adds a section to the cinder backends definitions
to allow creating a custom backend define for a region.

The region's values/cinder.yml would contain

backends:
    custom_backend: |
    [custom_backend_name]
    volume_backend_name = my_custom_backend
    volume_driver = cinder.volume.drivers.something.custom

This will allow us to add custom backends against qa-de-1 for example to enable the pure storage driver.

### DIFF
--- a/openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
+++ b/openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
@@ -48,6 +48,7 @@ template: |
       vmware_storage_profile = {{ .Values.backends.standard_hdd.vmware_storage_profile }}
       extra_capabilities = '{"vcenter-shard": "{= host.split(".")[0] =}"}'
 
+      {{ .Values.backends.custom_backend }}
 
       [vstorageobject]
       volume_driver = cinder.volume.drivers.vmware.fcd.VMwareVStorageObjectDriver
--- a/openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
+++ b/openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
@@ -48,7 +48,12 @@ template: |
       vmware_storage_profile = {{ .Values.backends.standard_hdd.vmware_storage_profile }}
       extra_capabilities = '{"vcenter-shard": "{= host.split(".")[0] =}"}'
 
-      {{ .Values.backends.custom_backend }}
+      {{- range $backend, $kvs := .Values.backends.additional_backends }}
+      [{{ $backend }}]
+      {{- range $key, $value := $kvs }}
+      {{ $key }} = {{ $value }}
+      {{- end }}
+      {{- end }}
 
       [vstorageobject]
       volume_driver = cinder.volume.drivers.vmware.fcd.VMwareVStorageObjectDriver
--- a/openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
+++ b/openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
@@ -48,12 +48,12 @@ template: |
       vmware_storage_profile = {{ .Values.backends.standard_hdd.vmware_storage_profile }}
       extra_capabilities = '{"vcenter-shard": "{= host.split(".")[0] =}"}'
 
-      {{- range $backend, $kvs := .Values.backends.additional_backends }}
+      {{ range $backend, $kvs := .Values.backends.additional_backends }}
       [{{ $backend }}]
       {{- range $key, $value := $kvs }}
       {{ $key }} = {{ $value }}
       {{- end }}
-      {{- end }}
+      {{ end }}
 
       [vstorageobject]
       volume_driver = cinder.volume.drivers.vmware.fcd.VMwareVStorageObjectDriver


openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
```diff
@@ -48,6 +48,7 @@ template: |
       vmware_storage_profile = {{ .Values.backends.standard_hdd.vmware_storage_profile }}
       extra_capabilities = '{"vcenter-shard": "{= host.split(".")[0] =}"}'
 
+      {{ .Values.backends.custom_backend }}
```
joker-official, line 51: How about a different structure, that would allow more backends. I haven't tested it but this code might work
```
{{- range $backend, $kvs := .Values.backends.additional_backends }}
[{{ $backend }}]
{{- range $key, $value := $kvs }}
{{ $key }} = {{ $value }}
{{- end }}
{{- end }}
```
and would look like this in the region's YAML:
```yaml
backends:
  additional_backends:
    pure_vmware:
      volume_driver: cinder.volume.drivers.vmware.vmdk.VMwareVcVmdkDriver
      vmware_storage_profile: cinder-pure
      extra_capabilities: '{"vcenter-shard": "{= host.split(".")[0] =}"}'
```

hemna: The custom_backend var as I have in this patch can be more than 1 backend, since it's a free form field.

```
backends:
custom_backend: |
[custom_backend_name]
volume_backend_name = my_custom_backend
volume_driver = cinder.volume.drivers.something.custom

[vmware_pure]
volume_backend_name = vmware_pure
volume_driver = cinder.volume.drivers.blah
vcenter-shard = something

Comment (joker-official) [reply openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml:51]: free-form fields have the disadvantage for me, that they are not YAML and thus not integrate so well into the rest of the config and it's often harder to spot errors in there.

But you're right. To indicate that, maybe we should rename it to `custom_backends`?

**Review by joker-official — APPROVED** [review]
Approving, since it seems to be a matter of taste to do it differently.

**Comment by hemna** [comment]
> Approving, since it seems to be a matter of taste to do it differently.

ok I redid this using your mechanism.

**Comment by joker-official** [comment]
Tested it with:
```diff
--- a/qa-de-2/values/cinder.yaml
+++ b/qa-de-2/values/cinder.yaml
@@ -14,6 +14,14 @@ volume_size_limit_gb: 1000
 
 backends:
   enabled: vmware,standard_hdd
+  additional_backends:
+    purestoragevvol:
+      volume_driver: cinder.volume.drivers.vmware.vmdk.VMwareVcVmdkDriver
+      volume_backend_name: vmware
+      vmware_storage_profile: 'cinder-pure'
+      extra_capabilities: '{"vcenter-shard": "{= host.split(".")[0] =}"}'
+
+
 
 mariadb:
   enabled: true
```
We need to remove the `-` in the first `range` command, because it looks like this:
```

      [standard_hdd]
      volume_driver = cinder.volume.drivers.vmware.vmdk.VMwareVcVmdkDriver
      volume_backend_name = standard_hdd
      vmware_storage_profile = cinder-standard-hdd
      extra_capabilities = '{"vcenter-shard": "{= host.split(".")[0] =}"}'
      [purestoragevvol]
      extra_capabilities = {"vcenter-shard": "{= host.split(".")[0] =}"}
      vmware_storage_profile = cinder-pure
      volume_backend_name = vmware
      volume_driver = cinder.volume.drivers.vmware.vmdk.VMwareVcVmdkDriver

      [vstorageobject]
```
It also looks like we need a `''` around the value part.

hemna:
> 
>       [standard_hdd]
>       volume_driver = cinder.volume.drivers.vmware.vmdk.VMwareVcVmdkDriver
>       volume_backend_name = standard_hdd
>       vmware_storage_profile = cinder-standard-hdd
>       extra_capabilities = '{"vcenter-shard": "{= host.split(".")[0] =}"}'
>       [purestoragevvol]
>       extra_capabilities = {"vcenter-shard": "{= host.split(".")[0] =}"}
>       vmware_storage_profile = cinder-pure
>       volume_backend_name = vmware
>       volume_driver = cinder.volume.drivers.vmware.vmdk.VMwareVcVmdkDriver
> 
>       [vstorageobject]
> ```
> 
> It also looks like we need a `''` around the value part.

I removed the - from the first range command.  Can we escape the ' in the secrets file, so that it shows up in the final rendering?   I'm not sure it's a good idea to blindly quote every value

joker-official:
> I removed the - from the first range command. Can we escape the ' in the secrets file, so that it shows up in the final rendering? I'm not sure it's a good idea to blindly quote every value

Why do you think it's a problem? Because we might end up with an Integer becoming a String? In the end, this is an INI. Does it support integers as such?
In any case, I can load the resulting YAML, so kubernetes can probably too and we don't need extra quoting. The name we would quote is generated from the vcenter and thus a hostname which doesn't support spaces.

For multiple backends, we need to get rid of another `-` in the last `{{- end }}`.
```
      {{ range $backend, $kvs := .Values.backends.additional_backends }}
      [{{ $backend }}]
      {{- range $key, $value := $kvs }}
      {{ $key }} = {{ $value }}
      {{- end }}
      {{ end }}
```

joker-official (approved):

PR #2173, opened 2021-01-13
[nova] Switch to sentry-operator

joker-official:
Instead of manually updating/adding the "sentry_dsn" key in/to the
secrets after adding its project to sentry, we now tell the
sentry-operator to add the project and provide the DSN to us via
k8s secret.

We also add the `SENTRY_DSN` environment variable to all services, that
didn't have it before.

### DIFF
--- a/openstack/nova/templates/api-deployment.yaml
+++ b/openstack/nova/templates/api-deployment.yaml
@@ -58,8 +58,13 @@ spec:
               value: "localhost"
             - name: STATSD_PORT
               value: "9125"
+            {{- if .Values.sentry.enabled }}
             - name: SENTRY_DSN
-              value: {{.Values.sentry_dsn | quote}}
+              valueFrom:
+                secretKeyRef:
+                  name: sentry
+                  key: {{ .Chart.Name }}.DSN.python
+            {{- end }}
 {{- if .Values.python_warnings }}
             - name: PYTHONWARNINGS
               value: {{ .Values.python_warnings | quote }}
--- a/openstack/nova/templates/api-metadata-deployment.yaml
+++ b/openstack/nova/templates/api-metadata-deployment.yaml
@@ -54,8 +54,13 @@ spec:
               value: "localhost"
             - name: STATSD_PORT
               value: "9125"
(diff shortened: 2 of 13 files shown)


openstack/nova/templates/api-deployment.yaml
```diff
@@ -58,8 +58,13 @@ spec:
               value: "localhost"
             - name: STATSD_PORT
               value: "9125"
+            {{- if .Values.sentry.enabled }}
             - name: SENTRY_DSN
-              value: {{.Values.sentry_dsn | quote}}
+              valueFrom:
+                secretKeyRef:
+                  name: sentry
+                  key: {{ .Chart.Name }}.DSN.python
```
grandchild, line 66: I was expecting to find something in the secrets that would define this value. I cannot. Where does it come from?

grandchild:
Is the CI failure expected? It doesn't seem to have a relation to this PR.

openstack/nova/templates/api-deployment.yaml
grandchild: nevermind, found it now

grandchild: Although it's still in the old format, `sentry_dsn`, so i guess there's a secret change necessary as well?

joker-official: I don't get your questions. Do you mean `{{ .Chart.Name }}`? The `sentry` secret? The second is described in the commit-message.

grandchild: Ah via k8s secret. What is the rationale behind doing it that way? Why not a regular secret? I'm asking because it adds another source of information, and increases spread.

joker-official: We have a `sentry-operator` which creates the projects in sentry for us and provides the DSN to us via secret. The point is, that one can reinstall sentry without having to do manual processing for every service (creating the project in sentry, updating the secrets).

grandchild: i see.

grandchild (approved):

PR #2177, opened 2021-01-13
[thanos] add query timeout config

notque:
Hey Arno, does this look good for you? I want to be able to tune the query timeout.

### DIFF
--- a/common/prometheus-server/templates/thanos/prometheus-thanosQueryDeployment.yaml
+++ b/common/prometheus-server/templates/thanos/prometheus-thanosQueryDeployment.yaml
@@ -33,6 +33,7 @@ spec:
         - --log.level={{ .Values.logLevel }}
         - --query.replica-label=prometheus_replica
         - --query.auto-downsampling
+        - --query.timeout={{ include ".Values.thanos.querier.timeout" }}:2m
         - --store={{ include "prometheus.fullName" . }}-thanos-store:{{ required ".Values.thanos.store.service.grpcPort missing" .Values.thanos.querier.service.grpcPort }}
         - --store={{ include "prometheus.fullName" . }}:10901
         - --web.route-prefix={{ required ".Values.thanos.querier.webRouteprefix missing" .Values.thanos.querier.webRouteprefix }}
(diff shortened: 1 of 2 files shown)


common/prometheus-server/templates/thanos/prometheus-thanosQueryDeployment.yaml
```diff
@@ -33,6 +33,7 @@ spec:
         - --log.level={{ .Values.logLevel }}
         - --query.replica-label=prometheus_replica
         - --query.auto-downsampling
+        - --query.timeout={{ include ".Values.thanos.querier.timeout" }}:2m
```
auhlig, line 36: Is `:2m` is intended? Why `include`?

auhlig (changes requested):
2 minutes is already the default query timeout. What's the use-case of making this configurable beyond this default?

common/prometheus-server/values.yaml
```diff
@@ -241,6 +241,8 @@ thanos:
     service:
       httpPort: 9090
       grpcPort: 10901
+    # Timeout for queries
+    #timeout: 2m
```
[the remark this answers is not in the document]
notque: I added it commented so that it was clear it was an available option. The intention is not to change the default for everyone, but allow the default to be configurable. If you prefer a different syntax for that idea, that's fine.

common/prometheus-server/templates/thanos/prometheus-thanosQueryDeployment.yaml
notque: The intention was if a value was not provided, to default to 2m, which is the regular default. The point was to allow it to be configurable, but not change the default. I thought include was the proper way to do that in this syntax.

auhlig: I get that. Sorry if I wasn't clear: This is technically not working as you intended. Values are not `included` like that and defaults, if that's what the `:2m` should be for, are not provided like that.

common/prometheus-server/values.yaml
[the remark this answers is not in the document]
auhlig: I have no preferences, but helm does. That's why the test failed :)
Which values did you have in mind? Still not getting why `2m` isn't suitable?

notque:
Closing this request, I tested manually the change, and it wasn't providing the benefit I was hoping for.

PR #2180, opened 2021-01-15
[Cinder] Fix rate limit

hemna:
This patch fixes the ratelimit.yml file.   The rate limiting wasn't
working with cinder on some urls, because the blocks declaring
the uris were getting re-declared for the different actions GET vs POST.
The actions need to be in the same declaration block for the uri.


galkindmitrii (approved):
LGTM, thanks!

PR #2185, opened 2021-01-18
[alertmanager] separate cinder service slack messages

geisslet:
cinder has alerts for tier os and vmware - generic service slack channel selection setup as we did it before doesn't work anymore
don't want to remove the tier completely
+ version bump

side-note: 
- don't like it, but to remove the tier seems even more problematic
-  we need a larger clean up mid term

### DIFF
--- a/global/prometheus-alertmanager-operated/Chart.yaml
+++ b/global/prometheus-alertmanager-operated/Chart.yaml
@@ -3,4 +3,4 @@ appVersion: "1.0"
 type: application
 description: Prometheus Alertmanager via operator.
 name: prometheus-alertmanager-operated
-version: 0.3.15
+version: 0.3.16
--- a/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
+++ b/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
@@ -99,12 +99,19 @@ route:
       tier: test-tier
       region: area51
 
+  - receiver: slack_cinder
+    continue: true
+    match_re:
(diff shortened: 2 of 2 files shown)


geisslet:
fyi @hemna @joker-official

auhlig:
Maybe the `tier` can be removed [here](https://github.com/sapcc/helm-charts/pull/2185/commits/338a08491d5a43086326f0b9b5c5b42bd395c7d4#diff-800d7f580eb6b7daf0f9afb2348581ac176a90a7d763e0f0acf6684694c272a8L105). Can you check whether alerts from openstack services actually have a different tier set here? I assume it's hardcoded to os and cinder is the 1st candidate to cover mult. tiers .

hemna:
The tier for the openstack-exporter alerts have been set to vmware:

https://github.com/sapcc/helm-charts/blob/master/prometheus-exporters/openstack-exporter/alerts/cinder.alerts

geisslet:
@auhlig yeah, thought about that, too - and agree on the Extrawurst-thing - but I skipped it because wasn't sure about side effects - like higher failure rate on sending messages because .. ah, no that's covered .. anyway I decided to go the safer (messier) way, because I want to discuss some general restructuring of the routing to make that easier to understand (would open an issue).

Summary: if you also think side effect risk are low (couldn't find any obvious risk by scanning the alerts (git grep -e "tier" --and -e "service") .. I would dump this PR and open a new.

geisslet:
replaced by new PR: https://github.com/sapcc/helm-charts/pull/2212

PR #2186, opened 2021-01-18
[elastic-query-exporter] Add alert for unexpected VM power down

max-len:
Loglevel INFO, playbook needs to be added, no VM info in alert

### DIFF
--- a/prometheus-exporters/elastic-query-exporter/alerts/logs.alerts
+++ b/prometheus-exporters/elastic-query-exporter/alerts/logs.alerts
@@ -41,3 +41,16 @@ groups:
         annotations:
           description: "VCSA logs missing in Octobus for vCenter {{ $labels.vcenter }}."
           summary: "VCSA logs missing in Octobus for vCenter {{ $labels.vcenter }}."
+      - alert: VMPoweredDownDuringFailedHostMigration
+        expr: elasticsearch_octobus_vmdown_doc_count
+        for: 5m
+        labels:
+          severity: info
+          tier: vmware
+          service: audit
+          context: logging
+          no_alert_on_absence: "true"
+          meta: "A VM on host {{ $labels.hostsystem }} unexpectedly powered down during a failed host migration (vMotion or Storage vMotion)."
+        annotations:
+          description: "A VM on host {{ $labels.hostsystem }} unexpectedly powered down during a failed host migration (vMotion or Storage vMotion)."
+          summary: "A VM unexpectedly powered down during a failed host migration (vMotion or Storage vMotion)."
--- a/prometheus-exporters/elastic-query-exporter/alerts/logs.alerts
+++ b/prometheus-exporters/elastic-query-exporter/alerts/logs.alerts
@@ -47,7 +47,7 @@ groups:
         labels:
           severity: info
           tier: vmware
-          service: audit
+          service: compute
           context: logging
           no_alert_on_absence: "true"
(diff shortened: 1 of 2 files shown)


prometheus-exporters/elastic-query-exporter/alerts/logs.alerts
```diff
@@ -41,3 +41,16 @@ groups:
         annotations:
           description: "VCSA logs missing in Octobus for vCenter {{ $labels.vcenter }}."
           summary: "VCSA logs missing in Octobus for vCenter {{ $labels.vcenter }}."
+      - alert: VMPoweredDownDuringFailedHostMigration
+        expr: elasticsearch_octobus_vmdown_doc_count
+        for: 5m
+        labels:
+          severity: info
+          tier: vmware
+          service: audit
```
christopherhans, line 50: Ne service is not `audit`, it is `compute`.
We used audit for the audit related alerts.

prometheus-exporters/elastic-query-exporter/files/exporter.cfg
```diff
@@ -156,3 +155,24 @@ QueryJson = {
...
+          { "exists": { "field": "node.nodename" }},
+          { "match_phrase": { "message": "An operation required the virtual machine to quiesce and the virtual machine was unable to continue running." }}
+        ],
+        "filter": [
+          { "range": { "@timestamp": { "gte": "now-1h" }}}
+        ]
+      }
+    },
+    "aggs": {
+      "hostsystem": { "terms": { "field": "node.nodename.keyword" }}
```
christopherhans, line 176: can't we just use `message` as field?

christopherhans (changes requested):

christopherhans:
If we could get the whole message in the label `message`, we could do something like
`{{ labels.message | regex_replace('.+', '(?<=on\s).+\s\(.+\)' }}`
this will result in something like `cdd23v010062 (c52b98d6-af94-404a-afb8-619734b062e4)`

christopherhans:
Playbook: https://operations.global.cloud.sap/docs/devops/alert/vcenter/#VM%20powered%20off%20during%20vMotion

prometheus-exporters/elastic-query-exporter/alerts/logs.alerts
max-len: Fixed

max-len:
> Playbook: https://operations.global.cloud.sap/docs/devops/alert/vcenter/#VM%20powered%20off%20during%20vMotion

Added

prometheus-exporters/elastic-query-exporter/files/exporter.cfg
max-len: Unfortunately the mapped keyword field doesn't behave consistently on ElasticDB side. In the sample alert I was testing with this field was "missing" and the aggregate returned no results.

christopherhans: I just don't think that only pointing out the host is a huge help, if there are dozens of powered off VMs.

Could you please reach out to OCB Team, why this field behaves differently?

max-len: Yes, I'm on it. I'd suggest to merge this to test the current state in QA in parallel.

christopherhans (approved):
Roll out for testing in info severity.

PR #2193, opened 2021-01-19
new flavor g_c12_m48

dhoeller:


openstack/sap-seeds/templates/flavor-seed.yaml
```diff
@@ -87,6 +87,15 @@ spec:
     extra_specs:
       "vmware:hv_enabled": "True"
       "hw_video:ram_max_mb": "16"
+  - name: "g_c11_m48"
```
mariusleu, line 90: Should be g_**c12**_m48?

dhoeller: Thanks!

PR #2203, opened 2021-01-25
[Octavia] Add tracing with OSProfiler

laerlingSAP:


openstack/octavia/templates/etc/_octavia.conf.tpl
```diff
@@ -20,6 +20,9 @@ api_v2_enabled = True
 # How should authentication be handled (keystone, noauth)
 auth_strategy = keystone
 
+# Tracing
+{{- include "osprofiler" . }}
+
```
notandy, line 25: This needs to be moved above a config segment, else the config options `enabled_provider_drivers` and `default_provider_driver` will be in the wrong config segment, e.g. how it looks rendered:

```ini
    # How should authentication be handled (keystone, noauth)
    auth_strategy = keystone

    # Tracing

    [profiler]
    connection_string = jaeger://localhost:6831
    enabled = true
    hmac_keys = ...
    trace_sqlalchemy = false

    # Dictionary of enabled provider driver names and descriptions
    enabled_provider_drivers = noop_driver: 'The No-Op driver.', f5: 'F5 BigIP driver.'

    # Default provider driver
    default_provider_driver = f5
```

notandy (changes requested):

openstack/octavia/templates/etc/_octavia.conf.tpl
laerlingSAP: Oops... Thanks!

PR #2206, opened 2021-01-25
[elastic-query-exporter] Fix alert expression

max-len:
@christopherhans please merge


christopherhans:
I see that you changed nothing at the deployment, but are you aware of: `[ERROR] templates/: template: elastic-query-exporter/templates/deployment.yaml:39:312: executing "elastic-query-exporter/templates/deployment.yaml" at <.Values.global.octobus.apikey>: nil pointer evaluating interface {}.octobus` ?

max-len:
Yes, Olaf confirmed that's a glitch in the CI. I suspect the secrets are not properly implemented there, but fixing this is out of my scope atm.

christopherhans (approved):

PR #2208, opened 2021-01-26
[vrops-exporter] only powered on VMs for alert VMsOnFailoverHost

richardtief:
+ better information on alerts (vcenter instead of datacenter)

### DIFF
--- a/prometheus-exporters/vrops-exporter/alerts/datastore.alerts
+++ b/prometheus-exporters/vrops-exporter/alerts/datastore.alerts
@@ -51,12 +51,12 @@ groups:
       tier: vmware
       service: storage
       context: "vrops-exporter"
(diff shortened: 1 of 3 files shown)


prometheus-exporters/vrops-exporter/alerts/datastore.alerts
```diff
@@ -51,12 +51,12 @@ groups:
       tier: vmware
       service: storage
       context: "vrops-exporter"
-      meta: "Datastore {{ $labels.datastore }} utilization > 80% ({{ $labels.datacenter }}, {{ $labels.vccluster }}, {{ $labels.hostsystem }})"
+      meta: "Datastore {{ $labels.datastore }} utilization > 80% ({{ $labels.vcenter }}, {{ $labels.vccluster }}, {{ $labels.hostsystem }})"
```
viennaa, line 54: Please either go with a `.` after the general statement or without it. Some examples do have a `.` at the end, some don't.

viennaa (changes requested):
Please don't just attach the information in the

viennaa (approved):

PR #2212, opened 2021-01-28
[alertmanager] rm tier from generic service routing

geisslet:
this is mainly to handle multi-tier alerts by service


geisslet:
fyi @hemna @joker-official

auhlig:
Kewl 🎉 
I'll deploy this on Monday morning to avoid unwanted alerts - not that I could foresee some- over the weekend

PR #2213, opened 2021-01-28
[elastic-query-exporter] Switch to search result based metric

max-len:
* adjust alert expression
* remove alert timeout as this is an event based alert
* add syslog message in alert description
* node.nodename removed from alert description
  as exporter can't parse out nested fields yet
  ESXi name will be in syslog message, as will affected VM
* remove aggs from query needed before for aggs based metric


max-len:
@christopherhans please review and merge. alert is still on info level and will be tested in qa.

christopherhans (approved):
lgtm

PR #2214, opened 2021-01-29
[neutron] Add NSXTExporterNoMetrics alert

joker-official:
The NSXTExporterNoData alert doesn't work, if the exporter stopped
exporting for a longer period of time, because prometheus can only see
things as absent, if they ever reported.

We now build a query matching all running nsxv3-agent pods to one
nsxv3_agent of the nsxv3_cluster_management_last_successful_data_fetch
metric. If we can't find a metric reported by a running pod, we fire an
alert.

### DIFF
--- a/openstack/neutron/alerts/nsxt.alerts
+++ b/openstack/neutron/alerts/nsxt.alerts
@@ -90,6 +90,18 @@ groups:
       description:  'NSX-T Exporter has no data'
       summary:  'NSX-T Exporter has no data'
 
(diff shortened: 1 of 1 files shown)


notandy (approved):

geisslet (approved):
Alert definition is fine.

pipeline error: 
engine.go:159: [INFO] Missing required value: .Values.global.registry is missing
Error: 1 chart(s) linted, 1 chart(s) failed

PR #2221, opened 2021-02-01
[swift] Deployment options for Proxy: Deployment, Daemonset or Both

reimannf:
Swift proxies can be deployed as
* Deployment
* Daemonset
* Both

New value sets:
* proxy_deployment_*
* proxy_daemonset_* 

Resource request and limits have to be specified per swift cluster

Other changes
* Remove `resources.enabled` value
* Move `proxy_on_nodes` to flatten


majewsky (approved):
Looks like a viable strategy overall. I assume the resource config is going to be a third argument for `swift_proxy_container`? (Which, by the way, should be called `swift_proxy_containers` in plural.)

majewsky (approved):

PR #2226, opened 2021-02-03
[Octavia] Add housekeeping

laerlingSAP:
The housekeeping pod exists but housekeeping is never configured in the config. That's why we have many thousands of deleted load balancers just lying around.

Edit: There are in fact default values, but the expiry age is one week.

### DIFF
--- a/openstack/octavia/requirements.lock
+++ b/openstack/octavia/requirements.lock
@@ -20,5 +20,8 @@ dependencies:
 - name: utils
   repository: file://../../openstack/utils
   version: 0.1.1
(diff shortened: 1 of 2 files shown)


notandy (changes requested):
Please add default values in chart's values.yaml (and check the helm linting)

notandy (approved):

laerlingSAP:
Defaults aren't necessary; For those regions where `house_keeping` is not defined in `octavia.yaml`, it will just not write any housekeeping config into `octavia.conf`. And the provider driver [already sets defaults anyway](https://github.com/sapcc/octavia/blob/stable/stein-m3/octavia/common/config.py#L452-L460).

PR #2229, opened 2021-02-04
[swift] proxy deplyoment per Availabilty Zone

reimannf:
Option to create swift proxy deployments per AZ.
The initial amount of replicas per AZ can be specified, otherwise default to 0.

Scheduling
* Only schedule on nodes current az
* Do not schedule to storage nodes (missing tolerations)
* Prefere to be scheduled on nodes not yet running this proxy from a deplyoment

### DIFF
--- a/openstack/swift/templates/proxy-deployment.yaml
+++ b/openstack/swift/templates/proxy-deployment.yaml
@@ -11,7 +11,7 @@ metadata:
     os-cluster: {{ $cluster_id }}
 spec:
   revisionHistoryLimit: 5
-  {{- $replicas_az := default 1.0 (index $cluster (printf "proxy_deployment_replicas_%s" $az)) }}
+  {{- $replicas_az := default 0.0 (index $cluster (printf "proxy_deployment_replicas_%s" $az)) }}
   replicas: {{ $replicas_az }}
   strategy:
     type: RollingUpdate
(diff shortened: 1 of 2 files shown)


openstack/swift/templates/proxy-deployment.yaml
```diff
@@ -1,32 +1,34 @@
...
+  name: swift-proxy-{{ $cluster_id }}-{{ $az }}
   labels:
-    release: "{{$.Release.Name}}"
-    os-cluster: {{$cluster_id}}
+    release: "{{ $.Release.Name }}"
+    os-cluster: {{ $cluster_id }}
 spec:
   revisionHistoryLimit: 5
-  replicas: {{ required "proxy_deployment_replicas is required" $cluster.proxy_deployment_replicas }}
+  {{- $replicas_az := default 1.0 (index $cluster (printf "proxy_deployment_replicas_%s" $az)) }}
```
majewsky, line 14: Default should be 0 in my opinion.

openstack/swift/templates/proxy-deployment.yaml
```diff
@@ -43,23 +45,42 @@ spec:
...
+            - matchExpressions:
+              - key: failure-domain.beta.kubernetes.io/zone
                 operator: In
                 values:
-                - swift-proxy-{{$cluster_id}}
-              - key: from
-                operator: In
+                - {{ $az }}
+          # Prefere non storage nodes for scheduling
+          preferredDuringSchedulingIgnoredDuringExecution:
```
majewsky, line 60: Would prefer to have this as a hard requirement, to avoid surprises.

openstack/swift/templates/proxy-deployment.yaml
reimannf: done

openstack/swift/templates/proxy-deployment.yaml
reimannf: done

majewsky (approved):

PR #2231, opened 2021-02-05
[osprofiler] clean up helper functions

chuan137:
This PR clears old osprofiler setup with flamegraph-redis.

Also in the old setup, both global value `global.osprofiler.enabled` and service local `osprofiler.enabled` are checked in the helper function definition. It is bit confusing to control the integration of osprofiler, since osporfiler is not enabled for all services and one has to edit different places. **With this PR, osprofiler integration can only be enabled explicitly by the service local value `osprofiler.enabled`.**

This change will affect number of services with osprofiler integration in helm chart: `keystone`, `cinder`, `nova`, `nuetron` and `ironic`. Therefore I am asking the service owners to review this PR and be aware of the changes.

### DIFF
--- a/openstack/utils/templates/_helpers.tpl
+++ b/openstack/utils/templates/_helpers.tpl
@@ -31,30 +31,22 @@ propagate=0
 {{- end }}
 
 {{- define "osprofiler" }}
-    {{- $options := merge .Values.osprofiler .Values.global.osprofiler }}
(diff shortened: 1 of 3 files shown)


laerlingSAP (approved):

notandy:
@chuan137 how the status, can it be merged?

PR #2240, opened 2021-02-12
[vrops-exporter] adding new sleep and timout defaults

viennaa:
inventory is way faster now and we can go all in


richardtief (approved):
adjustable parameters are always better

PR #2241, opened 2021-02-12
[alertmanager] add metal alert routing to OSD compute

geisslet:
add missing metal alerts - same PD service requested as vmware


geisslet:
@auhlig friendly reminder

geisslet:
@auhlig friendly reminder 
! OSD needs metal alerts, too ! - they and up in the same PD service - that's why the same key is used

auhlig (approved):

PR #2251, opened 2021-02-16
adding cronus simulator template

dhalimi:
Adding new template under openstack/cronus for cronus-simulator


openstack/cronus/values.yaml
```diff
@@ -187,3 +188,30 @@ ingress-nginx:
...
+    podManagementPolicy: OrderedReady
+    annotations: {}
+    labels: {}
+    headless:
+      annotations: {}
+      labels: {}
+      servicePort: 80
+  updateStrategy: RollingUpdate
+  replicas: 1
+  nebula:
```
kayrus, line 210: what should be under the nebula? Empty string?

openstack/cronus/values.yaml
```diff
@@ -187,3 +188,30 @@ ingress-nginx:
...
+
+simulator:
+  enabled: true
+  image:
+    name: cronus-exporter-repository
+    pullPolicy: IfNotPresent
+    tag: latest
+    imagePullSecrets: []
+  statefulSet:
+    enabled: true
```
kayrus, line 200: I think there is no need to use statefulsets here. You don't use locks or anything else.

openstack/cronus/values.yaml
dhalimi: Yes, it gives the option to use a local nebula for debugging

openstack/cronus/values.yaml
dhalimi: I've disabled in secrets

kayrus:
LGTM

kayrus (approved):

PR #2252, opened 2021-02-17
[kube-system-metal] fix antiAffinity label keys

reimannf:
Ensuring the podAntiAffnity with updated label key

### DIFF
--- a/system/kube-system-metal/Chart.yaml
+++ b/system/kube-system-metal/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Kube-System relevant Service collection for metal clusters.
 name: kube-system-metal
-version: 0.12.10
+version: 0.12.11
 dependencies:
   - name: cc-rbac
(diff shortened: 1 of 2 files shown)


reimannf:
Would it be easily possible to have ingress pods spread out to Availability Zones? E.g. one deployment per AZ like here https://github.com/sapcc/helm-charts/blob/master/openstack/swift/templates/proxy-deployment.yaml#L47-L48

auhlig:
Nice @reimannf!

Pods of a single Deployment can be spread across AZ would be possible via:
```
# See https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints .
topologySpreadConstraints:
  - maxSkew: 1
     topologyKey: topology.kubernetes.io/zone
     whenUnsatisfiable: ScheduleAnyway
```

So we would not have to introduce a 2nd Deployment.
However, the current ingress chart doesn't support that and needs updating. 
I do that in a separate step.

PR #2259, opened 2021-02-23
Add pvc exporter helm chart

GlaciErrDev:


prometheus-exporters/pvc-exporter/values.yaml
```diff
@@ -0,0 +1,18 @@
...
+  # Name of the Prometheus by which this exporter is scraped.
+  prometheus: collector-kubernetes
+
+resources:
+  requests:
+    memory: 40Mi
+    cpu: 10m
+  limits:
+    memory: 100Mi
+    cpu: 20m
```
galkindmitrii, line 18: I think that is too low, the process might not be able to start. Please set to either 100m or 200m.

galkindmitrii (approved):

PR #2270, opened 2021-02-26
[neutron] use neutron service-account for accessing nova and designate

notandy:
This changes neutron.conf to use it's own neutron service account for accessing foreign services instead of foreign ones. This makes requester easier identifiable and removes external dependency global service passwords.

Please ensure, that the neutron service user has all needed roles to access the respective service.


joker-official (changes requested):
I think `neutron` needs to become `cloud_dns_admin` for this to work.

joker-official:
Also, I would really prefer if the PR description would also be in the commit message. It's easier to find that way in a couple of years.

joker-at-work:
`neutron` is still missing the appropriate rights to access designate as admin, afaics.

notandy:
I added the missing role for neutron: https://github.com/sapcc/helm-charts/commit/4f1bd6bc9f2323afc955096d78e92962cad9d58e

going to merge this commit

galkindmitrii:
Thanks Andy!

PR #2278, opened 2021-03-02
[Cinder]  Remove the cinder nanny from seed

hemna:
This patch removes the cinder nanny from the cinder seed template
as it already exists in the nannies chart.


thgrs (approved):
lgtm

PR #2280, opened 2021-03-02
[Cinder] remove cinder_nanny seeder roles

hemna:
This patch removes the dependent cinder_nanny user role assignments
for the seeder.  These are done in the nannies charts now.


thgrs (approved):
lgtm

PR #2291, opened 2021-03-03
[cinder] Add configmap hashes as annotations

joker-at-work:
This should make the cinder-volume pods restart if one of their
configmaps changes, but the deployment doesn't. This follows the
principle already in use by Nova.


hemna (approved):
lgtm

PR #2297, opened 2021-03-04
[openstack-exporter] Update chart to not use global master_password

hemna:
This patch updates the chart for openstack-exporter to use the
.Values.global.exporter_master_password instead of the
global master_password


viennaa (changes requested):
Don't merge, let's clarify first, which pw to use.

joker-at-work (commented):
I don't think it's necessary to use a master password here. Just define the password explicitly in the secrets. We only have one host we generate it for and thus imho don't gain anything by using the master-password algorithm.

Bare in mind, that you need to change the password in keystone by hand when rolling this out. The seeder will not change the password for you.

viennaa:
Already talked to @hemna that he would need to change the seed file too and make sure it is created correctly in the regions. 

Static works for me too. If the user stays the same, the seed file needs to be updated and the user  thenmanually removed from all affected keystones, to have the seeder recreate it.

joker-at-work:
Using `os user set --domain <user-domain> --project <project-name> --project-domain <project-domain> openstack_exporter  --password <new_password>` seems simpler and faster than deleting and waiting for the seeder.

viennaa:
I still would add it to the seeder, otherwise it's going to be a problem in a new region.

joker-at-work:
Yes. We definitely need to change the seeder to seed the new password. It's just for deploying the change, that I find it easier to do as described.

BugRoger:
I saw updates on the Seeder. Can we move this issue along?

hemna:
so What I'm gleaning from this conversation is that

1. I have to manually create a new password for the openstack_exporter user for every region with a call to the openstack client
2. Update each region secrets file to add the password
3. update the helm-chart to pull the password from the region in the openstack-exporter-seed.yaml 

Do we have a password generator for completing step 1?

viennaa:
1. use the master password set in each region `values.global.exporter_master_password` to generate a pw and set it in the region you want to rollout.
2. then roll out the exporter with the pipeline for this region, after this PR is merged.
3. seeder chart need to be updated, to have the correct pw mapping.

BugRoger:
@viennaa @hemna Where are we on this topic? Can I help in any way to move it along?

hemna:
> @viennaa @hemna Where are we on this topic? Can I help in any way to move it along?

I haven't been able to spend any time on this as I've been stuck working on a critical failure with cinder backup.

hemna:
The changes in this PR have already been supeceeded.

PR #2299, opened 2021-03-04
[swift] increase nginx proxy timeout

reimannf:
Esp. during high load situations, e.g. replication, there can be situations
where a proxy request calling out to its backend servers running in the
node_timeout. This leads to long runners and nginx timeout kicked in too
early. The result was a 504 for the client and a swift proxy request finishing
after that timeout.


majewsky:
thanks

PR #2320, opened 2021-03-12
[nova] Get rid of master_password

joker-at-work:


Carthaca (approved):
lgtm

rajivmucheli:
We need to remove `identity.password_for_user` at https://github.com/sapcc/helm-charts/blob/master/openstack/nova/templates/etc/_nova-compute-ironic.conf.tpl#L22, right ?

joker-at-work:
> We need to remove `identity.password_for_user` at https://github.com/sapcc/helm-charts/blob/master/openstack/nova/templates/etc/_nova-compute-ironic.conf.tpl#L22, right ?

True. Forgot that that's in there, because we plan to switch away from the `ironic` user to the `nova` user when accessing Ironic. This needed rollout of additional roles, first. I'll prepare the PR to switch the user and link it here afterwards.

joker-at-work:
I've pushed the change directly: https://github.com/sapcc/helm-charts/commit/06d67d747358cfc3de3e4893b424c8220c8141a9

PR #2324, opened 2021-03-16
Esxi host exporter

vulrandier:
Adapted old esxi-exporter config to new esxi-host-exporter


prometheus-exporters/esxi-host-exporter/values.yaml
```diff
@@ -0,0 +1,13 @@
+enabled: false
+listen_port: 9666
+docker_repo: DEFINED-IN-REGION
+image_version: 1.0.0
```
christopherhans, line 4: You have this version in your secret file. So you can also write `DEFINED-IN-REGION` here.

christopherhans (changes requested):

PR #2328, opened 2021-03-17
[cinder] Update the default rpc_response_timeout to 600

hemna:
This patch doubles the rpc_response_timeout.   All regions have
already been updated manually to 600.


rajivmucheli (approved):
lgtm

PR #2344, opened 2021-03-19
[cinder] add name to resource filter

hemna:
This patch updates cinder's resource_filters.json
file to include name in the volume filter keys.
The train release removed a deprecated feature
that causes the openstack client to not filter
volumes correctly as it doesn't support microversions
in the cinder api.


joker-at-work (changes requested):
That's not enough. As mentioned in slack, `snapshots` key also has a `name~` and thus probably needs a `name`.

joker-at-work (approved):

PR #2345, opened 2021-03-21
add new flags to cronus-simulatior to support changes in version v0.0.3

dhalimi:
The change is in cronus-simulator, adding new flags to support v0.0.3


kayrus (approved):
LGTM

Issue #2293, opened 2021-03-03
[utils/identity] Replace Global Master Password 


BugRoger:
https://github.com/sapcc/helm-charts/blob/master/openstack/utils/templates/_hosts.tpl#L136

- Remove all references to `Values.global.master_password` 
- Prefer to not reference a global value. Instead require a password being passed to the chart

Prior Art:
https://github.com/sapcc/helm-charts/commit/a4c48f4160d81b1f3bbdb356f7526c431763ca4a

Take note that this utility function is being used in many other charts:

- `utils.password_for_fixed_user_and_host`
  - `identity.password_for_user`
    - https://github.com/sapcc/helm-charts/blob/master/openstack/barbican/templates/etc/_barbican.conf.tpl
    - https://github.com/sapcc/helm-charts/blob/master/openstack/cinder/templates/etc/_cinder.conf.tpl
    - https://github.com/sapcc/helm-charts/blob/master/openstack/glance/templates/etc/_glance-api.conf.tpl
    - https://github.com/sapcc/helm-charts/blob/master/openstack/glance/templates/etc/_swift-store.conf.tpl
    - https://github.com/sapcc/helm-charts/blob/master/openstack/ironic/templates/etc/_ironic.conf.tpl
    - https://github.com/sapcc/helm-charts/blob/master/openstack/ironic/templates/etc/_ironic_inspector.conf.tpl
    - https://github.com/sapcc/helm-charts/blob/master/openstack/ironic/templates/seed.yaml
    - https://github.com/sapcc/helm-charts/blob/master/openstack/neutron/templates/etc/_ironic_neutron_agent.ini.tpl
    - https://github.com/sapcc/helm-charts/blob/master/openstack/nova/templates/_console_deployment.yaml.tpl
    - https://github.com/sapcc/helm-charts/blob/master/openstack/nova/templates/etc/_nova-compute-ironic.conf.tpl
    - https://github.com/sapcc/helm-charts/blob/master/openstack/nova/templates/etc/_nova.conf.tpl
    - https://github.com/sapcc/helm-charts/blob/master/openstack/nova/templates/seed.yaml
  - `svc.password_for_user_and_service`  
    - `oslo_messaging_rabbit_url` (unused)

Coordinated deployments might be required. :/

joker-at-work:
I will remove autogenerated defaults in Nova and switch to explicit values. It just needs some more time as preparations are necessary. Once I got a PR up, I'll link it.

[...]

PR #2350, opened 2021-03-23
[Neutron] Remove master_password

rajivmucheli:
Hi @notandy,

Please validate this PR, its related to https://github.com/sapcc/helm-charts/issues/2293

Regards,
Rajiv

### DIFF
--- a/openstack/neutron/templates/etc/_ironic_neutron_agent.ini.tpl
+++ b/openstack/neutron/templates/etc/_ironic_neutron_agent.ini.tpl
@@ -10,6 +10,6 @@ auth_url = {{.Values.global.keystone_api_endpoint_protocol_internal | default "h
 project_name = {{ default "service" .Values.global.keystone_service_project }}
 project_domain_name = {{ default "Default" .Values.global.keystone_service_domain }}
 username = {{ $user }}
-password = {{ coalesce .Values.global.ironicServicePassword .Values.global.ironic_service_password  (tuple . $user | include "identity.password_for_user")  | replace "$" "$$" }}
+password = {{ coalesce .Values.global.ironicServicePassword .Values.global.ironic_service_password }}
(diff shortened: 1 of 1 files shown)


openstack/neutron/templates/etc/_ironic_neutron_agent.ini.tpl
```diff
@@ -10,6 +10,6 @@ auth_url = {{.Values.global.keystone_api_endpoint_protocol_internal | default "h
 project_name = {{ default "service" .Values.global.keystone_service_project }}
 project_domain_name = {{ default "Default" .Values.global.keystone_service_domain }}
 username = {{ $user }}
-password = {{ coalesce .Values.global.ironicServicePassword .Values.global.ironic_service_password  (tuple . $user | include "identity.password_for_user")  | replace "$" "$$" }}
+password = {{ coalesce .Values.global.ironicServicePassword .Values.global.ironic_service_password }}
```
notandy, line 13: As far as I see, only `ironic_service_password` is used in the globals, can we use that solely? Also, please use required macro, e.g. `password = {{ required ".Values.global.ironic_service_password required!" .Values.global.ironic_service_password }}`. Thanks!

notandy (changes requested):

notandy:
Oh btw, could we even change this to the neutron credentials and check if neutron user has the required roles?

openstack/neutron/templates/etc/_ironic_neutron_agent.ini.tpl
rajivmucheli: i wasnt aware which parameter is used for password, hence i did not specify required only for this PR. 

i dont think we can get rid of ironic service password in the globals, since its used in other ironic services https://github.com/sapcc/helm-charts/pull/2348/files

notandy (approved):
We anyway don't use the neutron-ironic-agent, but change is fine!

PR #2359, opened 2021-03-25
HostWithRunningVMsNotResponding decreasing wait to 5m

christopherhans:
Change requested by Martin P.


christopherhans:
@richardtief found bug in exporter code. After fix we will observe if the alert performs as expected.

PR #2364, opened 2021-03-29
[mariadb-backup] enables dynamic storage setups

stefanhipfel:
makes it possible to define a dynamic list storages per region


stefanhipfel:
do not merge. secrets need to be changed first

PR #2365, opened 2021-03-29
[keystone] Yaml policies fix

bbobrov:
Reintroduce rule:owner and rule:token_subject to fix credentials and token-validation endpoints

### DIFF
--- a/openstack/keystone/templates/etc/_policy.yaml.tpl
+++ b/openstack/keystone/templates/etc/_policy.yaml.tpl
@@ -14,7 +14,7 @@
 #"admin_or_owner": "rule:admin_required or rule:owner"
 
 #
-#"token_subject": "user_id:%(target.token.user_id)s"
+"token_subject": "user_id:%(target.token.user_id)s"
 
 #
 #"admin_or_token_subject": "rule:admin_required or rule:token_subject"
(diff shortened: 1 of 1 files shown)


bbobrov:
linter will be happy after #2366

openstack/keystone/templates/etc/_policy.yaml.tpl
```diff
@@ -8,13 +8,13 @@
 "service_or_admin": "rule:admin_required or rule:service_role"
 
 #
-#"owner": "user_id:%(user_id)s"
```
Carthaca, line 11: Can you please a comment with the default policy so we can compare our overwritten rules with the original ones (even if they are the same)?

openstack/keystone/templates/etc/_policy.yaml.tpl
```diff
@@ -8,13 +8,13 @@
...
+"owner": "user_id:%(user_id)s"
 
 #
 #"admin_or_owner": "rule:admin_required or rule:owner"
 
 #
-#"token_subject": "user_id:%(target.token.user_id)s"
+"token_subject": "user_id:%(target.token.user_id)s"
```
Carthaca, line 17: same like above

openstack/keystone/templates/etc/_policy.yaml.tpl
bbobrov: I can, but it will be different from other entries (see "service_role" or "admin_required" above on line 5). Do you still think i need to do it?

Carthaca: Indeed, I think it should be added for those lines, too, then :)

Carthaca (approved):
In the end we should not let the discussion about comments block the deployment of the fix

PR #2369, opened 2021-03-29
nova/flavors: Introduce new g/c/m naming scheme

grandchild:
**nova/flavors: Remove 15 discontinued flavors**

The discontinued BigVM flavors remain, until replaced with matching HANA flavors.

Some of the removed flavors will be replaced with new flavors in a later commit.

---

**nova/flavors: Move OS default flavors to sap-seed**

Move "m1.medium", "m1.large", "m1.xlarge" from nova to sap-seeds flavor-seed list, unifying the lists into one.

Two flavors where not moved but removed:
  - `m1.tiny` is discontinued
  - `m1.small` was a duplicate and already existed in the sap-seed.

---

**nova/flavors: Rename & alias flavors to g/c/m scheme**

Rename all public, non-hana-used flavors to the new naming scheme and add the old name(s) as "config:alias" extra_spec.

This effectively deprecates the old flavor names, however they will still appear in `openstack flavor list` but with a `flavorid` starting with `x_deprecated_`, and they will deploy as the new flavors. This allows automation to continue using the old names for now.

---

**nova/flavors: Add new HANA flavors with NUMA traits**

HANA flavors require a fitting `CUSTOM_NUMASIZE_*` trait, that can be applied to hosts that provide the appropriate NUMA node size.

`hana_c384_m5835` is commented out because it is dependent on an updated VSphere for supporting more than 256 vCPUs.

### DIFF
--- a/openstack/nova/templates/flavor-seed.yaml
+++ b/openstack/nova/templates/flavor-seed.yaml
@@ -4,50 +4,5 @@ kind: "OpenstackSeed"
 metadata:
   name: nova-flavor-seed
 spec:
-  flavors:
-  - name: "m1.tiny"
-    id: "10"
-    vcpus: 1
-    ram: 508
-    disk: 1
-    is_public: true
-    extra_specs:
-      "vmware:hv_enabled": "True"
-      "hw_video:ram_max_mb": "4"
-  - name: "m1.small"
-    id: "20"
-    vcpus: 2
-    ram: 2032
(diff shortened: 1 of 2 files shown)


grandchild:
Can we remove `openstack/nova/templates/flavor-seed.yaml`?

dhoeller (approved):
LGTM

PR #2377, opened 2021-03-30
[keystone] fix policies by copying them from json

bbobrov:
It is too difficult to set up the new policies so that they don't break anything. Restore access to the level it was before and to set the base for further improvements.

Use policies from the old json file to restore access to the service to pre-change level.


Carthaca (approved):
lgtm

PR #2394, opened 2021-04-07
[keystone] Use default MaxRequestsPerChild (0)

bbobrov:
Fixes CCM-18723

Listener process serves MaxRequestsPerChild requests, shuts down,
closes all open connections, we run into the race condition described
in the bugreport. Using the default value will remove this limitation


databus23 (approved):
Good find!

Carthaca (approved):
Interesting, this was just copied from upstream https://github.com/openstack/openstack-helm/blob/34c3cd4adf8bdfcc7f960620f667e1f43f4bb9b2/keystone/values.yaml#L984

PR #2396, opened 2021-04-08
Add parameter to provide internal region specific cache scrape interval

mithunmgore:


mithunmgore:
@viennaa @christopherhans 
want to have different internal cache scrape interval for each region. Problem is, currently the ucs-exporter getting restart in the regions where ucsm servers are more than approx 10 servers (eu-de-1, eu-de-2).

prometheus-exporters/ucs-exporter/templates/ucs-exporter-deployment.yaml
```diff
@@ -28,6 +28,8 @@ spec:
              - {{ required ".Values.ucs.user missing" .Values.ucs.user }}
              - -p
              - {{ required ".Values.global.exporter_master_password missing" .Values.global.exporter_master_password }}
+             - -s
+             - {{ .Values.ucs.internal_cache_scrape_interval }}
```
viennaa, line 32: Please add this mapping in `values.yaml` with a default value for "normal" regions.

viennaa (changes requested):

prometheus-exporters/ucs-exporter/templates/ucs-exporter-deployment.yaml
BerndKue: What happens if not defined in secrets?
How about writing therefor
{{ if .Values.ucs.internal_cache_scrape_interval }}{{ .Values.ucs.internal_cache_scrape_interval }}{{ else }}300{{ end }}

mithunmgore: This one is not specified as "required", so exporter will go with default value(300), I don't think loop is required to specify this. 
@viennaa please confirm.

viennaa: The whole setting starting with `-s` need to be in the if else block. Default is 300, so nothing need to be specified indeed: https://github.com/sapcc/ucs-exporter/blob/master/ucs-exporter.py#L51-L52

So I'd go with 

```
{{ if .Values.ucs.internal_cache_scrape_interval }}
- -s
- {{ .Values.ucs.internal_cache_scrape_interval }}
{{ end }}
```

See comment in values.yaml.

prometheus-exporters/ucs-exporter/values.yaml
```diff
@@ -16,6 +16,7 @@ ucs:
    query: "ucsa"
    status: "active"
    debug: "0"
+   internal_cache_scrape_interval: DEFINED-IN-REGION-SECRETS
```
viennaa, line 19:
```suggestion
   # defaults to 300, only use if needed
   # internal_cache_scrape_interval: DEFINED-IN-REGION-SECRETS 
```

BerndKue:
@mithunmgore If you make the changes like proposed by Tommy, I would be fine with merging.

mithunmgore:
Done.

viennaa (approved):

BerndKue:
@mithunmgore Thanks a lot!

PR #2401, opened 2021-04-12
[glance] set the number of threads for glance_store and set number of

galkindmitrii:
workers.

- needed for https://github.com/sapcc/glance_store/pull/2
- set 4 workers per replica as 10 by default caused issues
- increase number of replicas as number of workers was reduced.

### DIFF
--- a/openstack/glance/templates/etc/_glance-api.conf.tpl
+++ b/openstack/glance/templates/etc/_glance-api.conf.tpl
@@ -14,6 +14,7 @@ admin_role = ''
 
 rpc_response_timeout = {{ .Values.rpc_response_timeout | default 300 }}
 rpc_workers = {{ .Values.rpc_workers | default 1 }}
+workers = {{ .Values.workers | default 4 }}
 
 wsgi_default_pool_size = {{ .Values.wsgi_default_pool_size | default 100 }}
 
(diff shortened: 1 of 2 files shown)


rajivmucheli:
Hi, has the buffer scenario been tested ? we may end up consuming 4x memory (i.e. 4 workers are enabled)

(swift_store_large_object_chunk_size * workers * 1000)

https://docs.openstack.org/glance/latest/configuration/glance_api.html#glance.store.swift.store.swift_buffer_on_upload

galkindmitrii:
> Hi, has the buffer scenario been tested ? we may end up consuming 4x memory (i.e. 4 workers are enabled)
> 
> (swift_store_large_object_chunk_size * workers * 1000)
> 
> https://docs.openstack.org/glance/latest/configuration/glance_api.html#glance.store.swift.store.swift_buffer_on_upload

Hi. Yes, before there were 10 workers per pod. Even though the documentation suggests it should be limited to max 8.

> Provide a non-negative integer value to set the number of child process workers to service requests. By default, the number of CPUs available is set as the value for workers limited to 8. For example if the processor count is 6, 6 workers will be used, if the processor count is 24 only 8 workers will be used. The limit will only apply to the default value, if 24 workers is configured, 24 is used.

Tests made by @GlaciErrDev showed there were issues with threaded upload with 10 workers, but not with 8 or less.
IMO, it makes sense to reduce the number of workers and increase the number or replicas. This also means the max disk usage per pod should be less.

rajivmucheli (approved):
lgtm

PR #2404, opened 2021-04-12
[Keystone] Add keystone-mariadb pvc consumption alert

rajivmucheli:
Hi @Carthaca 

Please validate this PR. Its regarding [task](https://issues.wdf.sap.corp/browse/CCM-12503) 

Regards,
Rajiv

### DIFF
--- a/openstack/keystone/alerts/kubernetes/keystone.alerts
+++ b/openstack/keystone/alerts/kubernetes/keystone.alerts
@@ -19,7 +19,6 @@ groups:
 
   - alert: OpenstackKeystoneDatabaseSize
     expr: predict_linear(pvc_usage{persistentvolumeclaim="db-keystone-pvclaim"}[1h], 24 * 3600) * 100 >= 90
-    expr: 
     for: 5m
     labels:
       tier: os
--- a/openstack/keystone/alerts/kubernetes/keystone.alerts
+++ b/openstack/keystone/alerts/kubernetes/keystone.alerts
@@ -18,8 +18,8 @@ groups:
       summary: 'Keystone API ingress endpoint is not healthy'
 
   - alert: OpenstackKeystoneDatabaseSize
-    expr: predict_linear(pvc_usage{persistentvolumeclaim="db-keystone-pvclaim"}[1h], 24 * 3600) * 100 >= 90
-    for: 5m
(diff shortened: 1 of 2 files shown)


openstack/keystone/alerts/kubernetes/keystone.alerts
```diff
@@ -16,3 +16,18 @@ groups:
     annotations:
       description: 'The keystone ingress API endpoint is throwing 5xx for 5 min. Check if the keystone-api k8s pods are OK.'
       summary: 'Keystone API ingress endpoint is not healthy'
+
+  - alert: OpenstackKeystoneDatabaseSize
+    expr: predict_linear(pvc_usage{persistentvolumeclaim="db-keystone-pvclaim"}[1h], 24 * 3600) * 100 >= 90
```
Carthaca, line 21: Prediction is based on the data of the last hour and forecasting for the next 24 hours.

I would like to change two things:
1. increase the range of base data, e.g. to tell what might happen in the next 24 hours look back what happened in the last 6 hours to smoothen extreme fluctuations
2. increase the time to react: currently we would be warned 1 day upfront (and I bet this will happen on the weekend). Better warn 1 week ahead instead

Combined I imagine something like
`predict_linear(pvc_usage{persistentvolumeclaim="db-keystone-pvclaim"}[1d], 7 * 24 * 3600) * 100 > 90`

"If it grows for 7 days like it has grown for the past day, the pvc would be used more than 90 percent"

And maybe add a guard to prevent false positives:
`(pvc_usage{persistentvolumeclaim="db-keystone-pvclaim"} * 100 > 70) and predict_linear...`
"only warn if currently already 70 percent is used"

openstack/keystone/alerts/kubernetes/keystone.alerts
```diff
@@ -16,3 +16,18 @@ groups:
...
+    for: 5m
```
Carthaca, line 22: increase to avoid flapping, I suggest `for: 1h`

openstack/keystone/alerts/kubernetes/keystone.alerts
```diff
@@ -16,3 +16,18 @@ groups:
...
+    labels:
+      tier: os
+      service: keystone
+      severity: warning
+      context: availability
+      dashboard: mariadb-overview?var-host=keystone-mariadb
+      meta: "The keystone database is predicted to run out of space in the next 24h"
+      playbook: 'docs/devops/alert/keystone/#database_size'
```
Carthaca, line 30: the playbook text needs to match the alert in the end :)

Carthaca (changes requested):
Thanks for picking this up, I would like to make this alert less noisy, please see my inline comments

Carthaca:
superseded by #2415

PR #2408, opened 2021-04-14
vrops-exporter: HostWithRunningVMsNotResponding changed running VM metric

christopherhans:
Changed the count of powered on VMs (VM collector) to the amount of running VMs on the host (host collector). This metric only considers powered on VMs. 

https://docs.vmware.com/en/vRealize-Operations-Manager/8.3/com.vmware.vcom.core.doc/GUID-C3CAAE15-2E83-431F-8F2D-C5297A3B6EA9.html : `This excludes powered off VMs as they do not impact ESXi performance.`

In that way we don't miss the critical alert, in case the VM collector doesn't work as expected.


richardtief:
Ok, make sense to me.

viennaa (approved):

PR #2409, opened 2021-04-14
[neutron] Add subgroups to ACI config

sebageek:
Subgroups or direct-mode hostgroups (hostgroups with setting
direct_mode = True) can now be specified. They belong to a parent group
and allow single nodes to be switched between infrastructure mode and
baremetal mode inside the ACI driver.


defo89 (approved):
lgtm

PR #2413, opened 2021-04-15
Allow updating imagePullPolicy

hemna:
This patch parameterizes the imagePullPolicy to allow a region to
change the default of 'IfNotPresent' to something else.  This allows
for testing patches against qa-de-3 and have the deployment always pull
the image, thus avoiding stale cached images.


joker-at-work (approved):
I don't see this as necessary, but if you think it helps, go for it.

PR #2417, opened 2021-04-16
[cinder] Fix the cinder backup_deployment template

hemna:
After trying this in qa-de-3, the backup service didn't deploy. I
believe it's due to this templating format.


joker-at-work (approved):
I believe you're right. We should switch it to using other template-start strings for jinja2 like we did for the other VCenterTemplate objects.

PR #2420, opened 2021-04-19
[openstack-exporter][cinder] Alter cinder alerts no absense

hemna:
This patch adds the no_alert_on_absence: "true" to 2 of the cinder
alerts for thin provisioning.   Cinder has migrated many of the
volume types for regions to thick provisioning and those regions
will not have the thin provisioning metrics.  This patch ignores
those missing metrics as they are expected.

Once all the regions have migrated to thick provisioning, then these
2 alerts will be removed/replaced.


joker-at-work:
Sorry, Walt, I failed to look at this at the time. Do you think it's still necessary?

PR #2425, opened 2021-04-22
[alertmanager] merge inhibition rules

geisslet:
**issue**: alerts with different name, but same context (often used) are inhibit - alertname should key if severity is in

comment: ["Semantically, a missing label and a label with an empty value are the same thing. Therefore, if all the label names listed in equal are missing from both the source and target alerts, the inhibition rule will apply."](https://prometheus.io/docs/alerting/latest/configuration/#:~:text=An%20inhibition%20rule%20mutes%20an,names%20in%20the%20equal%20list)

issue-owner: @christopherhans @max-len

### DIFF
--- a/global/prometheus-alertmanager-operated/Chart.yaml
+++ b/global/prometheus-alertmanager-operated/Chart.yaml
@@ -3,4 +3,4 @@ appVersion: "1.0"
 type: application
 description: Prometheus Alertmanager via operator.
 name: prometheus-alertmanager-operated
-version: 1.0.4
+version: 1.0.5
--- a/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
+++ b/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
@@ -9,29 +9,13 @@ inhibit_rules:
       severity: 'critical'
     target_match:
       severity: 'warning'
-    equal: ['region', 'alertname', 'cluster']
+    equal: ['region', 'alertname', 'cluster', 'context']
 
   - source_match_re:
       severity: 'critical|warning'
     target_match:
       severity: 'info'
-    equal: ['region', 'alertname', 'cluster']
-
-  - source_match_re:
-      severity: 'critical'
-      context: '.+'
-    target_match_re:
-      severity: 'warning'
-      context: '.+'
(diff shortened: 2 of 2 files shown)


auhlig:
That looks like a problem introduced by using too generic alertnames. 
Can you give an example?

christopherhans:
"DataStoreCapacityWarning" is suppressing "VMsOnFailoverHost".

https://supernova.global.cloud.sap/?category=Productive&label_region=eu-nl-1&label_alertname=VMsOnFailoverHost&display=dashboard&layout=standard

max-len:
The current configuration suppresses alerts of higher severity, if they have the same 'context' (but still different alertnames).
We have a lot of different alerts coming coming from vrops-exporter, which is set as context. Other contexts seem also quite generic (nanny, availability)

For example a critical 'VM down' alert would suppress a warning 'datastore capacity' alert in the same region and cluster, because of same context. They're both needed, though.

auhlig:
This works as expected and described in our ops documentation.

Critical alerts per `region`, `context`, `cluster` would inhibit less criticals ones.
The idea of this logic is to keep important alerts visible and avoid a flood of lesser, simultaneously firing alerts at the same time. 

This seems to be a misconfiguration of the alert from the above example @christopherhans.
The alert `DataStoreCapacityWarning` inhibits`VMsOnFailoverHost`use the rather generic `context = vrops-exporter`.
I'd recommend using a more alert-specific `context` than the exporter name.

max-len:
Okay, understood.
In case we don't want to use this feature - i.e. an alert should always fire - we can just leave the 'context' tag out of the alert definition, as the inhibition rule matches an existing 'context' value, correct?

auhlig:
Yes. I guess setting no `context` might be better than a bad one.

Where the same context would make sense:
Given the warning alert `DataStoreCapacity80PercentFull` and critical alert `DataStoreCapacityFull` and both firing,
the 80% full alert would be inhibited by the critical one. 
Only the critical one reaches slack/pagerduty.

geisslet:
will close the PR.
update the alert cookbook to reference the detailed label description (missing).
guess will do some investigation on how the context is used in the near future .. guess that isn't something people are aware of (hadn't that in mind, too - to be honest).
//thanksAndTakeCare

PR #2426, opened 2021-04-22
[blackbox-exporter] + [documentation-exporter] enable scrapping blackbox docs

dhalimi:


prometheus-exporters/blackbox-exporter/templates/_blackbox.yaml.tpl
```diff
@@ -66,3 +66,27 @@ modules:
         - '"CeleryAppVersionCheck":true'
         - '"CeleryAliveCheck":true'
         - '"problems":\[\]'
+  docs_home_content:
+    prober: http
+    timeout: 10s
+    http:
+      valid_status_codes: [200]
+      method: GET
+      fail_if_body_not_matches_regexp:
```
auhlig, line 75: Ok from my side.
Just mind that this quite explicit and probe will fail and trigger an alert as soon as you add/change a single character on that page.
Was hoping the response code is enough, no? What do you think @artherd42?

dhalimi: Hi, thank you, the content alerts will be a warning as the text is static at doesn't expect to change.

PR #2428, opened 2021-04-23
[vcsa-exporter] more precise alert context

christopherhans:


viennaa (commented):
This is too many entries for basically the same thing. I would leave it as is. If you would apply a similar pattern to every exporter than it would always be:
* xyz-exporter-warning
* xyz-exporter-critical

The information is redundant and you've got the "service"  in the alert text already

christopherhans:
The pattern in my PR is exporter-component-severity, i.e. vcsa-service-critical. New alerts could be like vcsa-api-critical etc.

Either we remove the context or change it somehow.

viennaa (approved):
Looks good, thanks.

PR #2429, opened 2021-04-23
[esxi-exporter] more precise alert context

christopherhans:


prometheus-exporters/esxi-exporter/alerts/service.alerts
```diff
@@ -8,7 +8,7 @@ groups:
       severity: info
       tier: vmware
       service: compute
-      context: "esxi-exporter"
+      context: "esxi-service"
```
viennaa, line 11:
```suggestion
      context: "esxi-service-critical"
```

To follow our vcsa pattern i would go this route. This is still info, so it doesn't really matter, but we can prepare it already

auhlig:
Recommendation: A context would be something more general like `availability`.
Whenever there's something impacting esxi availability (flapping with warning severity, explosion with critical severity) the lesser alert can be inhibited.

PR #2434, opened 2021-04-29
nova/flavors: Forbid normal flavors on future HANA-exclusive hosts

grandchild:
In the future there will be HANA-VM-exclusive clusters. Forbid all
normal flavors to spawn on these by setting a new
CUSTOM_HANA_EXCLUSIVE_HOST host trait as forbidden.

The trait will then only have to be set on the appropriate clusters
when we do introduce them and openstack placement will do the rest.

Note: Exclusion is _not_ set for legacy BigVM flavors, since they are
mostly used for HANA VMs as well.


dhoeller (approved):
LGTM

PR #2438, opened 2021-04-29
[alertmanager] qa routing fix

geisslet:
First shot was a fail.

Guess there qa alerts without severity killing the CommonLabels approach.
This seems to me dangerous that this happens in prod, too -- so propose: if this works well we should always use .Labels.severity

In parallel I will do a alert inventory asap.

### DIFF
--- a/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
+++ b/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
@@ -591,7 +591,7 @@ receivers:
 
   - name: slack_qa
     slack_configs:
-      - channel: '#alert-qa-{{"{{ .severity }}"}}'
+      - channel: '#alert-qa-{{"{{ .Labels.severity }}"}}'
         api_url: {{ required ".Values.slack.webhookURL undefined" .Values.slack.webhookURL | quote }}
         username: "Pulsar"
         title: {{"'{{template \"slack.sapcc.title\" . }}'"}}


auhlig:
Unfortunately this won't help. An alert is missing the severity label, which can't be addressed here but on the alert. 

Per convention all alerts must have a small set of labels (region,cluster, severity,etc.) for routing to the respective receiver(s). 
The mandatory label `severity` of an alert indicates how serious the issue is.
Without it proper routing of the alert cannot be expected. This is not a bug in the alertmanger but misconfiguration of the alert. More details described in our ops documentation.

geisslet:
I know and fully agree, but:

1. it seems if one firing alert doesn't have the severity, it kills the .CommonLabels.severity (this is what I try to address, because it kills the routing/content)

2. in parallel we need to have a validation
  2a. you proposed to have a check on release, but this need to be ensured in all pipelines - here I do have a lack of idea how to fix that in a proper approach (teach me, I will do)
  2b. I will do an inventory asap and poke people / fix it with severity

3. still think we should replace the .CommonLabels.severity by .Labels.severity just to make it more robust, shouldn't we?

auhlig:
Ah. I see what you're thinking, but the alertmanger doesn't work like that.

In short: Alerts without the label are not routed. Properly configured ones are not impacted.
Detail: As per configuration alerts are grouped by alertname, region, cluster etc. - for this group we can use the CommonLabels. So the group consists of the "same" misconfigured alert but from many regions and/or clusters in a region.
What you see in the log is this group failing, which is what I'd expect. 
This seems to be limited to an alert lingering somewhere in our qa-de-1 region. 

For the future mandatory labels on alerts should be validated ideally before deployment.
The GitHub PR tests on this repo already check Prometheus alerts syntax, but not the labels and could be extended. One would have immediate feedback when opening a PR.

This would also be something for OPA to prevent malformed alerts from being deployed in the first place.
Can you reach out to @majewsky for this?

geisslet:
ok, will do @majewsky 
but we need to do something with the totally wrong setup now ( - channel: '#alert-qa-{{"{{ .severity }}"}}' ).
Do I get it right, that you propose to stay on the .CommonLabels? - if so: i do a new PR

auhlig:
I agree. Can you find out and fix the alert with the missing label?

There isn't anything that can or should be done on the alertmanager in that regard. 
The alert is missing the label `severity`. 
Thus using `.CommonLabels.severity` or `.Labels.severity` does not make any difference.
Proper handling of alerts required the mandatory labels to be configured.

geisslet:
thanks @auhlig (+ sorry)
new PR: https://github.com/sapcc/helm-charts/pull/2439
close this here

PR #2440, opened 2021-04-30
[swift-proxy] enable 'delete_container_retry_count' for bulk operations.

galkindmitrii:
Large Glance images may include 100s of segments per container.
Deletion of such images sometimes often fails with 'HTTP 409 Conflict'
response from Swift. Enabling this option seem to resolve the problem.

The delete_container_retry_count parameter is used during a bulk delete of objects
and their container. This would frequently fail because it is very likely that
all replicated objects have not been deleted by the time the middleware got a
successful response. It can be configured the number of retries. And the
number of seconds to wait between each retry will be 1.5**retry

### DIFF
--- a/openstack/swift/templates/etc/_proxy-server.conf.tpl
+++ b/openstack/swift/templates/etc/_proxy-server.conf.tpl
@@ -169,6 +169,7 @@ url_base = https:
 
 [filter:bulk]
 use = egg:swift#bulk
+delete_container_retry_count = 3
 
 [filter:container-quotas]
 use = egg:swift#container_quotas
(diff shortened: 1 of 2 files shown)


rajivmucheli (approved):
lgtm

reimannf (changes requested):
As discussed already in a PM, that will not solve `409` when `glance` deletes images as it does sequential deletes of the segments and not bulk delete.
But anyhow, please define a configurable value `bulk_delete_container_retry_count = 3` in https://github.com/sapcc/helm-charts/blob/master/openstack/swift/values.yaml und use this instead.

galkindmitrii:
> As discussed already in a PM, that will not solve `409` when `glance` deletes images as it does sequential deletes of the segments and not bulk delete.
> But anyhow, please define a configurable value `bulk_delete_container_retry_count = 3` in https://github.com/sapcc/helm-charts/blob/master/openstack/swift/values.yaml und use this instead.

Sequential deletion happens when image upload fails due to an error.

We've tested this setting in qa-de-1 and deletion of large images did not cause 409 errors anymore.

PR #2446, opened 2021-05-04
[neutron] run metadata agent as sidecar on apod nodes

defo89:
* with aPod setup amount of cp nodes will increase drastically, and it does not make sense to run metadata agents as DaemonSet there, hence it will run as a sidecar
* if apod is enabled in the region, metadata agents are not scheduled on apod nodes

### DIFF
--- a/openstack/neutron/templates/statefulset-network-agent-apod.yaml
+++ b/openstack/neutron/templates/statefulset-network-agent-apod.yaml
@@ -136,6 +136,37 @@ spec:
               mountPath: /etc/sudoers
               subPath: sudoers
               readOnly: true
+        - name: neutron-metadata-agent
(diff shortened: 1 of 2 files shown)


openstack/neutron/templates/statefulset-network-agent-apod.yaml
```diff
@@ -136,6 +136,37 @@ spec:
...
+          livenessProbe:
+            exec:
+              command: ["neutron-agent-liveness", "--config-file", "/etc/neutron/neutron.conf", "--config-file", "/etc/neutron/hostname.conf", "--agent-type", "Metadata agent"]
+            initialDelaySeconds: 30
+            periodSeconds: 30
+            timeoutSeconds: 10
+          resources:
+{{  toYaml $.Values.pod.resources.metadata_agent | indent 12 }}
+          volumeMounts:
+            - name: metadata-proxy
```
notandy, line 165: the unix domain socket provided by the metadata-agent should be only mapped into the filesystem namespace of the neutron-agent which runs in the same pod, not to the host anymore (else, multiple network-agents running on the same host would cause a conflict for the hostPath domainsocket)

notandy (changes requested):

PR #2447, opened 2021-05-04
[blackbox-exporter] doc content module - reduced text for regex check

dhalimi:


auhlig:
You need to bump the chart version to get this deployed @dhalimi

PR #2459, opened 2021-05-10
[vrops-exporter] Added VM diskspace metrics for mount /log

christopherhans:


richardtief (approved):
Looks good. But in this case you don't have the possibility to change the output metric. 

This is due to the special characters in the metric. These are normalized in the exporter. The result will be: 

```
vrops_virtualmachine_guestfilesystem_storage_log_usage
vrops_virtualmachine_guestfilesystem_storage_log_percentage
vrops_virtualmachine_guestfilesystem_storage_log_capacity
```

christopherhans:
Alright, good to know. Should adjust the config file, or leave it as it is?

richardtief (approved):
Thanks!

PR #2463, opened 2021-05-10
[vrops-exporter] Add metrics for cluster imbalance

max-len:
Export two additional cluster level metrics:
- Cluster Services | Total Imbalance
- Cluster Configuration | DRS Configuration | Tolerable imbalance threshold

vRops key for second metric is different in vRops UI / doc / API resource list. Last one is the only one working and has matching values, so exporting this with respective name.


richardtief (approved):
Thanks!

PR #2482, opened 2021-05-17
[cinder] Update to require cinder_service_password

hemna:
This patch removes a reference to the old identity.password_for_user
and forces a required global value for cinder_service_password


openstack/cinder/values.yaml
```diff
@@ -13,6 +13,8 @@ global:
     enabled: false
   imagePullPolicy: IfNotPresent
 
+  cinder_service_user: cinder
```
joker-at-work, line 16: What's that for? There's already a default in the chart.

hemna: ok I didn't see the value in values.yaml and was concerned it might be used elsewhere.  I was comparing the nova chart to see what it does.  I can clean that up later as I prefer explicit settings in values.yaml.

joker-at-work (approved):

PR #2499, opened 2021-05-19
Introduce kubernetes PriorityClass for scheduling control.

galkindmitrii:
- adds non-default openstack-service-critical PriorityClass.
- assigning it should prevent resource eviction and ensure a faster scheduling.

Enable for RabbitMQ, MariaDB charts.


system/priority-class/templates/openstack-priority-class.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: openstack-service-critical
+value: 100000000
```
Carthaca, line 5: to improve readability, I suggest to write as `100_000_000`

Carthaca (changes requested):
Hi, looks like a very useful addition to me :+1:

 `system/kube-system-metal/Chart.lock` needs an update, please, like our ci suggests

galkindmitrii:
> Hi, looks like a very useful addition to me 👍
> 
> `system/kube-system-metal/Chart.lock` needs an update, please, like our ci suggests

Thanks Maurice!

I'll update the lock for kube-system-metal after this is merged, otherwise have to split into multiple PRs to get priority class chart pushed to repository first.

Carthaca (approved):

PR #2505, opened 2021-05-20
Make cluster imbalance metrics available globally

max-len:
Add cluster imbalance metrics to federation for global dashboard


viennaa (approved):
lgtm

PR #2506, opened 2021-05-21
[neutron] add initcontainer to ensure ebtable_nat

defo89:
* This is to avoid issues on newer flatcar versions with 5.x kernel which does not load ebtable_nat module by default causing neutron-linuxbridge-agent container to fail with "The kernel doesn't support the ebtables 'nat' table." error.
* New nodes are being deployed with this module to load but this is an extra measure for existing nodes.


openstack/neutron/templates/statefulset-network-agent-apod.yaml
```diff
@@ -56,6 +56,22 @@ spec:
...
+          securityContext:
+            privileged: true
+          command:
+            - sh
+            - -c
+          args:
+            - |-
+              set -xe
+              chroot /host modprobe ebtable_nat
+              chroot /host lsmod | grep ebtable
```
velp, line 71: I think you can remove this additional check for the module because `modprobe` will return `1` exit code anyway. But it's up to you

velp (approved):
LGTM

openstack/neutron/templates/statefulset-network-agent-apod.yaml
defo89: Good point, I agree. Will remove the extra check.

notandy (approved):

PR #2510, opened 2021-05-26
[Cinder] Update cinder.conf to use admin endpoint

hemna:
This patch adds the nova section to cinder.conf to
tell cinder to use the nova admin interface when making
callbacks to nova.   This is so we don't get message notification
errors due to policy permissions.


openstack/cinder/templates/etc/_cinder.conf.tpl
```diff
@@ -106,3 +106,6 @@ region_name = {{.Values.global.region}}
 
 [coordination]
 backend_url = memcached://{{ .Chart.Name }}-memcached.{{ include "svc_fqdn" . }}:{{ .Values.memcached.memcached.port | default 11211 }}
+
+[nova]
+interface=admin
```
joker-at-work, line 111: We seem to use spaces around `=` in other places. We should keep the format the same through the file if possible.

joker-at-work (changes requested):

joker-at-work (approved):

PR #2512, opened 2021-05-27
[flavors] Restore disk size for m5.xlarge2hdd private flavor

grandchild:
I reduced the disk size by error during flavor renaming.


joker-at-work (approved):

dhoeller (approved):
LGTM

PR #2514, opened 2021-05-28
[Swift] Remove nginx

reimannf:
The container nginx running as a container in the proxy pod is unused and can be removed. Also all references to nginx config and similar is removed.

@majewsky I am unsure whether we need a readiness probe on the proxy container of the swift-proxy pod. That was former done by the nginx but rendered already useless with introduction of haproxy and checking the NodePort directly I think. What is your opinion?


majewsky (approved):
Yeah, the readiness probe on the haproxy level is enough IMO.

PR #2521, opened 2021-06-01
[nova] Set Openstack Request Id in Ingress

fwiesel:
Various openstack services support forwarding a global request id since Pike.
This sets the value in the ingress and log it with a g-prefix to differentiate
it from the normal request id.

### DIFF
--- a/openstack/nova/templates/etc/_nova.conf.tpl
+++ b/openstack/nova/templates/etc/_nova.conf.tpl
@@ -5,6 +5,7 @@
 {{- end }}
 
 log_config_append = /etc/nova/logging.ini
+logging_context_format_string = %(asctime)s.%(msecs)03d %(process)d %(levelname)s %(name)s [%(request_id)s g%(global_request_id)s %(user_identity)s] %(instance)s%(message)s
 state_path = /var/lib/nova
 
 # we patched this to be treated as force_resize_to_same_host for mitaka
--- a/openstack/nova/templates/placement-ingress.yaml
+++ b/openstack/nova/templates/placement-ingress.yaml
@@ -7,8 +7,10 @@ metadata:
     system: openstack
     type: api
     component: nova
-  {{- if .Values.use_tls_acme }}
   annotations:
+    ingress.kubernetes.io/configuration-snippet: |
+      {{- include "utils.snippets.set_openstack_request_id" . | indent 6 }}
+  {{- if .Values.use_tls_acme }}
     kubernetes.io/tls-acme: "true"
   {{- end }}
 spec:
--- a/openstack/nova/templates/api-ingress.yaml
+++ b/openstack/nova/templates/api-ingress.yaml
@@ -7,8 +7,10 @@ metadata:
     system: openstack
     type: api
     component: nova
-  {{- if .Values.use_tls_acme }}
   annotations:
+    ingress.kubernetes.io/configuration-snippet: |
+      {{- include "utils.snippets.set_openstack_request_id" . | indent 6 }}
+  {{- if .Values.use_tls_acme }}
     kubernetes.io/tls-acme: "true"
   {{- end }}
 spec:


openstack/nova/templates/etc/_nova.conf.tpl
```diff
@@ -5,6 +5,7 @@
 {{- end }}
 
 log_config_append = /etc/nova/logging.ini
+logging_context_format_string = %(process)d %(levelname)s %(name)s [%(request_id)s g%(global_request_id)s %(user_identity)s] %(instance)s%(message)s
```
joker-at-work, line 8: Do we need an additional ` ` bewetween instance and message or is this like before? Can you maybe point to where you got the string from? I assume you just extended it and it would be good to know for reference later on.

joker-at-work (changes requested):
We should increase the version of the utils Chart, because charts are now uploaded *once* once per version to our chart repo. If you update the chart and someone consumes it from the chart repo, they will not get the updated version. Therefore, we should update the utils chart outside of this PR, update its version and also update all charts currently consuming the utils chart via local filesystem reference - their deployments will break otherwise.

One question: How does this work for e.g. Cinder -> Nova direct calls (or the other way around later on). There's no ingress involved in that afaik. Does Nova always set the header and Cinder just doesn't log the global req-id, yet?

openstack/nova/templates/etc/_nova.conf.tpl
fwiesel: The original comes from here: https://github.com/openstack/oslo.log/blob/master/oslo_log/_options.py#L148-L150, it also has no white-space between `$(instance)s` and `$(message)s`
(Well, I got it from here: https://github.com/sapcc/helm-charts/blob/master/openstack/keystone/templates/etc/_keystone.conf.tpl#L9, but that is likely from the Oslo.log defaults).

joker-at-work: Can we reference this in the commit-message, please? :)

fwiesel:
How it works: The `global_request_id` environment variable is set either directly through the middleware `https://github.com/openstack/oslo.middleware/blob/stable/rocky/oslo_middleware/request_id.py#L46-L47` (or some similar implementation), but *only* if the header is set in the incoming request.

Nova and co are supposed to pass that variable then on to requests they do (e.g. `https://github.com/openstack/nova/search?q=global_request_id`). That is done in nova, cinder, and neutron AFAIK, glance not.

So, if a service does a request itself and it doesn't go over the ingress, then we are left with a `gNone` in the log-output.

> We should increase the version of the utils Chart, because charts are now uploaded once once per version to our chart repo. If you update the chart and someone consumes it from the chart repo, they will not get the updated version. Therefore, we should update the utils chart outside of this PR, update its version and also update all charts currently consuming the utils chart via local filesystem reference - their deployments will break otherwise.

Okay, I will create a pull request just for the utils and dependencies.

fwiesel:
I am just wondering if it makes sense to have separate pull requests anyway, because if I touch all projects, then I would need the okay anyway from everyone

joker-at-work:
> I am just wondering if it makes sense to have separate pull requests anyway, because if I touch all projects, then I would need the okay anyway from everyone

Not really. You can just add the functionality to the utils chart, update all their versions and nothing changes for the, because they don't use the new functionality.

joker-at-work (commented):
What about the placement Ingress and the console ingress? Do they not need the annotation or are they unable to use global request ids?

fwiesel:
Please correct me, if I am wrong. I was assuming that, the placement api is only called internally, so it would get the global request id from the request to the public nova endpoint.

The console endpoint looks to me like some proxy code, which does not pay any attention to openstack request id at all, presumably because they are proxying to services like VNC, etc...

joker-at-work:
IIRC, Nova also calls placement in reoccurring tasks, which don't come from the Ingress. At least nova-bigvm behaves like a client. We also have hammer, the seeder and admins calling into placement directly.

It might not be necessary to use a global request-id for those cases, though, as placement doesn't call out into other services and thus the local request-id should suffice.

If I read the code correctly, it uses a Nginx-local request-id to generate the one for OpenStack. So it might still help to be able to see which line logged by Nginx belongs to which line logged by placement at certain times.

fwiesel:
Personally, I was thinking of it as an advantage only to have global ids for requests coming from the public endpoints. That way we can easily identify them.
And wherever I was reasonably sure that the API didn't call out to other services, I have avoided adding the Lua script to the ingress in order to avoid the overhead/error source there (E.g. Keystone).

> If I read the code correctly, it uses a Nginx-local request-id to generate the one for OpenStack. So it might still help to be able to see which line logged by Nginx belongs to which line logged by placement at certain times.

Yes, but unfortunately, the Nginx request id is simply 32 hex characters, so we cannot simply search for the request-id string and then find the corresponding Nginx log entry.

But now that you say it, that could be solved in the log stash parser, and we could normalise there the request id to the 32-hex string.

joker-at-work:
We can also just use ```
$ hammer utils convert-uuid 586e4613-b7f8-4881-a6e3-d6cc4b166a9a -f compact
586e4613b7f84881a6e3d6cc4b166a9a
```
and an `OR` in the search query to find both Nginx and OpenStack logs.

**Comment by fwiesel** [comment]
Or
 `echo '586e4613-b7f8-4881-a6e3-d6cc4b166a9a'  | tr -d -`

There are hundred ways to do it, I just would have preferred to be able to paste the request id directly in a query and get all relevant information without intermediate step. But that is just venting my disappointment.

I am unclear now about what changes you are requesting. Shall I activate the request-id also on the other ingresses?

**Comment by joker-at-work** [comment]
> I am unclear now about what changes you are requesting. Shall I activate the request-id also on the other ingresses?

I'd say we should activate it for placement, too. As you said, the consoles probably don't even care logging them anywhere and if we see them lacking a global id, we can activate it later on.

File: openstack/nova/templates/console-ingress.yaml [hunk 2]
```diff
@@ -14,6 +14,7 @@ metadata:
       rewrite "(?i)/{{ $name }}/(.*)" /$1 break;
       rewrite "(?i)/{{ $name }}$" / break;
       {{- end }}
+      {{- include "utils.snippets.set_openstack_request_id" . | indent 6 }}
```
joker-at-work, line 17: Looks like you added the global id to the console ingress instead of the placement ingress by accident.

joker-at-work (changes requested):

openstack/nova/templates/console-ingress.yaml
fwiesel: Indeed, sorry about that.

openstack/nova/templates/placement-ingress.yaml
```diff
@@ -7,8 +7,9 @@ metadata:
     system: openstack
     type: api
     component: nova
-  {{- if .Values.use_tls_acme }}
   annotations:
+    {{- include "utils.snippets.set_openstack_request_id" . | indent 4 }}
```
joker-at-work, line 11: Comparing this to the `api-ingress.yaml`, it looks like it's missing
> ingress.kubernetes.io/configuration-snippet: |

and needs more indent then, too?

joker-at-work (changes requested):

openstack/nova/templates/placement-ingress.yaml
fwiesel: Indeed.

joker-at-work (approved):

PR #2522, opened 2021-06-01
[Cinder] Set Openstack Request Id in Ingress

fwiesel:
This sets the value in the ingress and log it with a g-prefix to differentiate
it from the normal request id. If the request comes over nova-api, nova will forward the request-id to cinder.
Cinder should forward the very same id then to other services (e.g. glance, or back to nova in case of call-backs)

It has been added after the normal logging-context string:
logging_context_format_string = %(process)d %(levelname)s %(name)s [%(request_id)s %(user_identity)s] %(instance)s%(message)s

If the request did not came in over ingress (e.g. internally by the service) it will render to gNone

### DIFF
--- a/openstack/cinder/templates/api-ingress.yaml
+++ b/openstack/cinder/templates/api-ingress.yaml
@@ -7,8 +7,10 @@ metadata:
     system: openstack
     type: api
     component: cinder
-  {{- if .Values.use_tls_acme }}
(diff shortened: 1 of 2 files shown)


hemna:
looks ok to me

joker-at-work (approved):

PR #2527, opened 2021-06-02
Revert "vrops-exporter: HostWithRunningVMsNotResponding changed running VM metric"

richardtief:
Reverts sapcc/helm-charts#2408 because vROps returns the last number of virtual machines on the host despite disconnected host.

### DIFF
--- a/prometheus-exporters/vrops-exporter/alerts/host.alerts
+++ b/prometheus-exporters/vrops-exporter/alerts/host.alerts
@@ -6,7 +6,7 @@ groups:
       vrops_hostsystem_runtime_connectionstate{state="notResponding"}
       and on (hostsystem) vrops_hostsystem_runtime_powerstate{state="Unknown"}
       and on (hostsystem) vrops_hostsystem_runtime_maintenancestate{state="notInMaintenance"}
-      and on (hostsystem) vrops_hostsystem_summary_running_vms_number > 0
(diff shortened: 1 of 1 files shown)


prometheus-exporters/vrops-exporter/alerts/host.alerts
```diff
@@ -6,7 +6,7 @@ groups:
       vrops_hostsystem_runtime_connectionstate{state="notResponding"}
       and on (hostsystem) vrops_hostsystem_runtime_powerstate{state="Unknown"}
       and on (hostsystem) vrops_hostsystem_runtime_maintenancestate{state="notInMaintenance"}
-      and on (hostsystem) vrops_hostsystem_summary_running_vms_number > 0
+      and on (hostsystem) count(vrops_virtualmachine_runtime_powerstate{state="Powered On"}) by (hostsystem) > 0
```
christopherhans, line 9: Do we really need a grouping here? The comparison is on hostsystem level anyways.

richardtief: No, we don't need it. But is is also not wrong.

richardtief: You could suggest a change if you like.

christopherhans (approved):

PR #2544, opened 2021-06-08
Change policy.json to show bgpvpn for the end-user and give him control for router associations.

velp:
This PR provides to the end-user simple access for BGPVPNs. The end-user should be able to:
1. get own bgpvpn by ID
2. get list of own bpgvpns
3. create/update/delete/get router associations for own bgpvpns

To get bgpvpn or association the end-user must have role `network_viewer` and be owner of this resource. To create/update/delete router association the end-user must have role `network_admin` and be owner of this resource.


sebageek (approved):
lgtm

sebageek:
neutron-bgpvpn-pr
For anyone else wondering what the policy looks like. project_id was added to owner rule as tenant_id has long been deprecated and bgpvpn doesn't have it. crud on router association doesn't allow any access to route targets (or other stuff that could get dangerous for us).

PR #2550, opened 2021-06-10
Add rules create_port:fixed_ips:ip_address and create_port:fixed_ips:subnet_id to fix 403 error in Ussuri release

velp:
This PR adds rules `create_port:fixed_ips:ip_address` and `create_port:fixed_ips:subnet_id` to fix 403 error when you try to create port in RBAC shared network:

```
HttpException: 403: Client Error for url: https://network-3.qa-de-1.cloud.sap/v2.0/ports, (rule:create_port and (rule:create_port:fixed_ips and (rule:create_port:fixed_ips:subnet_id and rule:create_port:fixed_ips:ip_address))) is disallowed by policy
```

These rules exist before but only in Ussuri release they were fixed and start working. As result we had problem with default value of rule `create_port:fixed_ips:ip_address` because by default it does not allow `ip_address` (https://github.com/openstack/neutron/blob/master/neutron/conf/policies/port.py#L124)


Scsabiii (approved):
LGTM

PR #2558, opened 2021-06-15
vrops-exporter: raise nsx-t alerts to warning

christopherhans:


viennaa:
Why is the playbook for all of them `#nsxt_object_limit_exceeded`. This does not quite fit, does it?

christopherhans:
yeah kinda. The playbook is for all the same, because right now the action for all is the same. This is designed as requested. They will probably come up with more detailed playbooks in the future.

PR #2568, opened 2021-06-21
Update nova.alerts

artherd42:


hemna:
lgtm

joker-at-work:
Imho, the commit-message should contain _some_ explanation for the changes at least. E.g.
```
We're updating the nova alerts to be Inhibition alerts to make clear that they're not used by the Nova team, but only for inhibiting other alerts. That's also why we remove the tier: so the alerts don't end up in slack channels of Nova.
```

PR #2569, opened 2021-06-23
[vcenter-operator] Add helm3 annotations

fwiesel:
Add the annotations and labels as per migration guide.


fwiesel:
Not sure, if it is even necessary anymore.

joker-at-work:
We've migrated to `helm3` without those labels.

PR #2571, opened 2021-06-23
[vrops-exporter] added prefix info to config

richardtief:


prometheus-exporters/vrops-exporter/templates/collector-configmap.yaml
```diff
@@ -22,6 +22,7 @@ data:
     #    - 'VMStatsDefaultCollector'
 
     ClusterPropertiesCollector:
+    # Prefix: vrops_cluster_
```
viennaa, line 25: why is it commented out?

christopherhans (approved):
Good for those who are new to the exporter.

viennaa:
If it is an information, please specify it as such. looks like it is a configurable thing in the config. or is it?

prometheus-exporters/vrops-exporter/templates/collector-configmap.yaml
richardtief: Just an info. Added `INFO` to the comment.

PR #2578, opened 2021-06-25
[neutron] Add opt to disable controlplane agents

sebageek:
This option will allow us to distinctively disable the controlplane
agents. Required for regions where we have fully switched to the apod
design.

---
Dictated but not read.


defo89:
LGTM

PR #2579, opened 2021-06-25
[neutron] Don't override controlplane value

sebageek:
With a defaultfilter the controlplane will be set to true whenever it's
false, essentially taking away the option to set it to false. Therefore
we now add it to our default values andset it to true. Now we should be
able to override it.


defo89 (approved):
lgtm

PR #2581, opened 2021-06-25
[swift] add configurable 'delete_concurrency' for bulk deletions.

galkindmitrii:
As per doc:
To speed up the bulk delete process, multiple deletes may be executed in
parallel. Avoid setting this too high, as it gives clients a force multiplier
which may be used in DoS attacks. The suggested range is between 2 and 10.

Glance image deletion logic was rewritten to use bulk deletions along
with retry mechanism: https://github.com/sapcc/glance_store/commit/467a09a93ae1421949d0b6c62f060141307d8f71

While this has reduced the number of conflict errors, there are still
some cases happening in regions with the most load.


reimannf (approved):
Thanks

majewsky (approved):

PR #2585, opened 2021-06-28
[vrops-exporter] pull nsxt metrics to global

richardtief:


christopherhans (commented):
Do we need **all** NSXT metrics in global?

richardtief:
We need almost all NSX-T management cluster metrics. If we add more resources besides the clusters, like NSX-T management nodes, then I change it at that point. Memo is made.

christopherhans (approved):

PR #2589, opened 2021-06-28
[nginx-ingress] Log global_request_id variable

fwiesel:
The value can be set in each ingress where required, otherwise it will print a hypen '-'
It is currently set in various openstack services, and allows us to correlated more easily the request
between the nginx and the upstream service


auhlig:
Can you increment the [Chart version](https://github.com/sapcc/helm-charts/blob/master/system/kube-system-metal/Chart.yaml#L4) @fwiesel?

auhlig (approved):

auhlig:
Nevermind. I did it

PR #2593, opened 2021-06-29
[vrops-exporter] NSX-T 2.5 metric mapping

richardtief:


prometheus-exporters/vrops-exporter/templates/collector-configmap.yaml
```diff
@@ -405,6 +405,13 @@ data:
         key: "System Capacity|Groups Based on IP Sets|MaxSupportedCount"
       - metric_suffix: "sys_capacity_groups_based_in_ip_usage_count"
         key: "System Capacity|Groups Based on IP Sets|UsageCount"
+        
+     # NSX-T 2.5
+      - metric_suffix: "sys_capacity_system_wide_logical_switch_usage_count"
```
christopherhans, line 410: Why do we dont have 'ports' in the suffix? (Same in the 3.0 port usage)

richardtief: Added it. But it has not been noticed so far, because the exporter has not found the requested key in the API answer. Then a normalisation takes place, in which `ports` is included. 

`vrops_nsxt_system_capacity_system_wide_logical_switch_ports_usagecount`

christopherhans (approved):

PR #2600, opened 2021-07-02
[alertmanager] inhibition extension based on host

geisslet:
add `inhibition_host` to your alert to inhibit it, if a inhibition-alert is firing

### DIFF
--- a/global/prometheus-alertmanager-operated/Chart.yaml
+++ b/global/prometheus-alertmanager-operated/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 type: application
 description: Prometheus Alertmanager via operator.
 name: prometheus-alertmanager-operated
-version: 1.2.1
+version: 1.2.2
 
 dependencies:
   - alias: prometheus-alertmanager
--- a/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
+++ b/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
@@ -39,6 +39,12 @@ inhibit_rules:
       alertname: 'PodNotReady|ManyPodsNotReadyOnNode'
     equal: ['node']
 
+  - source_match_re:
+      context: "inhibition"
+    target_match_re:
+      alertname: '.*'
+    equal: ['tier', 'inhibition_host']
+
 route:
   group_by: ['region', 'service', 'alertname', 'cluster']
   group_wait: 1m


geisslet:
rough idea description:

- during build up (for instance) we have false positive critical alerts for a not ready BB 
- it's a known condition, so we can set an alert indicating that condition also in context that it is for inhibition
- by adding now the "inhibition_host" label to the original false positive alerts we can mitigate false paging without manual silence setup as we now need to do today and which doesn't work well
- that this need a review of the existing alerts is by intention

artherd42 (approved):

geisslet:
additional comment: 
- we use the latest AM, but use deprecated syntax
- will update that in a extra PR later; just wanted to have it the same for now

auhlig (commented):
Can you elaborate on why context `inhibition` and label `inhibition_host` are needed and what they are used for?
I understand one once to silence alerts during the build-up, which can be done already for an entire region, cluster, service, etc. .
In which case creating a silence with a comment like `xyz in build-up` would be easier and more transparent.
Removing it is done with a single click once it's expected to work without additional deployments of each service to update the alerts.

With this change, every alert leveraging this would need to have the label `inhibition_host` specified, correct?
This might take a while to deploy per region as the alerts reside with the application. 
One would need to deploy a service all the way to the target cluster.

auhlig:
> additional comment:
> 
> * we use the latest AM, but use deprecated syntax
> * will update that in a extra PR later; just wanted to have it the same for now

Yess. Thanks!

geisslet:
@auhlig - here some details on our thoughts.

> In which case creating a silence with a comment like xyz in build-up would be easier and more transparent.

The predefined silences exists in supernova as today, but people forget it. 
In addition it doesn't cover all alerts and you need several silences (label names aren't equal in metrics, e.g server vs server_name.)

We do have metrics indicating a build up and also other conditions.
Hence we do not know a mechanism to create silences on condition with board tooling nor another tool, the idea was to inhibit.

> Can you elaborate on why context `inhibition` and label `inhibition_host` are needed and what they are used for?

Context is `inhibition` to clearly mark an alert as inhibition root - a classification so to speak.
Label `inhibition_host` is just a defined matcher to map different label names which are used in the metrics but with fitting content.

Yes it's manual effort. We also thought in a initiative to do same label renaming .. but in the end it seems more flexible to have a separate label.

We also want do reuse the idea for maintenance (etc).
Since the different teams also interpret `host` in a different way it would be generic and hopefully evolve over time.

But you are right, it's not a low hanger we can bring in quickly - nonetheless it seems to us the easiest approach.
We can have a short real talk, if you have a better idea - we hadn't :-/

auhlig:
> The predefined silences exists in supernova as today, but people forget it.
> In addition it doesn't cover all alerts and you need several silences (label names aren't equal in metrics, e.g server vs server_name.)

Sounds to me like the current process works and should be enforced and documented.

Several silences are what I'd expect. 
Not all services and their dependencies are ready at the same time.

The different labels is another problem. 
We need a guideline on how the label for a thing should be named.
Subsequently, this needs fixing in the exporters we control and for which we don't we introduced normalization rules to address that and ensure consistent labels, see  [prior art](https://github.com/sapcc/helm-charts/blob/master/prometheus-exporters/kube-state-metrics-exporter/aggregations/aggregations.rules).

> We also want do reuse the idea for maintenance (etc).

That's another use-case for the silence IMHO and make on-going maintenances transparent.

PR #2608, opened 2021-07-07
[prometheus-infra] Added hardware model and bios version to global federation

christopherhans:


richardtief (changes requested):
Pls put it in the right category. This will make it easier to read.

richardtief (approved):
Thanks

PR #2622, opened 2021-07-13
[Octavia] Introduce new pending LB alert to replace old one

laerlingSAP:
This should do away with sampling issues.


notandy (approved):

laerlingSAP:
Leaving the branch open for removal of the old alert later down the line

PR #2624, opened 2021-07-13
Revert "[Cinder] Update cinder.conf to use admin endpoint"

hemna:
Reverts sapcc/helm-charts#2510


joker-at-work (approved):
We're reverting this, because we're not sure it's the right cause. We might have to change Nova's policies instead to allow Cinder to push stuff instead of using the admin endpoint - it should really use the internal endpoint.

PR #2629, opened 2021-07-14
[cinder] Configure access to Nova

joker-at-work:
To use the os-server-external-events endpoint of Nova, e.g. for
informing Nova about the extension of an attached volume, Cinder needs
to use proper credentials and also needs the cloud_compute_admin role in
the token's scope.


hemna:
lgtm

PR #2634, opened 2021-07-16
Label all vcenter objects by vcenter/datacenter/cluster

fwiesel:
This should allow us to more easily associated kubernetes resources with vcenter resources.
Now we can decommision a cluster or a vcenter and remove all corresponding resources in kubernetes.
In prometheus, we can associate the alert with the state of the object in netbox, disable the alerting
of resources still in build-up


fwiesel:
I just named the labels as they are named in the vcenter, which doesn't line up 100% with what happens in netbox. It also may look different in prometheus.

openstack/cinder/templates/vcenter_datacenter_cinder_deployment.yaml
```diff
@@ -34,6 +36,8 @@ template: |
           application: cinder
           component: volume-vmware
           name: cinder-volume-vmware-{= name =}
+          vcenter: {= host =}
```
joker-at-work, line 39: Why don't we use `{= name =}` here? Looking at `name:` above, this should be the short-form `vc-a-1` or something like that, what we want, I would say.

openstack/neutron/templates/vct-nsxv3-nanny.yaml
```diff
@@ -18,6 +18,9 @@ template: |
       system: openstack
       type: backend
       component: neutron
+      vcenter: {= vcenter_name =}
```
joker-at-work, line 21: It's `{= vcenter_name =]` here, but `{= host =}` below. We have to make up our mind which one we take.

openstack/cinder/templates/backup_configmap.yaml
```diff
@@ -14,6 +14,8 @@ template: |{{`
       system: openstack
       type: configuration
       component: cinder
+      vcenter: {{ host }}
+      datacenter: {= availability_zone =}
```
joker-at-work, line 18: This is mixing two different ways of templating. I think it has to be `{{ availability_zone }}`

fwiesel: Indeed,  missed that one.

openstack/cinder/templates/vcenter_datacenter_cinder_deployment.yaml
fwiesel: I would have preferred that too, but both the alerts defined for the `vcsa-exporter` and `vrops-exporter` have the fqdn as a prometheus label.

We can probably also take care with some prometheus relabelling actions to transform the Kubernetes label to fqdn in prometheus. Not sure, which of the two options is better.

openstack/neutron/templates/vct-nsxv3-nanny.yaml
fwiesel: I've changed it for now everywhere to `host`

joker-at-work (approved):

openstack/cinder/templates/backup_configmap.yaml
grandchild: ah, this was confusing, the line above changed.

PR #2638, opened 2021-07-20
federate image usage metrics

DEiselt:
IDE kicked in deleting some whitespaces.


system/prometheus-infra/templates/_prometheus-infra.yaml.tpl
```diff
@@ -10,6 +10,7 @@
     'match[]':
       - '{app="cloudprober-exporter"}'
       - '{app="thousandeyes-exporter"}'
+      - '{app="image-usage-exporter"}'
```
IvoGoman, line 13: In Prometheus the job is called `infra-monitoring/image-usage-exporter`. I have looked into the PodMonitor spec documentation and it might be an error in how we defined the JobLabel.

> The label to use to retrieve the job name from.

[https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#podmonitorspec]( https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#podmonitorspec )

I assume it defaults to 'namespace/podname' because currently the jobLabel has the actual name of the job . https://github.com/sapcc/helm-charts/blob/86ac34916d38c240c889f567634e9c0720fdce2b/prometheus-exporters/image-usage-exporter/templates/metrics-podmonitor.yaml#L11

IvoGoman (changes requested):

IvoGoman (approved):
LGTM

PR #2642, opened 2021-07-26
[Barbican] update context of OpenstackBarbicanUserCertificateExpiresI…

rajivmucheli:
…n30days alert

### DIFF
--- a/openstack/barbican/alerts/openstack-barbican.alerts
+++ b/openstack/barbican/alerts/openstack-barbican.alerts
@@ -39,7 +39,7 @@ groups:
       severity: info
       tier: os
       service: 'barbican'
-      context: 'barbican'
+      context: 'barbican-cert-expiry'
       dashboard: barbican
       meta: 'Certificate {{ $labels.certificate_name }} will expire in {{ $value | humanizeDuration }}'
     annotations:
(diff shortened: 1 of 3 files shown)


openstack/barbican/alerts/openstack-barbican.alerts
```diff
@@ -39,7 +39,7 @@ groups:
       severity: info
       tier: os
       service: 'barbican'
-      context: 'barbican'
+      context: 'barbican-cert-expiry'
```
auhlig, line 42: This is a bit redundant as the service is already `barbican`. 
Preferably something broader or use `certificate` if you expect more alerts here, in which case automating the renewal might be useful.

auhlig:
+ you'd need a route like
```
- receiver: slack_barbican_certificate
    continue: false
    match_re:
      service: barbican
      context: certificate
      severity: critical|warning|info
      region: ap-ae-1|ap-au-1|ap-cn-1|ap-jp-1|ap-jp-2|ap-sa-1|ap-sa-2|eu-de-1|eu-de-2|eu-nl-1|eu-ru-1|la-br-1|na-ca-1|na-us-1|na-us-2|na-us-3
```

and the corresponding channel (copy&paste from existing)

auhlig:
Cool. Now only this is missing
> In addition, My Helm charts are versioned. The Alertmanager ChartVersion should be incremented accordingly.

PR #2646, opened 2021-07-28
[Neutron] Restart neutron-(rpc-)server on neutron-vendor change

fwiesel:
If we do not restart the server, drivers living only in the server
won't know about the changes of the settings (e.g. arista,manila)


notandy (approved):
Lgtm

PR #2656, opened 2021-07-30
[esxi-host-exporter] changed configmap mount

christopherhans:


prometheus-exporters/esxi-host-exporter/templates/deployment.yaml
```diff
@@ -50,7 +50,7 @@ spec:
             - mountPath: /atlas
               name: atlas-output
               readOnly: true
-            - mountPath: /
+            - mountPath: /exporter/
```
viennaa, line 53:
```suggestion
            - mountPath: /exporter/esxi-exporter-config.yaml
```

viennaa (changes requested):

prometheus-exporters/esxi-host-exporter/templates/deployment.yaml
christopherhans: why is this necessary?

viennaa: It is indeed not needed, just make sure you are using `/exporter/config.yaml` as the file to read in the exporter.

viennaa (approved):

prometheus-exporters/esxi-host-exporter/templates/deployment.yaml
christopherhans: yes thats the case:
https://github.com/sapcc/esxi-exporter/blob/main/modules/configuration/ExporterConfig.py#L24

viennaa: well, ok ^^

PR #2657, opened 2021-07-30
[esxi-host-exporter] corrected mount

christopherhans:


prometheus-exporters/esxi-host-exporter/templates/deployment.yaml
```diff
@@ -47,10 +47,10 @@ spec:
               memory: {{ required ".Values.resources.requests.memory missing" $.Values.resources.requests.memory }}
               cpu: {{ required ".Values.resources.requests.cpu missing" $.Values.resources.requests.cpu }}
           volumeMounts:
-            - mountPath: /atlas
+            - mountPath: /esxi-exporter/atlas
               name: atlas-output
               readOnly: true
-            - mountPath: /exporter/
+            - mountPath: /esxi-exporter/
```
viennaa, line 53: this would overwrite the existing directory

PR #2659, opened 2021-07-30
[esxi-host-exporter] Changed esxi to vmware-esxi in Prometheus config

christopherhans:


system/infra-monitoring/templates/_prometheus-infra-collector.yaml.tpl
```diff
@@ -976,14 +976,13 @@
...
+  - role: service
+    namespaces:
+      names:
+      - infra-monitoring
   metrics_path: /
   relabel_configs:
-    - source_labels: [job]
-      regex: esxi
-      action: keep
     - target_label: __address__
```
viennaa, line 986: please use respective variable from the available meta services. `address` is not containing the service name.
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#service

system/infra-monitoring/templates/_prometheus-infra-collector.yaml.tpl
```diff
@@ -976,14 +976,13 @@
...
       replacement: esxi-host-exporter:9666
```
viennaa, line 987: Port is not needed, the service will be queried on its exposed port.

viennaa (changes requested):
Almost there :)

viennaa (approved):

PR #2660, opened 2021-08-02
HostWithRunningVMsNotResponding alert set to warning

richardtief:
We must set HostWithRunningVMsNotResponding alert to info for now to solve an inconsistency.


christopherhans:
Might be better to change it to warning. No paging, but it doesn't get lost in info channel. In the end this is still a critical alert.

christopherhans (approved):
lgtm. Make sure to increase the severity, as soon as the issue in vrops is resolved. Thank you.

PR #2661, opened 2021-08-02
[vrops-exporter] set time to fire for HostNotResponding alert

richardtief:
set time to fire for `HostNotResponding` alert


richardtief:
No need to change. Instant is intentional.

PR #2669, opened 2021-08-05
[manila][policy] allow to force share and snapshot deletion only to cloud admins

galkindmitrii:
(CCM-22705)

"share:force_delete": "rule:context_is_admin"


frr2018 (approved):
LGTM.

PR #2680, opened 2021-08-16
[Keystone] update app version to train

rajivmucheli:
@bbobrov shall we update the chart version to 0.3.0 maybe ?


bbobrov:
sounds good, but where should it happen?

bbobrov (approved):

PR #2682, opened 2021-08-16
[kube-system-metal] Add maintenance controller

Nuckal777:


system/flatcar-linux-update-agent/templates/update-agent.yaml
```diff
@@ -1,35 +1,38 @@
...
+  selector:
+    matchLabels:
+      app: flatcar-linux-update-agent
   template:
     metadata:
       labels:
-        app: container-linux-update-agent
-        container-linux-update.v1.coreos.com/agent-version: v0.7.0
+        app: flatcar-linux-update-agent
+        agent-version: v0.7.3
```
databus23, line 16: Not sure this label (and annotation) adds much other than to drift from the image that is updated eventually while this value is forgotten.

system/flatcar-linux-update-agent/templates/update-agent.yaml
```diff
@@ -1,35 +1,38 @@
...
     spec:
       tolerations:
       - key: "species"
         operator: "Exists"
         effect: "NoSchedule"
       containers:
       - name: update-agent
-        image: quay.io/coreos/container-linux-update-operator:v0.7.0
+        image: quay.io/kinvolk/flatcar-linux-update-operator:v0.7.3
```
databus23, line 26: Maybe this should be a value that can be updated without requiring a a new chart version

system/maintenance-controller/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: maintenance-controller
+description: A controller to manage node maintenance
+type: application
+version: 0.1.0
+appVersion: "0.1"
+home: https://github.com/sapcc/maintenance-controller
+dependencies:
+  - name: flatcar-linux-update-agent
```
databus23, line 9: The flatcar-linux-update-agent is not a hard dependency of the maintenance controller. So I suggest to make it at optional by adding `condition: flatcar-linux-update-agent.enabled` and having the `enabled` value default to `true` in the charts values.yaml.

system/maintenance-controller/maintenance.yaml
```diff
@@ -0,0 +1,55 @@
...
+    check:
+    - hasAnnotation:
+        name: reboot_needed
+        config:
+            key: flatcar-linux-update.v1.flatcar-linux.net/reboot-needed
+            value: "true"
+    - maxMaintenance:
+        name: check_count
+        config:
+          max: 1
```
databus23, line 32: Maybe this should be a configurable value?

system/maintenance-controller/maintenance.yaml
```diff
@@ -0,0 +1,55 @@
+# see https://github.com/sapcc/maintenance-controller for documentation
+intervals:
+    requeue: 60s
+    notify: 5h
```
databus23, line 4: Maybe this should be a value

system/maintenance-controller/templates/manager.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: maintenance-controller-config
+  namespace: kube-system
+type: Opaque
+stringData:
+{{ tpl (.Files.Glob "maintenance.yaml").AsConfig . | indent 2 }}
```
databus23, line 8: As an alternative to the included config template it would be nice of the chart allows to specify the complete content of the `maintenance.yaml` as a value. This would allow reusing this chart with a config that is completely different (e.g. no flatcar OS involved)

databus23 (changes requested):

databus23 (approved):
LGTM

PR #2686, opened 2021-08-18
Create apicmcpfault.alerts

rahul61185:


prometheus-exporters/apic-exporter/alerts/apicmcpfault.alerts
```diff
@@ -0,0 +1,15 @@
+groups:
+- name: apicmcpfault.alerts
+  rules:
+  - alert: NetworkApicMcpFaults
+    expr: (network_apic_mcp_fault_counter_total) by (fault_desc, fault_lifecyle, fault_summary)
```
artherd42, line 5: This is not a valid expression. See https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/ for defining alerting rules and https://prometheus.io/docs/prometheus/latest/querying/basics/ about querying Prometheus. You can test your query in the https://prometheus-infra-collector.na-us-1.cloud.sap/graph?g0.expr=&g0.tab=1&g0.stacked=0&g0.range_input=1h for example.

PR #2690, opened 2021-08-23
[esxi-host-exporter] Increased scrape time

christopherhans:
Query times out in Prometheus. Tested the scrape time locally via port-forwarding.


artherd42 (approved):
lgtm

PR #2703, opened 2021-08-25
add a sentry link to NetworkNS alert

B1Sandmann:
We learned today that the InconsistentModelException can sometimes give more hints when there is a NetworkNamespaceProbesFailed issue


B1Sandmann:
@sebageek FYI

PR #2718, opened 2021-08-31
[common/promServer] Enables podMonitorSelector by default

IvoGoman:
In our control plane Infra-Collector prometheus PodMonitors are not being scraped. Enabling the pod service discovery does not work, as we already have a job for this. Enabling the SD would cause pods being scraped twice.
Thus I would like to enable the podMonitorSelector by default.

### DIFF
--- a/common/prometheus-server/CHANGELOG.md
+++ b/common/prometheus-server/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.2.2
+
+* Enables podMonitorSelector by default. Previously it was toggled by `.Values.serviceDiscoveries.pods.enabled` or `.Values.serviceDiscoveries.kubeDNS.enabled`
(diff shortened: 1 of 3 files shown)


auhlig:
Hey @IvoGoman,
The common/prometheus-server chart provides service discovery for various k8s resources of-of-the-box following best-practices esp. for labels.
While I would encourage to use them, some of them are disabled by default and need to be explicity enabled via the values: https://github.com/sapcc/helm-charts/blob/0c8cfa8558ab26544f9a403a1fd0d13e1b169d96/common/prometheus-server/values.yaml#L274-L359

In your instance of the infra-collector you could do so by setting `serviceDiscoveries.pods.enabled: true`.

auhlig (approved):

PR #2719, opened 2021-09-01
blackbox metrics moved control plane to scaleout

developer-abhi:
@majewsky : please approve this.


majewsky:
Where do I get my alerts now, if they're not delivered to the cc-os-limes Slack channel via this alert definition?

developer-abhi:
here : https://github.com/sapcc/helm-charts/blob/master/openstack/cc3test/alerts/ccloud-limes.alerts.

PR #2721, opened 2021-09-02
[alertmanager] blackbox mv to scaleout - adopt routing

geisslet:
request by @developer-abhi


auhlig (changes requested):
Why is this needed?
The region will be `eu-de-1` as well, though the `cluster` will change.

geisslet:
sorry - my fault - didn't thought about this - thanks!

PR #2732, opened 2021-09-09
Adding new flavor for S4 g_c6_m24

dhoeller:


joker-at-work (approved):
Running my check script and looking manually, I cannot find a problem with that.

PR #2746, opened 2021-09-15
blackbox metrics moved control plane to scaleout

developer-abhi:
@joker-at-work : Please approve this.


joker-at-work (commented):
What about all the other blackbox metrics? The commit states, that the "blackbox metrics moved", but I clearly see some being still used. Could you extend the commit-message to explain which metrics moved and maybe the plan for the other metrics?

developer-abhi:
blackbox metrics for api disabled in canary region.(not deployed everywhere to avoid bombarding of absent metrics alert in info channel for every service every alert) right now we have duplicated the same metrics in different prom openstack-prometheus and infra-scaleout-prometheus and we defined alert here for compute(nova): https://github.com/sapcc/helm-charts/blob/master/openstack/cc3test/alerts/compute.alerts.

joker-at-work:
I still don't get it. We replace 6 Alerts with 2? Is this PR just about the API checks? If yes, it should clearly state so.

developer-abhi:
other 4 metrics alert is be part of neutron/jump service alerts : https://github.com/sapcc/helm-charts/blob/master/openstack/cc3test/alerts/network.alerts and https://github.com/sapcc/helm-charts/blob/master/openstack/cc3test/alerts/ccloud-jump-server.alerts as these alert belong to neutron service like (metadata/dhcp/jump). This PR belong to API/Datapath checks.

PR #2752, opened 2021-09-20
add aws-ecr-creds-helper

defo89:


system/aws-ecr-creds-helper/templates/cronjob.yaml
```diff
@@ -0,0 +1,51 @@
...
+spec:
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    metadata:
+      name: ecr-creds-helper
+    spec:
+      backoffLimit: 3
+      template:
+        metadata:
+          name: ecr-creds-helper
```
databus23, line 14: If we add a label to the pod spec (e.g. `cronjob=aws-ecr-creds-helper`) we can query more easily in prometheus. if we do our own alert rule we might as well make it easier for us

databus23: hmm or maybe this needs to go as a label on the `jobTemplate` above...

system/aws-ecr-creds-helper/alerts/ecr-creds-helper.alerts
```diff
@@ -0,0 +1,17 @@
+groups:
+- name: ecr-creds-helper.alerts
+  rules:
+
+    ### ecr-creds-helper ###
+
+    - alert: EcrCredsHelperJobFailed
+      expr: sum(kube_job_status_failed{job_name=~"aws-ecr-creds-helper-.+"}) by (job_name) > 2
```
databus23, line 8: summarizing by `job_name` is not correct as every job created by the cronjob has its own name. I think we want to summarise over multiple cronjob runs.
Maybe its also better to turn the query around and alert if there are no successful runs in the recent past.
E.g.:
```
sum(increase(kube_job_status_succeeded{job_name=~"aws-ecr-creds-helper-.+"}[8h])) == 0 
```

system/aws-ecr-creds-helper/templates/cronjob.yaml
defo89: I will keep them added for the sake of completeness although as we found out they still do not show up under job in Prometheus

PR #2753, opened 2021-09-21
kube-system-*: Remove pull secret injector

databus23:
The payed docker hub accounts are expired and we are not supposed to use docker hub anymore.


defo89 (approved):
lgtm

PR #2760, opened 2021-09-21
Adding 4 TB flavor for 8 Socket Lenovo Servers

dhoeller:


openstack/sap-seeds/templates/flavor-seed.yaml
```diff
@@ -664,3 +664,16 @@ spec:
...
+    vcpus: 288
+    ram: 4481528
+    disk: 64
+    extra_specs:
+      "vmware:hv_enabled": "True"
+      "hw_video:ram_max_mb": "16"
+      "resources:CUSTOM_MEMORY_RESERVABLE_MB": "4481528"
+      "resources:CUSTOM_BIGVM": "2"
+      "trait:CUSTOM_NUMASIZE_C48_M729": "required"
+      "hw:cpu_cores": "2"
```
joker-at-work, line 678: This needs to be 3, I'd guess, because we're above 2 * 128 with the `vcpus` set to `288`.

joker-at-work (changes requested):

PR #2772, opened 2021-09-30
[vrops-exporter] Add cpu|workload to crawled vROps metrics

ofrendo:


prometheus-exporters/vrops-exporter/templates/collector-configmap.yaml
```diff
@@ -274,6 +274,8 @@ data:
         key: "cpu|usage_average"
       - metric_suffix: "cpu_usage_average_mhz"
         key: "cpu|usagemhz_average"
+      - metric_suffix: "cpu_workload"
```
richardtief, line 277: Thank you. Please change the suffix of the metric to `cpu_workload_percentage` to make it more descriptive.

richardtief (changes requested):

prometheus-exporters/vrops-exporter/templates/collector-configmap.yaml
ofrendo: Thanks for your feedback! I've changed the Prometheus metric name from `cpu_workload` to `cpu_workload_percentage`.

richardtief (approved):
Thanks.

PR #2773, opened 2021-10-01
Removal of deprecated flavors

dhoeller:


joker-at-work (approved):
The blog post should have mentioned, that we also remove remove the *_cpu flavors of that type.

Looks good otherwise.

PR #2784, opened 2021-10-04
[elastic-query-exporter] Disable log shipping alerts for apods

max-len:
This will be reverted as soon as syslog config is corrected via automation


christopherhans (approved):
lgtm. Keep in mind to remove this again, as soon as syslog is configured correctly on apods.

PR #2786, opened 2021-10-05
clair: use up to date upstream image to fix scanning of rpm based distros

SuperSandro2000:


majewsky (approved):
Can be merged when the build failures have been addressed.

majewsky (approved):

PR #2789, opened 2021-10-05
Change neutron-rpc-server container name to match pod

sebageek:
@notandy at your discretion :)


notandy (approved):
This looks right to me!

PR #2798, opened 2021-10-06
openstack/swift: add resource limits for health-exporter

talal:


majewsky (approved):
Let's give this a shot in QA.

PR #2819, opened 2021-10-18
Revert "[nova] Increase minimum large VM size"

joker-at-work:
This reverts commit a52588b095fd1b7d3c963b7bbaecd79da7d2158f.

We don't want to enable those smaller large VMs in DRS afterall, because
we are going to enable Cerebro for balancing those large VMs in
production and there shouldn't be two balancing algorithms fighting over
where a VM is supposed to be running.


fwiesel:
That has been decided? I am unclear about that or how the integration is supposed to look like.

PR #2832, opened 2021-10-22
Adding Swift api availability metrics

reimannf:
As Swfit is not using the systems nginx-ingress-controller, but provides
the same information through haproxy metrics, they got added here as well.
`relabel_values` ensures that same set of labels.


artherd42:
add swift-api to global API Availability Dashboard https://github.wdf.sap.corp/cc/grafana-content/commit/58ccf08409190b8d536849002d03f7d1e8689e88

PR #2862, opened 2021-11-01
system/gatekeeper: add GkHighCPURequest to prod ui

talal:


majewsky (approved):
It seems you have found your way around Rego's unusual control flow quite aptly. Very nice, keep it up!

majewsky (changes requested):
Oh wait, actually, there's one thing: We should only deploy this policy to the relevant clusters, e.g. to baremetal, but not to scaleout. I don't know for sure about the other clusters; can you check with Dmitri where it should be applied?

Then add a condition to the config object to only apply it in clusters with the right `.Values.cluster_type`. This value is given in the pipeline with `--set`, so we can rely on it being available.

talal:
Talked to Marian, he says we only need this in baremetal.

majewsky (approved):

PR #2863, opened 2021-11-02
[MariaDB] Uses dockerHubMirror for MariaDB image

IvoGoman:


galkindmitrii:
Keppel itself depends on Keystone and this adds "chicken-egg" problems.
Because all Openstack services including Keystone use this DB chart.
Unless there is an automatic failover to another region and a fire-drill completed, I'd rather not have this merged.

IvoGoman:
Was not aware of this issue. Closing the PR for now

joker-at-work:
Don't we have `dockerHubMirrorAlternateRegion` for exactly those chicken-egg problems?

galkindmitrii:
> Don't we have `dockerHubMirrorAlternateRegion` for exactly those chicken-egg problems?

We do, but that is still a hardcoded region with no automatic failover. There is currently a discussion to move those to ECR.

PR #2867, opened 2021-11-02
[cerebro] Initial commit cerebro helm chart

ofrendo:


openstack/cerebro/cerebro-main/values.yaml
```diff
@@ -0,0 +1,20 @@
+image:
+  pullPolicy: IfNotPresent
+  tag: SET-AS-HELM-PARAMETER # This should be set via --set to a git commit hash
+  registry: SET-AS-HELM-PARAMETER
```
viennaa, line 4: Please link the production ready image here already. Unless we need to deploy different versions in different regions, this is the place to control it.

openstack/cerebro/cerebro-main/values.yaml
```diff
@@ -0,0 +1,20 @@
+image:
+  pullPolicy: IfNotPresent
+  tag: SET-AS-HELM-PARAMETER # This should be set via --set to a git commit hash
+  registry: SET-AS-HELM-PARAMETER
+
+vernemq:
+  vernemq:
+    image:
+      repository: SET-AS-HELM-PARAMETER
+      tag: SET-AS-HELM-PARAMETER  # This should be set via --set to a git commit hash
```
viennaa, line 10: As above. Do you maintain this image or is this something public? If the ladder, we've got to fetch it with keppel too.

openstack/cerebro/ci/test-values.yaml
```diff
@@ -0,0 +1,50 @@
...
+    value: "Parameter for dc-state-service"
+
+
+secrets:
+  fileSecrets:
+  - metadata:
+      name: my-file-secret
+    data:
+      content: |
+        Contents of a secret file
```
viennaa, line 39: I'd add two examples, since you are looping over this it should improve the testing of the range.

openstack/cerebro/cerebro-main/values.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+vernemq:
+  vernemq:
+    image:
+      repository: SET-AS-HELM-PARAMETER
+      tag: SET-AS-HELM-PARAMETER  # This should be set via --set to a git commit hash
+
+
+migrationRecommenderService:
+  prometheusClientPort: 8092
```
viennaa, line 14: These ports should be greater than `9000` to match the prometheus exporter pattern.

openstack/cerebro/cerebro-main/templates/dc-state-service.yaml
```diff
@@ -0,0 +1,76 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: dc-state-service
+  labels:
+    app: dc-state-service
+  annotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/port: "{{ .Values.dcStateService.prometheusClientPort }}"
```
viennaa, line 9: which prometheus is scraping this? If it is the same as for vrops, we've got to define it in the region secrets.

`prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}`

add a default for like

```yaml
alerts:
  enabled: false
  prometheus: some-random-prometheus-url
```

Don't bother the alerts yet. If we have alerts at some point, we've got to enable them in the regions and need to do respective routing beforehand. It doesn't hurt, to have it in here tho.

openstack/cerebro/cerebro-main/templates/dc-state-service.yaml
```diff
@@ -0,0 +1,76 @@
...
+      port: 80
+  selector:
+    app: dc-state-service
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: dc-state-service
+  labels:
+    app: dc-state-service
```
viennaa, line 23: is this missing the `data` part?

openstack/cerebro/cerebro-main/templates/dc-state-service.yaml
```diff
@@ -0,0 +1,76 @@
...
+kind: Service
+metadata:
+  name: dc-state-service
+  labels:
+    app: dc-state-service
+  annotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/port: "{{ .Values.dcStateService.prometheusClientPort }}"
+spec:
+  type: NodePort
```
viennaa, line 11: What's the reason for using a `NodePort` over `ClusterIP` here?

viennaa (changes requested):
Thanks for developing this Chart. I have some comments and questions. Since the 3 services you are introducing are pretty much the same, config wise, I only added the comments to `dc-state-service.yaml`, they apply to the other two as well.

You also should consider to rename `dc-state-service.yaml` to something more generic like `dc-state.yaml`. `-service` indicates only the definition of a service, at a first superficial look.

openstack/cerebro/cerebro-main/values.yaml
ofrendo: The actual image (`/cerebro/dc-state-service/`) is defined in each microservice  (see for example the [dc-state-service](https://github.com/ofrendo/helm-charts/commit/e1291b8d23c52fda9664e6f8e8e8de5407e21d86#diff-2c1ab44d7cdb73e54b3fa545079ac486574004447ab802002d4ce13337e9d11fR43). The `values.yaml` only defines the tag. 

- We can of course use `latest` for tag as the default. Would you prefer using `latest` for `tag` by default?`
- The registry should be Keppel. Should we pass the (internal) Keppel URL as default for `registry`?

openstack/cerebro/cerebro-main/values.yaml
ofrendo: Yes, we maintain this image. All images are stored on Keppel.

openstack/cerebro/ci/test-values.yaml
ofrendo: Added another example for a file secret.

openstack/cerebro/cerebro-main/values.yaml
ofrendo: Changed `8092` to `9092` for all three microservices.

openstack/cerebro/cerebro-main/templates/dc-state-service.yaml
ofrendo: I'm not sure which `data` part you are referring to. The helm chart is able to run locally

openstack/cerebro/cerebro-main/templates/dc-state-service.yaml
ofrendo: Changed `NodePort` to `ClusterIP` for the `dc-state-service` and the `optimization-service`.

The `migration-recommender-service` should be reachable from the Nanny. 

Can the Nanny reach the `migration-recommender-service` if we change `NodePort` to `ClusterIP` there?

openstack/cerebro/cerebro-main/templates/dc-state-service.yaml
viennaa: You are embedding this in your deployment with `envFrom`, but there is no data in here, where you can derive an environment variable. That's why I've asked for a `data` part.

PR #2871, opened 2021-11-02
VCenterHasMultipleGreyVMs Severity to warning

christopherhans:


christopherhans:
Alert is handled by TARS automatically. Playbook as fallback option, if TARS fails. InfraOps is aware.

PR #2872, opened 2021-11-02
[MariaDB] Adds backup keeping binlog files for 60 Minutes

IvoGoman:
Currently binlog files are cleaned after rotation to the next file. With this change those binlog files are kept until they are 60 Minutes old and purged afterwards.

This change is required for the replication into the DatahubDB. In some cases the restore/replication of the full dump into the DatahubDB takes up to 40 minutes. At this point in time the binlog file has already aged out.
Keeping the binlog files longer ensures that the replication can work for databases with larger backup sizes.


IvoGoman:
As discussed in slack with @galkindmitrii. I expect no more than 50mb of additional disk space usage by holding the binlog files for 60 minutes. Also we will get warned by [HighPVCUsagePredicted](https://github.com/sapcc/helm-charts/blob/master/prometheus-exporters/pvc-exporter/templates/alerts/_pvc-usage.alerts.tpl#L4) if the db volume should start to fill up.

PR #2880, opened 2021-11-03
Notifier add mail

NiranB:
added myself to recipients list


dhalimi (commented):
This is not the place to add your user. 
Template can't has values, use @Secrets

PR #2883, opened 2021-11-03
[Keystone] Bumps mariadb requirement to increase binlog file retention

IvoGoman:
Binlog files are now purged after 60 minutes instead of directly after rotation. This is required for the database replication into the datahubdb to work.


rajivmucheli (approved):
lgtm

PR #2885, opened 2021-11-03
[Nova] Bumps mariadb requirement to increase binlog retention

IvoGoman:
Binlog files are now purged after 60 minutes instead of directly after rotation. This is required for the database replication into the datahubdb to work.

Also brings 
* [mariadb] reduce readiness probe initial delay
* [mariadb] change lifecycle pod script to sleep 1 between db pings.

### DIFF
--- a/openstack/nova/requirements.lock
+++ b/openstack/nova/requirements.lock
@@ -1,10 +1,10 @@
 dependencies:
 - name: mariadb
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.3.22
+  version: 0.3.26
 - name: mariadb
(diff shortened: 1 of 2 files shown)


joker-at-work:
I don't want to restart the mariadb, yet. I'd rather wait for the next version of the mariadb chart, that also pulls docker images from our keppel-mirror instead of directly from docker. If you'd change the image sources, too, that would speed up the process.

IvoGoman:
I wanted to add that as well but there are some points that have to be clarified before: https://github.com/sapcc/helm-charts/pull/2863#issuecomment-957232611

PR #2892, opened 2021-11-08
added custom flavor for cc3test server create(regular/hana)

developer-abhi:
@dhoeller @joker-at-work cc3test required custom flavor for server create (regular/hana) reasons : 
1. with hana flavor required to increase cc3test project quota as we spawning large VMs for checking new server create. 
2. with hana flavor putting load on BB with new test server create with large VMs ( which will got double after adding mount_volume check for BB) 
3. with hana/regular flavour lots of logic to put in code to pickup as smallest flavor as possible.

### DIFF
--- a/openstack/sap-seeds/templates/flavor-seed.yaml
+++ b/openstack/sap-seeds/templates/flavor-seed.yaml
@@ -38,6 +38,16 @@ spec:
       "hw_video:ram_max_mb": "16"
       "catalog:alias": "m1.small,m1.smallhdd"
       "trait:CUSTOM_HANA_EXCLUSIVE_HOST": "forbidden"
+  - name: "c_c2_m2_cc3test"
+    id: "21"
+    vcpus: 2
+    ram: 2032
+    disk: 64
+    is_public: false
+    extra_specs:
+      "vmware:hv_enabled": "True"
+      "hw_video:ram_max_mb": "16"
+      "catalog:alias": "m1.xsmall_cc3test"
   - name: "g_c2_m4"
     id: "22"
     vcpus: 2
--- a/openstack/sap-seeds/templates/flavor-seed.yaml
+++ b/openstack/sap-seeds/templates/flavor-seed.yaml
@@ -38,15 +38,6 @@ spec:
       "hw_video:ram_max_mb": "16"
       "catalog:alias": "m1.small,m1.smallhdd"
       "trait:CUSTOM_HANA_EXCLUSIVE_HOST": "forbidden"
-  - name: "hana_cc3test"
-    id: "21"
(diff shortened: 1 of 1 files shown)


openstack/sap-seeds/templates/flavor-seed.yaml
```diff
@@ -38,6 +38,16 @@ spec:
       "hw_video:ram_max_mb": "16"
       "catalog:alias": "m1.small,m1.smallhdd"
       "trait:CUSTOM_HANA_EXCLUSIVE_HOST": "forbidden"
+  - name: "c_c2_m2_cc3test"
```
joker-at-work, line 41: I think `hana_cc3test` would be a better name, as this is what it's supposed to be used for. Thus, it should live at the end of the file and get a new id.

We should also add a comment (in the next line?) to very shortly say why we need this flavor.

openstack/sap-seeds/templates/flavor-seed.yaml
```diff
@@ -38,6 +38,16 @@ spec:
...
+    id: "21"
+    vcpus: 2
+    ram: 2032
+    disk: 64
+    is_public: false
+    extra_specs:
+      "vmware:hv_enabled": "True"
+      "hw_video:ram_max_mb": "16"
+      "catalog:alias": "m1.xsmall_cc3test"
```
joker-at-work, line 50: I don't think we need an alias for it. The alias was supposed to be used for migrating away from old flavor names.

openstack/sap-seeds/templates/flavor-seed.yaml
developer-abhi: flavor name changes to "hana_cc3test" and moved to end of file with comment why its needed.

openstack/sap-seeds/templates/flavor-seed.yaml
developer-abhi: removed alias as its new flavor

openstack/sap-seeds/templates/flavor-seed.yaml
```diff
@@ -621,3 +621,13 @@ spec:
       "trait:CUSTOM_NUMASIZE_C48_M729": "required"
       "hw:cpu_cores": "3"
       "vmware:hw_version": "vmx-18"
+  # HANA flavor for testing
+  - name: "hana_cc3test"
+    id: "21"
```
joker-at-work, line 626: Please give this flavor a new id from the hana range ("312" seems to be the next free)

developer-abhi: set flavor id to 312

PR #2898, opened 2021-11-10
Add GkIngressAnnotations to prod UI

talal:


majewsky (approved):
I don't know how much effort you had getting into Rego, but you certainly make it look effortless. Great work as always.

PR #2899, opened 2021-11-10
system/gatekeeper: add GkPodLabels to prod UI

talal:


majewsky (approved):
Looks good. I have one nitpick, then this can be deployed:

PR #2903, opened 2021-11-11
Update openstack-manila.alerts

artherd42:
since cc3tests are on scaleout cluster this alert move to scaleout promethus too. https://github.com/sapcc/helm-charts/blob/master/openstack/cc3test/alerts/file-shares.alerts#L60-L88


chuan137 (approved):
lgtm

PR #2907, opened 2021-11-12
system/gatekeeper: validate value of service label on Pods

talal:
Since there is [some Slack routing](https://github.com/sapcc/helm-charts/blob/master/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl) that matches on `tier` or `service` only. I have made the policy such that only one of these labels is required.

However, when `tier == os` then service is also required and its value must be valid.

### DIFF
--- a/system/doop-central/templates/configmap.yaml
+++ b/system/doop-central/templates/configmap.yaml
@@ -134,7 +134,7 @@ data:
       <dl>
         <dt>Why is this a problem?</dt>
         <dd>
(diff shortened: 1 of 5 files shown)


majewsky (approved):
I was going to object that `service` is only really well-defined for `tier: os` (as far as the Alertmanager config goes), so we should not require it for other tiers, but upon surveying existing PrometheusRules in prod, it appears that `service` gets set most of the time regardless of tier. So I'm fine with how this check works.

majewsky:
Oh, upon rereading your intro comment on this PR, it sounds like you wanted to make it such that only `tier` **or** `service` is required? I don't think that's what the code does right now. `count(missing_labels_on_pod) > 1` sounds like it would trigger when either is missing, even if the other one is there.

talal:
Since there are only two required labels, `count(missing_labels_on_pod) > 1` would mean that both the labels are missing.

I know realise that the statement: `count(missing_labels_on_pod) == 2` would convey the intent better so I have changed it.

PR #2908, opened 2021-11-12
[nova] Increase max_concurrent_live_migrations

fwiesel:
Only one migration per vcenter is quite conservative and it would take too much time
to empty a whole cluster when draining a cluster.

### DIFF
--- a/openstack/nova/values.yaml
+++ b/openstack/nova/values.yaml
@@ -39,6 +39,9 @@ cell2:
   conductor:
     config_file: {}
 
+default:
+  max_concurrent_live_migrations: 4
+
 defaults:
   default:
     graceful_shutdown_timeout: 60


joker-at-work (commented):
This would mean up to 16 per vCenter assuming 4 BBs in a VC?
```
Concurrent operations
Storage vMotion operations per host
2
Concurrent operations
Storage vMotion operations per datastore
8
```
I know this is a single config option and we cannot build ^ into it, but with cross-vc vmotion, we have at least 1 storage-vmotion, too for the ephemeral storage. Just found this, but not sure how this would show in production and how we could consider this.

fwiesel:
> This would mean up to 16 per vCenter assuming 4 BBs in a VC?

As far as I know, the concurrent operation limits are configuration values, and those are the defaults. 
Ff the limits are reached, the corresponding tasks do not fail, but are queued, so an instance is in migration state
longer than necessarily. Not ideal, but I am not terribly concerned about choosing here the "right" value.

We look therefor at a "performance regression" (task taking longer than needed), if we are choosing the same datastore more than eight times, and the same host more often then twice given that we can draw from those "bags" `max_concurrent_live_migrations` times (combinatorics).

The hosts are purely per cluster so have four chances to draw more than twice the same host from a bag of (at least) twelve host. If I calculated it correctly the probability for that is around 23.6%

The ephemeral storage is segregated by clusters, so we are drawing there four times from a bag with just one color, and have to get more than eight times the same... that's not going to happen, the probability is zero until eight migrations.

Now SVMotion with attached volumes is another thing, which is actually vcenter wide. So, your given example needs to be modified by the fact that we are looking at volumes, and that makes the probabilities more difficult.

I see roughly 1 volumes per vm in average, which would put us 16 vmotions for only 4 datastores (@8 ops) that puts us closer to 50%. So you are right, some work is needed here to avoid getting into a bottleneck here.
I would say, for that we need https://github.com/sapcc/cinder/pull/97 merged and then our migration tool could take that into account.

Until then, I would suggest to instruct operators to limit the number of cross-vcenter migration with volumes to a reasonable limit.

joker-at-work:
Given the recent learning that at least a cross-VC migration holds a lock on the cluster, for cross-vc migrations away from a cluster the maximum parallelism seems to be 1. Still only a "performance regression", but mentioning it for completeness sake.

fwiesel:
Closing it. We can revisit at another time.

PR #2910, opened 2021-11-15
Create snmp exporter module for OS independent MIBs

swagner-de:
We are getting IOS XR devices in PX. Some MIBs are equal over NXOS, IOS
XE and IOS XR. I am creating a new SNMP exporter module that covers
these MIBs, move all alerts and dashboard to the new metric and then
keep only the OS specific MIBs to the according OS specific snmp
exporter modules.

### DIFF
--- a/prometheus-exporters/snmp-exporter/templates/alerts.yaml
+++ b/prometheus-exporters/snmp-exporter/templates/alerts.yaml
@@ -16,15 +16,17 @@ spec:
 
 {{- end }}
 {{- end }}
-{{- $values := .Values -}}
-{{- if $values.snmp_exporter.pxdlrouternxos.alerts.enabled }}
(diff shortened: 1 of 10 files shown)


prometheus-exporters/snmp-exporter/templates/alerts.yaml
```diff
@@ -17,8 +17,8 @@ spec:
 {{- end }}
 {{- end }}
 {{- $values := .Values -}}
-{{- if $values.snmp_exporter.pxdlrouternxos.alerts.enabled }}
-{{- range $path, $bytes := .Files.Glob "pxdl.alerts/*.alerts" }}
+{{- if $values.snmp_exporter.px_snmp.alerts.enabled }}
```
Kuckkuck, line 20: This should be .px. and not px_snmp right? Otherwise the default has to be disabled. I think the variable should be named like the module name, otherwise it's hard to find out, where the variable is used.

PR #2912, opened 2021-11-16
metrics moved from cp to scaleout cluster

developer-abhi:
@rajivmucheli @geisslet : Please approve this, as metrics moved from cp to scaleout cluster


rajivmucheli (approved):
lgtm

PR #2923, opened 2021-11-18
Use tier and service labels from Pod for KubernetesPodRestartingTooMuch

talal:
I checked the helm template rendering using:

```sh
helm template --dry-run prometheus-kubernetes-rules ./prometheus-rules/prometheus-kubernetes-rules
```

and it gave the expected output:

```yaml
labels:
  tier: "{{ if $labels.label_tier }}{{ $labels.label_tier}}{{ else }}k8s{{ end }}"
  service: "{{ if $labels.label_service }}{{ $labels.label_service}}{{ else }}resources{{ end }}"
```

/cc @majewsky

### DIFF
--- a/prometheus-rules/prometheus-kubernetes-rules/Chart.yaml
+++ b/prometheus-rules/prometheus-kubernetes-rules/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: "1.0"
 description: A collection of Prometheus alerting and aggregation rules for Kubernetes.
 name: prometheus-kubernetes-rules
-version: 1.2.9
+version: 1.2.10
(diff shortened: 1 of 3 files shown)


databus23 (commented):
Given that we will want to use this for more then a single alert. I would suggest to create a template helper.

So that this alert can be simplified to something like this:

```
labels:
  tier: {{ include "tierLabelOrDefault" .Values.tier }}
  service: {{ include "serviceLabelOrDefault" "resources" }}
```

databus23 (changes requested):
In order to make the alert template as clean as possible I would not wrap the default value in a `dict`. I'm not seeing a benefit and the downside of more visual noise.

databus23 (approved):
LGTM. 👍

majewsky (approved):

PR #2924, opened 2021-11-18
ResMgmt: add tier and service labels to Pods

talal:


majewsky (approved):
Nice. We have the bimonthly dependency update coming up soon, so we'll roll this out together with that wave of deployments.

PR #2928, opened 2021-11-21
[Manila] Changes to support Xena release

kpsap:
Based on reading from
https://docs.openstack.org/releasenotes/manila/wallaby.html and https://docs.openstack.org/releasenotes/manila/xena.html

@galkindmitrii  check

### DIFF
--- a/openstack/manila/templates/etc/_manila.conf.tpl
+++ b/openstack/manila/templates/etc/_manila.conf.tpl
@@ -47,15 +47,16 @@ migration_ignore_scheduler = True
 migration_wait_access_rules_timeout = 3600
 
 # all default quotas are 0 to enforce usage of the Resource Management tool in Elektra
-quota_shares = 0
-quota_gigabytes = 0
-quota_snapshots = 0
-quota_snapshot_gigabytes = 0
-quota_share_networks = 0
-quota_share_groups = 0
-quota_share_group_snapshots = 0
-quota_share_replicas = 0
-quota_replica_gigabytes = 0
+[quota]
+shares = 0
+gigabytes = 0
+snapshots = 0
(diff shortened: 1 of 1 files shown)


openstack/manila/templates/etc-configmap.yaml
```diff
@@ -14,7 +14,7 @@ data:
 {{ include (print .Template.BasePath "/etc/_manila.conf.tpl") . | indent 4 }}
   api-paste.ini: |
 {{ include (print .Template.BasePath "/etc/_api-paste.ini.tpl") . | indent 4 }}
-  policy.json: |
+  policy.yaml: |
 {{ include (print .Template.BasePath "/etc/_manila-policy.json.tpl") . | indent 4 }}
```
galkindmitrii, line 18: manila-policy template should be renamed and converted to yaml
+ there seem to be a couple of changes in the policies according to: https://docs.openstack.org/releasenotes/manila/de/wallaby.html

Carthaca:
The policy changes have been done in https://github.com/sapcc/helm-charts/pull/2934, but the changes to the `manila.conf` are still relevant, right?

kpsap:
yes, this PR was created first and then realised that policy.yaml is acceptable in victoria as well, so create another PR for just policy.yaml. After 2934 merged, only quota will go for xena.

kpsap:
since we are not using quota from manila, we can merge PR after Xena rollout just to avoid lot of warnings from various pods.

Carthaca:
Ok, it is only a deprecation warning, so the config setting should still work the old way on xena, too

But it really is important for us, we explicitly want to set all quota to 0 by default for all resources

Carthaca (approved):

PR #2929, opened 2021-11-21
[Neutron] Add interconnection Neutron plugin

velp:


openstack/neutron/templates/etc/_neutron-policy.json.tpl
```diff
@@ -332,7 +332,14 @@
     "update_bgpvpn_router_association": "rule:context_is_editor or rule:shared_bgpvpns",
     "delete_bgpvpn_router_association": "rule:context_is_editor or rule:shared_bgpvpns",
     "get_bgpvpn_router_association": "rule:context_is_viewer or rule:shared_bgpvpns",
-    "get_bgpvpn_router_association:tenant_id": "rule:context_is_viewer or rule:shared_bgpvpns"
-
-
+    "get_bgpvpn_router_association:tenant_id": "rule:context_is_viewer or rule:shared_bgpvpns",
```
notandy, line 335: If .Values.interconnection.enabled is disabled, this json list would end with a trailing comma, which is not supported by json and would probably fail to be parsed.

notandy (changes requested):

openstack/neutron/templates/seed.yaml
```diff
@@ -75,10 +75,10 @@ spec:
         role: cloud_network_admin
       - project: l3admin
         role: cloud_compute_admin
-{{- if and (.Values.interconnection.enabled) (.Values.interconnection.service_password) }}
+{{- if and (.Values.interconnection.enabled) }}
```
notandy, line 78: Probably doesn't need the and (..) prefix anymore

notandy (changes requested):

openstack/neutron/templates/etc/_neutron-policy.json.tpl
velp: Good catch, will fix.

openstack/neutron/templates/seed.yaml
velp: yes, exactly

notandy (approved):

PR #2938, opened 2021-11-24
DHCP agent config update

braunmatt:
clean up and remove overwrite


notandy:
@galkindmitrii How the test went? Any issues/differences for upcoming vms?

galkindmitrii:
@notandy - I had to prepare another patch for Neutron dhcp agent to honor the dns_domain setting on network ahead of defaults set in config. I assume this PR won't be needed.

PR #2945, opened 2021-11-26
Adjust API versions as per v1.22 deprecations

talal:
/cc @jknipper


autofix-beta-apiversions.go
```diff
@@ -67,7 +123,11 @@ func main() {
 				os.Exit(1)
 			}
 			fmt.Printf("%s: autofixing %s/%s -> %s/%s\n", path, apiVersion, kind, newAPIVersion, kind)
-			lines[idx] = strings.Replace(line, apiVersion, newAPIVersion, 1)
+			newLine := fmt.Sprintf("apiVersion: %s", newAPIVersion)
+			if shouldCheckAPICapabilities[kind] {
+				newLine = wrapInCapabilitiesCheck(apiVersion, newAPIVersion)
```
majewsky, line 128: Is there anything that prevents an already-wrapped API version from being wrapped again on subsequent runs?

majewsky (commented):
According to the [deprecation documentation](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22), these API version upgrades come with structural changes to the respective objects. That's basically the same as for the previous wave of deprecations in 1.16, but this time around, the changes are much more significant. For example, every Ingress uses `serviceName` which is renamed to `service.name`, so we would need to account for these syntax changes as well. And if we have the `kind:` in a conditional, we need to have the syntax changes wrapped as well.

That's something that I did not anticipate when initially scoping the ticket; it only occured to me now when checking this PR and reviewing the deprecation notice.

How do we want to proceed with this? Asking especially towards @jknipper. I would like to avoid making a huge mess with conditional syntaxes, but can we afford to wait for the final 1.20 upgrades before proceeding with these changes?

autofix-beta-apiversions.go
talal: There is now 😅

jknipper:
As @majewsky pointed out there are structural changes in some resources. Also I don't know what will happen if we update cloud backed resources, eg. ingresses. I think this is something the service owner needs to tackle down. I think populate the doop dashboard with deprecation information is a good first step to bring this issue to a broader audience.

majewsky:
Yeah, let's only do the adjustments to the autofixer in this PR and then push the policy out.

PR #2946, opened 2021-11-26
system/gatekeeper: use a set for deprecated API versions

talal:


majewsky:
Looks good, but I will hold off on the ✔️ for now because of the same concern as in #2945.

majewsky (approved):

PR #2947, opened 2021-11-26
swift: upgrade memcached

SuperSandro2000:
See commit messages for release notes


reimannf (approved):

majewsky (approved):
I double-checked the release notes and found no backwards-incompatible changes, so we should be good to go. Still, I would like to let this simmer in QA for a few days before going to prod. ~Please wait for Falk's approval before merging to ensure that no prod deployments are blocked.~ **Edit:** Falk ninja'd me. :)

Also, when deploying to QA, please check that the metric delivery still works as expected. We don't use the memcached metrics for alerts, but there is a Memcached dashboard in the regional Grafana that should be checked.

PR #2970, opened 2021-12-08
[manila] allow filer's individual setting: netapp_volume_snapshot_reserve_percent

chuan137:
Need to set the field to 0 for the hypervisor_storage filers.


kpsap:
@chuan137 How nanny behave if snap_reserve is 0% ?

galkindmitrii:
this will be set only for hypervisor storage and nanny will ignore those filers

kpsap (approved):

PR #2972, opened 2021-12-08
[neutron] Set search domain as local in dnsmasq

joker-at-work:
We don't expose openstack.$region.cloud.sap or its subdomains for
customers to register, but only use it locally in the current network
and on the local dnsmasq. Therefore, we want to configure dnsmasq to
also not recurse queries for this domain if they cannot be answered
locally.

This is supposed to fix a problem one of gardener's customers has, as
they use the search-domain for everything with less than 5 dots (i.e.
ndots 5 setting) and thus do multiple queries to the network's DNS
server for a "normal" domain like "www.googleapis.com".

### DIFF
--- a/openstack/neutron/templates/etc/_dnsmasq.conf.tpl
+++ b/openstack/neutron/templates/etc/_dnsmasq.conf.tpl
@@ -6,3 +6,4 @@ no-negcache
 {{- range .Values.dnsmasq.dhcp_options }}
 dhcp-option={{ . }}
 {{- end }}
+local=/{{required "A valid .Values.dns_local_domain is required!" .Values.dns_local_domain }}/


openstack/neutron/templates/etc/_dnsmasq.conf.tpl
```diff
@@ -6,3 +6,4 @@ no-negcache
 {{- range .Values.dnsmasq.dhcp_options }}
 dhcp-option={{ . }}
 {{- end }}
+local=/openstack.{{ .Values.global.region }}.cloud.sap/
```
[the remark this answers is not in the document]
joker-at-work: Sounds like a really good idea :+1:

openstack/neutron/templates/etc/_dnsmasq.conf.tpl
```diff
@@ -6,3 +6,4 @@ no-negcache
 {{- range .Values.dnsmasq.dhcp_options }}
 dhcp-option={{ . }}
 {{- end }}
+local=/{{ .Values.dns_local_domain }}/
```
notandy, line 9: Please use a `required` helm function for the value to ensure that local is not defined if the value is empty.

notandy (changes requested):

openstack/neutron/templates/etc/_dnsmasq.conf.tpl
joker-at-work: Are you sure we want to require this or should we add an `if` around it, that only sets `local` if we have a `dns_local_domain` set to something?

notandy: If we plan to roll it out worldwide on not specifically for some regions (I don't see a reason why have this "feature" only in selected regions), then it's fine to go with required.
Just wanted to ensure that we don't roll out this configuration without a valid value (e.g. due to missing --values or so)

notandy (approved):

PR #2975, opened 2021-12-08
[keystone] Enable authenticate.failed notifications

bbobrov:
Keystone should emit notifications about failed authn attempts so that
the notifications could be consumed by an auditing software

### DIFF
--- a/openstack/keystone/values.yaml
+++ b/openstack/keystone/values.yaml
@@ -69,7 +69,6 @@ api:
       # ignore all authentication events since the scope (project_id/domain_id) is missing
       # Note: exception for *authenticate* event types: the outcome has to be added as suffix to make this work
       # see https://github.com/openstack/keystone/blob/af4e98c770d771144463e6dd49cb4b559d48c403/keystone/notifications.py#L743
-      - identity.authenticate.failed
       - identity.authenticate.pending
 
   token:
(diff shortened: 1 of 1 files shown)


rajivmucheli (approved):
lgtm

notque:
Please comment the line out instead of removing it for context in that section given it's non-obvious what the paths as it doesn't match the documentation, and instead refers to the link.

notque:
I'd like to get this merged on Monday so we can get the change deployed, and I can start work on logstash pipeline changes that we require. Thanks!

bbobrov:
> Please comment the line out instead of removing it for context in that section given it's non-obvious what the paths as it doesn't match the documentation, and instead refers to the link.

Interesting, there is difference with the documentation indeed. I agree, lets comment it out instead.

bbobrov:
so i decided to dig a little deeper. The correct name is `identity.authenticate.`**`failure`**, and in the list in this change there is  `identity.authenticate.`**`failed`**. Which means that `identity.authenticate.`**`failure`** events have been emitted all this time. This means several things:
1. There is a bug in upstream keystone - by default it emits some events even though it should not.
2. There is a minor bug in our system - we have been emitting events even though we did not want it.
3. The events are already available for consuming. I have checked and they are in hermes' elasticsearch already indeed - check out event  `d2472531-5dae-533f-be81-8a1dbd4ee1a1` in the qa.

This PR needs to be reworked to comment out the right name, but it is a cosmetic change that hermes is not depending on.

PR #2976, opened 2021-12-09
Revert "[vrops-exporter] make vm tool version metric globally available"

richardtief:
Reverts sapcc/helm-charts#2939


richardtief:
Too many metrics. This broke prometheus global.

PR #2977, opened 2021-12-09
Added support for asr9k (not peers) in ssh-exporter

dimtass:
@swagner-de can you please have a look at this PR. This works only for the router stats (not peers) using the `show ntp stats` results. This doesn't need any changes to the ssh-exporter.

### DIFF
--- a/system/infra-monitoring/vendor/atlas/templates/configmap.yaml
+++ b/system/infra-monitoring/vendor/atlas/templates/configmap.yaml
@@ -92,12 +92,11 @@ data:
                 credential: "default"
                 device: "cisco-ios-xr"
               metrics_label: "corerouter"
-              target: 1
-              manufacturer: "cisco"
-              region: "{{ .Values.global.region }}"
               role: "core-router"
+              platform: "cisco-ios-xr"
               tenant: "converged-cloud"
-              q: "rt"
+              target: 1
+              region: "{{ .Values.global.region }}"
               status: "active"
 
             - custom_labels:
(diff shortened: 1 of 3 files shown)


system/infra-monitoring/vendor/atlas/templates/configmap.yaml
```diff
@@ -86,6 +86,20 @@ data:
...
+                job: "network/ssh"
+                credential: "default"
+                device: "cisco-ios-xr"
+              metrics_label: "corerouter"
+              target: 1
+              manufacturer: "cisco"
+              region: "{{ .Values.global.region }}"
+              role: "core-router"
+              tenant: "converged-cloud"
+              q: "rt"
```
swagner-de, line 100: Is there a reason for this query string? Cause it letls our the routers on the old naming convention.

swagner-de (changes requested):

system/infra-monitoring/vendor/atlas/templates/configmap.yaml
dimtass: This exporter meant to be targeting only the ASR9K routers that don't support SNMP OIDs. The networking team provided me with this filter in netbox which will target those routers. If the naming convention is changed then they will provide me with the new fiter.

dimtass: Field filters were changes as requested after the proper fields were added in netbox.

PR #2980, opened 2021-12-10
[keystone][hermes] enable authenticate events success fail

notque:
Enable success, and failure events for authenticate. Unsure if we want to do pending.


bbobrov:
conflicting with #2975

notque:
Closing as duplicate, handling in #2975

PR #2986, opened 2021-12-14
openstack/backup-replication, openstack/content-repo: statsd 0.8.1 -> 0.22.4

SuperSandro2000:


majewsky (approved):
Let's give this a try in QA. Please check in the cc-os-backup and cc-os-swift Slack channels whether any alerts arise after deploying this. Since this is about a rather small number of metrics, it should also be possible to validate directly in prometheus-openstack that the metrics are correctly scraped in the same format as before.

PR #2999, opened 2021-12-16
ResMgmt: add owner info

talal:
/cc @reimannf @SuperSandro2000


openstack/clair/values.yaml
```diff
@@ -3,6 +3,13 @@ global:
   region: DEFINED_IN_VALUES_FILE
   registry: DEFINED_IN_VALUES_FILE
 
+owner-info:
+  maintainers:
+    - Stefan Majewsky
+    - Falk Reimann
+    - Sandro Jäckel
```
SuperSandro2000, line 10:
```suggestion
    - Sandro Jäckel
    - Falk Reimann
```

openstack/backup-etcd/values.yaml
```diff
@@ -1,3 +1,10 @@
+owner-info:
+  maintainers:
+    - Falk Reimann
+    - Stefan Majewsky
+    - Sandro Jäckel
+    - Muhammad Talal Anwar
+
```
majewsky, line 7: backup-etcd has not been used for several years (ever since we moved the controlplane to Kubernikus). @reimannf Any objections to deleting this chart entirely?

openstack/swift-utils/values.yaml
```diff
@@ -1,9 +1,16 @@
+# NOTE: This chart usually consumes the same values.yaml as the "swift" chart.
+
```
majewsky, line 2: They don't anymore. We split the values files several years ago, so this is a good opportunity to remove this note.

majewsky (changes requested):
Please pull the chart via `repository: https://charts.eu-de-2.cloud.sap` instead of `repository: file://../../common/owner-info`. We will want everyone else to use the chart as well, and some teams insist on using Chartmuseum for all dependencies. Since it causes friction when a chart gets pulled via Chartmuseum in some charts and via directory reference in other charts, we should follow the crowd here right away.

openstack/backup-etcd/values.yaml
reimannf: Yes it can be deleted

talal: Deleted ☠️

majewsky (approved):

majewsky:
CI says:
```
[ERROR] Chart.yaml: dependencies are not valid in the Chart file with apiVersion 'v1'. They are valid in apiVersion 'v2'
```
Which is true. I was surprised that it works in Chart.yaml for v1, I thought it has to be in `requirements.yaml` for v1. But judging from backup-replication in QA, it works?

talal:
Yeah backup-replication seems to work fine. I guess it's more of soft warning.

I have migrated the charts to `v2`.

PR #3003, opened 2021-12-17
[sap-seeds] Support 2 different list of hana* flavors

joker-at-work:
We need to deploy hana* flavors with different settings in some regions,
because we want to switch to HANA-only clusters. Since switching is not
possible everywhere at once, we now use 2 different lists of hana*
flavors in this chart and make it possible to switch between both via
secrets value "use_hana_exclusive".


dhoeller (approved):
LGTM

PR #3006, opened 2021-12-20
[sap-seeds] Deprecated bigVM flavors in non-exclusive list

joker-at-work:
We only want the deprecated bigVM flavors availabe in regions, where we
don't have BBs marked as exclusively for HANAs, because we don't run
nova-bigvm in those regions anymore and thus those flavor, which depend
on the CUSTOM_BIGVM resource, will never be deployable.


dhoeller (approved):
LGTM

PR #3013, opened 2021-12-23
[Cinder] Update rabbitmq from 0.2.0 to 0.2.6

hemna:
This patch updates the cinder helm chart to use rabbitmq chart 0.2.6


joker-at-work (approved):
I'd also write into the commit-message, why we want that update i.e. what we're trying to achieve here other than updating the version.

PR #3018, opened 2022-01-05
[Swift] do not expire async pending metrics

reimannf:
With `statsd_exporter` 0.9.0 TTLs per mapping were introduced. This leads now to expiration of counters, which was not the case previously although we had the global TTL setting.
https://github.com/prometheus/statsd_exporter/blob/master/CHANGELOG.md#090--2019-03-11

To avoid absent alert flapping in underutilized regions, the async pending metrics do not expire with this setting.

### DIFF
--- a/openstack/swift/etc/statsd-exporter.yaml
+++ b/openstack/swift/etc/statsd-exporter.yaml
@@ -117,8 +117,12 @@ mappings:
   name: swift_object_replicator
   labels:
     action: $1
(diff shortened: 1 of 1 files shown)


majewsky (approved):

reimannf:
@majewsky: fyi https://github.com/sapcc/helm-charts/pull/3018/commits/2782232c2f203756df2cc7c4cd7ed551e105dcce

PR #3021, opened 2022-01-05
[manila] reduce default number of volumes per svm

Carthaca:
20 is the sweet spot for us.
Roughly calculated out of:
max 5000 volumes (NetApp limit on certain models)
max 256 SVMs (due to 2 lifs per svm and max 512 lifs NetApp limit)

@frr2018 fyi

to be deployed after xena upgrade


galkindmitrii:
Maybe documentation update + blog post would be a good idea?
-> Nobody expects 5x times more ports in their networks all of a sudden.

Carthaca:
The current behaviour is in the FAQ [How many neutron ports are created by manila?](https://documentation.global.cloud.sap/compute/filestore-faq-general)

PR #3025, opened 2022-01-07
[manila] helm outdated-dependencies update

Carthaca:
most likely to be done after xena upgrade

[rabbitmq] increase timeout for readiness probe to 3 sec
[rabbitmq] harmonize component label across charts
[mysql_exporter] remove mysql_database_size default metric as there is
pvc_exporter and respective alert now
[sql_exporter] allow to define additional custom metric sources in
values
[common][mysql_exporter] an attempt to add nova cell2 as a second
source for querying.


galkindmitrii:
thanks @Carthaca, would be nice to grab https://github.com/sapcc/helm-charts/pull/3029 as well when merged

PR #3029, opened 2022-01-10
[common][rabbitmq] use keppel for rabbit images

galkindmitrii:
images.

Alternate region is used in a similar way like for keystone or mariadb
charts. In case Keppel down in the region - the service won't be able to
start.

Also bump chart.

### DIFF
--- a/common/rabbitmq/templates/deployment.yaml
+++ b/common/rabbitmq/templates/deployment.yaml
@@ -38,7 +38,7 @@ spec:
       hostname: {{ template "fullname" . }}
       containers:
       - name: rabbitmq
-        image: "{{ .Values.image }}:{{.Values.imageTag }}"
+        image: "{{include "dockerHubMirror" .}}/{{ .Values.image }}:{{.Values.imageTag }}"
         imagePullPolicy: {{ default "IfNotPresent" .Values.imagePullPolicy | quote }}
         command:
           - bash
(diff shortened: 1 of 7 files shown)


Carthaca (changes requested):
I would prefer to have a switch like introduced for mariadb in 922f101559c1daacf5717e29fb1ffde341e8b5c3

some services may want to use the alternative region (e.g. keystone, swift, designate) to prevent a circular dependency for keppel, but other services probably not

majewsky:
I agree with Maurice. Usage of `AlternateRegion` should be restricted to the charts that need it. It is absolutely correct that Keystone and Designate use `AlternateRegion` to avoid a deadlock during regional cold-start, but for all other services, adding the DC interconnect to the alternate region as an additional deployment dependency is not really useful.

common/rabbitmq/templates/deployment.yaml
```diff
@@ -38,7 +38,7 @@ spec:
       hostname: {{ template "fullname" . }}
       containers:
       - name: rabbitmq
-        image: "{{ .Values.image }}:{{.Values.imageTag }}"
+        image: {{include "dockerHubMirror" .}}/{{ .Values.image }}:{{.Values.imageTag }}"
```
Carthaca, line 41: uneven number of `"`

openstack/designate/values.yaml
```diff
@@ -7,6 +7,7 @@ debug: "True"
 
 global_setup: false
 tempest_enabled: false
+use_alternate_registry: true
```
Carthaca, line 10: This needs to go under the `rabbitmq` key(s)

openstack/keystone/values.yaml
```diff
@@ -27,6 +27,9 @@ skipRegionCheck: false
 # current keystone release
 release: train
 
+# fetch keppel images from another region
+use_alternate_registry: true
```
Carthaca, line 31: same

Carthaca (changes requested):

Carthaca (approved):
lgtm

PR #3040, opened 2022-01-12
Define PX-Switch Alerts for EVPN and OSPF

swagner-de:


nmilovancevic:
Shouldn't we set this as critical alert:
`alert: PXSwitchBgpDown`
Because this would mean that most probably the whole AZ is disconnected.

PR #3042, opened 2022-01-14
[Rabbitmq] Add ability to unset priority-class

fwiesel:
The default-function interpretes empty strings also as unset,
thereby not differentiating between an unset value and a priority-class
set explicity to an empty string.


fwiesel:
The cronus service needs to deploy to the scaleout cluster, and there is only the system default, which we do not want to use.

PR #3045, opened 2022-01-14
[manila] enforce max share size

Carthaca:
16TB for normal types
64TB for hypervisor storage

DRAFT: we need to check what we have currently in the system already
documentation already states this limit, but it was not enforced


Carthaca:
is already in master

PR #3046, opened 2022-01-14
[manila] allow shares from snapshot up to 70% fillgrade

Carthaca:
- other shares still only up to 50% fillgrade
- shares from snapshot need to be created on the same host ideally


kpsap:
looks good

galkindmitrii:
just keep in mind that share creation from snapshots will start failing on some filers

Carthaca:
was added by https://github.com/sapcc/helm-charts/commit/b563a832003abd51c39f35a8e9eb4d66e811cd6e

PR #3048, opened 2022-01-17
[cerebro] Add project seed

ofrendo:


openstack/cerebro/regions/ap-ae-1/values/cerebro.yaml
```diff
@@ -0,0 +1 @@
+technicalUserPerRegion: TCC_D059373_01
```
viennaa, line 1: please remove the users, I wouldn't put them in the sapcc repository

Add a file `openstack/cerebro/values.yaml` with
```
technicalUserPerRegion: null
```

I'm going to push the user you specified in our cmdb.

viennaa (changes requested):

openstack/cerebro/regions/ap-ae-1/values/cerebro.yaml
ofrendo: Done

viennaa: I meant completely removing the `regions` path and putting a `values.yaml` in the root of this like `openstack/cerebro/values.yaml`

PR #3054, opened 2022-01-19
[prometheus-exporters] increase PVC full severity to Warning.

galkindmitrii:
Alert on INFO level was missed recently, causing all openstack service
metrics go missing in eu-de-1 and no alerting possible.


rajivmucheli:
lgtm

PR #3056, opened 2022-01-19
drop metrics with very high cardinality filling the tsdb

artherd42:


thgrs (approved):
looks good to me

databus23:
I think this metric belongs to a histogram there should be `openstack_http_response_time_sum` and `openstack_http_response_time_count` metrics that kind of belong together. Is it maybe drop all metrics starting with `openstack_http_response_time_`

PR #3059, opened 2022-01-20
drop some metrics from pod-sd and endpoint-sd

artherd42:
prep to drop high cardinality openstack_http_response_* metrics

### DIFF
--- a/common/prometheus-server/templates/_helpers.tpl
+++ b/common/prometheus-server/templates/_helpers.tpl
@@ -135,3 +135,9 @@ prometheus-{{- (include "prometheus.name" .) -}}
   regex: ^({{ . | join "|" }})$
   action: keep
 {{- end -}}
+
+{{- define "prometheus.drop-metrics.metric-relabel-config" -}}
+- sourceLabels: [ __name__ ]
+  regex: ^({{ . | join "|" }})$
+  action: drop
+{{- end -}}
--- a/common/prometheus-server/CHANGELOG.md
+++ b/common/prometheus-server/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.5.4
+
+* [endpoint/service SD] Exclude set of metrics from being scrapped.
+* [pod/service SD] Exclude set of metrics from being scrapped.
+
 ## 4.5.3
 
 * [endpoint/service SD] Fix scraping of ports declared by the pod but not the service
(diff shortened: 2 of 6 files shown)


common/prometheus-server/templates/_helpers.tpl
```diff
@@ -135,3 +135,9 @@ prometheus-{{- (include "prometheus.name" .) -}}
   regex: ^({{ . | join "|" }})$
   action: keep
 {{- end -}}
+
+{{- define "prometheus.drop-metrics.metric-relabel-config" -}}
+- sourceLabels: [ __name__ ]
+  regex: ^({{ . | join "|" }})$
```
majewsky, line 141: List items should be escaped with `regexQuoteMeta` to avoid accidental overmatching or mismatching. Since the templating syntax does not have a `map` or similar function to apply the quoting across all list elements, I'd rather suggest passing in the forbidden metrics as one regex from the get-go.

artherd42: As per existing ServiceMonitor prometheus-collector-kubernetes-cadvisor, which uses the same syntax for rendering metricRelabelings, thei shoudn't be an issue:
```
    - action: drop
      regex: POD;(container_cpu_cfs_periods_total|container_cpu_cfs_throttled_seconds_total|container_cpu_cfs_throttled_periods_total|container_cpu_usage_seconds_total|container_cpu_user_seconds_total|container_cpu_system_seconds_total|container_fs_inodes_total|container_fs_limit_bytes|container_fs_usage_bytes|container_last_seen|container_memory_usage_bytes|container_memory_working_set_bytes|container_start_time_seconds)
      sourceLabels:
      - container
      - __name__
```

auhlig:
Cool!
Can you increment the [Chart version](https://github.com/sapcc/helm-charts/blob/33dcbaa9575e2cd5f28a2d50b319935de5f01b1a/common/prometheus-server/Chart.yaml#L4) and add a small note to the [changelog](https://github.com/sapcc/helm-charts/blob/33dcbaa9575e2cd5f28a2d50b319935de5f01b1a/common/prometheus-server/CHANGELOG.md) @artherd42 ?
Feel free to merge after that

artherd42:
Done! Please review and merge @auhlig.

auhlig (approved):

auhlig:
`prometheus-server@4.5.4` should be usable within the next minute

PR #3061, opened 2022-01-20
[nova] Add CpuInfoMigrationFilter

fwiesel:
This filter checks for cpu compatability for live-migrations.
It requires that the corresponding filter is in the image
used for the deployment.


joker-at-work (approved):
commit-message: "compatability" -> "compatibility"

fwiesel:
Fixed the spelling

PR #3067, opened 2022-01-21
[vrops-exporter] VCenterHasMultipleGreyVMs: consider host state

stanislav-zaprudskiy:


christopherhans (commented):
Your expression works, although it is more complicated than needed here. If the hostsystem of a VM and host metric is the same, datacenter, vccluster and vcenter are also the same. Or is there some unusual behaviour with grey VMs?

`count by(vcenter) (
(vrops_virtualmachine_health == -1) 
and on(hostsystem) vrops_hostsystem_runtime_powerstate{state="Powered On"}
and on(hostsystem) vrops_hostsystem_runtime_connectionstate{state="connected"}
and on(hostsystem) vrops_hostsystem_runtime_maintenancestate{state="notInMaintenance"})` would also do the trick.

stanislav-zaprudskiy:
@christopherhans , thank you for the feedback, I updated the PR as suggested.

PR #3083, opened 2022-01-26
[alerts] rm VmfsDataStoreCapacity* for B1

geisslet:
@richardtief @rmueller-b1 

`VmfsDataStoreCapacity` alerts are turned off because a change is implemented in Openstack so that `Cinder` now has control over scheduling new volumes. It is also set so that a datastore is only filled to about 80%. We still monitor the average utilization of all datastores in a vCenter, but no longer monitor the datastores themselves.


richardtief (approved):
Looks fine.

PR #3102, opened 2022-02-01
introduce enable-pamd-tty to enable the init container

SchwarzM:


system/auditbeat/templates/auditbeat-daemonset.yaml
```diff
@@ -26,6 +26,7 @@ spec:
       hostNetwork: true
       hostPID: true  # Required by auditd module
       dnsPolicy: ClusterFirstWithHostNet
+{{- if .enable_pamd_tty }}
```
jknipper, line 29: I think it need to be `{{- if .Values.enable_pamd_tty }}`.

jknipper:
You also need to bump chart version here and in audit-logs chart.

jknipper:
Please do a rebase on master branch., chart versions already changed.

PR #3112, opened 2022-02-02
Prometheus: metric to measure L3 port utilisation

rhalina:
Adding a check to generic-ssh-exporter in prometheus to measure the number of
reserved transport ports on ASR routers in qa-de-* regions.

### DIFF
--- a/prometheus-exporters/network-generic-ssh-exporter/templates/_config.yaml.tpl
+++ b/prometheus-exporters/network-generic-ssh-exporter/templates/_config.yaml.tpl
@@ -455,6 +455,30 @@ metrics:
     command: show ntp config
     timeout_secs: 5
 
+  port_utilization_TCP:
+    regex: >-
+      \s*(\d+)\s{3}TCP\sIPv4
+    multi_value: true
+    metric_type_name: gauge
+    value: $1
+    labels:
+        port: $1
+    command: show ip ports reservation | ex TPM
+    description: Plot the number of reserved TPA ports (UDP/TCP) to check if unused ports get released again.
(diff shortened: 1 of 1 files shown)


prometheus-exporters/network-generic-ssh-exporter/templates/_config.yaml.tpl
```diff
@@ -547,6 +547,30 @@ metrics:
...
+    value: $1
+    labels:
+        port: $1
+    command: show ip ports reservation | ex TPM
+    description: Plot the number of reserved TPA ports (UDP/TCP) to check if unused ports get released again.
+    timeout_secs: 3
+
+  port_utilization_UDP:
+    regex: >-
+      \s*(\d+)\s{3}.*\s+UDP\sIPv4
```
swagner-de, line 565: This Regex matches only once for the whole block.
I could get a result with `^\s*(\d+)\s{3}.*?\s+UDP\sIPv4.*?$`.

prometheus-exporters/network-generic-ssh-exporter/templates/_config.yaml.tpl
```diff
@@ -588,8 +612,10 @@ batches:
     - firewall_vrf_stats_half_open_udp
     - firewall_vrf_stats_half_open_tcp
     - firewall_vrf_stats_half_open_icmp
-
-
+  {{- if (hasPrefix "qa-de-".Values.global.region) }}
```
swagner-de, line 615: You are missing a space here.

swagner-de (changes requested):

PR #3122, opened 2022-02-04
[mariadb-backup] removes unnecessary prometheus port scrape

stefanhipfel:
to remove multiple targets warnings.

Also put scrape annotations only to the deployment


artherd42 (approved):
lgtm

PR #3128, opened 2022-02-07
[maia] drop some technical labels for prometheus-openstack

Carthaca:
similar to 46834805462b35b09c7a0bce4428fb7961d96b29


artherd42 (commented):
thanos labels can't be dropped as required by thanos.

PR #3130, opened 2022-02-08
[vrops-exporter] Add source vrops collector to nsxt alerts

christopherhans:
The operations team requested to have to source vROps listed at every NSX-T alert. I picked the collector label, cause it is the closest to what we need without much label handling needed.


richardtief (approved):
Looks good. Thanks.

PR #3133, opened 2022-02-08
drop certain auditbeat fields in the config

SchwarzM:


IvoGoman (approved):
Processor looks good as far as I can tell. Thanks for taking care of dropping these fields

PR #3140, opened 2022-02-09
[cinder] use formatstring template

notandy:


hemna (commented):
lgtm

hemna (approved):

joker-at-work:
... and now we lost the `%(resource)s` from the format-string again :)

notandy:
> ... and now we lost the `%(resource)s` from the format-string again :)

could you elaborate?

joker-at-work:
> could you elaborate?

sure. we added `%(resource)s` in https://github.com/sapcc/helm-charts/pull/3248 which is not part of the format-string in `utils` and thus, with merging this PR, we lost that part of the format-string again.

notandy:
ah ok, this of course would break the unified formatstring effort.

joker-at-work:
Would it? We have `%(instance)s` in there, too, so we could just add `%(resource)s` in addition, right?

notandy:
> Would it? We have `%(instance)s` in there, too, so we could just add `%(resource)s` in addition, right?

Ah you mean it would be a "replacement" to `%(instance)s`
I am fine to add globally the `%(resources)s` formatstring - it should be ignored for all other openstack services?

joker-at-work:
It's not a replacement but an addition. A service sets either `resource` or `instance` or non of them. If the service doesn't set anything, it doesn't matter - same as with `instance`.

joker-at-work:
https://github.com/sapcc/helm-charts/pull/3355 something like this should work

PR #3150, opened 2022-02-11
Added prom scrap config for vault

dimtass:
Added vault scrap config


databus23 (changes requested):
This is problematic in two ways. First and foremost behind the url are three vault replicas so scraping the load balancer url will randomly pick any of the threee replicas giving inconsistent metrics.
Second, instead of modifying the prometheus infra collector scrape config we should use a  `ServiceMonitor` resource that we can ship within our `vault` chart.

PR #3160, opened 2022-02-16
[vrops-exporter] no alerting for hosts in maintenance

richardtief:
Most alerts are for hosts in maintenance. This PR should reduce the noisy alert channels.


viennaa (approved):
Good catch

PR #3163, opened 2022-02-17
removed tls annotations for ingress statsd exporter to probe

developer-abhi:
@artherd42 : please approve this


prometheus-exporters/statsd-exporter/templates/ingress.yaml
```diff
@@ -11,11 +11,6 @@ metadata:
     vice-president: {{ required ".Values.ingress.vicePresident missing" .Values.ingress.vicePresident | quote }}
     kubernetes.io/tls-acme: {{ required ".Values.ingress.certManager missing" .Values.ingress.certManager | quote }}
     disco: {{ required ".Values.ingress.disco missing" .Values.ingress.disco | quote }}
-  {{- if .Values.ingress.authentication.enabled }}
```
artherd42, line 14: You don't need to remove the whole code block. authentication could be disabled in the chart values https://github.com/sapcc/helm-charts/blob/master/prometheus-exporters/statsd-exporter/values.yaml#L34-L36 or per region in the secrets.

PR #3183, opened 2022-02-23
[keystone] Tolerate more errors in the lifesaver

bbobrov:
Currently the lifesaver middleware grants 100 credits to a user and
takes 5-10 per failed operation, which allows only 10-20 failed
operations per minute. We can easily handle more than that, and some
failed operations are generated by cli, non-perfect terraform scripts
and just by simple small errors, that users should not be punished for.

Raise the amount of credits that the user is granted and make refill of
the credits faster. This will allow 5 times failed operations per minute
more.

### DIFF
--- a/openstack/keystone/templates/etc/_keystone.conf.tpl
+++ b/openstack/keystone/templates/etc/_keystone.conf.tpl
@@ -203,11 +203,11 @@ domain_allowlist = {{ .Values.lifesaver.domain_allowlist | default "Default, tem
 user_allowlist = {{ .Values.lifesaver.user_allowlist | default "admin, keystone, nova, neutron, cinder, glance, designate, barbican, dashboard, manila, swift" }}
 user_blocklist = {{ .Values.lifesaver.user_blocklist | default "" }}
 # initial user credit
-initial_credit = {{ .Values.lifesaver.initial_credit | default 100 }}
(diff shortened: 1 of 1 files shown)


rajivmucheli (changes requested):
could you update the values here as well to avoid confusion :

https://github.com/sapcc/helm-charts/blob/master/openstack/keystone/templates/etc/_keystone.conf.tpl#L212

bbobrov:
Sorry, i do not understand. How should they look like? My understanding is that we change only the initial credit that the user has and the refill speed, and do change how much is deducted.

rajivmucheli:
pasted the wrong link, its :

https://github.com/sapcc/helm-charts/blob/master/openstack/keystone/templates/etc/_keystone.conf.tpl#L206
https://github.com/sapcc/helm-charts/blob/master/openstack/keystone/templates/etc/_keystone.conf.tpl#L208

rajivmucheli:
we could add the new values here, instead:

initial_credit = {{ .Values.lifesaver.initial_credit | default 500 }}
refill_seconds = {{ .Values.lifesaver.refill_seconds | default 60 }}

bbobrov:
But why change it in 2 places? We use variables specifically so that we don't have to edit the helm charts when we want to change some parameters

rajivmucheli (approved):
lgtm, values are in 1 place now.

PR #3189, opened 2022-02-25
[Cinder] Set Openstack Request Id in Ingress

fwiesel:
If the request did not came in over ingress (e.g. internally by the service) it will render to gNon


rajivmucheli (approved):
thanks.

PR #3198, opened 2022-03-01
[octavia] Remove .html from playbook links

laerlingSAP:
With the new documentation system (hugo), playbook links with html at the end
usually work, but for some reason they don't for the stuck_pending link.

In order to enforce a uniform format (and not confuse future contributors), I
remove .html on every octavia playbook link.


killermoehre (approved):
LGTM

PR #3199, opened 2022-03-01
remove all the .html endings for playbooks

killermoehre:
```bash
git --no-pager grep --name-only -E 'playbook:.*\.html' \
| while read -r file; do
    sed -i 's|\(playbook:.*\)\.html|\1|g' "$file";
done
```
```bash
perl -pi -e 's|(playbook: .+?)\.html(.*)|\1\2|g' $(git grep -l -P '(playbook: .+?)\.html(.*)')
```


laerlingSAP:
LGTM, but since I adapted Octavia playbook links in #3198 already, I'm the last person to have a say in this.

geisslet:
### I vote against in the meanwhile!

1. the check done doesn't validate the risk we take with that massive change
2. it needs a rollout everywhere

Please let's understand first.

PR #3205, opened 2022-03-02
[hermes] add local user table into enrichment

bbobrov:


notque:
That is adding a user table, but doing nothing with it. Do you want me to add the fields?

PR #3213, opened 2022-03-07
[common][rabbitmq] add rabbitmq_node_reinstall_affinity

Carthaca:
limit number of rabbitmq restarts during kubernetes node reinstall


Carthaca:
@galkindmitrii fyi

PR #3217, opened 2022-03-08
maia: drop labels "alert_tier" and "alert_service" on ingestion

majewsky:
These labels are maintained at the pod level to direct k8s alerts to the service-specific Slack channel, so scraping via pod SD will pick them up. They are an internal implementation detail of our alerting setup and should not be exposed to customers.


thgrs (approved):
lgtm

PR #3220, opened 2022-03-09
[Cinder] Reduce cinder-volume restarts during k8s node maintenance

hemna:
We bump the utils chart here to get the
"kubernetes_maintenance_affinity" helper, so we can easily define the
affinity to operational nodes (compared to nodes waiting to upgrade) for
starts/restarts during k8s node maintenance.

This adds affinity for cinder-volume pods.

### DIFF
--- a/openstack/cinder/templates/vcenter_datacenter_cinder_deployment.yaml
+++ b/openstack/cinder/templates/vcenter_datacenter_cinder_deployment.yaml
@@ -40,6 +40,7 @@ template: |
       spec:
         hostname: cinder-volume-vmware-{= name =}
 {{ tuple "{= availability_zone =}" | include "kubernetes_pod_az_affinity" | indent 8 }}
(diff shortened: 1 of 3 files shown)


openstack/cinder/templates/vcenter_datacenter_cinder_deployment.yaml
```diff
@@ -44,6 +44,7 @@ template: |
         hostname: cinder-volume-vmware-{= name =}
 {{ tuple "{= availability_zone =}" | include "kubernetes_pod_az_affinity" | indent 8 }}
 {{ include "utils.proxysql.pod_settings" . | indent 8 }}
+{{- include "kubernetes_maintenance_affinity" . | indent 4 }}
```
joker-at-work, line 47: This needs to go one line up, because it extends the affinity dict created by `include "kubernetes_pod_az_affinity"`

joker-at-work (changes requested):

PR #3223, opened 2022-03-10
Nova: Switch for standalone placement

fwiesel:
Unless specifically opted-in (by setting placement.enabled=False)
this PR doesn't change single line / checksum.

To migrate, deploy first the standalone placement chart
(different changeset).
This will update the endpoint in keystone, so all request
should be routed through the new deployment.

Then deploy this chart with placement.enabled=False,
which will undeploy nova internal placement.

The dependency of the data in the nova_api database needs
to be solved separately.

### DIFF
--- a/openstack/nova/templates/seed.yaml
+++ b/openstack/nova/templates/seed.yaml
@@ -56,6 +56,7 @@ spec:
     - interface: public
       region: '{{ .Values.global.region }}'
       url: 'https://{{include "nova_api_endpoint_host_public" .}}:{{.Values.global.novaApiPortPublic}}/v2/%(tenant_id)s'
+{{- if .Values.placement.enabled }}
   - name: placement
     type: placement
     description: Openstack Placement API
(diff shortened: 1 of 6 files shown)


fwiesel:
It depends on #3224 being deployed first

openstack/nova/templates/seed.yaml
```diff
@@ -69,7 +70,7 @@ spec:
     - interface: public
       region: '{{ .Values.global.region }}'
       url: 'https://{{include "placement_api_endpoint_host_public" .}}:{{.Values.global.placementApiPortPublic}}'
-
+{{ end }}
   domains:
   - name: Default
     users:
```
joker-at-work, line 76: The role-assignments down below don't have a `if .Values.placement.enabled`. Is that by choice? If yes, please add the reasoning to the commit-message.

joker-at-work (commented):
commit-message:

> This split out the placement config, and adds an option

"splits"

joker-at-work (approved):

PR #3235, opened 2022-03-15
sentry: update pgmetrics chart to 0.3.0

majewsky:
Besides cleaning up the chart a bit, the primary motivation is to upgrade the postgres-exporter. There are two separate metrics pods in Sentry, both on ancient postgres-exporter versions:

- the pgmetrics chart includes the sentry-pgmetrics deployment using postgres-exporter v0.4.6 (the current version is v0.10.1)
- the postgres chart includes a sidecar container in the sentry-postgresql deployment at postgres-exporter v0.1.1 (!)

Since it's a bad idea anyway to stuff too many sidecars into the Postgres pod (since then every change to the sidecar requires restarting the DB), this PR gets rid of the second postgres-exporter entirely. Note that this causes a DB restart at deploy time.

There will be a separate PR in the secrets repo for the respective values.yaml change.


rajivmucheli (approved):
thanks

PR #3240, opened 2022-03-16
[ElasticQueryExporter] Adds alerts for audit logs

IvoGoman:
- removes file suffix from PrometheusRule name
- includes audit alerts as template


artherd42 (approved):
lgtm

PR #3241, opened 2022-03-16
limes: add pgmetrics for scrape failure

talal:


majewsky (changes requested):
I don't like these metric names a lot. We already have the `limes_failed_{,rate_}scrapes` counters, and these new gauges are called `limes_failed_project_{,rate_}scrapes_gauge`, which does not really clarify the distinction. I'll counter-propose `limes_failing_{,rate_}scrapes_gauge`, with the main distinction being the word "failed" for the counters (since they count events that have happened) vs. "failing" for the gauges (since they count projects that are currently having problems).

Also, to avoid multiple alerts being fired about the same thing, please downgrade the existing failed-scrapes alerts to severity `info`. That way, they are going to get inhibited by the new alert since the new alert will have the same context label, but a higher severity. We can still keep the old alerts as a backup in case the new alerts fail to fire.

majewsky (approved):

PR #3253, opened 2022-03-22
[prometheus-openstack] update prometheus-server; add api-only ingress

auhlig:
This PR updates the openstack prometheus in preparation for the mandated OAUTH protection of the UI.
A second, internal ingress is added limited to the federation API endpoint.
Once this is rolled out, Prometheis further up in the federation chain must switch to the API ingress.
Please check this PR and deploy


auhlig:
+ Switch to API-only URL for Alertmanagers

PR #3257, opened 2022-03-22
[octavia] Add MySQL InnoDB configuration for timeout

laerlingSAP:
Similar to [Neutron](https://github.com/sapcc/helm-charts/blob/d8a505f73d9ce00849731302adacbdba5ab3ed10/openstack/neutron/values.yaml#L308-L311)

### DIFF
--- a/openstack/octavia/values.yaml
+++ b/openstack/octavia/values.yaml
@@ -38,6 +38,11 @@ mariadb:
       - octavia.session_persistence
     oauth:
       client_id: octavia
+  extraConfigFiles:
+    octavia.cnf: |+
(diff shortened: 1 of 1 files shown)


galkindmitrii:
Please test well that this change won't make things worse. For Neutron I came to those values based on multiple experiments and the fact that innodb would be aggressively rolling back transactions.

laerlingSAP:
The problem this PR was supposed to alleviate seems to have been of other causes, and thus already alleviated by other changes.
Closing this PR, as it is better to not change such fundamental values if not necessary.

PR #3258, opened 2022-03-23
Rabbitmq: Option for statefulSet without clustering

fwiesel:
By default, this change doesn't change anything, except it
will raise an error when a deployment with more than one
replica with persistence enabled is requested, as this doesn't work.

One has to use a statefulset for that, but it is *not* a clustered
setup. The use-case is a special one for notifications:

We currently create for each service a single independent
rabbitmq, and the consumer connects to each of them.

If one goes down, no notifications can be send (over rabbitmq).

This allows us to spin up a statefulset with several independent
on the consumer side, let the producers connect to any of them,
and the consumer to all of them.

When one instance is down, the producer(s) can reconnect to
any of the remaining ones.
A Pod disruption budget ensures that at least one is up.

Clustering is not needed, as we do not need a consistent
state within rabbitmq, as we didn't have that before either.
We only need to ensure, that all messages put on the queues
will be delivered to the consumer.

### DIFF
--- a/common/rabbitmq/templates/headless-service.yaml
+++ b/common/rabbitmq/templates/headless-service.yaml
@@ -1,0 +1,19 @@
+{{- if eq .Values.kind "statefulset" }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "fullname" . }}-pods
+  labels:
+    app: {{ template "fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+    type: rabbitmq
+    component: {{ .Release.Name }}
+    system: openstack
+spec:
+  selector:
+    app: {{ template "fullname" . }}
+  publishNotReadyAddresses: true  # See https://www.rabbitmq.com/cluster-formation.html#peer-discovery-k8s
+  clusterIP: None
+{{- end }}
--- a/common/rabbitmq/Chart.yaml
+++ b/common/rabbitmq/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rabbitmq
-version: 0.2.20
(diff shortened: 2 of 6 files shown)


common/rabbitmq/templates/headless-service.yaml
```diff
@@ -0,0 +1,19 @@
...
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+    type: rabbitmq
+    component: {{ .Release.Name }}
+    system: openstack
+spec:
+  selector:
+    app: {{ template "fullname" . }}
+  publishNotReadyAddresses: true  # See https://www.rabbitmq.com/cluster-formation.html#peer-discovery-k8s
```
Carthaca, line 17: Where do you enable the `rabbitmq_peer_discovery_k8s` plugin or why would we need this besides that?

fwiesel: I don't enable it, as it isn't supposed to be clustered setup. I think that is likely better solved with the rabbitmq operator.
It the proposal for a small iterative step consolidating the indepenedent set of rabbitmqs used for notifications in a statefulset with poddisruptionbudget.

I just set the head-less service up to make the dns records available like in the clustered setup. So that a consumer can access all the individual pods via dns and get the messages from there.
Getting the messages from the "normal" service wouldn't work as we would connect to a random host, having only a subset of the messages.

Carthaca: My question aimed not at the headless service in general, but rather to that specific `publishNotReadyAddresses` option.
Why are you interested to be able to connect to pods that are not ready - is this something the consumers should be able to do on graceful shutdown or something?

Carthaca (approved):

notque (approved):
LGTM, when it's merged happy to start trying it out.

PR #3261, opened 2022-03-24
[prometheus-vmware] initial charts created

richardtief:


system/prometheus-vmware/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
auhlig, line 1: Can you make this a `v2` Helm Chart?

system/prometheus-vmware/requirements.yaml
```diff
@@ -0,0 +1,12 @@
+dependencies:
```
auhlig, line 1: The dependencies block can go into the Chart.yaml as Helm does not use `requirements.yaml` and `requirements.lock` anymore. ([Example](https://github.com/sapcc/helm-charts/blob/master/system/kube-monitoring-metal/Chart.yaml#L7))

system/prometheus-vmware/templates/_prometheus.alerts.tpl
```diff
@@ -0,0 +1,181 @@
+groups:
```
auhlig, line 1: You'll get these alerts for free as they are already part of the [common/prometheus-server Helm Chart](https://github.com/sapcc/helm-charts/blob/master/common/prometheus-server/templates/alerts/_prometheus.alerts.tpl).

system/prometheus-vmware/values.yaml
```diff
@@ -0,0 +1,89 @@
...
+additionalScrapeConfigs: {}
+
+scrapeInterval: "60s"
+
+configMaps:
+  - atlas-netbox-sd
+
+alertmanagers:
+  hosts:
+    - alertmanager.scaleout.eu-de-1.cloud.sap
```
auhlig, line 21: Can you use `alertmanager-internal.scaleout.eu-de-1.cloud.sap` and `alertmanager-internal.scaleout.eu-nl-1.cloud.sap` as we want to enable 2FA for the other endpoints.

system/prometheus-vmware/templates/_prometheus-vmware-scrape-config.yaml.tpl
```diff
@@ -0,0 +1,368 @@
+# Scrape config for service endpoints.
```
auhlig, line 1: Ideally, the below config can be reduced to a minimum using [ServiceMonitors](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#servicemonitor) and [PodMonitors](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#podmonitor).
This allows decoupling the Prometheus server from the scrape configuration: Each scrape target, e.g. exporter can define a ServiceMonitor/PodMonitor with its specific configuration eliminating the need to adjust any central configuration and re-deploy the Prometheus itself.
Most of the targets should be scraped via the generic endpoint, pod SD.
You get them for free as well for any service/pod labeled `prometheus.io/scrape: "true"` and  `prometheus.io/targets: "vmware"` if you do in the values.yaml:
```
prometheus-server:
  serviceDiscoveries:
    endpoints:
      enabled: true
    pods:
      enabled: true
```

system/prometheus-vmware/templates/_prometheus-vmware-scrape-config.yaml.tpl
```diff
@@ -0,0 +1,368 @@
...
+      target_label:       __metrics_path__
+      replacement:        /api/v1/nodes/${1}:4194/proxy/metrics
+
+# Scrape config for exporters
+#
+# * `prometheus.io/scrape`: Only scrape pods that have a value of `true`
+# * `prometheus.io/path`: If the metrics path is not `/metrics` override this.
+# * `prometheus.io/port_1`: Scrape the pod on the indicated port instead of the default of `9102`.
+
+{{- $values := .Values.vrops_exporter -}}
```
auhlig, line 342: .. except this. Here you'd need a separate Service-/PodMonitor.

PR #3272, opened 2022-03-25
MariaDB Log level configuration

dhoeller:


galkindmitrii (approved):
just please keep in mind that kubernetes will kill mariadb pod if this log file gets too big.
that means that mariadb will do a crash recovery on restart which might take some time...

PR #3274, opened 2022-03-28
Rabbitmq Online Credential Update

fwiesel:
The original attempt was first starting rabbitmq, then configuring it, shutting it down and start it again. A change of the script was triggering a re-deployment.

That had a couple of downsides:
- Longer startup times: We had to start up rabbitmq twice.
- Incompatible with later rabbitmq versions: The init script used has been removed
- Restart on credential change: This makes credential changes a bit cumbersome

Now the configuration code is executed on upgrade whenever the job code changes.


notandy (changes requested):
Please check the staging/rabbitmq implementation - it directly injects the credentials without the need to run either - a side-container or restarting the rabbitmq.

fwiesel:
Could you elaborate how that supposed to happen, I cannot get it from the code.

As I read it, it still restarts the container thanks to the [annotation](https://github.com/sapcc/helm-charts/blob/0abf92a6dd85174212c99037cdf3e36e3863f7b2/common/staging/rabbitmq/templates/statefulset.yaml#L26).
And as I read the documentation, an import after boot requires a [command](https://www.rabbitmq.com/definitions.html#import-after-boot) to be executed.

notandy:
> Could you elaborate how that supposed to happen, I cannot get it from the code.
> 
> As I read it, it still restarts the container thanks to the [annotation](https://github.com/sapcc/helm-charts/blob/0abf92a6dd85174212c99037cdf3e36e3863f7b2/common/staging/rabbitmq/templates/statefulset.yaml#L26). And as I read the documentation, an import after boot requires a [command](https://www.rabbitmq.com/definitions.html#import-after-boot) to be executed.

you've linked the answer:
[Definition Import at Node Boot Time](https://www.rabbitmq.com/definitions.html#import-on-boot)

notandy (changes requested):
> Could you elaborate how that supposed to happen, I cannot get it from the code.
> 
> As I read it, it still restarts the container thanks to the [annotation](https://github.com/sapcc/helm-charts/blob/0abf92a6dd85174212c99037cdf3e36e3863f7b2/common/staging/rabbitmq/templates/statefulset.yaml#L26). And as I read the documentation, an import after boot requires a [command](https://www.rabbitmq.com/definitions.html#import-after-boot) to be executed.

you've linked the answer:
[Definition Import at Node Boot Time](https://www.rabbitmq.com/definitions.html#import-on-boot)

fwiesel:
"Boot time" means for me restart. That seems to be consistent with the chart, which does restart the pod on a configuration change, which triggers the import of the `policy.json` on import. But if I may point out the last paragraph of the section you have linked:

> The definitions in the file will not overwrite anything already in the broker. However, if a blank (uninitialised) node imports a definition file, it will not create the default virtual host and user.

I don't think, that is solving the same problem I am trying to. Unless you are saying the documentation is mistaken.
And I am still missing how that is happening during runtime without a restart

notandy:
Yes it doesn't have to create the default virtual host and user, since the definition includes specific host and user.

It solves all your "downsides" except the online credentials changes. True, that doesn't work with the definition-on-startup solution, but in a cluster approach this is also not needed.

Your solution requires a python3 installation in the same image as rabbitmq, a side car container, ships python code in an config map and imposes a much more complexity. Doesn't pay imho.

fwiesel:
> It solves all your "downsides" except the online credentials changes. True, that doesn't work with the definition-on-startup solution, but in a cluster approach this is also not needed.

True, it addresses the first two items, but as the title of the PR might suggest, the main point was credentials change. Something we likely have to address (as an upgrade to a later version) rather sooner then later. I would guess in a couple of weeks.

It is intentionally not an attempt at a cluster solution, but an intermediary step before we get a "proper" clustered solution, which my understanding is still months down the line. 

> Your solution requires a python3 installation in the same image as rabbitmq, 

Which is part of the images, hence the language choice of implementation. Some time in the future that may not be the case anymore, but until then we have a clustered setup.
We could also create a go-binary in a different image/container, but I do not see there a significant advantage, except for decoupling the images, which in my view is a non-issue for the targeted life-time. It would also mean having to figure out to interface with rabbitmq in a fashion as `rabbitmqctl` so we can do that without credentials (see next point, why)

> a side car container, 

To my knowledge, there are only two options for changing credentials without risking locking yourself out (which I rather want to avoid).
As side-car container or the same container. Best practices suggest going with a side-car container. I can also put it in the same one, I wouldn't say that is an improvement.

> ships python code in an config map

It ships the same way we ship our bash scripts... We could build that also in the images, but to reiterate, I think that doesn't make sense for an intermediary solution. 

>  and imposes a much more complexity.

I think the ~120 lines of python script on top of ~40 lines of bash is hardly a different degree of complexity.
Especially if the alternative is a clustered setup. I do believe a clustered setup is the right way, but that requires more learning on our side.
I would expect that 120 lines of python script to be understandable by most people in a couple of minutes. I do not have the same expectation with a clustered setup.

notandy:
> True, it addresses the first two items, but as the title of the PR might suggest, the main point was credentials change. Something we likely have to address (as well as an upgrade to a later version) rather sooner then later. I would guess in a couple of weeks.

In general, I'd like to see online pw change mechanism (which could also done by a side-container), but the arguments are not valid.
This PR requires a restart of the rabbitmq anyway and I expect so many upcoming changes that the rabbitmq restarts saved by this pr as neglectable. 

> Which is part of the images, hence the language choice of implementation. Some time in the future that may not be the case anymore, but until then we have a clustered setup. We could also create a go-binary in a different image/container, but I do not see there a significant advantage, except for decoupling the images, which in my view is a non-issue for the targeted life-time. It would also mean having to figure out to interface with rabbitmq in a fashion as `rabbitmqctl` so we can do that without credentials (see next point, why)

In case of a side container, I rather liked to see an immutable image with some generic software than monitors and injects new credentials than following the anti-pattern of injecting code.

> It ships the same way we ship our bash scripts... We could build that also in the images, but to reiterate, I think that doesn't make sense for an intermediary solution.

This is no argument to continue with this anti-pattern - nothing is more permanent than an intermediary solution.

> I think the ~120 lines of python script on top of ~40 lines of bash is hardly a different degree of complexity. Especially if the alternative is a clustered setup. I do believe a clustered setup is the right way, but that requires more learning on our side. I would expect that 120 lines of python script to be understandable by most people in a couple of minutes. I do not have the same expectation with a clustered setup.

I agree that we disagree.

galkindmitrii (commented):
Just my two cents:

- there is an official rabbitmq operator:
https://www.rabbitmq.com/kubernetes/operator/operator-overview.html

- and a part of it is a topology operator: https://www.rabbitmq.com/kubernetes/operator/using-topology-operator.html

Together they allow to do pretty much anything with RabbitMQ.
Create clusters, create vhosts, users, define permissions, etc.
All in a very straightforward, kubernetes-way.

fwiesel:
I am all for the rabbitmq-operator, I just didn't want to go that way, because it is (in my eyes) a proper solution, I thought that how that should look like is left to whoever will get the mandate and time to do it properly, while evaluating all the options. 
As said before, this is a sub-optimal solution by choice, to replaced by something better.

If we have consensus on using the rabbitmq-operator and someone is willing to take up the work for the necessary changes, I will happily close this PR.

fwiesel:
I've added those changes. But in general, I would say that this change doesn't need to include all improvements. We can add more of them later.
I think, a change should just make things better than before, and I hope that applies here.

PR #3283, opened 2022-03-29
[nova] Use Placement filter instead of Availability Zone Filter

fwiesel:
Please have a look at the individual commits for more details.

### DIFF
--- a/openstack/nova/templates/etc-configmap.yaml
+++ b/openstack/nova/templates/etc-configmap.yaml
@@ -20,8 +20,6 @@ data:
 {{ include "util.helpers.valuesToIni" .Values.api_metadata.config_file | indent 4 }}
   nova-conductor.conf: |
 {{ include "util.helpers.valuesToIni" .Values.conductor.config_file | indent 4 }}
-  nova-scheduler.conf: |
-{{ include (print .Template.BasePath "/etc/_nova-scheduler.conf.tpl") . | indent 4 }}
   policy.json: |
 {{ include (print .Template.BasePath "/etc/_nova-policy.json.tpl") . | indent 4 }}
   logging.ini: |
--- a/openstack/nova/templates/etc/_nova-scheduler.conf.tpl
+++ b/openstack/nova/templates/etc/_nova-scheduler.conf.tpl
@@ -9,6 +9,7 @@ statsd_enabled = {{ .Values.scheduler.rpc_statsd_enabled }}
 discover_hosts_in_cells_interval = 60
 workers = {{ .Values.scheduler.workers | default 1 }}
 driver_task_period = {{ .Values.scheduler.driver_task_period | default 60 }}
+query_placement_for_availability_zone = True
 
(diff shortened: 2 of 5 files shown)


joker-at-work (commented):
I currently feel very uncomfortable with pushing more stuff into placement. I known it's the way to go forward, but there are two concerns I have:
1) every scheduling-decision we push from nova-scheduler into placement hides stuff for debugging. I myself sometimes forget, why I don't see much filtered hosts (because placement didn't return them) and having less visibility for filtering in the logs people find is causing more support effort on L4 instead of L2 level. Do you know of a way to get more insight into placements decision-process like we have in the nova-scheduler? Something like "found 3 hosts matching the requested $ram, $cpu, $disk, $custom_bla and $az" would help, if it comes back with the same `greq-`.
2) we've built tooling (e.g. Elektra, hammer at least) around the aggregates and not placement. not a big concern, just something we have to make sure checks placement for AZs instead of aggregates - they might get synced over, but I don't know the process there.

fwiesel:
1. Funnily enough, that is why I looked into it. I saw only one host coming back, and it was in the wrong availability-zone, which made me wonder, why on Earth is filtering out everything else *but* the one with the wrong availability zone. There is no logging output of the kind in placement, but the output of placement is input for the scheduler, so it is a bit redundant. The output of the scheduler can be improved, but personally, I think it is improved by removing hosts in the wrong availability-zone, as from my perspective, having multiple lines of `nova.scheduler.filters.availability_zone_filter` is not particularly insightful. We can add an log output (either placement, or nova) which gives such a summary. I rather see it in nova, because there is where most people already look.
2. The aggregates seem to be synced in multiple places. Once explicitly in our [job](https://github.com/sapcc/helm-charts/blob/master/openstack/nova/templates/bin/_db-online-migrate.tpl#L12), but then also whenever the traits are synced in `_refresh_associations`.
3.

joker-at-work:
I think our current setup for shard-aggregates will break this filtering. If I read [this](https://github.com/sapcc/nova/blob/2f7393ca03a1ee02284ea3932f1cf626a25fb6cd/nova/scheduler/request_filter.py#L83-L84) correctly, it will require all the aggregates that have the appropriate AZ set to match. This would be multiple shards in our case, which contradict each other. We need to change the way we create shard-aggregates and the current shard aggregates, before we can change away from the `AvailabilityZoneFilter`.

fwiesel:
That is not the case, the list of aggregates is [OR](https://github.com/sapcc/nova/blob/2f7393ca03a1ee02284ea3932f1cf626a25fb6cd/nova/objects/request_spec.py#L841-L844). 

I am a bit conflicted about how the availability zones are in the vcenter aggregate, as well in an availability-zone aggregate. But that is luckily a separate topic of this one.

joker-at-work (approved):
Ah, thanks for pointing that out. Then it should work also for ironic-deployments, that are not in a shard-aggregate.

PR #3291, opened 2022-03-31
VcsaSslCertificateExpiry closing small gaps

christopherhans:
We encountered a small gap in the vrops_vcenter_summary_certificate_expiry_date metric, which caused a false resolve message.


richardtief (approved):
Thanks.

PR #3312, opened 2022-04-01
[cinder]: migrate ingress to v1

databus23:
In preparation for the next upgrade of our kubernetes clusters we
need to remove references to api versions that are no longer supported.

The next version of kubernetes - 1.22 - removes support for the `Ingress` resource
in api version `networking.k8s.io/v1beta1`.

There are actually some minor structural changes to the `Ingress` object that
need to happen when migrating to `networking.k8s.io/v1`.

This blog post goes into some detail about what changes are required:
https://awstip.com/upgrading-kubernetes-ingresses-from-v1beta1-to-v1-7f9235765332

This PR should contain the necesarry changes for the chart in question.

Please review and merge at your own discretion. Happy hacking!

**Note: This PR was mass created, please validate and test the changes before rolling it to production.**

### DIFF
--- a/openstack/cinder/templates/api-ingress.yaml
+++ b/openstack/cinder/templates/api-ingress.yaml
@@ -1,4 +1,4 @@
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 
 metadata:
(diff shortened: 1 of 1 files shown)


hemna (commented):
I guess this looks ok.  I don't know how to test this or really have time at the moment.

databus23:
This is a rather small change. I suggest you merge it, deploy it to qa and if the cinder service is still reachable via its url all should be good.
We need to upgrade to 1.22 for compliance reasons before June and in the coming weeks we will reject deprecated api versions via gatekeeper. So getting this rolled out for your service is in your own interest to keep the cinder deployments working.

PR #3329, opened 2022-04-01
[nova]: migrate ingress to v1

databus23:
In preparation for the next upgrade of our kubernetes clusters we
need to remove references to api versions that are no longer supported.

The next version of kubernetes - 1.22 - removes support for the `Ingress` resource
in api version `networking.k8s.io/v1beta1`.

There are actually some minor structural changes to the `Ingress` object that
need to happen when migrating to `networking.k8s.io/v1`.

This blog post goes into some detail about what changes are required:
https://awstip.com/upgrading-kubernetes-ingresses-from-v1beta1-to-v1-7f9235765332

This PR should contain the necesarry changes for the chart in question.

Please review and merge at your own discretion. Happy hacking!

**Note: This PR was mass created, please validate and test the changes before rolling it to production.**

### DIFF
--- a/openstack/nova/templates/console-ingress.yaml
+++ b/openstack/nova/templates/console-ingress.yaml
@@ -1,5 +1,5 @@
 kind: Ingress
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
 
 metadata:
   name: nova-console
@@ -30,9 +30,12 @@ spec:
       {{- range $name, $config := .Values.consoles }}
         {{- if $config.enabled }}
       - path: /{{ $name }}/?(.*)
+        pathType: Prefix
         backend:
-          serviceName: nova-console-{{ $name }}
-          servicePort: {{ $name }}
+          service:
+            name: nova-console-{{ $name }}
+            port:
+              name: {{ $name }}
         {{- end }}
       {{- end }}
 
--- a/openstack/nova/templates/api-ingress.yaml
+++ b/openstack/nova/templates/api-ingress.yaml
@@ -1,5 +1,5 @@
 kind: Ingress
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
 
 metadata:
   name: nova-api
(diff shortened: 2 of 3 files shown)


openstack/nova/templates/console-ingress.yaml
```diff
@@ -30,9 +30,12 @@ spec:
       {{- range $name, $config := .Values.consoles }}
         {{- if $config.enabled }}
       - path: /{{ $name }}/?(.*)
+        pathType: Prefix
```
fwiesel, line 33: I think, that needs to be `ImplementationSpecific`, but quite honestly, I do not know why we even have the query parameters in the url-match

joker-at-work: You're misreading this. These are not query parameters but an optional `/`. We use this match in the annotations above to write a custom config into Nginx.

joker-at-work: IIRC, the console ingress was quite tricky in that we have multiple backends for it listening on different path-prefixes, but the backends themself don't understand these prefixes and need a rewrite to remove them. That's what the annotations do at least. But iirc, they stopped working with some nginx-ingress version and for that one, we might need the match here to get the actual path for the backends.

fwiesel: Ah, you are right. And looking at the [examples](https://kubernetes.github.io/ingress-nginx/examples/rewrite/), `pathType: Prefix` is correct, but we might want to change the annotation to simply:
`nginx.ingress.kubernetes.io/rewrite-target: /$1`

joker-at-work: We have to be careful in changing the syntax here. @auhlig and I had some back-and-forth a while ago about an upgrade plan for the nginx-ingress and Nova's console ingress. I'm not sure where we are right now and if this is compatible. https://github.com/sapcc/helm-charts/commit/1c3ccd789bb5d13852e42278dbbc337a44ec5205 has some details

joker-at-work: Prefix works in qa-de-2. Fabian, do you want to discuss that other part with Arno and me or should we just merge this and move the discussion about the annotations to later?

auhlig: Sorry. Missed the 1st mention. But +1 for moving ahead to qa and fixing (if needed) the annotation there

fwiesel: Let's move forward with this as it is. We can address the annotation in a different change (if needed)

PR #3357, opened 2022-04-01
[kube-metrics-exporter] configurable namespace

viennaa:
Namespace was always set to `any`. This leads to every prometheus pulling in the whole metrics of the cluster, even if they aren't needed. Instead of dropping it multiple times we should make this configurable here.

Still defaults to `any`, if nothing is set.

### DIFF
--- a/prometheus-exporters/kube-state-metrics-exporter/values.yaml
+++ b/prometheus-exporters/kube-state-metrics-exporter/values.yaml
@@ -6,6 +6,10 @@ global:
 # The name of the Prometheus instance which shall evaluate the alert and recording (aggregation) rules.
 prometheusName:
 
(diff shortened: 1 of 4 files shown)


prometheus-exporters/kube-state-metrics-exporter/values.yaml
```diff
@@ -6,6 +6,9 @@ global:
 # The name of the Prometheus instance which shall evaluate the alert and recording (aggregation) rules.
 prometheusName:
 
+# Select specific namespace, otherwise it will grab all namespaces
+namespaces:
```
auhlig, line 10:
```suggestion
namespaces: []
```

auhlig (commented):
Small nit for default value, but LGTM otherwise

PR #3366, opened 2022-04-04
[Nova] Shutdown delay to avoid connection resets

fwiesel:
change to the kube-proxy.

### DIFF
--- a/openstack/nova/templates/api-deployment.yaml
+++ b/openstack/nova/templates/api-deployment.yaml
@@ -70,6 +70,15 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+          lifecycle:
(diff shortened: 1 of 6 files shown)


joker-at-work (commented):
> at the same time as the as the endpoint-controller,

in the commit-message

Is there no better way than delaying the rollout of e.g. eu-de-2 for 60s? Can we maybe tell the serving process to drain and wait for that (which might wait longer, but will by that at least answer all connections)?

joker-at-work (approved):
Except for the commit-message, I'm fine with this. I missed the point in my first review.

fwiesel:
Will open another PR taking https://github.com/sapcc/helm-charts/pull/3374#discussion_r841559234 into account

PR #3367, opened 2022-04-04
[Barbican] Shutdown delay to avoid connection resets

fwiesel:
change to the kube-proxy.


rajivmucheli (approved):
lgtm

PR #3371, opened 2022-04-04
[Ironic] Shutdown delay to avoid connection resets

fwiesel:
change to the kube-proxy.

### DIFF
--- a/openstack/ironic/Chart.lock
+++ b/openstack/ironic/Chart.lock
@@ -17,5 +17,5 @@ dependencies:
 - name: utils
   repository: https://charts.eu-de-2.cloud.sap
   version: 0.7.0
-digest: sha256:08af657cd9858e299fa93cd055829c9ed964717104bbe4f34e219b8c82a0eedd
-generated: "2022-12-22T09:25:37.758145-05:00"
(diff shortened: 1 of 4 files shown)


fwiesel:
Will open another PR taking https://github.com/sapcc/helm-charts/pull/3374#discussion_r841559234 into account

fwiesel:
Done, now it should work.

stefanhipfel:
Do we also take the other delays into consideration? eg: deregistrationDelay(LB) and terminationGracePeriod (POD)?

fwiesel:
This delay is purely for the nginx, which speaks directly with the pods, so in a way the `deregistrationDelay`. 

The `terminationGracePeriod`  defines how long kubernetes waits for for sending the KILL signal after sending the TERM signal, so that is already after the pod is terminating.
But it's true, we potentially need to extend `terminationGracePeriod` , if the delay of the termination becomes too long. I think the default for `terminationGracePeriod` is 60 seconds, while the shutdown delay is around [10 seconds by default](https://github.com/sapcc/helm-charts/blob/master/openstack/utils/templates/snippets/_pre_stop_graceful_shutdown.tpl#L7), so unless we do something extraordinary here, we should be fine with the defaults.

stefanhipfel (approved):

PR #3374, opened 2022-04-04
[Neutron] Shutdown delay to avoid connection resets

fwiesel:
change to the kube-proxy.

### DIFF
--- a/openstack/neutron/values.yaml
+++ b/openstack/neutron/values.yaml
@@ -138,6 +138,7 @@ api:
   # minimum number of workers to keep at all times
   cheaper: 0
   uwsgi: false
+  shutdownDelaySeconds: 10
 
 service_plugins: asr1k_l3_routing
 default_router_type: ASR1k_router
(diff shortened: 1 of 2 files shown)


openstack/neutron/values.yaml
```diff
@@ -138,6 +138,7 @@ api:
   # minimum number of workers to keep at all times
   cheaper: 0
   uwsgi: false
+  shutdownDelaySeconds: 10
```
notandy, line 141: Hi, 
could we reduce we delay to 5 seconds which should be more than enough too process all existing connections, since neutron-api via uWSGI rarely needs more than 200ms to [process a connection](https://grafana.qa-de-1.cloud.sap/d/uwsgi/uwsgi?orgId=1&refresh=5s&from=now-6h&to=now)?

Also, neutron doesn't maintain long-lived connection and closes every connection after processing it, so no need for waiting longer.

notandy (changes requested):

openstack/neutron/values.yaml
notandy: Thanks for clarification,

if this value is depending on the kubernetes cluster and not on the service, can we have this value either defined in the global secrets or maybe even better to define it as a helm function so it can globally changed? I'd like to unclutter the secrets.
Thanks

fwiesel:
I will open another PR

PR #3382, opened 2022-04-05
[Nova] Use common snippet with shutdown delay to avoid connection refused

fwiesel:
…lures


fwiesel:
That requires still an updated dependency (#3387) as the default isn't set reliably

fwiesel:
Done, ready for review/merging

joker-at-work (approved):
I'm fine with it, if this renders correctly (I haven't checked).

PR #3384, opened 2022-04-05
[Barbican] Use shutdown delay snippet to avoid connection failures

fwiesel:


fwiesel:
That requires still an updated dependency (#3387) as the default isn't set reliably

fwiesel:
Done, now it should work

rajivmucheli (approved):

PR #3385, opened 2022-04-05
[Cinder] Use shutdown delay snippet to avoid connection failures

fwiesel:


hemna (approved):
LGTM

PR #3389, opened 2022-04-05
[kube-parrot] support global region

defo89:
if globalRegion is enabled, creates Daemonset per region with specific BGP AS values


SchwarzM (approved):
lgtm

PR #3390, opened 2022-04-05
dns lookup on sap name servers

artherd42:
related to https://github.wdf.sap.corp/cc/cc3test/issues/111


developer-abhi:
lgtm

galkindmitrii:
just my IMO - info level alert won't ever be looked at :/

artherd42:
as long as no team is assigned to the alert, info level is correct - https://operations.global.cloud.sap/docs/devops/alert/alert_cookbook/

PR #3394, opened 2022-04-06
[Keystone] Shutdown delay to avoid connection failures

fwiesel:
change to the kube-proxy.


bbobrov:
thanks a lot for the change. Does it make sense to use the same way that was used in #3386 (with pre_stop_graceful_shutdown)?

fwiesel:
The problem is, that keystone has its own `preShutDown` script, that needs to be executed, particularly the command `apachectl -k graceful-stop`, the snippet doesn't do that.

PR #3395, opened 2022-04-06
Migrate all PX k8s manifests to helm

swagner-de:
PX deployments, configmaps, nads, etc. were rendered by handcrafted
python scripts, in some cases k8s manifests were written by hand. In
order to abandon that script infrastructure, we move everything over to
helm and secrets.

Previous rendered manifests where these templates are based on can be
found here:
https://github.wdf.sap.corp/cc/px-cfg/tree/4da52beca51319b4ef2959751a72d881f3dd9388/kubernetes

Old pythonic templates are to be found here:
https://github.wdf.sap.corp/cc/px-cfg/tree/4da52beca51319b4ef2959751a72d881f3dd9388/kubernetes/template

### DIFF
--- a/px/bird/Chart.yaml
+++ b/px/bird/Chart.yaml
@@ -1,0 +1,4 @@
+apiVersion: v2
+description: A Helm chart for a PX Bird deployment
+name: px-bird
+version: "0.1"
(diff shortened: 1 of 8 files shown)


rhalina (approved):
Looks fine to me.

PR #3396, opened 2022-04-06
Revert "avoid de-scheduling of databases"

fwiesel:
An annotation change means a restart, which is currently not
well handled.

This reverts commit 6dbbc0d7c2accad0cb158de1c49a31566eadda68.


auhlig:
Annotation changes don't require a restart AFAK

PR #3397, opened 2022-04-06
[nova] Optionally enable RPC metrics collection

joker-at-work:
We've added a patch to oslo.messaging, that allows us to track the
number of RPC calls per called function and also if this call errored
out. This functionality can be enabled by setting statsd_enabled in the
DEFAULT section.

This change adds the possibility to enable this functionality for
nova-scheduler, nova-conductor and nova-compute pods (both ironic and
vmware).


joker-at-work:
I have not applied this to a cluster, but only verified that it "looks right" in the output of `helm template`.

fwiesel (approved):

PR #3398, opened 2022-04-06
[infra-monitoring] cross scrape prometheus-vmware

richardtief:


system/infra-monitoring/templates/_prometheus-infra-collector.yaml.tpl
```diff
@@ -809,3 +809,18 @@
     - action: labeldrop
       regex: "metrics_label"
 {{ end }}
+
+- job_name: 'prometheus-vmware'
+  scheme: https
```
viennaa, line 814: Sure `https` works without the tls-config? Even if it is internal, this is TLS.

richardtief: You are right. This must be http to resolve it internally.

viennaa (approved):

PR #3404, opened 2022-04-07
[common/rabbitmq] add clustered functionality to statefulset rabbitmq

notandy:
Automatically uses statefulset if replicas > 1 cluster.enabled = true
   enables zero-conf rabbitmq clustering with autofederation

   Uses RabbitMQ 3.9 when statefulset is used.

   Prometheus-metrics are currently done by the old plugin, but the new
   inbuilt prometheus-exporter is enabled by default if using statefulset

   Don't changes anything when used with classic deployment-based rabbitmq.

   Statefulset got rid of nasty bash scripts, thus no restarts required
   -> faster startup (about 2 seconds)


common/rabbitmq/templates/etc/_definitions.json.tpl
```diff
@@ -13,14 +13,28 @@
...
 {{ toJson $users | indent 8 }},
 {{- end }}
 {{- if .Values.metrics.enabled }}
-        {"name":"{{.Values.metrics.user}}","password":"{{tuple . .Values.metrics.user .Values.metrics.password | include "rabbitmq.pass"}}","tags":"monitoring"},
+        {"name":"{{.Values.metrics.user}}","password":"{{tuple . .Values.metrics.user .Values.metrics.password | include "rabbitmq_pass"}}","tags":"monitoring"},
 {{- end }}
-        {"name":"guest","password":"{{ tuple . .Values.users.default.user .Values.users.default.password | include "rabbitmq.pass" }}","tags":"monitoring"}
+        {"name":"guest","password":"{{ tuple . .Values.users.default.user .Values.users.default.password | include "rabbitmq_pass" }}","tags":"monitoring"}
     ],
+{{- if .Values.cluster.ha }}
```
fwiesel, line 24: I would prefer if we can set the policy via the values more fine grained. In particular, for RPC it might sense to set a TTL on the messages.

common/rabbitmq/values.yaml
```diff
@@ -82,12 +86,28 @@ alerts:
...
 
-kind: deployment  # needs to be statefulset with persistence.enabled and ...
-replicas: 1       # replicas more than one. Clustering is currently not implemented
+# replicas. Uses statefulset for multiple replicas.
+# Cluster formation is enabled with cluster.enabled = true.
+replicas: 1
+cluster:
+  enabled: false
+  ha: true # Enables HA
+  erlangCookie: "DefaultErlangSecretCookie"
```
fwiesel, line 95: I would prefer an empty default, and failing installation than quietly using a insecure default.

fwiesel (commented):
I've added my comments, but I think it is not up for me to approve it. @notque or @galkindmitrii are probably more fitting, since they continue the work.

notque:
Updating Rabbitmq is good, it needs tested but seems reasonable.
Clustering adds additional complexity without a benefit I can see from the setup we are using at the moment. That may change in the future, but I don't see implementing that in the near term.

You're using replicas > 1 to mean using clustering, when that should be a clustering value. For example, if I don't want to move to cluster, but do want 3 replicas, there are unintended consequences like "clusterIP: None"

It also changes values around in their meaning, and breaks what I'm currently running, and moving into production.

Adding the clustering seems like a great idea to me, but it needs to be if you turn on clustering, not just if you're using more than 1 replica.

PR #3410, opened 2022-04-08
Multiple fixes for tempest:

misamoylov:
* Fix cleanup for keystone
* Fix cleanup for cinder
* Add api version for charts
* Fix cleanup for nova
* Add primary roles


openstack/tempest/tempest-base/templates/etc/_tempest_extra_options.tpl
```diff
@@ -76,15 +76,16 @@ max_microversion = latest
...
+unified_limits = False
+cold_migration = False
+live_migration = False
+live_migrate_back_and_forth = False
+vnc_console = True
 vnc_server_header = WebSockify
-serial_console = true
-spice_console = true
-attach_encrypted_volume = false
+serial_console = True
```
galkindmitrii, line 86: I'm not sure about the consoles, I think neither of those is enabled.

galkindmitrii (commented):
nice! as discussed, not sure of api version is needed when interacting with cinder api.

PR #3421, opened 2022-04-14
[manila] disable nfs v4.0, enable nfs v4.1

Carthaca:
requires https://github.com/sapcc/manila/pull/31 to not disable v4.0
on existing share servers

SVM migrate is only working for v3 and v4.1, so we discourage the future use
of v4.0 to be able to move data


Carthaca:
not gonna work, too much existing customers on 4.0 to take that option away

PR #3427, opened 2022-04-19
Add looking glas charts

rhalina:
Move bird looking glas deployment to helm


px/lg/templates/configmap-communities.yaml
```diff
@@ -0,0 +1,10 @@
+{{- $config_path := printf "%s%s.json" $.Values.lg_config_path "lg_px_communities_v2" -}}
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+    name: cfg-{{ .Values.region }}-pxrs-lg-communities
+data:
+#    "{{ $.Values.global.region }}-pxrs-lg-communities.cfg": |
```
swagner-de, line 8: Can we remove the comment please?

px/lg/templates/configmap-communities.yaml
```diff
@@ -0,0 +1,10 @@
...
+    "lg_px_communities_v2.json": |
+{{ $.Files.Get $config_path | indent 8}}
```
swagner-de, line 10: It would be a good safe-guard to have a check that the file exists and not have helm secretly put an empty string there. Check the example here:
https://github.com/sapcc/helm-charts/blob/master/px/bird/templates/_configmap-bird.tpl#L14-L15

px/lg/values.yaml
```diff
@@ -0,0 +1,20 @@
...
+    proxy_port:
+    authenticate: false
+    subdomain: px
+  lgadmin:
+    authenticate: true
+    proxy_port:
+    subdomain: pxadmin
+
+probe_interval_ms: 5000
+probe_timeout_ms: 3000
```
swagner-de, line 20: Those do not belong in this chart.

px/lg/values.yaml
```diff
@@ -0,0 +1,20 @@
+region: __DEFINED_AS_ARG__
+apod:
```
swagner-de, line 2: The lg does not need an apod indicator, it can be placed anywhere in k8s.

px/lg/values.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+looking_glass:
+  lg:
+    proxy_port:
+    authenticate: false
+    subdomain: px
+  lgadmin:
+    authenticate: true
+    proxy_port:
+    subdomain: pxadmin
```
swagner-de, line 17: Might be a good idea to have these default in line with the defaults of the bird deployment:
https://github.com/sapcc/helm-charts/blob/master/px/bird/values.yaml#L25-L33

swagner-de (changes requested):

swagner-de (approved):
lgtm

PR #3432, opened 2022-04-20
alertmanager: deliver alerts to Slack channel #cc-k8s-gatekeeper

majewsky:
The new `slack_by_k8s_service` is a direct copy of `slack_by_os_service`. I want to split out Gatekeeper alerts that currently only end up in #cc-k8s-info since that channel is has so much noise in it that it's completely useless for me.

### DIFF
--- a/global/prometheus-alertmanager-operated/Chart.yaml
+++ b/global/prometheus-alertmanager-operated/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 type: application
 description: Prometheus Alertmanager via operator.
 name: prometheus-alertmanager-operated
-version: 2.0.2
+version: 2.0.3
 
 dependencies:
   - alias: prometheus-alertmanager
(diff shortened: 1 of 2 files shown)


geisslet:
If your intention is to get it out of cc-k8s you need to move your route before `- receiver: slack_k8s_*`
And you need to set `continue: false` - otherwise it ends up in `#cc-k8s-*`, too.

**Notes:**
- you need to increase the helm version, too
- @ auhlig will do the merge and rollout afterwards


Bonus comment which is just for notice: we use a deprecated generic slack token approach to have that kind of convenient service channel templating - maybe we need to adopt in the future.

majewsky:
I don't care if it also gets delivered into cc-k8s-info, so I'll leave it as-is to keep it consistent with `slack_by_os_service`.

> we use a deprecated generic slack token approach to have that kind of convenient service channel templating - maybe we need to adopt in the future.

Why is this approach deprecated? Is Slack switching to channel-local tokens?

auhlig (approved):

PR #3440, opened 2022-04-25
gatekeeper: remove GkImagesFromDockerhub

talal:


majewsky:
If we will end up deleting this policy entirely, the `if false` step is not going to be necessary, no? As far as I remember, we needed this step for GkHighCpuRequests because we did a rename and did not want to delete the file entirely as a middle step.

talal:
I think we need this because if you look at the [GkHighCpuRequests PR](https://github.com/sapcc/helm-charts/pull/2862/files) that resulted in the issue, you can see that both the config and `ConstraintTemplate` files were renamed but we still ran into this issue.

I've created a PR for stage-2 too, so if you refer to [this commit](https://github.com/sapcc/helm-charts/pull/3443/commits/f89f4532f5a0af8b4aa6d94864ee0444a9c34ce8), the file renames look similar to the GkHighCpuRequests PR.

I wanted to do this in two stages just to be on the safe side.

majewsky:
I played this through in my head, and you're more right than I am. However, we must only delete the config object, not the template object. Since gatekeeper is deployed before gatekeeper-config, deleting the template in this step will cause problems because the existing config object contradicts the deletion of the template. But if we delete only the config object, deploy that entirely, then we can replace the template object with the renamed one and add the config object for the renamed template all within the second phase.

PR #3445, opened 2022-04-26
[nova] Use ingress rewrite-target

fwiesel:
Instead of using a configuration-snippet,
we can use the rewrite-target as in the
example in the documentation:
https://kubernetes.github.io/ingress-nginx/examples/rewrite/

We need to replace nginx.ingress.kubernetes.io
with ingress.kubernetes.io as the ingress is configured
to take the latter non-default annotation prefix

### DIFF
--- a/openstack/nova/templates/console-ingress.yaml
+++ b/openstack/nova/templates/console-ingress.yaml
@@ -9,13 +9,7 @@ metadata:
     component: nova
   annotations:
     ingress.kubernetes.io/use-regex: "true"
-    ingress.kubernetes.io/configuration-snippet: |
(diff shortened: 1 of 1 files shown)


fwiesel:
Continuation from  #3329. It works at least in qa-de-2, but I understood there might be incompatibilities with some nginx ingress controller versions.

auhlig:
Cool. Yes, the configuration snippet was required during migration and upgrade of the ingress controller - which is now done. The rewrite annotation is now the preferred way

auhlig (approved):

joker-at-work (approved):
I'm happy that we can close this topic now. It's been going on since 2 years :)

PR #3447, opened 2022-04-27
[ironic][flavors] correct the resources:vcpu

BerndKue:


sandzwerg (commented):
Correct according to the documentation: https://docs.openstack.org/ironic/latest/install/configure-nova-flavors.html
I assume that won't cause side effects for running instances

sandzwerg (approved):

PR #3449, opened 2022-04-27
[blackbox]: remove ingress spec from template

developer-abhi:
@databus23 : please approve this


thgrs (approved):
lgtm

databus23:
I think the spec can be fixed by changed `backend` to `defaultBackend`.

developer-abhi:
ok, Will change `spec.backend` to `spec.defaultBackend`

PR #3459, opened 2022-05-02
[sap-seeds] Optionally add legacy bigvm flavors

joker-at-work:
When deploying the sap-seeds with use_hana_exclusive set, we did not
include the legacy flavors that are not applicable to
CUSTOM_HANA_EXCLUSIVE_HOST-enabled hosts. But there's at least one
region where we now need to also include those flavors, because one of
the AZs will never get a CUSTOM_HANA_EXCLUSIVE_HOST-enabled host.

Therefore, we move the legacy flavors to their own template for
inclusion, include them unconditionally if the hana_no_exclusive list is
used and based on the setting
hana_exclusive_contains_legacy_bigvm_flavors also in the hana_exclusive
case.


joker-at-work:
hm ... have to rework that a little. we don't want the `CUSTOM_BIGVM` resource in the hana-exclusive case.

PR #3463, opened 2022-05-02
remove old apic exporter from helm chart

artherd42:


IvoGoman (changes requested):
There is another entry for `apic_exporters` in the values.yaml on line 495

IvoGoman (approved):

PR #3464, opened 2022-05-02
[designate] policy switch to yaml format as json is deprecated since

galkindmitrii:
Wallaby release.


kpawar-sap:
openstack/designate/bin/manage-pools missing reference of pools.yaml ?

PR #3466, opened 2022-05-02
[Nova] Set appVersion in chart

fwiesel:
Practically, we only support one version in the chart,
so we might as well be specific about it, which makes
the release tracking also more consistent with other
charts


joker-at-work (commented):
How does that help?

fwiesel:
It is the metadata field which defines the version of the software in the chart.
Other charts keep track of the releases that way, such as barbican, glance, neutron and manila.

Yes, there is an overlap with us tracking versions in the image tags, but we are only testing the chart with exactly one version, so we might as well be explicit about it.

That way, the question, which release we are running is then answered with `helm list`.

joker-at-work (approved):

PR #3468, opened 2022-05-02
[Nova] Bumps mariadb chart dependency

IvoGoman:
- Fixes OOMKill for backup pod during full restore
    Due to VPA Butler setting lower mem limits


joker-at-work (changes requested):
Nova chart has 3 references to the mariadb chart. Please update accordingly.

PR #3472, opened 2022-05-03
[openstack/utils] Change mysql db default settings

fwiesel:
Follow in the defaults the settings from ansible.
Creating connections to mariadb is not terribly costly,
but keeping a "large" pool of connections around
consumes a portion of max_connections of the server.

### DIFF
--- a/openstack/utils/Chart.yaml
+++ b/openstack/utils/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: utils
-version: 0.3.8
+version: 0.3.9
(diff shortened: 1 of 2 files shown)


fwiesel:
@viktor-kovacs I understand, those are the recommended settings (from ansible?), so I thought, it would make sense to take them as the default.

@galkindmitrii Do you have a different opinion?

galkindmitrii:
Sorry, missed the PR.
I'm against default _global.max_overflow | default 1000_ because this is per worker/process, some services have 16-20 worker processes, meaning if something goes wrong they will attempt to open 16-20 of connections. We've seen a few times maxed out max_overflow connections.

PR #3473, opened 2022-05-03
Do Not Merge YET [Cinder] Increase default max volume size to 10Tib

hemna:
This patch increases the default for the max volume size
to 10Tib.


dhoeller (approved):
LGTM

PR #3479, opened 2022-05-04
[keystone] Do not remove orphaned role assignments via cron

bbobrov:
Although removed role assignments sit in the database without use,
practice shows that the users often reappear, and it is not nice to
leave them without their prior assignments.


rajivmucheli (approved):
lgtm

bbobrov:
We decided not to proceed with it

PR #3482, opened 2022-05-04
GkVulnerableImages: use image ID

talal:
I've left the policy message to use `image` instead of `imageID`. Let me know if you want ID in the message.


majewsky (approved):
Sometimes changes are simpler than one would expect. 🙂

PR #3484, opened 2022-05-06
[Swift] enable ccloud conform swift roles in Yoga

reimannf:
Old roles remain valid for now


openstack/swift/templates/seeds.yaml
```diff
@@ -19,7 +20,12 @@ spec:
   roles:
   - name: swiftoperator
   - name: swiftreseller
-
+{{- if le $swift_release "yoga" }}
```
majewsky, line 23: Should this not be `ge` instead of `le`, as in: from Yoga onwards?

majewsky (approved):

PR #3487, opened 2022-05-09
[PX] Resolve most of DOOP warnings

swagner-de:


rhalina (approved):
The changes look sane to me. 

In case we make substantial changes to the service, i.e. connect members with a large number of diverse routes we may want to adjust the memory limits as BGP ADD-PATH significantly increases the number of routes kept in memory.

PR #3494, opened 2022-05-10
[nova] Use single statsd-exporter.yaml

joker-at-work:
We recently introduced an additional statsd-rpc-exporter.yaml, which did
not include the default settings, but only the mappings for
oslo.messaging.* metrics. When introducing the new config file, the
vmware hypervisor pods did not get an update. Additionally, we missed to
set the right file in the conductor pod's "args".

Since we want both the default settings and the mappings in all statsd
exporters we run, we now use a single file for both in all the pods.


joker-at-work:
> Metrics that don't match any mapping in the configuration file are translated into Prometheus metrics without any labels and with any non-alphanumeric characters, including periods, translated into underscores.

https://github.com/prometheus/statsd_exporter

stefanhipfel (approved):

PR #3500, opened 2022-05-12
[nannies][manila] add snapshot-missing nanny

chuan137:
Extra secrets for netapp api user/password. And extra configmap for
netapp filers list.


chuan137:
The nanny implementation is in https://github.com/sapcc/openstack-nannies/pull/56

PR #3505, opened 2022-05-13
[px/lg] Add kubernetes resource limits

rhalina:
To address DOOP warnings the looking glas deployments now have cpu and
memory resource limits

### DIFF
--- a/px/lg/templates/deployment.yaml
+++ b/px/lg/templates/deployment.yaml
@@ -33,6 +33,13 @@ spec:
       - name: {{ $.Values.global.region }}-pxrs-{{ $lg }}
         image: keppel.{{ $.Values.registry }}.cloud.sap/{{ $.Values.lg_image }}
         imagePullPolicy: Always
+        resources:
+          requests:
+            cpu: {{ $.Values.resources.lg.requests.cpu }}
+            memory: {{ $.Values.resources.lg.requests.memory }}
+          limits:
+            cpu: {{ $.Values.resources.lg.limits.cpu }}
+            memory: {{ $.Values.resources.lg.limits.memory }}
         volumeMounts:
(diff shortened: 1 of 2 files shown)


px/lg/templates/deployment.yaml
```diff
@@ -33,6 +33,13 @@ spec:
       - name: {{ $.Values.global.region }}-pxrs-{{ $lg }}
         image: keppel.{{ $.Values.registry }}.cloud.sap/{{ $.Values.lg_image }}
         imagePullPolicy: Always
+        resources:
+          requests:
+            cpu: {{ $.Values.resources.lg.requests.cpu }}
+            memory: {{ $.Values.resources.lg.requests.memory }}
+          limits:
+            cpu: {{ $.Values.resources.lg.limits.cpu }}
+            memory: {{ $.Values.resources.lg.limits.memory }}
```
swagner-de, line 42: If the structure of the values file matches the structure of the manifest, you can just include the values like here:
https://github.com/sapcc/helm-charts/blob/master/px/bird/templates/_deployment-bird.tpl#L63

No need to translate each individual attribute.

px/lg/values.yaml
```diff
@@ -24,3 +24,12 @@ looking_glass:
     proxy_port: 5005
     subdomain: pxadmin
     privileged: true
+
+resources:
+  lg:
+    requests:
+      cpu: "25m"
+      memory: "250Mi"
```
swagner-de, line 32: The VPA upperbound recommendation is actually more in that range. I would use that as a limit. `requests` is what is allocated on container start at the machine. I don't think we need to allocate that much.

swagner-de (changes requested):

PR #3510, opened 2022-05-18
bump prometheus-server in prometheus-openstack

artherd42:


openstack/prometheus-openstack/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: "1.0"
 description: A helm chart for OpenStack monitoring tools.
 name: openstack-prometheus
-version: 0.1.2
+version: 6.2.0
```
viennaa, line 5: This should rather be increased in small steps, as it is chart agnostic. Since we're bumping a major version, we can increment a little further.

```suggestion
version: 1.0.0
```

viennaa (changes requested):
Thanks for this PR. I would be a little bit more conservative with bumping the chart version of this chart in general.

viennaa (approved):

PR #3519, opened 2022-05-19
adding owner info and upgrading chart version

viennaa:
upgrading multiple charts to `v2` and adding owner dependency as requested by doop.

```Check: GkOwnerInfoOnHelmReleases```


richardtief (approved):
Thanks. Looks good.

PR #3521, opened 2022-05-19
 Bump utils version

bbobrov:
This is required because jaeger-agent was pointed to another image in the previous commit


bbobrov:
self-merging since this is a minor change

PR #3525, opened 2022-05-19
GkImagesFromNonKeppel: fix Rego

talal:


majewsky:
I agree with the reasoning in principle, but I would like to ask if this is based on a theoretical concern or on actual errors from gatekeeper-audit?

Here's my thinking on this: `podSpec.containers` can definitely not be null, that *surely* must be rejected outright by the k8s API. For `initContainers`, it is true that we have lots of pods without any of these. But if that caused `podSpec.initContainers` to be null, the evaluation errors would have caused gatekeeper-audit to choke, but we have not seen any alerts relating to that. That makes me think that the k8s API normalizes `initContainers = null` into `initContainers = []`.

I'm not opposed to the change in general, but if we put in extra complexity, I want to have it properly justified.

majewsky:
I have tested with Sandro's pipeline and my own objects. For `initContainers`, your assessment is correct and we do in fact need `object.get` there. For `containers`, I've removed the `object.get` because it is in fact impossible to set this to null or even empty array because the k8s API complains about it:

```
The Deployment "stefantest" is invalid: spec.template.spec.containers: Required value
```

Therefore `object.get` is superfluous there.

PR #3526, opened 2022-05-20
[alertmanager] pagerduty network split by alert type

geisslet:
there is a managed service split within tier net - this setup isn't final and the sequence hack is dirty, but necessary


geisslet:
vault and secret are adopted

databus23 (approved):

databus23 (changes requested):
The `admin` region does not exist anymore for a long time and should be removed to reduce clutter by a few percent.

databus23 (approved):

PR #3528, opened 2022-05-20
castellum, keppel, limes, tenso: set anti affinity for api deployment

SuperSandro2000:


majewsky (changes requested):
Do we really need three different affinities? The keystone-api only has one. Since you copied from swift-proxy, I'm guessing there's extra complexity in there because swift-proxy can be deployed both as a daemonset and as a deployment in parallel. (Also, there may be some superfluous rules in there.)

Please spend some more time understanding what we really need here.

openstack/castellum/templates/api-deployment.yaml
```diff
@@ -26,6 +26,26 @@ spec:
...
+                  values:
+                  - castellum
+                - key: application
+                  operator: In
+                  values:
+                  - neutron
+                - key: component
+                  operator: In
+                  values:
+                  - api
```
majewsky, line 46: Besides the copy-paste error, this does not match the label structure that we have in this deployment. This labelSelector should probably match the one used by the Deployment (see `.spec.selector.matchLabels` in this same file).

Analogously for the other deployments.

majewsky (approved):

PR #3531, opened 2022-05-23
[mariadb] Prevent trible scraping of metrics

stefanhipfel:
let prometheus find the container with the port "metrics" by itself and scrape the pod once instead 3 times (for each container)


fwiesel (approved):

galkindmitrii (approved):
thx!

PR #3533, opened 2022-05-23
Fix endpoints/service service discoery by port name

databus23:
turns out the ` __meta_kubernetes_service_port_name` is not available when using endpoints sd.
We need to use `__meta_kubernetes_endpoint_port_name` instead


geisslet (approved):
lgtm

artherd42:
@databus23 Please update versions in CHANGELOG.md with your recent changes.

PR #3534, opened 2022-05-23
[manila] prevent double scraping of api

Carthaca:
prom is looking at the port named `metrics` and gets confused by
specifying a port annotation if the service has more than 2 ports

needs a go from @databus23, needs an underlying prom 'chart rodeo'

### DIFF
--- a/openstack/manila/templates/api-service.yaml
+++ b/openstack/manila/templates/api-service.yaml
@@ -9,7 +9,6 @@ metadata:
     component: manila
   annotations:
     prometheus.io/scrape: "true"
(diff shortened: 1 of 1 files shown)


databus23:
For reference this PR needs to be merged: https://github.com/sapcc/helm-charts/pull/3533

Once its merged the umbrella-charts for prometheus-collector and prometheus-openstack need to be bumped and deployed to actually roll this out

databus23:
https://github.com/sapcc/helm-charts/pull/3533 is rolled to scaleout and metal control planes in all regions. So this can be merged from my pov.

PR #3542, opened 2022-05-24
cronus - adding new flags that enable new queued message wait and retry

dhalimi:


bassel-z:
Hi @dhalimi , you need to add the jobQueue, Waiting Queue config as well

bassel-z:
LGTM

PR #3545, opened 2022-05-24
[manila] adjust audit according to new share access api

Carthaca:


notque (approved):
Looks reasonable to me.

PR #3547, opened 2022-05-25
Prometheus-Openstack: Allow -* suffix for named metrics ports

fwiesel:
The name of a port needs to be unique in a pod,
but there might be more than container with metrics in a pod.
This extends the convention for ports named 'metrics' also
to ports with a suffix separated by hyphen ('metrics-foo'),
so one doesn't have to resort to a pod annotation.


fwiesel:
I am not sure about what I have done here, especially how the two jobs interact.
I fear now we would match the same port twice with the two jobs.

databus23:
I'm confused about the location of the change. I made changes to `common/prometheus-server` to change the service discovery which is also used by this chart:

https://github.com/sapcc/helm-charts/blob/master/common/prometheus-server/templates/podmonitors/pod-sd.yaml

fwiesel:
Can we then remove those configs? They seem to me then either obsolete or redundant.

PR #3555, opened 2022-05-26
[mariadb] Pre-upgrade hook for db/user creation

fwiesel:
The idea is to roll-out new users with new credentials in a pre-upgrade hook
so that they are available to the actual deployment making use of it.

Users (or databases) cannot be deleted this way (much less rolled back).
The cleanup is left to a later iteration, which may make use of an expiry policy.

Instead of having custom init scripts with slightly different patterns,
the fields `databases` and `users` are there to create and update them.

This also allows for uniform access to other macros, such as side-car
containers, which need knowledge about the users and the databases as well.

By using a job which executes the change within the container
of the database, we can change the credentials of any user
without fearing a lockout.

The job will be executed on upgrade and rollback and since it
gets the values from the previous deployment, it not only
will set the new credentials on upgrade, it also restores the old ones
on rollback.

### DIFF
--- a/common/mariadb/templates/pre-change-job.yaml
+++ b/common/mariadb/templates/pre-change-job.yaml
@@ -1,0 +1,41 @@
+{{- if or .Values.databases .Values.users }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ include "fullName" . }}-pre-change"
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    app.kubernetes.io/version: {{ default "unknown" .Chart.AppVersion }}
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+  annotations:
+    "helm.sh/hook": pre-upgrade,pre-rollback
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
+  template:
+    metadata:
+      name: "{{ include "fullName" . }}-pre-change"
+      labels:
+        app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
+        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ .Values.name }}-db-exec
+      containers:
+      - name: pre-change-job
+        image: "{{ required ".Values.global.dockerHubMirrorAlternateRegion is missing" .Values.global.dockerHubMirrorAlternateRegion }}/{{ .Values.pre_change_hook.image }}:{{ .Values.pre_change_hook.image_version }}"
+        command:
+        - kubectl-v{{ .Values.pre_change_hook.kubectl_version }}
+        - exec
+        - -c
+        - mariadb
+        - deploy/{{ include "fullName" . }}
+        - --
+        - mariadb
+        - -uroot
+        - --batch
+        - -e
+        - |
+          {{- include (print .Template.BasePath "/initdb/_init.sql.tpl") . | trim | nindent 10 }}
+{{- end }}
--- a/common/mariadb/values.yaml
+++ b/common/mariadb/values.yaml
@@ -36,6 +36,10 @@ users:
 #    grants:
 #    - ALL ON example.*
 
+pre_change_hook:
+  image: "sapcc/unified-kubernetes-toolbox"
+  image_version: "20220613150453"
+  kubectl_version: "1.23"
 
 # name of priorityClass to influence scheduling priority
 priority_class: "openstack-service-critical"
--- a/common/mariadb/Chart.yaml
+++ b/common/mariadb/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: mariadb
-version: 0.4.0
+version: 0.5.0
--- a/common/mariadb/templates/db-exec-rbac.yaml
+++ b/common/mariadb/templates/db-exec-rbac.yaml
@@ -1,0 +1,46 @@
+{{- if or .Values.databases .Values.users }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.name }}-db-exec
+  annotations:
+    "helm.sh/hook": pre-upgrade,pre-rollback
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    "helm.sh/hook-weight": "-50"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: monsoon3
+  name: {{ .Values.name }}-db-exec
+  annotations:
+    "helm.sh/hook": pre-upgrade,pre-rollback
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    "helm.sh/hook-weight": "-40"
+rules:
+- apiGroups: ["extensions", "apps"]
+  resources: ["deployments"]
+  verbs: ["get", "watch", "list"]
+- apiGroups: [""] # "" indicates the core API group
+  resources: ["pods", "services"]
+  verbs: ["get", "watch", "list"]
+- apiGroups: [""]
+  resources: ["pods/exec"]
+  verbs: ["create"]
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ .Values.name }}-db-exec
+  annotations:
+    "helm.sh/hook": pre-upgrade,pre-rollback
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    "helm.sh/hook-weight": "-30"
+subjects:
+- kind: ServiceAccount
+  name: {{ .Values.name }}-db-exec
+roleRef:
+  kind: Role
+  name: {{ .Values.name }}-db-exec
+  apiGroup: rbac.authorization.k8s.io
+{{- end }}


common/mariadb/templates/initdb/_init.sql.tpl
```diff
@@ -1,3 +1,25 @@
...
+CREATE DATABASE IF NOT EXISTS {{ .Values.name }};
+{{- else }}
+    {{- range .Values.databases }}
+CREATE DATABASE IF NOT EXISTS {{ . }};
+    {{- end }}
+{{- end }}
+
+{{- if and .Values.global.dbUser .Values.global.dbPassword }}
+CREATE USER IF NOT EXISTS {{ .Values.global.dbUser }};
+GRANT ALL PRIVILEGES ON {{ .Values.name }}.* TO {{ .Values.global.dbUser }} IDENTIFIED BY {{ include "db_password" . }};
```
galkindmitrii, line 11: Do I understand that correct that in case of multiple Values.databases we are not adding the permissions?

common/mariadb/values.yaml
```diff
@@ -18,10 +18,29 @@ join_buffer_size: "4M"
 binlog_format: "MIXED"
 expire_logs_days: 10
 
+
+pre_change_hook:
+  image: "sapcc/unified-kubernetes-toolbox"
+  image_version: "20220525082919"
```
galkindmitrii, line 24: an extra image dependency, I assume the job will be failing on k8s upgrade and we would need extra coordination for deployment of all openstack services...

common/mariadb/templates/pre-change-job.yaml
```diff
@@ -0,0 +1,41 @@
...
+        app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
+        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ .Values.name }}-db-exec
+      containers:
+      - name: pre-change-job
+        image: "{{ required ".Values.global.dockerHubMirrorAlternateRegion is missing" .Values.global.dockerHubMirrorAlternateRegion }}/{{ .Values.pre_change_hook.image }}:{{ .Values.pre_change_hook.image_version }}"
+        command:
```
galkindmitrii, line 28: Is there any chance we can do that easier?
e.g. by adding k8s lifecycle command with some failsafe sql? https://github.com/sapcc/helm-charts/blob/master/common/mariadb/templates/deployment.yaml#L55-L58

fwiesel:
I've added PR  #3595 with the reasoning behind the decisions on how it has been realized.

common/mariadb/templates/initdb/_init.sql.tpl
fwiesel: Yes, `.Values.databases` only creates the databases.
The users and their grants are created & maintained by `.Values.users` and `.Values.users[*].grants` respectively.

We can add some defaults, which avoids repetitive defaults, the issue though here is, that such a logic internal to this chart would be needed to replicated to anyone trying to evaluate the settings (i.e. the proxysql side-car,  PR #3554).

Certainly doable, but I think having to be explicit once in the parent chart is not a lot of effort. I would take care of it.

common/mariadb/values.yaml
fwiesel: I agree some extra work is needed on adapting the pipeline to make it work more seamlessly.
But I wouldn't say it requires extra coordination.  It uses the same image repository as in the ci-pipeline itself.

We need to update the pipelines (manually) when the image changes. That is exactly also the time we need to change this image version. So no extra coordination, but yes, extra manual effort.

We can change the pipeline-definition(s) to use exactly the same version, when the pipeline is applied.
Or we can change the pipeline to transport the latest version through the pipeline the same way all other images are transported.
Personally, I have a slight preference to the latter, but how it is done is a different discussion, I would say.

common/mariadb/templates/pre-change-job.yaml
fwiesel: I listed all the options in #3595. We probably move the discussions of alternatives there.

Just as a TLDR:
* `postStart` means restart. I want just want to get a user deployed out without a restart.
* While `postStart` running, the `Service` is down, as the container is [not marked running](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-state-running), so it adds to the downtime

common/mariadb/values.yaml
joker-at-work: Just a remark: Not all users of the shared pipeline use image version transporting and not all users of this chart (I would assume) use the shared pipeline. Additionally, people tend to only update the shared tasks for a new image-version and not every reference, so Dmitry has a good point here.

joker-at-work:
I don't understand that sentence in commit "Pre-upgrade hook for db/user creation"

> Instead of having custom init scripts with sligthly different patterns,
> the fiels databases and users are there create and update the users.

same commit. this means "users" and not "uses", right? both could make sense, but create pretty different expectations

> which need knowledge about the uses

common/mariadb/values.yaml
fwiesel: > Just a remark: Not all users of the shared pipeline use image version transporting and not all users of this chart (I would assume) use the shared pipeline

Well, but also the hook will only trigger if the "managed" database / user fields are used, so I don't think we have to cover all use-cases. Personally, I am fine with just the "openstack" services, or even nova and ironic, as they have more than one database & schema.

> Additionally, people tend to only update the shared tasks for a new image-version and not every reference, so Dmitry has a good point here.

I do not follow... 

I understand Dmitry the following way:
We upgrade k8s, so we need to ensure that the toolbox used everywhere is current enough.

There are 31 references to the `sapcc/unified-kubernetes-toolbox` in our pipeline definitions, 10 of those are under "shared", except in one location for ironic, all openstack services use the shared ones.

If we *need* to update the references in a pipeline then someone has to manually change those references,
and manually apply the pipeline. 

I am saying, the very procedure could result in the very same image being used the next time the pipeline is triggered inside this hooks as well in the pipeline job. 
Without effort of improving the situation, it just means that one has to change 32 places instead of 31.

Personally, I would opt for transporting the image version through the pipeline, and leave it to the teams which opt-out of image transport, but want to opt-in into the new functionality how they want to do it.

fwiesel:
> I don't understand that sentence in commit "Pre-upgrade hook for db/user creation"
> 
> > Instead of having custom init scripts with sligthly different patterns,
> > the fiels databases and users are there create and update the users.

Sorry, that was badly written, I've changed it to:
```
Instead of having custom init scripts with slightly different patterns,
the fields "databases" and "users" are there to create and update them.
```

> same commit. this means "users" and not "uses", right? both could make sense, but create pretty different expectations
> 
> > which need knowledge about the uses

Correct, I changed it.

common/mariadb/templates/initdb/_init.sql.tpl
```diff
@@ -1,3 +1,25 @@
-CREATE DATABASE IF NOT EXISTS {{.Values.name}} CHARACTER SET utf8 COLLATE utf8_general_ci;
-GRANT ALL PRIVILEGES ON {{.Values.name}}.* TO {{.Values.global.dbUser}}@localhost IDENTIFIED BY '{{include "db_password" .}}';
-GRANT ALL PRIVILEGES ON {{.Values.name}}.* TO {{.Values.global.dbUser}}@'%' IDENTIFIED BY '{{include "db_password" .}}';
+{{- if not .Values.databases }}
+CREATE DATABASE IF NOT EXISTS {{ .Values.name }};
```
IvoGoman, line 2: The `CHARACTER SET utf8 COLLATE utf8_general_ci` should be left in.
Otherwise the default character set of the server will be used.
Checked this for Nova in QA and the server default is `latin1` but the nova db is created with `utf8`:
```
show variables like '%character_set_server%';
+----------------------+--------+
| Variable_name        | Value  |
+----------------------+--------+
| character_set_server | latin1 |
+----------------------+--------+
```
```suggestion
CREATE DATABASE IF NOT EXISTS {{ .Values.name }} CHARACTER SET utf8 COLLATE utf8_general_ci;
```

common/mariadb/templates/initdb/_init.sql.tpl
```diff
@@ -1,3 +1,25 @@
...
+{{- else }}
+    {{- range .Values.databases }}
+CREATE DATABASE IF NOT EXISTS {{ . }};
```
IvoGoman, line 5: Same as above. `CHARACTER SET utf8 COLLATE utf8_general_ci` should be added to the create DB statement
```suggestion
CREATE DATABASE IF NOT EXISTS {{ . }} CHARACTER SET utf8 COLLATE utf8_general_ci;
```

common/mariadb/templates/initdb/_init.sql.tpl
fwiesel: In order to only have a single place where the character-set and collation is set, I added this in the front:
```
SET character_set_server = '{{ .Values.character_set_server }}';
SET collation_server = '{{ .Values.collation_server }}';
``` 

I hope, that's fine with you as well.

common/mariadb/templates/initdb/_init.sql.tpl
fwiesel: See [before](https://github.com/sapcc/helm-charts/pull/3555/files#r889121676)

common/mariadb/templates/initdb/_init.sql.tpl
fwiesel: No, I added it after your proposal.

fwiesel:
Hi, this PR is open now for three weeks. If there is no more feedback, I would assume that there is no strong objections, and I can merge it.

common/mariadb/templates/pre-change-job.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - kubectl-v{{ .Values.pre_change_hook.kubectl_version }}
```
joker-at-work, line 29: Isn't `kubectl` in the image  wrapper that automatically uses the appropriate version? Or is this to guard against an old image being used and thus not having the version yet?

common/mariadb/templates/pre-change-job.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - exec
+        - -c
+        - mariadb
+        - deploy/{{ include "fullName" . }}
```
joker-at-work, line 33: Why is this `deploy`? Is that short for `deployment`? If yes, why did you chose the short-form instead of the full name?

common/mariadb/templates/pre-change-job.yaml
```diff
@@ -0,0 +1,41 @@
+{{- if or .Values.databases .Values.users }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ include "fullName" . }}-pre-change"
```
joker-at-work, line 5: Don't we need some special naming (a variable in the name e.g. chart version), because we're unable to update Jobs with helm3?

common/mariadb/templates/pre-change-job.yaml
fwiesel: The wrapper is for a different use-case ("out of cluster") and doesn't work with an in-cluster setup.

It requires a user and a password so it can do its magic (version discovery) and log-on to any kubernetes api endpoint and works with a kubeconfig.
The script relies on the in-cluster config and the credentials provided by the service-account.

Since we are now more strict on following upstream versions of kubernetes, I do not expect a large drift between versions across the pipeline, which requires an automatic version detection. So I think, we can avoid injecting the credentials and having an additional step (version discovery), which might fail. Still, it can be overwritten in each region, should the need arise.

common/mariadb/templates/pre-change-job.yaml
fwiesel: No, the annotation `"helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded` takes care of that.
The job will be deleted when:
- the hook will be executed (so before the upgrade), in case it exists
- the hook has been successfully executed

In other words, helm should only leave the job around, if the hook didn't succeed, so one can debug it.
So one can simply retry the execution without having to manually delete the job, I've added `before-hook-creation` .

(Something which I think should also be added to https://github.com/sapcc/helm-charts/blob/master/common/region_check/templates/region-check-job.yaml#L16)

common/mariadb/templates/pre-change-job.yaml
fwiesel: > Why is this `deploy`? Is that short for `deployment`?

Yes:
```
$ kubectl api-resources | grep deploy
deployments                        deploy           apps/v1                                true         Deployment
```
> If yes, why did you chose the short-form instead of the full name?

That is the way it is done in the [official documentation](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#exec):
> kubectl exec deploy/mydeployment -- date

joker-at-work:
I'm fine with merging this, but would like to see @viktor-kovacs and @bashar-alkhateeb opinion.

fwiesel:
@viktor-kovacs @bashar-alkhateeb What is your view on this PR?

viktor-kovacs:
> @viktor-kovacs @bashar-alkhateeb What is your view on this PR?

I'm fine with this PR too.

bashar-alkhateeb:
somehow it sounds odd, to spin up a container (in a job) and give it a service account to configure another container in runtime using kubectl command.
i would prefer to have a precess inside the mariadb container to reconfigure it in case the config map got updated. (watching the configmap update)

fwiesel:
> i would prefer to have a process inside the mariadb container to reconfigure it in case the config map got updated. (watching the configmap update)

Well, I tried that with rabbitmq and that had some negative feedback.
See the [ADR](https://github.com/sapcc/helm-charts/pull/3595) for the reasoning.
I want to avoid to have the same discussion in multiple places so I would like to focus that there.

For me more important which option to chose it to have an implementation in place.
We can still improve on that.

fwiesel:
@bashar-alkhateeb Is that more an remark, or are you opposing this change?

PR #3556, opened 2022-05-26
[mariadb] Undefine name field

fwiesel:
The field is required to be set in the fullName macro,
but defining the value to 'DEFINED-IN-COMPONENT-CHART' will
simply cause that value to be used if it isn't defined.

Rather fail with a clean error message, then with a garbage
output

### DIFF
--- a/common/mariadb/values.yaml
+++ b/common/mariadb/values.yaml
@@ -3,7 +3,7 @@
 # Declare name/value pairs to be passed into your templates.
 # name: value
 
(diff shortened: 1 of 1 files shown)


majewsky:
I prefer `name: null` in these cases, since it documents the expected keys more robustly. A null value will still behave like you expect, both with `if` and with `require`.

fwiesel:
I've changed it and all other places to `null`. I think, we should at least in one chart follow one convention. I have no strong opinion on which.

PR #3557, opened 2022-05-26
[mariadb] Require root_password to be set

fwiesel:
Prior to this change, helm would not complain about the value
to be unset, but mariadb would never start up, as the preStart
hook would never complete.

Hence require the password, as it won't work without it.


majewsky (approved):
Yikes, good catch.

PR #3558, opened 2022-05-26
[Cinder] update helm chart requirements

hemna:
This patch updates the requirements.yml for
utils
mariadb
rabbitmq
redis

### DIFF
--- a/openstack/cinder/requirements.lock
+++ b/openstack/cinder/requirements.lock
@@ -1,10 +1,10 @@
 dependencies:
 - name: utils
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.3.7
+  version: 0.3.11
 - name: mariadb
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.3.50
+  version: 0.3.55
 - name: mysql_metrics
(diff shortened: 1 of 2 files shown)


joker-at-work:
Could you please specify why we're updating the requirements and what features we get by that, i.e. what's the goal of this PR?

hemna:
For mariadb:

```
- Prevent trible scraping of metrics (#3531) - let prometheus find the container with the port "metrics" by itself and scrape the pod once instead 3 times (for each container).
- adjust init.sql: create DB only if not exist
- Bumps backup image, reduces memory usage during full backup (#3465)
- Adds flag to enable readiness sidecar (#3295)
```

for rabbitmq
```
make probes configurable
```

for redis
```
- exporter 1.0.3 -> 1.37.0
- update to redis 7.0.0
- bump exporter
```

for utils
```
- disable oslo_messaging_rabbit heartbeat_in_pthread
- openstack.utils.db_url_mysql: Use dbUser credentials instead of root
- Fix mariadb name
- ```

joker-at-work:
Would you mind adding that to the commit-message, so anybody could find this information without knowing there was a PR and searching for it?

PR #3560, opened 2022-05-27
[unbound] config updates:

galkindmitrii:
- log servfail responses
- max ttl for negative cache to 100 sec vs. 3600 by default
- cache max ttl to 8 hours vs. 24 hours by default


dhoeller (approved):
LGTM

PR #3562, opened 2022-05-30
Pin PX on specific apods that are cabled to PX switches

swagner-de:
We used to pin PX to certain AZs where we knew the apods were cabled to
PX, however if more apods come in within an AZ and not all are cabled
to PX, this is going to break.

We now specify explicitly which apod to run on. We also moved some
logic regarding the placement in AZs from the chart to the k8s
scheduler. The intention is that we have one instance from each PX
domain running in each AZ. Subsequently, if we have more than 1 AZ in a
region we define an podAntiAffinity on the combined service and domain
label, so a service's domain will never be placed in the same AZ.


defo89 (approved):
lgtm

PR #3563, opened 2022-05-30
[nova] Disable console-auth

fwiesel:
Console-auth is deprecated, so we disable it via
setting the replica count to zero (before removing
it eventually completely)

### DIFF
--- a/openstack/nova/templates/consoleauth-deployment.yaml
+++ b/openstack/nova/templates/consoleauth-deployment.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.pod.replicas.consoleauth }}
 kind: Deployment
 apiVersion: apps/v1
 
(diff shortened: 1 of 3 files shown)


openstack/nova/templates/consoleauth-deployment.yaml
```diff
@@ -74,6 +75,7 @@ spec:
           mountPath: /etc/nova/logging.ini
           subPath: logging.ini
           readOnly: true
+{{- end }}
 {{ $envAll := . }}
 {{- range $name, $config := .Values.consoles }}
   {{- if $config.enabled }}
```
joker-at-work, line 81: Do you plan to remove this part when we finally remove nova-consoleauth from the helm-chart?

joker-at-work (approved):

openstack/nova/templates/consoleauth-deployment.yaml
fwiesel: Line 1-78? Yes, that should go. Anything between `if .Values.pod.replicas.consoleauth` and `end`

PR #3565, opened 2022-06-01
[prometheus] Allow -* suffix for named metrics ports

fwiesel:
In case there is more than one port with metrics,
we need to give them distinct names. This change takes all
ports either named `metrics` (as before), but also ports with a suffix separated by a hyphen as prometheus metrics ports.

### DIFF
--- a/common/prometheus-server/templates/podmonitors/pod-sd.yaml
+++ b/common/prometheus-server/templates/podmonitors/pod-sd.yaml
@@ -41,7 +41,7 @@ spec:
             - __meta_kubernetes_pod_container_port_number
             - __meta_kubernetes_pod_container_port_name
             - __meta_kubernetes_pod_annotation_prometheus_io_port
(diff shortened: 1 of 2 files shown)


majewsky:
@fwiesel I've made two amendments to this PR:

1. The non-controversial part is that the same change is now also made to the `ServiceMonitor` object, so the new naming scheme will also work for ports on services, not just for ports on pods.
2. The possibly controversial part is that I'm restricting the regex from `metrics-.*` to `metrics-alt\d+`. I want to keep the rest of the `metrics-...` namespace open, to give us some room for defining new naming conventions if we ever rework this whole service discovery stuff entirely. (I already have a ticket for that in the backlog, but since Arno is OoO for now, we will probably not be working on this right away.)

majewsky:
If there are no objections until then, I will merge this tomorrow.

PR #3571, opened 2022-06-01
[Cinder] Enable SAP Affinity filters

hemna:
This patch enable the new affinity/anti-affinity filters.

SAPSameBackendFilter
SAPDifferentBackendFilter


joker-at-work (commented):
Is there a reason why we move those SAP-flavored filters to the end? Are they especially expensive?

hemna:
> 

no

joker-at-work (approved):

PR #3579, opened 2022-06-02
[nova] Add ProxySQL side-cars

fwiesel:
This makes use of the proxysql side-car snippets.

### DIFF
--- a/openstack/nova/templates/bin/_db-online-migrate.tpl
+++ b/openstack/nova/templates/bin/_db-online-migrate.tpl
@@ -12,3 +12,5 @@ $nova_manage db online_data_migrations
 if echo "${available_commands_text}" | grep -q -E '[{,]placement[},]'; then
   $nova_manage placement sync_aggregates
 fi
+
+{{ include "utils.proxysql.proxysql_signal_stop_script" . }}
--- a/openstack/nova/templates/cell2-conductor-deployment.yaml
+++ b/openstack/nova/templates/cell2-conductor-deployment.yaml
@@ -29,29 +29,43 @@ spec:
         alert-service: nova
 {{ tuple . "nova" "conductor" | include "helm-toolkit.snippets.kubernetes_metadata_labels" | indent 8 }}
       annotations:
-        {{- if .Values.cell2.conductor.config_file.DEFAULT.statsd_enabled }}
+        {{- if or .Values.cell2.conductor.config_file.DEFAULT.statsd_enabled .Values.proxysql.mode }}
         prometheus.io/scrape: "true"
         prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
         {{- end }}
         configmap-etc-hash: {{ include (print $.Template.BasePath "/etc-configmap.yaml") . | sha256sum }}
     spec:
 {{ tuple . "nova" "conductor" | include "kubernetes_pod_anti_affinity" | indent 6 }}
+{{ include "utils.proxysql.pod_settings" . | indent 6 }}
       terminationGracePeriodSeconds: {{ .Values.defaults.default.graceful_shutdown_timeout | add 5 }}
       hostname: nova-conductor
+      initContainers:
+        - name: dependencies
+          image: {{ required ".Values.global.registry is missing" .Values.global.registry}}/ubuntu-source-nova-conductor:{{.Values.imageVersionNovaConductor | default .Values.imageVersionNova | default .Values.imageVersion | required "Please set nova.imageVersion or similar"}}
+          imagePullPolicy: IfNotPresent
+          command:
+            - kubernetes-entrypoint
+          env:
+            - name: COMMAND
+              value: "true"
+            - name: NAMESPACE
+              value: {{ .Release.Namespace }}
+            - name: DEPENDENCY_SERVICE
+              value: "{{ .Release.Name }}-{{ .Values.cell2.name }}-rabbitmq,{{ .Release.Name }}-{{ .Values.cell2.name }}-mariadb"      
       containers:
         - name: nova-conductor
           image: {{ required ".Values.global.registry is missing" .Values.global.registry}}/ubuntu-source-nova-conductor:{{.Values.imageVersionNovaConductor | default .Values.imageVersionNova | default .Values.imageVersion | required "Please set nova.imageVersion or similar"}}
           imagePullPolicy: IfNotPresent
           command:
             - dumb-init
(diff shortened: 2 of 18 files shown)


openstack/nova/templates/cell2-conductor-deployment.yaml
```diff
@@ -49,6 +50,7 @@ spec:
               name: nova-etc
               subPath: nova-cell2.conf
               readOnly: true
+            {{- include "utils.proxysql.volume_mount" . | indent 12 }}
```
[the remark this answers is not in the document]
fwiesel: I completely overlooked that...
Yes, that isn't going to work (with unix sockets).
I'll extract it to a job, and create a dependency for this deployment

openstack/nova/values.yaml
```diff
@@ -706,13 +706,28 @@ db_name: nova
...
+  - nova_cell0
+  users:
+    nova:
+      name: nova
+      grants:
+      - "ALL PERMISSIONS on nova.*"
+    nova_cell0:
+      name: nova_cell0
+      grants:
+      - "ALL PERMISSIONS on nova_cell0.*"
```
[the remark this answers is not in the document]
fwiesel: It is simply a [grant statement]( https://mariadb.com/kb/en/grant/) without `GRANT` and `TO user_specification [ user_options ...]`
So:
```
    priv_type [(column_list)] [, priv_type [(column_list)]] ...
    ON [object_type] priv_level
``` would be okay.

File: openstack/nova/templates/bin/_db-online-migrate.tpl [hunk 3]
```diff
@@ -12,5 +10,7 @@ $nova_manage db online_data_migrations
 {{- end }}
 
 if echo "${available_commands_text}" | grep -q -E '[{,]placement[},]'; then
-  $nova_manage placement sync_aggregates
+  $nova_manage placement sync_aggregates || true
```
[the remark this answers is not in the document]
fwiesel: Ah, sorry. That is a left-over from testing when I didn't want to debug why that was failing at a particular time. I removed it.

openstack/nova/values.yaml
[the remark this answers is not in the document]
fwiesel: And it is now [ALL PRIVILEGES](https://mariadb.com/kb/en/grant/#the-all-privileges-privilege), not `PERMISSIONS`

openstack/nova/templates/cell2-conductor-deployment.yaml
```diff
@@ -29,40 +29,44 @@ spec:
...
             - name: COMMAND
-              value: "nova-conductor --config-file etc/nova/nova.conf --config-file /etc/nova/nova-conductor.conf --config-file /etc/nova/nova-cell2.conf"
+              value: "true"
             - name: NAMESPACE
               value: {{ .Release.Namespace }}
             - name: DEPENDENCY_SERVICE
-              value: "{{ .Release.Name }}-{{ .Values.cell2.name }}-rabbitmq,{{ .Release.Name }}-{{ .Values.cell2.name }}-mariadb"
+              value: "{{ .Release.Name }}-{{ .Values.cell2.name }}-rabbitmq,{{ .Release.Name }}-{{ .Values.cell2.name }}-mariadb"      
+            - name: DEPENDENCY_JOB
+              value: "nova-update-cells2-{{ .Release.Revision }}"
```
joker-at-work, line 56: This means, we cannot restart the pod anymore, if the Job `nova-update-cells2-$rev` got deleted, right? We don't have the same restriction in the non-cell2 conductor deployment. Would it be possible to integrate the cell2 syncing into the main `db-migration` Job?

I'm also wondering why we name this `nova-update-cells2` and not `nova-db-migrate-cell2` as what we're doing here is the `db-migration`'s action for cell2.

fwiesel: Now actually [reading the code](https://github.com/sapcc/nova/blob/stable/rocky-m3/nova/cmd/manage.py#L470-L483), I think the `db-migrate` job already updates all cells...
We just have to ensure that the cells are created before the db-migration is run.

joker-at-work: Then there's not point in having that init-container anymore, is there?

fwiesel: I was jumping a bit the gun at reading the help text, and missed that
>  Currently fanout only looks for cell0.

So, back to refactoring that. I'll open a separate PR for that

openstack/nova/values.yaml
```diff
@@ -706,13 +706,28 @@ db_name: nova
 max_pool_size: 10
 max_overflow: 5
 
+proxysql:
+  mode:
```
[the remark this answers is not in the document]
fwiesel: Fix is in #3829

openstack/nova/templates/cell2-conductor-deployment.yaml
fwiesel: PR is in #3768

openstack/nova/templates/cell2-conductor-deployment.yaml
```diff
@@ -29,29 +29,42 @@ spec:
...
+        - name: dependencies
           image: {{ required ".Values.global.registry is missing" .Values.global.registry}}/ubuntu-source-nova-conductor:{{.Values.imageVersionNovaConductor | default .Values.imageVersionNova | default .Values.imageVersion | required "Please set nova.imageVersion or similar"}}
           imagePullPolicy: IfNotPresent
           command:
-            - dumb-init
             - kubernetes-entrypoint
           env:
             - name: COMMAND
-              value: "nova-conductor --config-file etc/nova/nova.conf --config-file /etc/nova/nova-conductor.conf --config-file /etc/nova/nova-cell2.conf"
+              value: "true"
```
joker-at-work, line 50: Why is the cell2 conductor so different from the cell1 conductor in that it has an init-container for waiting for the dependencies instead of in the "real" container and no `dumb-init` anymore for the conductor itself?

joker-at-work: Could you please explain why we have that init-container to wait for the dependencies instead of waiting in the real container? It seems like nova-conductor is the only Deployment doing that now. What's the adavantage in doing it in the init-container and why do others not need it?

joker-at-work: Thank you.

joker-at-work (approved):

PR #3591, opened 2022-06-02
system/gatekeeper: add test for policies

SuperSandro2000:


majewsky (commented):
Nice. Can you also add some tests that pass? I'm thinking one for a pod with CPU requests below the maximum, and one for a pod with no CPU requests at all. Maybe also one for a pod with CPU requests on _some_ containers only.

majewsky (approved):

PR #3594, opened 2022-06-03
[prometheus-global] memory requests for global prometheis

viennaa:
to allow proper scheduling k8s should be aware of the memory load expected by these global prometheis instances.
Otherwise prometheus could end up running on a node with a very limited amount of memory available causing it to permanently crash.


richardtief (approved):
cc @auhlig

PR #3595, opened 2022-06-03
ADR-0001: Using Helm-Hooks with Kubectl Pod for Configuration Changes

fwiesel:


docs/decisions/0001-using-helm-hook-with-kubectl-pod-for-configuration-changes.md
```diff
@@ -0,0 +1,138 @@
...
+* Bad, because we need extra pipelines to add out configuration logic to the images
+* Bad, because a change of the configuration logic would/should always trigger a deployment, ie. downtime
+
+### Different Image
+
+The opposite of "Same Image".
+
+### Helm-Chart Hook
+
+We could make use of helm-chart hooks for upgrade and rollback.
```
majewsky, line 79: May be good to link to https://helm.sh/docs/topics/charts_hooks/ here. I was confused what was meant by this since I had not heard of this feature before.

majewsky (commented):
My main question would be, how many downtimes are actually caused by configuration changes vs. required pod restarts for OS upgrades? If the majority of the problems ends up being unavoidable pod restarts, would the time spent working on this mitigation not be better spent on preparing the HA setup?

docs/decisions/0001-using-helm-hook-with-kubectl-pod-for-configuration-changes.md
fwiesel: I've added the link

fwiesel:
Now that the conferences are over, I hope I can get you to have a look at it.
Failing that, I have to assume that you are okay with it.

bashar-alkhateeb:
I think we can use the same solution that the rabbitmq operator is using to configure custom configuration, the operator uses a sidecar container to configure the cluster via API calls.
the solution in my point of view for rabbitmq can be by spinning up a side car container whenever we have configuration change the container will make some API calls against the container/cluster to deploy the new state, calls need management port (which we configure) ,a default user and a default password (which we have as well, configured in the startup script).
with this solution we will avoid creating new service accounts and give a container permissions to the K8S cluster, in addition to that we will avoid multiple restarts of the pods if there is only a configuration change 

API Calls can be done by a bash script run in the sidecar container e.g.
`curl -i -u rabbitmq:default_password -H "content-type:application/json"  -XPUT -d'{"password":"bkh1bkh1","tags":"testapi"}' "http://localhost:15672/api/users/bkh1/"
HTTP/1.1 201 Created
content-length: 0
content-security-policy: script-src 'self' 'unsafe-eval' 'unsafe-inline'; object-src 'self'
date: Mon, 18 Jul 2022 13:19:48 GMT
server: Cowboy
vary: accept, accept-encoding, origin
`

PR #3598, opened 2022-06-07
[mariadb] Common user/db creation script

fwiesel:
Instead of having repetitive free-form init.sql scripts,
provide the fields `databases` and `users`, which allow
the user of the script to create the databases and users
uniformly.


galkindmitrii (approved):
Looks good. As long as existing DBs/services don't need to do changes should be fine.

fwiesel:
Yes, it is completely backwards compatible and needs to be opt-in by version change and actually setting the values `database` and `users`

PR #3601, opened 2022-06-07
[Keystone] CCloud conform swiftreseller role

reimannf:
Swift roles will be migrated to a CCloud conform naming approach and
`swiftreseller` will be replaced by `cloud_objectstore_admin`. To allow
a smooth transition, we will have both roles for a period of time.
This change will blacklist the new role `cloud_objectstore_admin` as well.
The old role `swiftreseller` will be removed after the transition.


bbobrov (approved):

rajivmucheli:
@bbobrov arent we using policy.yaml instead of policy.json ? the commit is in policy.json.

PR #3602, opened 2022-06-07
Re - federate cpu metrics

SchwarzM:
after https://github.com/sapcc/helm-charts/commit/beb675c3eb0ba237d50c2cf956f0d4a1dbf81472 we do not aggregate cpu metrics anymore but also do not federate them thus they are not showing up in the prom.
this should re-enable the federation of the cpu metrics.


BugRoger (approved):

BugRoger:
LGTM

majewsky (approved):

PR #3605, opened 2022-06-08
[nova] Split out placement database

fwiesel:
Based on #3223

Moves all tables from `nova_api` to the one specified in `placementdbName` as soon as a `placementdbPassword` is specified.
The operations are idempotent, and can be reverted with a deployment with the value `placement.unsplit=true`.

No downtime required, as the alias is there practically immediately.

The backup still needs to be adjusted.

### DIFF
--- a/openstack/nova/templates/placement-configmap.yaml
+++ b/openstack/nova/templates/placement-configmap.yaml
@@ -10,6 +10,14 @@ metadata:
 data:
   nova.conf: |
 {{ include (print .Template.BasePath "/etc/_nova.conf.tpl") . | indent 4 }}
+    [placement_database]
+    connection = mysql+pymysql://
+    {{- if .Values.placementdbPassword -}}
+    {{ .Values.placementdbUser }}:{{ .Values.placementdbPassword | urlquery }}@nova-api-mariadb/{{ .Values.placementdbName }}?charset=utf8
+    {{- else -}}
+    {{ .Values.apidbUser }}:{{ .Values.apidbPassword | urlquery }}@nova-api-mariadb/{{ .Values.apidbName }}?charset=utf8
+    {{- end }}
+    {{- include "ini_sections.database_options_mysql" . | indent 4 }}
   policy.json: |
 {{ include (print .Template.BasePath "/etc/_nova-policy.json.tpl") . | indent 4 }}
   logging.ini: |
--- a/openstack/nova/templates/placement-configmap.yaml
+++ b/openstack/nova/templates/placement-configmap.yaml
@@ -1,3 +1,4 @@
+{{- if and .Values.placement.enabled }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -10,6 +11,7 @@ metadata:
 data:
   nova.conf: |
 {{ include (print .Template.BasePath "/etc/_nova.conf.tpl") . | indent 4 }}
+{{- if not .Values.placement.unsplit }}
     [placement_database]
     connection = mysql+pymysql://
     {{- if .Values.placementdbPassword -}}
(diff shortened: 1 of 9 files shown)


openstack/nova/templates/placement-db-split-job.yaml
```diff
@@ -0,0 +1,52 @@
...
+      labels:
+        app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
+        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ .Release.Name }}-pod-exec
+      containers:
+      - name: placement-db-split
+        image: "{{ required ".Values.global.dockerHubMirrorAlternateRegion is missing" .Values.global.dockerHubMirrorAlternateRegion }}/{{ .Values.pre_change_hook.image }}:{{ .Values.pre_change_hook.image_version }}"
```
joker-at-work, line 29: Why are we using the alternate region's dockerhub mirror here?

openstack/nova/templates/placement-db-unsplit-job.yaml
```diff
@@ -0,0 +1,48 @@
...
+        - -uroot
+        - --batch
+        - -e
+        - |
+          {{- range mustReverse .Values.placement.tables }}
+            {{- list $envAll . | include "helpers.table_dealias" | indent 10 }}
+          {{- end }}
+
+          DROP DATABASE IF EXISTS {{ .Values.placementdbName }};
+          DROP USER IF EXISTS {{ .Values.placementdbUser }};
```
joker-at-work, line 47: We don't check for those variables to be set, while we do check for them when splitting. Is it save to run this with null values or empty strings?

openstack/nova/templates/pod-exec-rbac.yaml
```diff
@@ -0,0 +1,46 @@
+{{- if and .Values.placement.enabled (or .Values.placement.unsplit (and .Values.placementdbUser .Values.placementdbPassword .Values.placementdbName)) }}
```
joker-at-work, line 1: We need this, too, if `unsplit` is `true`.

openstack/nova/templates/placement-db-split-job.yaml
fwiesel: Mostly because I copied and pasted it. I changed it to `.Values.global.registry`

openstack/nova/templates/placement-configmap.yaml
```diff
@@ -10,6 +10,14 @@ metadata:
 data:
   nova.conf: |
 {{ include (print .Template.BasePath "/etc/_nova.conf.tpl") . | indent 4 }}
+    [placement_database]
+    connection = mysql+pymysql://
+    {{- if .Values.placementdbPassword -}}
+    {{ .Values.placementdbUser }}:{{ .Values.placementdbPassword | urlquery }}@nova-api-mariadb/{{ .Values.placementdbName }}?charset=utf8
```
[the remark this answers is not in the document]
fwiesel: It moves the tables to its own database (or schema) within the same mariadb instance.
There is no automatic way to move it to another mariadb instance (yet).
My understanding was, we have an agreement to leave that to a later time after we have split the services.
So, I was not seeing the need to provide the means to point it to another mariadb-instance, as placement within nova is legacy anyway.

openstack/nova/templates/placement-db-unsplit-job.yaml
fwiesel: Yes, it results in a syntax error.
> Query Error: Error: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 1

openstack/nova/templates/pod-exec-rbac.yaml
fwiesel: It is. As long as placement is enabled and either unsplit is true, or the credentials are set:
>  if and .Values.placement.enabled (or **.Values.placement.unsplit**

joker-at-work: I don't seem to be able to read. You're absolutely right m)

joker-at-work (approved):

PR #3608, opened 2022-06-09
[nova] Finalize placement-db split

fwiesel:


fwiesel:
This should be merged and rolled ideally shortly after #3605 has deployed, but not before

joker-at-work (approved):

PR #3610, opened 2022-06-09
Use pod label to denote F5 model name for Limes

BenjaminLudwigSAP:
This PR will be merged after the `model` helm value is added.
This code change exposes the `model` value as an Octavia worker pod label.
Limes will iterate over Octavia worker pods and map model names to capacity values, then sum them up per region.


majewsky (approved):

openstack/octavia/templates/octavia-worker-deployment.yaml
```diff
@@ -10,6 +10,7 @@ metadata:
     helm.sh/chart: {{ include "octavia.chart" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
+    f5-device-model: {{ index .Values.workers $name "model" }}
```
majewsky, line 13: Can be simplified:
```suggestion
    f5-device-model: {{ $v.model }}
```

PR #3611, opened 2022-06-09
[Cinder] remove prometheus.io/port

hemna:
This removes the prometheus.io/port from the chart and changes it to
a spec named metrics.
This is related to discussion on slack:

https://operations.global.cloud.sap/docs/support/playbook/kubernetes/target_scraped_multiple_times/#usage-or-prometheusioport-annotation-with-multiple-ports


richardtief (approved):
Looks good.

PR #3628, opened 2022-06-15
Cerebro LB quota adjustment

chrismcveigh:
- Loadbalancer adjustment due to Quota failures with pool_members

### DIFF
--- a/openstack/cerebro/initial-quota.yaml
+++ b/openstack/cerebro/initial-quota.yaml
@@ -1,13 +1,13 @@
 ccadmin/cerebro:
     compute:
         cores: at least 80
-        instances: at least 20 
+        instances: at least 20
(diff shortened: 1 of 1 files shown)


openstack/cerebro/initial-quota.yaml
```diff
@@ -1,26 +1,22 @@
...
         subnet_pools: at least 2
         ports: at least 128
         rbac_policies: at least 5
         security_group_rules: at least 60
-        security_groups: at least 10
-        loadbalancers: at least 5
-        healthmonitors: at least 10
-        l7policies: at least 0
-        listeners: at least 10
-        pools: at least 20
```
viennaa, line 23: they need to stay, please just increase where needed

viennaa (changes requested):

PR #3631, opened 2022-06-15
[rabbitmq] hermes openstack users

notque:
Start for adding users to rabbitmq for hermes openstack. We should probably be using a secrets file instead of inline placement of secrets.


fwiesel:
I rather would like to see it as 
```
{{- range $k, $v :=  .Values.users }}
 upsert_user {{$v.user | include "rabbitmq.shell_quote" }} {{ required (print ".Values.users." $k ".password missing")  $v.password | include "rabbitmq.shell_quote" }}
{{- end }}
```

PR #3645, opened 2022-06-20
[keystone] Option for central hermes rabbitmq

fwiesel:
This provides the option of using the HA setup rabbitmq from hermes
instead of the single non-ha one.

To ensure that no messages are dropped, one first needs to
supply audit.user and audit.password to enable
sending out the new events and after ensuring that
no messages are on the queue of the old rabbitmq,
one can disable the internal one.

### DIFF
--- a/openstack/keystone/templates/etc/_keystone.conf.tpl
+++ b/openstack/keystone/templates/etc/_keystone.conf.tpl
@@ -162,12 +162,26 @@ disable_user_account_days_inactive = {{ .Values.disable_user_account_days_inacti
 {{- end }}
 
 [oslo_messaging_notifications]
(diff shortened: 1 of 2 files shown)


rajivmucheli:
@bbobrov this looks fine by me, maybe you could review as well ?

bbobrov:
This looks good, but i do not fully understand this part:

> one first needs to supply audit.user and audit.password to enable sending out the new events

Is a user supposed to provide the values manually? Can these values be fetched from Vault? Or is it just for testing?

rajivmucheli:
an internal PR is raised and credentials are stored in Vault.

rajivmucheli (approved):
lgtm

PR #3650, opened 2022-06-21
[rabbitmq] Configure rabbitmq in Job

fwiesel:
Instead of starting rabbitmq twice,
and reconfiguring it only on startup,
move the reconfiguration to a Job.

The job uses kubectl to execute the configuration
script in the context of the rabbitmq container.

This makes the deployment compatible with
rabbitmq versions later than 3.6 which removed
the startup scripts used.


bashar-alkhateeb (approved):
looks fine for me .. but still think we need to find another solution for reconfiguring rabbitmq without interrupting the service .. but for now it is fine

PR #3651, opened 2022-06-21
[octavia] Pending LBs metric: Consolidate field names, add network ID

BenjaminLudwigSAP:
Change octavia_seconds_since_last_nonpending_status in two ways:
- Add `vip.network_id` field, so that it's more easily noticeable when one stuck LB clogs an entire tenant (tenants map to networks).
- Consolidate field names: Rename `octavia_host` to `loadbalancer_host` and `current_status` to `loadbalancer_status`, so that these related fields appear together in the metrics overview.

As far as I know the `octavia_seconds_since_last_nonpending_status` metric is not consumed by any automation.

### DIFF
--- a/openstack/octavia/values.yaml
+++ b/openstack/octavia/values.yaml
@@ -133,17 +133,18 @@ mysql_metrics:
         - "count_gauge"
     - name: octavia_seconds_since_last_nonpending_status
       labels:
-        - "current_status"
-        - "octavia_host"
         - "loadbalancer_id"
+        - "loadbalancer_status"
         - "loadbalancer_name"
+        - "loadbalancer_host"
+        - "loadbalancer_network"
       help: Seconds since last non-pending state of LB
       query: |
         (SELECT
           UNIX_TIMESTAMP(now()) - UNIX_TIMESTAMP(coalesce(lb.updated_at, lb.created_at)) AS elapsed_time,
-          lb.name AS loadbalancer_name,
-          lb.provisioning_status AS loadbalancer_status,
           lb.id AS loadbalancer_id,
+          lb.provisioning_status AS loadbalancer_status,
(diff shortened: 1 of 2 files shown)


openstack/octavia/values.yaml
```diff
@@ -140,14 +140,16 @@ mysql_metrics:
...
-          id AS loadbalancer_id,
-          server_group_id AS octavia_host
-        FROM load_balancer
-        WHERE provisioning_status LIKE "PENDING_%")
-        UNION (select '0', '', '', '', '');
+          UNIX_TIMESTAMP(now()) - UNIX_TIMESTAMP(coalesce(lb.updated_at, lb.created_at)) AS elapsed_time,
+          lb.name AS loadbalancer_name,
+          lb.provisioning_status AS loadbalancer_status,
+          lb.id AS loadbalancer_id,
+          lb.server_group_id AS loadbalancer_host,
```
notandy, line 147: If you rename the variable `octavia_host`, you need to adapt the labels to be exported by the exporter (see line #137), also this change would not be consistent with line 123 of octavia_loadbalancers_count_gauge metric, which still uses `octavia_host` label.
As last, if you change labels, please also change the octavia alerts in the same PR since else you would introduce a window in which the alerts would be broken.

All in all, I would recommend to stay with octavia_host, since it's purely cosmetic.

Regards the network_id, I am fine, but it needs to be added to the labels.

notandy (changes requested):

openstack/octavia/values.yaml
laerlingSAP: Thanks, I totally missed the labels..

notandy (approved):
lgtm, please don't forget to adapt the grafana dashboards for octavia after this change

PR #3655, opened 2022-06-21
system/gatekeeper: run tests for policies that require doop-image-checker

SuperSandro2000:


system/gatekeeper/templates/constrainttemplate-outdated-image-bases.yaml
```diff
@@ -11,24 +11,39 @@ spec:
...
+              type: string
 
   targets:
     - target: admission.k8s.gatekeeper.sh
       rego: |
         package outdatedimagebases
 
+        # check each pod container
+        obj := input.review.object
+        container := obj.spec.containers[_]
```
majewsky, line 29: I see why you're pulling this out of the `violation[]{ ... }` block, but this won't work as far as I can tell. My understanding would be that this will pick one `container` at random and not consider the rest. You will need to construct an object on the top-level that iterates over all containers and then contains all the responses within itself. Something like:

```rego
checks[{ "image": container.image, "response": resp }] {
  container := input.review.object.spec.containers[_]
  resp := ...
}
```

Then you can iterate over that object in the `violation` blocks.

system/gatekeeper/templates/constrainttemplate-vulnerable-images.yaml
```diff
@@ -18,25 +18,40 @@ spec:
...
+              type: string
 
   targets:
     - target: admission.k8s.gatekeeper.sh
       rego: |
         package vulnerableimages
 
+        # check each pod container
+        obj := input.review.object
+        container := obj.status.containerStatuses[_]
```
majewsky, line 36: Same concern as above.

system/gatekeeper/tests/fixtures/vulnerable-images/pod-low.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Pod
+metadata:
+  name: dummy
+status:
+  containerStatuses:
+    - image: keppel.example.com/vulnerability:low
+      imageID: docker-pullable://keppel.example.com/vulnerability:low@sha256:0000000000000000000000000000000000000000000000000000000000000000
+      name: dummy
+      ready: true
+      restartCount: 0
```
majewsky, line 11: As a test case for the aforementioned problem, please combine some of the test cases into a single Pod with multiple containers, to prove that the policy actually checks all of the containers in it, not just one of them.

majewsky (changes requested):

majewsky (changes requested):
```
=== RUN   vulnerable-images
    === RUN   pod-vulnerable
    --- FAIL: pod-vulnerable    (0.004s)
        open //x/src/github.com/sapcc/helm-charts/system/gatekeeper/tests/fixtures/vulnerable-images/pod-vulnerable.yaml: no such file or directory
```
I'm guessing that you forgot to commit this file?

majewsky (approved):

PR #3660, opened 2022-06-22
[nova] Remove placement from nova chart

fwiesel:
We have now a separate chart, so we might as well drop it


fwiesel:
True, missed that one. Now we only check that we are not on rocky anymore

joker-at-work (approved):

PR #3661, opened 2022-06-22
[nova] Fix placement db migration

fwiesel:
Until we use a version with a proper split,
the migration still need to run on both databases


fwiesel:
That is one of two options. Either we run the migrations like this (meaning we have to upgrade nova before placement).
Or we disable the dbmigrations, if placement has been disabled.

fwiesel:
Closing in favour of #3662

PR #3662, opened 2022-06-22
[nova] Disable online migrations with placement

fwiesel:
If placement gets deployed by its own service,
we can't have nova mess with its tables

### DIFF
--- a/openstack/nova/templates/bin/_db-online-migrate.tpl
+++ b/openstack/nova/templates/bin/_db-online-migrate.tpl
@@ -6,7 +6,10 @@ set -e
 nova_manage="nova-manage --config-file /etc/nova/nova.conf"
 available_commands_text=$(nova-manage --help | awk '/Command categories/ {getline; print $0}')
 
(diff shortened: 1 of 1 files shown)


fwiesel:
Alternative to #3661 , and my preferred option, as it allows us to upgrade placement already.
Third option is patching out all placement related migrations.

openstack/nova/templates/bin/_db-online-migrate.tpl
```diff
@@ -5,8 +5,10 @@ set -e
 
 nova_manage="nova-manage --config-file /etc/nova/nova.conf"
 available_commands_text=$(nova-manage --help | awk '/Command categories/ {getline; print $0}')
+{{- if .Values.placement.enabled }}
```
joker-at-work, line 8: Are we able to access the release version here to check that if we're on rocky? I fear we might forget this setting until we upgrade and then wonder ...

fwiesel: I am now checking for rocky in `.Values.imageVersion`  (as done in other places)

joker-at-work (approved):
Tested it with different values and seems to work:

PR #3663, opened 2022-06-22
[common/rabbitmq] updating rabbitmq to rabbitmq 3.10.5

bashar-alkhateeb:
update rabbitmq to rabbitmq 3.10.5
- replacing rabbitmq-start script with rabbitmq-config script to config rabbitmq after startup
- using lifecycle postStart hook to trigger rabbtimq-config script after building rabbitmq container


common/rabbitmq/templates/statefulset.yaml
```diff
@@ -42,10 +42,10 @@ spec:
...
         image: "{{include "dockerHubMirror" .}}/{{ .Values.image }}:{{.Values.imageTag }}"
         imagePullPolicy: {{ default "IfNotPresent" .Values.imagePullPolicy | quote }}
-        command:
-          - bash
-        args:
-          - /container.init/rabbitmq-start
+        lifecycle:
+          postStart:
+            exec:
+              command: ["bash", "/container.init/rabbitmq-config"]
```
fwiesel, line 48: See comment in "deployment"

PR #3667, opened 2022-06-23
[common/rabbitmq] Optimise startup script

fwiesel:
Change the script to call rabbitmq less often. All the changes except on-line reconfiguration.

### DIFF
--- a/common/rabbitmq/ci/test-values.yaml
+++ b/common/rabbitmq/ci/test-values.yaml
@@ -11,10 +11,10 @@ ports:
 
 users:
   default:
-    user: test
+    user: test1
     password: test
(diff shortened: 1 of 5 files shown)


bashar-alkhateeb (commented):
looks reasonable for me

fwiesel:
It certainly sends `SIGTERM`, but to my knowledge that triggers the `EXIT` handler in bash (not in other shells though).

notandy (changes requested):
~~I assume the designate change is not intended~~
It seems the github client is broken and shows deprecated changes, sorry

notandy (approved):

PR #3669, opened 2022-06-23
thanos latest chart update

viennaa:
moved storegateway back into single thanos crd
use default cluster domain to avoid hard coded store targets
naming shortened to not run into not deployable services (name was too long)


richardtief (approved):
Great! Thanks.

PR #3676, opened 2022-06-24
[Cinder] Stack volumes instead of spreading out

hemna:
This patch changes the default Capacity Weigher to
stack volumes on a datastore instead of spreading them out.
This will help prevent smaller volumes from landing on the
most free datastores in a shard.  This will help keep the
newer created/most free datastores for the largest of volumes.


hemna:
tested this in qa-de-1 and volumes will stack on a single datastore until that datastore is getting full.

PR #3678, opened 2022-06-27
[Thanos] adding optional ingress and fixing values

viennaa:
values are now accessed correctly
ingress with helpers added analog to prometheus-server


richardtief (approved):
Thanks! Looks good to me.

PR #3680, opened 2022-06-27
Bump thanos and introduce regional thanos

viennaa:


system/thanos-regional/.helmignore
```diff
@@ -0,0 +1,23 @@
...
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
```
richardtief, line 23: No need to push your `.helmignore` file.

richardtief (changes requested):
All good. Just one comment.

PR #3681, opened 2022-06-27
[thanos] operator chart

viennaa:


richardtief (approved):
Looks good. Let's make the operator work for us.

PR #3684, opened 2022-06-28
prefix with thanos instead of suffix

viennaa:
components are now all starting with thanos-
matches file structure and looks cleaner


richardtief (approved):
Nice! Thanks.

PR #3687, opened 2022-06-29
[common/rabbitmq] update rabbitmq to 3.10.5-management version 

bashar-alkhateeb:
- Updating rabbitmq docker image to 3.10.5-management 
        (release note: https://github.com/rabbitmq/rabbitmq-server/releases/tag/v3.10.5).
 - using TCP socket check for the readinessProbe instead of deprecated rabbitmqctl node_health_check 
         (resone: https://github.com/rabbitmq/cluster-operator/blob/43e7c05c4e5cf47322ef11b1b98b2bb70bf32a12/internal/resource/statefulset.go#L586-L604)
 - adding CHANGELOG

### DIFF
--- a/common/rabbitmq/templates/deployment.yaml
+++ b/common/rabbitmq/templates/deployment.yaml
@@ -71,8 +71,8 @@ spec:
         {{- end }}
         {{- if .Values.readinessProbe.enabled }}
         readinessProbe:
-          exec:
-            command: ["rabbitmqctl", "node_health_check"]
+          tcpSocket:
+            port: public
           failureThreshold: {{ .Values.readinessProbe.failureThreshold }}
           initialDelaySeconds: {{ .Values.readinessProbe.initialDelaySeconds }}
           periodSeconds: {{ .Values.readinessProbe.periodSeconds }}
--- a/common/rabbitmq/CHANGELOG.md
+++ b/common/rabbitmq/CHANGELOG.md
@@ -1,0 +1,13 @@
+Rabbitmq CHANGELOG
+==============
+
+This file is used to list changes made in each version of the common chart rabbitmq.
+
+0.4.0
+-----
+b.alkhateeb@sap.com
+ - Updating rabbitmq docker image to 3.10.5-management (release note: https://github.com/rabbitmq/rabbitmq-server/releases/tag/v3.10.5).
+ - using TCP socket check for the readinessProbe instead of deprecated rabbitmqctl node_health_check (resone: https://github.com/rabbitmq/cluster-operator/blob/43e7c05c4e5cf47322ef11b1b98b2bb70bf32a12/internal/resource/statefulset.go#L586-L604)
+ - adding CHANGELOG
+
+==============
--- a/common/rabbitmq/Chart.yaml
+++ b/common/rabbitmq/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rabbitmq
-version: 0.3.3
+version: 0.4.0
 description: A Helm chart for RabbitMQ
 sources:
 - https://github.com/sapcc/helm-charts/common/rabbitmq
--- a/common/rabbitmq/templates/bin/_rabbitmq-start.tpl
+++ b/common/rabbitmq/templates/bin/_rabbitmq-start.tpl
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -exuo pipefail
+set -euo pipefail
 
 LOCKFILE=/var/lib/rabbitmq/rabbitmq-server.lock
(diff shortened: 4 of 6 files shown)


common/rabbitmq/templates/deployment.yaml
```diff
@@ -71,8 +71,8 @@ spec:
         {{- end }}
         {{- if .Values.readinessProbe.enabled }}
         readinessProbe:
-          exec:
-            command: ["rabbitmqctl", "node_health_check"]
+          tcpSocket:
+            port: {{ default "5672" .Values.ports.public }}
```
fwiesel, line 75: I would go here with the name `"public"`, as it avoids having to change it in multiple places.

fwiesel: Sure: You can refer to the port by its number and then it should be the same value as in the [port definition](https://github.com/sapcc/helm-charts/pull/3687/files/410b8b4dcc17b2ef6556a1e9dae208be30b11631#diff-41bc547432263860bb872966d794df676bc675faeac1a21bb3e7e68155c2d5b4R85-R86), or you can refer to it by the name given in the same definition.

My personal preference would be to assign a name in the port-definition and refer everywhere else in kubernetes to it by its name.

bashar-alkhateeb: OK i think got it, you mean using the port definition, which is defined in the container ports, correct ?
```
 tcpSocket:
      port: public
```

fwiesel: Exactly

fwiesel:
Another stylistic remark regarding the commit history. Personally, I would prefer if all the changes would be a single commit. For me they only make sense together.

To put it differently, for me a commit history (in a PR) is not the history of what you actually have done, but a story you tell the reviewer of how your changes fit together.

common/rabbitmq/templates/deployment.yaml
bashar-alkhateeb: Fixed!

bashar-alkhateeb:
> Another stylistic remark regarding the commit history. Personally, I would prefer if all the changes would be a single commit. For me they only make sense together.
> 
> To put it differently, for me a commit history (in a PR) is not the history of what you actually have done, but a story you tell the reviewer of how your changes fit together.

hmm i am confused a little bit now, as i thought we are creating commits per change not per feature and the commit scope between merge commits defines the feature, did i misunderstand it ?

I split normally commits to make it easier for any kind of rollback by dropping the corresponding commits in the PR, e.g if the change is fine for deployment but some one reported that it is not suitable for statefulset ..

but if you are used to create commits per feature then i am fine with the approach as well ?

fwiesel:
> hmm i am confused a little bit now, as i thought we are creating commits per change not per feature and the commit scope between merge commits defines the feature, did i misunderstand it ?

Well, we do not have a well-defined development model, so it is hard to give a definite answer here.

Personally, I am aiming for a trunk-based development model, and having long-living feature branches are not a goal, but sometimes a necessary evil.
A PR is therefor not tied to a feature, but consists of one or more changes and will eventually deliver a feature. Ideally it should be small and self-contained. So, a single commit should be fine.

Unfortunately, for various reasons a PR may result in a very large delta, that it makes it difficult to understand (unfortunately everyone has here probably a different view on when that happens). Then we need to break it down into different commits that it can be understood. We can also make different PRs out of it, but that again is taste.

notque:
Read over the upgrade notes, I'm not seeing anything relevant or concerning. 

Readiness probe seems reasonable. I'm not sure if there are any weird interactions, so I'd want to run it in QA to test a bit. 

Seems reasonable to me.

bashar-alkhateeb:
> > hmm i am confused a little bit now, as i thought we are creating commits per change not per feature and the commit scope between merge commits defines the feature, did i misunderstand it ?
> 
> Well, we do not have a well-defined development model, so it is hard to give a definite answer here.
> 
> Personally, I am aiming for a trunk-based development model, and having long-living feature branches are not a goal, but sometimes a necessary evil. A PR is therefor not tied to a feature, but consists of one or more changes and will eventually deliver a feature. Ideally it should be small and self-contained. So, a single commit should be fine.
> 
> Unfortunately, for various reasons a PR may result in a very large delta, that it makes it difficult to understand (unfortunately everyone has here probably a different view on when that happens). Then we need to break it down into different commits that it can be understood. We can also make different PRs out of it, but that again is taste.

commits squashed in one commit as recommended

bashar-alkhateeb:
@fwiesel I removed the x option from the set command in the start script to prevent showing passwords in the debugging logs 
are you fine with this change ?

fwiesel:
Sure, fine by me.

PR #3689, opened 2022-06-30
[ironic] Update utils

fwiesel:
In preparation for central rabbitmq logging and proxysql sidecar


sandzwerg:
In the meantime utils was update to `0.5.0` which should be enough I assume?

stefanhipfel:
> In the meantime utils was update to `0.5.0` which should be enough I assume?

yes can be closed!

PR #3707, opened 2022-06-30
[prometheus-server] update ingress to match thanos query service

richardtief:


viennaa (changes requested):
please bump the chart version

viennaa (approved):

PR #3709, opened 2022-07-01
[barbican] Enable proxysql-side-car for the api

fwiesel:
As the upgrade job runs as a helm-chart hook,
it cannot be combined with the unix_socket mode:
The hook runs before the required configuration change
is rolled out

This change requires the credentials for the
database user to be set under ".Values.mariadb.users..."


rajivmucheli:
commit looks good, could this branch be updated to the latest commit ? rabbitmq requirements are still on 0.3.3.

rajivmucheli (approved):
lgtm

PR #3710, opened 2022-07-01
[designate] Add proxysql side-cars

fwiesel:
This change should route all db queries through a pod-local
side-car to handle the connection resets of a restarting
mariadb

It requires the credentials in .Values.mariadb.users


galkindmitrii (approved):
lgtm!

PR #3717, opened 2022-07-04
Octobus query exporter

max-len:
As discussed we take Elastic Stack processing time into account until they scaled up.


christopherhans (approved):
Accomodates OctoBus ingestion delay and ensures audit save alerting.

PR #3730, opened 2022-07-06
[rabbitmq] Make volumeClaimTemplate labels fix

fwiesel:
The statefulset only allows the template to be changed,
but not the volumeClaimTemplate. So we hard-code it to the
old value in order to avoid having to recreate everything


bashar-alkhateeb (approved):
Looks reasonable !

PR #3731, opened 2022-07-06
[cerebro] Export additional metrics and properties

nikatza:
Notes:
- The new metrics and properties appear in the documentation manual of vRealize Operations 8.6 
- The metrics _cpu|numpackages_ and _config|hardware|num_Cpu_ appear to be equivalent, but they are currently both included (as requested in the original internal ticket).


prometheus-exporters/vrops-exporter/templates/collector-configmap.yaml
```diff
@@ -230,6 +230,8 @@ data:
         key: "sys|build"
       - metric_suffix: "config_diskSpace_bytes"
         key: "config|diskSpace"
+      - metric_suffix: "config_hyperthreading_available"
```
ofrendo, line 233:
```suggestion
      - metric_suffix: "config_hyperthread_available"
```
Would try to stick as close to the original metric as possible

richardtief (approved):
All good.

PR #3734, opened 2022-07-07
[Cinder] Fixed a problem with backup_deployment template

hemna:
This patch fixes a templating problem with the backup_deployment
template.  The template had a template inside a template which caused
rendering of the inner template to fail.


joker-at-work (commented):
Please remove all the
```
{{`
```
and
```
`}}
```
so we don't run into this ever again.

Please also update `backup_configmap.yaml`, so we don't have the old format anymore.

PR #3735, opened 2022-07-07
[keystone] Add ProxySql side-car

fwiesel:
This needs to be explicitly enabled via
proxysql.mode=unix_socket


rajivmucheli (approved):
lgtm

PR #3737, opened 2022-07-08
[cinder] Added proxysql side-cars

fwiesel:
This PR adds the necessary snippets to enable proxysql side-car pods, without enabling them by default.
Functionally, it should behave the same after the change has been merged, until the mode is activated by setting `proxysql.mode=unix_socket`

### DIFF
--- a/openstack/cinder/templates/migration-job.yaml
+++ b/openstack/cinder/templates/migration-job.yaml
@@ -1,7 +1,7 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: cinder-migration-{{.Values.imageVersion | required "Please set cinder.imageVersion or similar"}}
+  name: cinder-migration-{{.Values.imageVersion | required "Please set cinder.imageVersion or similar"}}{{- if .Values.proxysql.mode}}-{{ .Values.proxysql.mode | replace "_" "-" }}{{ end }}
   labels:
     system: openstack
     type: configuration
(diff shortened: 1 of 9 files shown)


fwiesel:
Based on top of  #3734

openstack/cinder/templates/migration-job.yaml
```diff
@@ -10,15 +10,16 @@ spec:
...
+{{ include "utils.proxysql.job_pod_settings" . | indent 6 }}
+      initContainers:
+        - name: dependencies
           image: {{required ".Values.global.registry is missing" .Values.global.registry}}/loci-cinder:{{.Values.imageVersionCinderApi | default .Values.imageVersion | required "Please set cinder.imageVersion or similar" }}
           imagePullPolicy: IfNotPresent
           command:
             - kubernetes-entrypoint
           env:
             - name: COMMAND
-              value: "cinder-manage db sync"
```
hemna, line 21: how is the migration/upgrade from one version to another going to work without this call to the command that does the db sync/upgrade ?

fwiesel: It is still called. I just split the container into two.
You are seeing just the init-container, which calls the `kubernetes-entrypoint` with command true.
The actual command is still executed in the normal container [here](https://github.com/sapcc/helm-charts/pull/3737/files/35c83eac79b1fd1c2eb2aff79c0b921e9f954eab#diff-4220c1b8749757535904420c740ccaad6f1aec9cd9a5c750f24773cc3cc4302eR39-R40) without kubernetes-entrypoint.

The main motivation was that this way I can easily introduce a shell script with the termination hook without having to create a configmap with a script in it.

It also has the benefit that aiting for the dependencies won't show up as restarts of the actual pod, but as an init-retry.

hemna (approved):

PR #3738, opened 2022-07-08
[ironic] Add proxysql side-cars snippets

fwiesel:
Based on #3689

### DIFF
--- a/openstack/ironic/templates/etc/_ironic.conf.tpl
+++ b/openstack/ironic/templates/etc/_ironic.conf.tpl
@@ -52,7 +52,7 @@ dhcp_provider = neutron
 [api]
 host_ip = 0.0.0.0
 public_endpoint = https://{{ include "ironic_api_endpoint_host_public" .}}
-api_workers = {{ .Values.api.api_workers | default 8}}
+api_workers = {{ .Values.api.api_workers }}
 
 [database]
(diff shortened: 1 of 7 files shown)


openstack/ironic/templates/etc/_ironic.conf.tpl
```diff
@@ -52,7 +52,7 @@ dhcp_provider = neutron
 [api]
 host_ip = 0.0.0.0
 public_endpoint = https://{{ include "ironic_api_endpoint_host_public" .}}
-api_workers = {{ .Values.api.api_workers | default 8}}
```
stefanhipfel, line 55: is there a specific reason the default has been removed? 
In the past the default was: workers = cpu cores (which lead to very high memory usage)
Now default is max 4 or just 1 if the numbers of cpu cannot be determined.

fwiesel: The default here is (semi-)redundant with the value [set to 4 in values.yaml](https://github.com/sapcc/helm-charts/blob/master/openstack/ironic/values.yaml#L121).
So, this default will only apply, if someone sets the value explicitly to a falsy value.

sandzwerg:
You need to rebase, but after that I think you can merge.

PR #3745, opened 2022-07-11
[glance] use new objectstore_admin role

reimannf:
https://documentation.global.cloud.sap/blog/objectstorage-new-roles


reimannf:
Unsure what you referring regarding Elektra, but this has been updated: https://github.com/sapcc/elektra/blob/master/plugins/compute/app/views/compute/instances/new_snapshot.html.haml#L31

PR #3751, opened 2022-07-11
[openstack/utils] Add snippet for CORS annotations

joker-at-work:
Our dashboard wants to start talking to OpenStack services' APIs
directly from the user's browser. Since browsers don't allow
cross-origin requests without the server explicitly setting certain
headers, we configure our Nginx ingress to enable CORS headers and
configure them.

We use k8s annotations [0] and Nginx ingress instead of the
cors-middleware from oslo.middleware, because not having to call the
backend seems much more efficient.

[0] https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#enable-cors

### DIFF
--- a/openstack/utils/templates/snippets/_set_ingress_cors_annotations.tpl
+++ b/openstack/utils/templates/snippets/_set_ingress_cors_annotations.tpl
@@ -1,0 +1,10 @@
+{{ define "utils.snippets.set_ingress_cors_annotations" }}
+{{- if .Values.cors.enabled }}
+ingress.kubernetes.io/enable-cors: "true"
+ingress.kubernetes.io/cors-allow-headers: "{{ .Values.utils.cors.allow_headers }}{{ if .Values.cors.additional_allow_headers }},{{ .Values.cors.additional_allow_headers }}{{ end }}"
+ingress.kubernetes.io/cors-expose-headers: "{{ .Values.cors.expose_headers | default "*" }}"
+ingress.kubernetes.io/cors-allow-origin: "{{ required ".Values.utils.cors.allow_origin is missing" .Values.utils.cors.allow_origin }}{{if .Values.cors.additional_allow_origin }},{{ .Values.cors.additional_allow_origin }}{{ end }}"
+ingress.kubernetes.io/cors-allow-credentials: "{{ .Values.cors.allow_credentials | default "false" }}"
+ingress.kubernetes.io/cors-max-age: "{{ .Values.cors.max_age | default "1728000" }}"
+{{- end }}
+{{- end }}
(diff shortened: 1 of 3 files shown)


openstack/utils/templates/snippets/_set_ingress_cors_annotations.tpl
```diff
@@ -0,0 +1,10 @@
+{{ define "utils.snippets.set_ingress_cors_annotations" }}
+{{- if .Values.cors.enabled }}
```
fwiesel, line 2: I would suggest to change this to 
```
{{- if and .Values.cors.enabled .Values.utils.cors.allow_origin }}
```
as I think it probably doesn't make much sense to set the annotations without `allow_origin` set

openstack/utils/templates/snippets/_set_ingress_cors_annotations.tpl
```diff
@@ -0,0 +1,10 @@
...
+ingress.kubernetes.io/enable-cors: "true"
+ingress.kubernetes.io/cors-allow-headers: "{{ .Values.utils.cors.allow_headers }}{{ if .Values.cors.additional_allow_headers }},{{ .Values.cors.additional_allow_headers }}{{ end }}"
+ingress.kubernetes.io/cors-expose-headers: "{{ .Values.cors.expose_headers | default "*" }}"
+ingress.kubernetes.io/cors-allow-origin: "{{ .Values.utils.cors.allow_origin }}{{if .Values.cors.additional_allow_origin }},{{ .Values.cors.additional_allow_origin }}{{ end }}"
```
fwiesel, line 6: Since we set now the `allow_origin` variable, I would remove `additional_allow_origin`. While it doesn't hurt to have additional origins, I am a fan of having only one way to achieve the desired result to keep the variance smaller.

fwiesel (approved):

openstack/utils/templates/snippets/_set_ingress_cors_annotations.tpl
joker-at-work: The point with `additional_allow_origin` was that a service might want more than the default. To achieve this, the service would have to duplicate the default into its own values files, which makes it harder to update the elektra development origins. Therefore, I wanted to provide a way to specify more for the parent chart like we do with `additional_allow_headers`.

openstack/utils/templates/snippets/_set_ingress_cors_annotations.tpl
joker-at-work: I think that's confusing to the user. The user would expect CORS to be enabled, but it isn't because of a misconfiguration (on their side?). Instead, I'd rather add a `required` to using the `allow_origin` - what do you think?

fwiesel: True, that sounds better.

PR #3758, opened 2022-07-12
owner-info: add "support-group" and "service" fields

majewsky:
To support the introduction of our new support structure.

@onuryilmaz I'm including you as reviewer since you will likely be the one consuming these fields. @BugRoger can tell you more when you talk next time.


majewsky:
I was holding off on this PR because I was waiting for a decision which support groups are going to be set up. In my call today with @BugRoger, we decided that a central decision on the set of support groups is unlikely, so we will move ahead with preparing infrastructure like this and leaving it to the teams to decide how they will structure their support groups.

PR #3759, opened 2022-07-12
[prometheus] use new objectstore_admin role

viennaa:
Introducing the use of the new swift user before 7.0.0 change


artherd42:
Required to adjust with new roles for swift -> https://documentation.global.cloud.sap/blog/objectstorage-new-roles

auhlig (approved):

PR #3763, opened 2022-07-14
Consult cc/support-group and cc/service pods labels for pod alerts

databus23:
I propose that we use `cc/support-group` and `cc/service` label keys when annotating kubernetes resources to avoid any clashes with existing (pod) labels.
Those labels can then be used in alert definitions to populate `support_group` and `service` labels in alerts that can be used for alert routingin alertmanager

I adapted the generic alerts to support old `alert-service` and `alert-tier` labels as well with preference for the newer label keys.

Wdyt?


majewsky (approved):
Given the backwards compatibility, this looks safe to me. Label names are good.

databus23:
rebased on master

PR #3775, opened 2022-07-14
[neutron] Update rabbitmq-chart

fwiesel:


sebageek:
Superseeded by https://github.com/sapcc/helm-charts/pull/4619

PR #3780, opened 2022-07-14
[ironic] Update rabbitmq-chart

fwiesel:


stefanhipfel:
already done in other pr

PR #3781, opened 2022-07-14
[hermes] Update rabbitmq-chart

fwiesel:


notque (approved):
LGTM, merging and deploying in qa to test.

PR #3791, opened 2022-07-18
[neutron] Add proxysql sidecars

fwiesel:


fwiesel:
If I understanding it correctly, the agents do not connect to the database, but use RPC calls to the neutron-server/rpc instances.

At least when doing a short test, the agents were not raising any error, despite me setting the mode to unix sockets, which where not available to them.

sebageek (approved):
Let's try it

PR #3793, opened 2022-07-19
fix false-positive multiple scrape alerts

artherd42:


richardtief:
Thank you for this PR. It is not necessary to include this in every configuration line. Once per scrape job is enough to filter all time series. Like this: 


```
- job_name: 'foo'
  params:
     'match[]':
       - '{__name__!~"^(up|ALERTS.*|scrape.+)"}'
```

IvoGoman (approved):
Looks good.

richardtief:
> > Thank you for this PR. It is not necessary to include this in every configuration line. Once per scrape job is enough to filter all time series. Like this:
> > ```
> > - job_name: 'foo'
> >   params:
> >      'match[]':
> >        - '{__name__!~"^(up|ALERTS.*|scrape.+)"}'
> > ```
> 
> That will match every metric which does not match the regex, won't it? I believe we need to be this restrictive to avoid scraping too much.

Yeah, probably. Not sure if there is also a issue with `vector selector must contain at least one non-empty matcher`. Ok, then it is good to go.

PR #3808, opened 2022-07-19
[kos-operator] Switch to Chart v2 format

fwiesel:


stefanhipfel:
already migrated to v2

PR #3812, opened 2022-07-19
[designate] Switch to Chart v2 format and fix rabbitmq deps

fwiesel:
Chart v1 format is deprecated with Helm 3, and we are only running
helm 3 now.

Also the requirements.lock was not properly updated with the
switch to the rabbitmq chart 0.4.2


fwiesel:
Sorry, the removal of the lock was unintentional, I've added it again.

The point is consistency across the charts, and getting rid of the deprecation warning in helm3.
Consistency eases doing changes if you have to touch more than one chart.

galkindmitrii:
Thanks!

galkindmitrii (approved):

PR #3815, opened 2022-07-19
[cinder] Switch to Chart v2 format

fwiesel:


hemna (approved):
Looks ok to me, but I don't know helm.  I presume helm3 can handle this?

fwiesel:
Yes, v1 is for helm2, helm3 handles both v1 and v2, but issues a warning for v1.

PR #3826, opened 2022-07-19
Private flavors for HXM SQL Servers

frr2018:
Just to let you know. There was no way around it.


joker-at-work (approved):
I want to stress that I don't like this at all.

PR #3827, opened 2022-07-19
[placement] Stamp initial db version if not stamped

fwiesel:
This corresponds to the placement db stamp step
in the migration step: https://github.com/openstack/placement/blob/master/placement_db_tools/mysql-migrate-db.sh#L302

The stamp job won't overwrite existing stamps,
so it can be removed after the first 'proper' deployment

It won't do that automatically by looking at the version
in order to allow to combine this change with an upgrade
in a later change

### DIFF
--- a/openstack/placement/templates/stamp-alembic-job.yaml
+++ b/openstack/placement/templates/stamp-alembic-job.yaml
@@ -1,0 +1,33 @@
+{{- if and (not .Values.mariadb.enabled) .Values.api_db.initial_stamp -}}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "placement-stamp-alembic"
+  labels:
+{{ tuple . "placement" "stamp-alembic" | include "helm-toolkit.snippets.kubernetes_metadata_labels" | indent 4 }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
+spec:
+  template:
+    metadata:
+      name: "placement-stamp-alembic"
+      labels:
+{{ tuple . "placement" "stamp-alembic" | include "helm-toolkit.snippets.kubernetes_metadata_labels" | indent 8 }}
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: stamp
+        image: {{ required ".Values.global.dockerHubMirror is missing" .Values.global.dockerHubMirror }}/{{ .Values.mariadb.image | default "library/mariadb:10.5.12" }}
+        command:
+        - /usr/bin/env
+        - bash
+        - -c
+        - |
+          mariadb -v -hnova-api-mariadb -u{{ .Values.api_db.user }} --password='{{ .Values.api_db.password }}' '{{ .Values.api_db.name }}' -e '
+          CREATE TABLE IF NOT EXISTS alembic_version (version_num varchar(32) NOT NULL, PRIMARY KEY (version_num));
+          INSERT INTO alembic_version (version_num)
+              SELECT "b4ed3a175331" AS version_num
+              WHERE NOT EXISTS (SELECT * FROM alembic_version);
+          '
+{{- end }}
--- a/openstack/placement/ci/test-values.yaml
+++ b/openstack/placement/ci/test-values.yaml
@@ -2,7 +2,8 @@ global:
   region: regionOne
   domain: evil.corp
   registry: keppel.regionOne.cloud
-  dockerHubMirrorAlternateRegion: dockerhubmirror.example.com
+  dockerHubMirror: dockerhubmirror.example.com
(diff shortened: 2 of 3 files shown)


openstack/placement/templates/stamp-alembic-job.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if and (not .Values.mariadb.enabled) .Values.api_db.initial_stamp -}}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "placement-stamp-alembic"
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    app.kubernetes.io/version: {{ default "unknown" .Chart.AppVersion }}
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
```
joker-at-work, line 10: What do these do? I don't remember seeing them being set explicitly before.

And why don't we have any of the standard labels we usually set e.g.
```
  labels:
    system: openstack
    type: api
    component: placement
```

openstack/placement/templates/stamp-alembic-job.yaml
```diff
@@ -0,0 +1,38 @@
...
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
+spec:
+  template:
+    metadata:
+      name: "placement-stamp-alembic"
+      labels:
+        app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
+        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
```
joker-at-work, line 21: Why don't we add any of the standard labels here like for the api with `{{ tuple . "placement" "api" | include "helm-toolkit.snippets.kubernetes_metadata_labels" | indent 8 }}`?

openstack/placement/templates/stamp-alembic-job.yaml
```diff
@@ -0,0 +1,38 @@
...
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: stamp
+        image: {{ required ".Values.global.dockerHubMirrorAlternateRegion is missing" .Values.global.dockerHubMirrorAlternateRegion }}/{{ .Values.mariadb.image | default "library/mariadb:10.5.12" }}
```
joker-at-work, line 26: Why does this come from an the alternate region? Is that by choice or copy&paste like we had before?

openstack/placement/templates/stamp-alembic-job.yaml
```diff
@@ -0,0 +1,38 @@
...
+        command:
+        - /usr/bin/env
+        - bash
+        - -c
+        - |
+          mariadb -v -hnova-api-mariadb -u{{ .Values.api_db.user }} --password='{{ .Values.api_db.password }}' '{{ .Values.api_db.name }}' -e '
```
joker-at-work, line 32: `nova-api-mariadb` is hardcoded, because we only need to run this once and don't know how/when if ever we split off placement from there, right?

joker-at-work (commented):
I'm a little confused by the whole labeling thing. Actual code to be run looks fine though.

openstack/placement/templates/stamp-alembic-job.yaml
fwiesel: Exactly, it is because it is a one-time thing. Any new deployment will (hopefully) be on a newer version of placement anyway.

openstack/placement/templates/stamp-alembic-job.yaml
fwiesel: Good point, I'll change it

openstack/placement/templates/stamp-alembic-job.yaml
fwiesel: No, not by conscious choice, it is from the source.

openstack/placement/templates/stamp-alembic-job.yaml
fwiesel: > What do these do? I don't remember seeing them being set explicitly before.

As I understand it, they are the [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels). 
The labels you mention are unfortunately not uniformly set across the services, so I abandoned them in favour of the external standard. Or to me more correct: I used a file as template, where someone else has taken that decision and I just went with it.

Also, one note: Technically, it isn't part of the helm-chart installation, so it raises the question, if those labels even apply.
But also, since it is not part of the installation, I consider that question rather low priority.

joker-at-work (approved):

Issue #3762, opened 2022-07-13
[utils/shared] proxysql.mode=somewrongvalue does not fail


fwiesel:
The chart should not accept wrong values but fail with a clear error.

PR #3829, opened 2022-07-20
[utils] ProxySql side-car cleanup in prep for placement

fwiesel:
- A couple of macros were duplicated, they got removed.
- The check for valid modes (unix_socket, host_alias) was fixed
- The version of proxysql gets a patch version update
- Removed the check for Dependency.Enabled, as disabled ones do not
  appear among the dependencies
- The proxysql setting can now point with proxysql.force_enable
  to a list of database settings which are disabled.
  This is needed for proxysql as long as it runs
  with the nova-api database


fwiesel:
Fixes #3762

openstack/utils/templates/_proxysql.tpl
```diff
@@ -34,7 +34,7 @@
 {{- end }}
 
 {{- define "utils.proxysql.volumes" }}
-  {{- if .Values.proxysql }}
+  {{- if .Values.proxysql -}}
```
joker-at-work, line 37: Is this change necessary? I would have assumed the whitespace is consumed by the next `if` anyways.

joker-at-work (approved):

openstack/utils/templates/_proxysql.tpl
fwiesel: Ah, that one not... I've removed it

PR #3832, opened 2022-07-21
[Prometheus-server] enabling global Thanos structure

viennaa:
* Enabled Thanos sidecar by default. Look at templates/thanos/README.md for instructions, swiftStorageConfig needs to be set
* Removed deployment charts for Query, Compactor, Store.
* Removed Thanos all thanos related alerts, leaving sidecar in place
* incorporated tier and service template of prometheus-kubernetes-rules to alerts, since it is used by ALL Prometheis

[This change](https://github.com/sapcc/helm-charts/search?q=alertservicelabelordefault) from @databus23 needed to go in too, otherwise other Prometheis wouldn't work, since they neither know `alertServiceLabelOrDefault` nor `alertTierLabelOrDefault`


richardtief (changes requested):
Finally, we are putting this together. I have seen two little improvements.

auhlig:
Can you adjust the [test values](https://github.com/sapcc/helm-charts/blob/73db3714aaa167d5fb8f34ae7e3c757f894fd5bc/common/prometheus-server/ci/test-values.yaml) making the ci pass?

auhlig (approved):

PR #3835, opened 2022-07-21
[region_check] Delete before creation too

fwiesel:
If another hook fails, the region check won't get deleted
despite having succeeded, despite the hook-succeeded annotation.
A subsequent deployment will then fail due to the existing
region check job. Adding 'before-hook-creation' takes care of that.


galkindmitrii (approved):
seen that happening, thx!

PR #3837, opened 2022-07-21
[region_check] Use helm lookup instead of python job

fwiesel:
Helm provides a lookup function to query
the kubernetes api. That only works
in non-dry runs, but that also applies to the job.

Doing it with the helm functions eliminates
the need to spin up a pod, which is faster
and also doesn't require an image.


fwiesel:
This may be a bit more contentious than the previous PR (#3835).
While  looking at it, I thought we could do away with all the spinning up of a pod to query the api, considering that functionality does exist in the helm client.

majewsky:
I have a task in my backlog to prevent misplaced deployments via Gatekeeper policy. I'll push that up in prio and try to get it implemented to save you guys extra work.

fwiesel:
Implemented as policy

PR #3842, opened 2022-07-22
[rabbitmq] create cluster by using operator

bashar-alkhateeb:
**Draft to be used for further testing & to clarify the concept of the first option of using the operator in clustered mode**

- adding metrics sidecar to the cluster pods
- create the needed users using the topology operator
- activate metrics sidecar
- i didn't find a way to override the default-container for that override rabbitmq container at container[0] to be the default one


notandy:
Sorry but why we using the deprecated metrics sidecontainer instead the inbuild upstream prometheus metrics?

common/rabbitmq_cluster/templates/cluster.yaml
```diff
@@ -76,3 +66,34 @@ spec:
               labelSelector:
                 matchLabels:
                   app.kubernetes.io/name: {{ template "fullname" . }}
+            containers:
+            - name : rabbitmq
```
notandy, line 70: whats that entry for?

bashar-alkhateeb: if we didn't add this list item the metrics container will be the default one, which is not the current setup
 i thought that may confuse debugging in case of direct access is need to the containers
unfortunatly we cannot override default-container option as it is not supported in this resource type

bashar-alkhateeb:
> Sorry but why we using the deprecated metrics sidecontainer instead the inbuild upstream prometheus metrics?

a very good question! 
i only used the same image that was used for the rabbitmq chart, until we evaluate the imbedded prometheus which rabbitmq cluster offers (lets hope that we will get ride of the the side car container)
but for now I will try to test v0.29.0 and update this PR

bashar-alkhateeb:
metrics image updated to v0.29.0

fwiesel:
Hi, you added ha-policies for reply queues and fan-outs, but I fear that it might be contra-productive. oslo_messaging limits the ha features to rpc queues and I assume that is not an oversight but conscious decision.
If you see it differently, it would be good to add the reasons to the commit message.

bashar-alkhateeb:
Hi fabian ,
when using quorum queues only requests queues are clustered which showed in my tests message lost in case of any node failure 
we had the same tests in HA mode where all queues are set to be HA .. and we didnt have this issue in case of node failure

```
Quorum Clustered Queues:
=================================== summary ===================================
client: 
duration: 1527.31 count: 600000 (392.8 msg/sec) bytes: 1988019600 (1301650 bps)

**error:** 
**duration: 913.59 count: 16 (0.0 msg/sec) bytes: 55859 (61 bps)**

round-trip: 
duration: 1526.30 count: 599984 (393.1 msg/sec) bytes: 1987963741 (1302470 bps) latency: 0.014 min: 0.005 max: 8.112
```


```
Classic Clustered Queues:
=================================== summary ===================================
client: 
duration: 1449.05 count: 600000 (414.1 msg/sec) bytes: 1938040200 (1337452 bps)

**error: 
duration: 0.00 count: 0 (0.0 msg/sec) bytes: 0 (0 bps)**

round-trip:
duration: 1448.07 count: 600000 (414.3 msg/sec) bytes: 1938040200 (1338358 bps) latency: 0.013 min: 0.004 max: 1.0
```

for that i am working on building a kind of hybrid mode solution as a third option.

in all cases I will present my results in a small round once i finished the split brain tests then we can discuss which solution to use.

fwiesel:
My point here is a bit, that each choice is a trade-off and the option does exists for certain queues, while it does not exists for others. I am assuming that not providing that option for fan-out and reply queues is done with a reason.
Presumably the people implementing oslo_messaging are aware of those options, and *chose* not even to offer the option.
Ideally, I would like you to find out, why isn't that option provided by openstack in the first place.

You are focusing on the metric of message loss, and HA does indeed improves that, but my concern is that the other queues have different priorities, as they might be robust against message loss from the application side, so other metrics (throughput/latency) may be more important here.

bashar-alkhateeb:
I agree with you, and i'll try to find out the reason of not configuring the reply and fanout queues as classic queues when setting oslo.messeging to use quorum queues 

for the other metrics which we need to measure, they are already covered in the tests we are performing , but surprisingly using quorum queues showed a degraded performance than classic HA depending on the resources limits we assign to the pods, with 1 GB RAM limit , we faced up to 15% performance degradation when using quorum queues for requests queues only, that  for RPC calls.
and between 2% to 5% when repeating the same experiment but with using limits of 2 GB 
(still need to test for hybrid setup but i don't expect performance enhancements than pure quorum )

the reason as i could find was putting the connections in blocked mode until freeing up some memory, because quorum queues are in-memory queues and the simulator was flooding the rabbitmq cluster with messages (which is maybe not a real-world use-case, and maybe it is better to test the cluster with real-world like utilization), please find bellow the test results .

**RPC call  without any node failure, 100.000 messages per 6 parallel threads, 3 replicas cluster** 
```
limits:
   cpu: 5000m 
   memory: 1024Mi  
   

Classic Clustered Queues:
=================================== summary ===================================
client: 
duration: 1488.40 count: 600000 (403.1 msg/sec) bytes: 1949328000 (1309677 bps)

error: 
duration: 0.00 count: 0 (0.0 msg/sec) bytes: 0 (0 bps)

round-trip:
duration: 1487.11 count: 600000 (403.5 msg/sec) bytes: 1949328000 (1310817 bps) latency: 0.013 min: 0.004 max: 0.236


Quorum Clustered Queues:
=================================== summary ===================================
client:
duration: 1703.01 count: 600000 (352.3 msg/sec) bytes: 1994831400 (1171358 bps)

error:
duration: 0.00 count: 0 (0.0 msg/sec) bytes: 0 (0 bps)

round-trip:
duration: 1702.18 count: 600000 (352.5 msg/sec) bytes: 1994831400 (1171929 bps) latency: 0.015 min: 0.005 max: 1.081   
```

PR #3843, opened 2022-07-22
[mariadb] Handle unset dbUser

fwiesel:
dbUser may be unset, especially if the mariadb.users
type of initialisation is used

### DIFF
--- a/common/mariadb/Chart.yaml
+++ b/common/mariadb/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: mariadb
-version: 0.5.3
+version: 0.5.4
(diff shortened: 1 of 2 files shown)


joker-at-work (commented):
Is that really necessary? If I render Cinder currently while removing the `global.dbUser`, it also doesn't render this block:
```
data:
  init.sql: |
    CREATE DATABASE IF NOT EXISTS cinder CHARACTER SET utf8 COLLATE utf8_general_ci;
    GRANT ALL PRIVILEGES ON cinder.* TO @localhost IDENTIFIED BY '_REPLACED_VAULT_REFERENCE';
    GRANT ALL PRIVILEGES ON cinder.* TO @'%' IDENTIFIED BY '_REPLACED_VAULT_REFERENCE';
```

fwiesel:
I think, you have the wrong version of the mariadb chart. The superfluous grants for localhost have been removed some time ago.
My best guess is, you need to delete the charts folder in cinder.
And while you are at it, also the superfluous `openstack/cinder/templates/db-init-configmap.yaml`.

joker-at-work (approved):
You're right, I forgot to update the dependencies before rendering.

PR #3844, opened 2022-07-22
[openstack/utils] Use first dbname & user as default

fwiesel:
All services where setting a dbUser and dbName in the current
chart. That works for a single instance, but not in general.
Instead use the logic that the first database listed
under mariadb.databases is the default database,
and that there is user with the same key (but not necessarily
the same name) as the database.

### DIFF
--- a/openstack/utils/templates/_hosts.tpl
+++ b/openstack/utils/templates/_hosts.tpl
@@ -23,7 +23,13 @@ postgresql+psycopg2://{{$user}}:{{$password | urlquery}}@{{.Chart.Name}}-postgre
 
 {{- define "db_credentials" }}
     {{- if kindIs "map" . }}
-        {{- coalesce .Values.dbUser .Values.global.dbUser "root" }}:{{ coalesce .Values.dbPassword .Values.global.dbPassword .Values.mariadb.root_password | required ".Values.mariadb.root_password is required!" | urlquery }}
+        {{- if and .Values.mariadb.users .Values.mariadb.databases }}
+            {{- $db := first .Values.mariadb.databases }}
+            {{- $user := get .Values.mariadb.users $db | required (printf ".Values.mariadb.%v.name & .password are required (key comes from first database in .Values.mariadb.databases)" $db) }}
+            {{- $user.name | required (printf ".Values.mariadb.%v.name is required!" $db ) }}:{{ $user.password | required (printf ".Values.mariadb.%v.password is required!" $db ) }}
+        {{- else }}
+            {{- coalesce .Values.dbUser .Values.global.dbUser "root" }}:{{ coalesce .Values.dbPassword .Values.global.dbPassword .Values.mariadb.root_password | required ".Values.mariadb.root_password is required!" | urlquery }}
+        {{- end }}
     {{- else }}
         {{- $user := index . 2 }}
         {{- $password := index . 3 }}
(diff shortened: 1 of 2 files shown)


openstack/utils/templates/_hosts.tpl
```diff
@@ -46,7 +52,13 @@ postgresql+psycopg2://{{$user}}:{{$password | urlquery}}@{{.Chart.Name}}-postgre
 
 {{- define "db_url_mysql" }}
     {{- if kindIs "map" . }}
-        {{- tuple . (coalesce .Values.dbName .Values.db_name) (coalesce .Values.dbUser .Values.global.dbUser "root") (coalesce .Values.dbPassword .Values.global.dbPassword .Values.mariadb.root_password | required ".Values.mariadb.root_password is required!") .Values.mariadb.name | include "db_url_mysql" }}
+        {{- if and .Values.mariadb.users .Values.mariadb.databases }}
+            {{- $db := first .Values.mariadb.databases }}
+            {{- $user := get .Values.mariadb.users $db | required (print "User with key " $db " required") }}
+            {{- list . $db $user.name (required (printf "User with key %v requires password" $db) $user.password) | include "db_url_mysql" }}
```
joker-at-work, line 58: Why do we switch from `tuple` to `list` in this call vs. the previous/else call?

openstack/utils/templates/_hosts.tpl
```diff
@@ -46,7 +52,13 @@ postgresql+psycopg2://{{$user}}:{{$password | urlquery}}@{{.Chart.Name}}-postgre
 
 {{- define "db_url_mysql" }}
     {{- if kindIs "map" . }}
-        {{- tuple . (coalesce .Values.dbName .Values.db_name) (coalesce .Values.dbUser .Values.global.dbUser "root") (coalesce .Values.dbPassword .Values.global.dbPassword .Values.mariadb.root_password | required ".Values.mariadb.root_password is required!") .Values.mariadb.name | include "db_url_mysql" }}
+        {{- if and .Values.mariadb.users .Values.mariadb.databases }}
+            {{- $db := first .Values.mariadb.databases }}
+            {{- $user := get .Values.mariadb.users $db | required (print "User with key " $db " required") }}
```
joker-at-work, line 57: Can we align the help-text with the one above? `"User with key of the first database in .Values.mariadb.databases (" $db ") required"`

openstack/utils/templates/_hosts.tpl
```diff
@@ -23,7 +23,13 @@ postgresql+psycopg2://{{$user}}:{{$password | urlquery}}@{{.Chart.Name}}-postgre
 
 {{- define "db_credentials" }}
     {{- if kindIs "map" . }}
-        {{- coalesce .Values.dbUser .Values.global.dbUser "root" }}:{{ coalesce .Values.dbPassword .Values.global.dbPassword .Values.mariadb.root_password | required ".Values.mariadb.root_password is required!" | urlquery }}
+        {{- if and .Values.mariadb.users .Values.mariadb.databases }}
+            {{- $db := first .Values.mariadb.databases }}
+            {{- $user := get .Values.mariadb.users $db | required (print "User with key of the first database in .Values.mariadb.databases (" $db ") required") }}
+            {{- $user.name | required (print ".Values.mariadb.%v.name is required!" $db ) }}:{{ $user.password | required (print ".Values.mariadb.%v.password is required!" $db ) }}
```
joker-at-work, line 29: I think this need to be `printf` instead of `print`.

joker-at-work (changes requested):

openstack/utils/templates/_hosts.tpl
fwiesel: No particular reason. I had lists in my previous change... I'll change it to tuple for consistency

openstack/utils/templates/_hosts.tpl
fwiesel: True, fixed it.

openstack/utils/templates/_hosts.tpl
fwiesel: I am not quite clear what you mean with align?

The second one is simply a missing field, and we usually do something like `.xyz is required/missing`, which I think is sufficient.
But the first one has to explain the interdependency of the values. I tried to reformulate it... hopefully that fits the request.

joker-at-work: ok, let me explain `"User with key of the first database in .Values.mariadb.databases (" $db ") required"` in the first `get .Values.mariadb.users $db` on line 28 is more verbose than `"User with key " $db " required"` on line 57. Both take the user automatically after looking at the first value in `.databases`. I just wanted the `"User with key " $db " required"` to be `"User with key of the first database in .Values.mariadb.databases (" $db ") required"`, because to me the additional explanation of how we come to needing that specific user is a good hint for understanding it all.

fwiesel: Got it, they are now the same.

joker-at-work (approved):

PR #3845, opened 2022-07-22
[nova] remove custom-initdb configmaps

fwiesel:


joker-at-work (approved):
Looking at the rendered output for qa, this seems fine, as long as placement sets its own user/grants while using nova-api DB.

PR #3852, opened 2022-07-26
[cinder] Integrate locking via pvc

fwiesel:
to restarts of the nodes

### DIFF
--- a/openstack/cinder/Chart.lock
+++ b/openstack/cinder/Chart.lock
@@ -1,7 +1,7 @@
 dependencies:
 - name: utils
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.4.2
+  version: 0.7.0
 - name: mariadb
   repository: https://charts.eu-de-2.cloud.sap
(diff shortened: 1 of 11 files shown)


fwiesel:
Requires #3851 to be merged in order to update the chart dependencies, otherwise good to go

fwiesel:
Depends on #3851

hemna:
Can you explain this a bit more?   I'm not sure I understand the change away from memcached to a file ?

fwiesel:
I've now added an explanation to #3851, as the PRs for the other services may raise the same question.

fwiesel:
By default, this change should only differ in the proxysql version (if that is enabled, as in qa). Otherwise, the diff should be empty.
It will allow us to enable taking locks on a share instead in memcached, so a restart of memached would not result in locks getting lost.

fwiesel:
Merged via 3c9dc7d89557eb0bc829fa040ecbf47105e794e0

PR #3853, opened 2022-07-26
Gatekeeper owner info prefix

SuperSandro2000:


majewsky (approved):
Instead of giving you another round of notes, I've done a bunch of changes myself in new commits. Can you please counter-review that my changes make sense to you?

As you can see from the added FIXME note, there is still some work to be done on GkPodLabels, but I'd like to not overextend the scope of this PR. Let's get this stuff merged first.

SuperSandro2000:
LGTM

majewsky:
Note to self: Before bringing this into QA, take a backup of the existing doop-agent reports to check the differences between before and after these changes.

PR #3855, opened 2022-07-27
[glance] Remove notification rabbitmq

fwiesel:
We migrated to a central one for notifications,
so we might as well remove the integrated one


rajivmucheli (approved):
lgtm

PR #3859, opened 2022-07-27
[keystone] Remove notification rabbitmq

fwiesel:
We migrated to a central one for notifications,
so we might as well remove the integrated one


rajivmucheli:
rabbitmq pods seem to be already removed,

FYI : @bbobrov

PR #3860, opened 2022-07-27
[ironic] Remove superflous coordination section

fwiesel:
to restarts of the nodes


fwiesel:
Only draft, as the proper work only makes sense after merging all the other PRs

fwiesel:
Depends on #3851

stefanhipfel (approved):
hm just wondering why the inspector needs locking and not the conductors

fwiesel:
I just blindly replaced the configuration... but good point. I almost suspect it is a superfluous configuration.

fwiesel:
It is... maybe it was used once, but now it does locking via the database and uses a hashring for coordination.

PR #3862, opened 2022-07-27
Enable osprofiler for manila

misamoylov:


openstack/manila/templates/api-deployment.yaml
```diff
@@ -190,6 +190,7 @@ spec:
           resources:
             {{- toYaml .Values.pod.resources.api_statsd | trim | nindent 12 }}
           {{- end }}
+ {{- include "jaeger_agent_sidecar" . | indent 8 }}
```
galkindmitrii, line 193: you don't want to include it in all regions. This needs to be wrapped in in/end condition.

openstack/manila/templates/etc/_manila.conf.tpl
```diff
@@ -120,3 +120,5 @@ service_type = sharev2
 {{- if .Values.memcached.enabled }}
 {{- include "ini_sections.cache" . }}
 {{- end }}
+
+{{- include "osprofiler" . }}
```
galkindmitrii, line 124: Same, only add it if osprofiler is enabled in the region.

galkindmitrii (changes requested):

galkindmitrii (approved):

PR #3863, opened 2022-07-28
[manila] Integrate locking via pvc

fwiesel:
to restarts of the nodes


fwiesel:
Depends on #3851

Carthaca (approved):
Thanks for doing the integration work!

I double checked the other deployments: manila-share is using coordination, too. But is configured to use the file driver, since we only run one replica.
And our manila ensure does not require coordination.

PR #3866, opened 2022-07-28
[neutron] Integrate locking via pvc

fwiesel:
to restarts of the nodes

This is only needed for the arista driver


fwiesel:
Depends on #3851

fwiesel:
The default setting is memcached, so it won't change anything unless set explicitly.

fwiesel:
Now we have nfs v4.1 in all regions, so no manual lock breaking necessary.
The only change this PR will have unless the coordination backend is manually changed, is upgrading proxysql, which is only enabled in qa.

sebageek (approved):

PR #3880, opened 2022-08-03
[Domain-Seeds] disable project seeding of a particular domain.

rajivmucheli:
disables projects seeding in a particular region


majewsky (approved):

openstack/domain-seeds/values.yaml
```diff
@@ -35,3 +35,6 @@ tempestAdminPassword: ""
 # cc3test domain - user credentials
 cc3testAdminUser: DEFINE_IN_REGION_VALUES
 cc3testMemberUser: DEFINE_IN_REGION_VALUES
+
+domain:
+  deploy: true
```
majewsky, line 40: Please use more descriptive field names, so that we can place the region-specific values into `globals.yaml` and also consume them from other service's charts:
```suggestion
domain_seeds:
  skip_hcm_domain: true
```
```

**Review by majewsky — CHANGES_REQUESTED** [review]

File: openstack/domain-seeds/templates/domain-hcm-seed.yaml [hunk 2]
```diff
@@ -16,7 +16,7 @@ spec:
   - name: hcm
     description: HCM
 
-{{- if .Values.globals.seeds.skip_hcm_domain }}
+{{- if ne .Values.global.domain_seeds.skip_hcm_domain true }}
```
majewsky, line 19:
```suggestion
{{- if not .Values.global.domain_seeds.skip_hcm_domain }}
```

majewsky (approved):

PR #3890, opened 2022-08-09
[Nova][Seeds]disable project seeding of a particular domain

rajivmucheli:


openstack/nova/values.yaml
```diff
@@ -9,7 +9,9 @@ global:
   novaApiMetadataPortInternal: '8775'
   novaConsolePortPublic: '443'
   master_password: null
-
+  domain_seeds:
+    skip_hcm_domain: false
```
joker-at-work, line 13: Why is this setting added in `global`? IIRC, things in global are supposed to be read by other charts or something - is there another use-case I don't know about or are we striving for that use-case or can this also be non-global?

joker-at-work (commented):
Could we do that more generic i.e. something along the lines of
```
domain_seeds:
    skip_domains:
        - hcm
 ```
and then `{{- if not has "hcm" .Values.global.domain_seeds.skip_domains }}`?

joker-at-work:
We cannot do it differently, because all services are doing it the same way with the same setting and thus we need to align. I've also been told that this will not happen for another domain and thus we don't need to build a more generic way.

joker-at-work (approved):

PR #3896, opened 2022-08-09
[Arc][Seeds]disable project seeding of a particular domain

rajivmucheli:
we will set it in globals.yaml for all services


ArtieReus (approved):
lgtm

PR #3901, opened 2022-08-09
[MariaDB] Enable Performance Schema for RCA and further perf improvements

viktor-kovacs:
Enable performance schema required to get performance data from mariadb, and requires to do analysis.

### DIFF
--- a/common/mariadb/Chart.yaml
+++ b/common/mariadb/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: mariadb
-version: 0.5.5
+version: 0.5.6
--- a/common/mariadb/values.yaml
+++ b/common/mariadb/values.yaml
@@ -9,7 +9,7 @@ imagePullPolicy: IfNotPresent
 port_public: 3306
 max_connections: 1024
 buffer_pool_size: "1024M"
-performance_schema: "OFF"
+performance_schema: "ON"
 log_file_size: "256M"
 log_warnings: "2"
 query_cache_size: "0"


IvoGoman:
This will enable the performance schema for all DBs relying on this chart. I would propose to make this opt-in by creating a var in the values.yaml which is set to OFF by default. Then this can be en-/disabled for each service independently in their chart values or even only for single regions via the regional secrets.
What do you think?

joker-at-work:
It doesn't say so in the commit message, but Viktor ensured us that this has no negative side-effects like performance impacts. Therefore, I would vote for setting the default to ON - but still make it configurable.

viktor-kovacs:
> This will enable the performance schema for all DBs relying on this chart. I would propose to make this opt-in by creating a var in the values.yaml which is set to OFF by default. Then this can be en-/disabled for each service independently in their chart values or even only for single regions via the regional secrets.
> What do you think?

VK: The Performance schema should be enabled by default. There could not be any performance impact on the database. It has been improved a lot in past few years.  But without it we can not perform root cause analysis, for instance why a SQL statement cause issue what has not been there before. or hard to check anything why the DB behaves abnormally. Restarting the POD could not be the solution always. As you know, there could be certain level what causes issues like wrong execution plan, outdated statistic after changes, or huge deletes etc. I'm fine with the approach to have configuration on service level, but I should know what is your exact preferences.

If you need further details please let me know.

IvoGoman:
Thanks for the details, this helped me understand the reason for enabling it. I assumed this was primarily for analyzing the busier DBs. Also I was too quick and missed that the change is in the values.yaml and not the config itself. Also the docs mentioned `The performance schema is disabled by default for performance reasons.` But since this is well tested I am happy with activating it.

galkindmitrii:
+1, I'm not against it enabling it, but I've not enabled it in prod regions before exactly due to:

> The performance schema is disabled by default for performance reasons.

viktor-kovacs:
> +1, I'm not against it enabling it, but I've not enabled it in prod regions before exactly due to:
> 
> > The performance schema is disabled by default for performance reasons.

It is bit catch-22 situation. We have very basic data from exporter and we don't have the same load in QA. I'm fine to do additional benchmarking, but we can not replicate the current usage fully. 
If you check almost everybody has the same result. https://www.percona.com/blog/2011/04/25/performance-schema-overhead/ , https://engineering.linecorp.com/en/blog/mysql-research-performance-schema-instruments/.
Based on the service separation the overhead should be much lower.

If I take one example about cinder: Cinder uses query cache what is single threaded. The current query utilisation is 8K. If I disable it the utilisation goes up to 10-11K. 
If a take the -10% 
8K - 800 = 7.2K 
10K - 1K =  9K
Disabling the query cache introduce further queries in slow log. But there is no real data for improvement due to lack of data.
Would be nice to discuss these topics on meeting like shared service meeting and clarify them in order to create the proper agreed PRs.

viktor-kovacs:
This Pull Request will requeseted with advanced options

PR #3902, opened 2022-08-09
[Maia] fallback prometheus-server chart some Prometheis can't be ugpraded to 7.0.1, hence the operator can't be upgraded. To avoid being stuck on an old operator version these Prometheis can use this chart until they are phased out or renovated.

richardtief:


notque (approved):
Seems reasonable to me. I can attempt to deploy it after the merge.

PR #3903, opened 2022-08-09
[prom-infra-global] fallback prometheus-server chart

richardtief:
Some Prometheis can't be ugpraded to 7.0.1 since they are relying on the old Thanos setup. Hence the prometheus operator can't be upgraded. To avoid being stuck on an old prometheus operator version these Prometheis can use this chart until they are phased out or renovated.


richardtief:
Obsolete now.

PR #3907, opened 2022-08-11
[infra-monitoring] bump prometheus server; thanos added

richardtief:
Note: Thanos will not be used until activated in the regional secrets.


richardtief:
Obsolete now.

PR #3912, opened 2022-08-12
[neutron] Add osprofiler for qa-de-1

misamoylov:


galkindmitrii (commented):
lgtm, but maybe @notandy has something to add

notandy:
Does ist really work? If yes, add a sidecontainer for the rpc-server.

PR #3913, opened 2022-08-12
[Cinder] Disable incremental backups

hemna:
This patch adds the ability to set the cinder option
to disable incremental backups.  The patch defaults
the option to On.


joker-at-work (approved):
I would have preferred the commit-message to tell me why we default to ON i.e. why we're disabling incremental backups in our env.

PR #3916, opened 2022-08-16
Update jaeger operator to 1.30.0

misamoylov:


system/jaeger-operator/templates/role.yaml
```diff
@@ -54,95 +55,84 @@ rules:
...
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
 - apiGroups:
-  - autoscaling
+  - route.openshift.io
```
galkindmitrii, line 104: I doubt that this has any effect. We don't use openshift and that is openshift-specific group.

system/jaeger-operator/templates/role.yaml
```diff
@@ -54,95 +55,84 @@ rules:
...
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
 - apiGroups:
-  - monitoring.coreos.com
+  - console.openshift.io
```
galkindmitrii, line 116: same as above, this is openshift-specific

system/jaeger-operator/templates/operator.yaml
```diff
@@ -16,11 +17,11 @@ spec:
       serviceAccountName: jaeger-operator
       containers:
       - name: jaeger-operator
-        image: {{ .Values.global.dockerHubMirrorAlternateRegion }}/jaegertracing/jaeger-operator:{{ $options.jaeger.version }}
+        image: {{ .Values.global.dockerHubMirrorAlternateRegion }}/jaegertracing/jaeger-operator:1.30.0
```
galkindmitrii, line 20: I'd personally prefer the old version with a value set to 1.30.0 in values.yaml

system/jaeger-operator/templates/operator.yaml
```diff
@@ -16,11 +17,11 @@ spec:
...
         ports:
         - containerPort: 8383
           name: metrics
-        args: ["start"]
+        args: ["start", "--jaeger-all-in-one-image=keppel.eu-de-1.cloud.sap/ccloud-dockerhub-mirror/jaegertracing/all-in-one"]
```
galkindmitrii, line 24: please use the common reference to keppel mirror, similar to https://github.com/sapcc/helm-charts/blob/master/openstack/designate/templates/api-deployment.yaml#L116

galkindmitrii (approved):

PR #3922, opened 2022-08-16
system/gatekeeper: add GkRegionValueMismatch

SuperSandro2000:


system/gatekeeper/tests/suite.yaml
```diff
@@ -266,6 +266,20 @@ tests:
...
 
+- name: region-value-mismatch
+  template: rendered-chart/gatekeeper/templates/constrainttemplate-region-value-mismatch.yaml
+  constraint: rendered-chart/gatekeeper-config/templates/constraint-region-value-mismatch.yaml
+  cases:
+  - name: helm-release-region-mismatch
+    object: fixtures/region-value-mismatch/helm-release-region-mismatch.out.yaml
+    assertions:
+    - violations:
+      message: '.Values.global.region does not match the deployed region'
```
majewsky, line 277: It would be nice if this message included the specific region values encountered.
```suggestion
      message: '.Values.global.region does not match the deployed region: expected qa-de-1, but found qa-de-9'
```

majewsky (changes requested):

system/gatekeeper/templates/constrainttemplate-region-value-mismatch.yaml
```diff
@@ -0,0 +1,44 @@
...
+              type: string
+            region:
+              type: string
+
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+        package gkregionvaluemissing
+
+        helm_release := input.review.object
```
majewsky, line 25: Please rename this to fit the common style established in #3853:

```suggestion
        iro := input.review.object
```

system/gatekeeper/templates/constrainttemplate-region-value-mismatch.yaml
```diff
@@ -0,0 +1,44 @@
...
+        parser_url := sprintf("%s/v3", [input.parameters.helmManifestParserURL])
+        parser_resp := http.send({"url": parser_url, "method": "POST", "raise_error": false, "raw_body": helm_release.data.release, "timeout": "15s"})
+
+        violation[{"msg": msg}] {
+          helm_release.kind == "Secret"
+          helm_release.type == "helm.sh/release.v1"
+          helm_release.metadata.labels.status == "deployed"
+
+          parser_resp.status_code == 200
+          input.parameters.region != parser_resp.body.values.global.region
```
majewsky, line 35: This check should only trigger when `.Values.global.region` contains a valid value, i.e. when `GkRegionValueMissing` does not trigger. In other words, you should use `object.get()` to traverse through the values and check if you're getting a string value, similar to how `GkRegionValueMissing` is implemented now.

Please add a testcase where `.Values.global.region` is not set at all, and also one where it is set to an invalid value like `null`.

system/gatekeeper/templates/constrainttemplate-region-value-mismatch.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+          msg := sprintf(".Values.global.region does not match the deployed region: expected %s, but found %s", [input.parameters.region, parser_resp.body.values.global.region])
```
majewsky, line 37: Please add the support-group and service labels to this message, by examining `.Values.owner-info` like in GkDeprecatedApiVersions etc.

majewsky (changes requested):

system/gatekeeper/templates/constrainttemplate-region-value-mismatch.yaml
```diff
@@ -0,0 +1,48 @@
...
+        import data.lib.add_support_labels
+        import data.lib.helm_release
+
+        iro := input.review.object
+        release := helm_release.parse_k8s_object(iro, input.parameters.helmManifestParserURL)
+        violation[{"msg": release.error}] {
+          release.error != ""
+        }
+
+        violation[{"msg": add_support_labels.from_k8s_object(iro, msg)}] {
```
majewsky, line 38: You need to examine the owner-info. `iro` is a Helm release which does not have the right labels.

system/gatekeeper/templates/constrainttemplate-region-value-mismatch.yaml
majewsky: This was marked as resolved, but has not been resolved. Since we now have the lib module setup, it may be useful to move `is_region_name()` from the other policy into a lib module to reuse this code here.

majewsky (changes requested):

system/gatekeeper/templates/constrainttemplate-region-value-mismatch.yaml
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        violation[{"msg": add_support_labels.from_helm_release(release, msg)}] {
+          release.error == ""
+
+          # do not complain about the contents of historical Helm releases
+          iro.metadata.labels.status == "deployed"
+
+          region := object.get(release, ["values", "global", "region"], "")
+          not validation.is_region_name(region)
```
majewsky, line 48:
```suggestion
          validation.is_region_name(region)
```

majewsky (changes requested):

system/gatekeeper/tests/suite.yaml
```diff
@@ -345,6 +345,29 @@ tests:
...
+  - name: helm-release-no-region
+    object: fixtures/region-value-mismatch/helm-release-no-region.out.yaml
+    assertions:
+    - violations:
+      message: 'support-group=none,service=none: .Values.global.region does not match the deployed region: expected qa-de-1, but found ""'
+  - name: helm-release-invalid-region
+    object: fixtures/region-value-mismatch/helm-release-invalid-region.out.yaml
+    assertions:
+    - violations:
+      message: 'support-group=none,service=none: .Values.global.region does not match the deployed region: expected qa-de-1, but found ""'
```
majewsky, line 365:
```suggestion
      message: 'support-group=none,service=none: .Values.global.region does not match the deployed region: expected qa-de-1, but found "qa-de-9"'
```

system/gatekeeper/tests/suite.yaml
```diff
@@ -345,6 +345,29 @@ tests:
...
+  cases:
+  - name: helm-release-region-mismatch
+    object: fixtures/region-value-mismatch/helm-release-region-mismatch.out.yaml
+    assertions:
+    - violations: no
+  - name: helm-release-no-region
+    object: fixtures/region-value-mismatch/helm-release-no-region.out.yaml
+    assertions:
+    - violations:
+      message: 'support-group=none,service=none: .Values.global.region does not match the deployed region: expected qa-de-1, but found ""'
```
majewsky, line 360:
```suggestion
    - violations: no
```

majewsky (changes requested):

PR #3926, opened 2022-08-17
[nova] Add jaeger side car agent

misamoylov:


joker-at-work (commented):
Why does only `nova-api` need the sidecar? Can you point me to some explanation how the setup is going to look like in the end?

joker-at-work:
> Jaeger-agent as a side car for nova-api and for scheduler (added in second commit) listen requests from osprofiler which use as driver jaeger-client installed in the same pod (we need to have a jaeger-client in pod for that) to send traces from the service.

Ok, now we have enabled it in 2 pods - what about all the other ones? Do you expect there to be only requests to API and scheduler to be profiled? I still don't get why we don't need the jaeger sidecar on all pods involved in a request.

PR #3927, opened 2022-08-17
system/gatekeeper: move support label rendering to a lib module

majewsky:
This serves as a proposal for how we may structure Rego library modules in this chart going forward.

- [x] new library method: `add_support_labels.from_k8s_object()`
- [x] new library method: `add_support_labels.from_helm_release()`

*Update:* In order to gauge how much we can do in library modules, I also moved the helm-manifest-parser calls into a module.

- [x] new library method: `helm_release.parse_k8s_object()`

### DIFF
--- a/system/gatekeeper/templates/constrainttemplate-vulnerable-images.yaml
+++ b/system/gatekeeper/templates/constrainttemplate-vulnerable-images.yaml
@@ -53,7 +53,7 @@ spec:
           msg := "doop-image-parser did not return HTTP status 200. Please retry in ~5 minutes."
         }
 
-        violation[{"msg": msg}] {
+        violation[{"msg": add_support_labels.from_k8s_object(iro, msg)}] {
           check := checks[_]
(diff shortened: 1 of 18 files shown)


system/gatekeeper/templates/constrainttemplate-vulnerable-images.yaml
```diff
@@ -59,8 +61,10 @@ spec:
...
           status != "Unknown"
-          msg := sprintf(
-            "support-group=%s,service=%s: image %s for container %q has vulnerability status %q",
-            [iro_support_group, iro_service, check.container.image, check.container.name, status]
+
+          msgBase := sprintf(
+            "image %s for container %q has vulnerability status %q",
+            [check.container.image, check.container.name, status]
           )
+          msg := add_support_labels.from_k8s_object(iro, msgBase)
```
SuperSandro2000, line 69: This should be changed to the new syntax

PR #3928, opened 2022-08-19
WIP [nova] Update policies for Xena upgrade

grandchild:
- Remove all "discoverable" policy rules.
- Split `os-server-groups`

WIP. More to come...


grandchild:
As it turns out [Openstack upstream is not going to go through with the Xena/Yoga-era scoped policies concept](https://governance.openstack.org/tc/goals/selected/consistent-and-secure-rbac.html). So we're going to keep using the old policy roles until the new "persona"-based RBAC is adopted in Zed or possibly the next release.

grandchild:
Mh 😦 I force-pushed the new commits (also fixed an earlier commit) first and then separately force-pushed the rebase onto master, but Github apparently merged them.

Here's the original force-push diff: https://github.com/sapcc/helm-charts/compare/665e0af36a797953be61d37333e0996674832c67..36be963db58cea63f48a029f403da0d1dc010701

In the future I should just add a "." separator comment :P

joker-at-work (approved):

PR #3930, opened 2022-08-22
bump prometheus-server version to switch swift user for OpenstackSeed

artherd42:
https://documentation.global.cloud.sap/blog/objectstorage-new-roles


richardtief (approved):
All good.

PR #3929, opened 2022-08-19
update (i)pxe/uefi templates

sandzwerg:

PR #3931, opened 2022-08-22
[ironic] remove reference to deleted template

sandzwerg:
was missed in #3929


sandzwerg:
@stefanhipfel can you take another look?

PR #3932, opened 2022-08-22
[barbican] audit map exclude payload member for secrets endpoint

notque:
@rajivmucheli here is the change requested from Anton. Please approve, then we can merge, deploy, and test. Thanks!


rajivmucheli (approved):
lgtm

PR #3935, opened 2022-08-23
fix playbook links

artherd42:
points to https://operations.global.cloud.sap/docs/support/playbook/cinder/alerts/


developer-abhi (approved):
lgtm

PR #3938, opened 2022-08-23
[kube-system-admin-k3s] allowCrossNameSpace for IngressRoute CRD

richardtief:
By using the Traefik TLSOption resource in Kubernetes, we setup a default set of options that should apply to all ingresses. To achieve that, we created a TLSOption resource with the name default. There may exist only one TLSOption with the name default (across all namespaces) - otherwise they will be dropped. To explicitly use this TLSOption (and using the Kubernetes Ingress resources) we'll have to allow IngressRoute to reference resources in namespaces other than theirs.

https://doc.traefik.io/traefik/routing/providers/kubernetes-crd/#serverstransport-reference


defo89 (approved):
lgtm, but please make sure it does not break qa ingresses before rolling this out to prod

PR #3942, opened 2022-08-24
[cronus]: set ndots to 1

kayrus:
to reduce DNS lookups


dhalimi (approved):
Need to be tested first in QA with a load test

PR #3943, opened 2022-08-24
[cinder] Enable osprofiler for volume and volume-backup

misamoylov:


joker-at-work (changes requested):
`_volume-deployment.yaml.tpl` is not currently used in production afaik. But in any case, you also need to add the same to `_netapp-deployment.yaml` and much more important to `vcenter_datacenter_cinder_deployment.yaml`

`cd openstack/cinder; git grep Deployment` finds you all the places ...

openstack/cinder/templates/etc/_volume.conf.tpl
```diff
@@ -6,6 +6,8 @@
 enabled_backends={{$volume.name}}
 storage_availability_zone={{$volume.availability_zone}}
 
+{{- include "osprofiler" . }}
+
```
joker-at-work, line 10: Why is this necessary if the Deployment already includes `cinder.conf`?

openstack/cinder/templates/etc/_volume-vmware.conf.tpl
```diff
@@ -6,6 +6,8 @@
 enabled_backends = vmware
 storage_availability_zone={{$volume.availability_zone}}
 
+{{- include "osprofiler" . }}
+
```
joker-at-work, line 10: Why is this necessary if the deployment already includes `cinder.conf`?

openstack/cinder/templates/etc/_volume-netapp.conf.tpl
```diff
@@ -6,6 +6,8 @@
 enabled_backends=netapp
 storage_availability_zone={{$volume.availability_zone}}
 
+{{- include "osprofiler" . }}
+
```
joker-at-work, line 10: Why is this necessary if the deployment already includes `cinder.conf`?

openstack/cinder/templates/etc/_volume-netapp.conf.tpl
```diff
@@ -6,6 +6,7 @@
 enabled_backends=netapp
 storage_availability_zone={{$volume.availability_zone}}
 
+
```
joker-at-work, line 9: You're adding newlines here now. Please don't. Same in the other files.

joker-at-work (approved):

PR #3956, opened 2022-08-29
[Nova] Update settings for Xena

joker-at-work:
This PR updates the Nova chart to add/remove/update settings required for Xena while keeping it compatible with Rocky.


grandchild (approved):
> It feels better to have a clean config in production though

And more importantly, it signals to future us to remove these sections!

joker-at-work:
> And more importantly, it signals to future us to remove these sections!

We could have done that with a comment, though ;)

PR #3957, opened 2022-08-30
[nova] Remove consoleauth support

joker-at-work:
Starting with Rocky, Nova doesn't use the consoleauth service anymore
unless explicitly enabled in the [workarounds] section. Instead, tokens
are saved in the DB and read from there. We only needed consoleauth
enabled during the migration to Rocky and have disabled it since.

Since we're updating to Xena, we now remove the consoleauth Deployment
altogether, because there the consoleauth service doesn't exist in Nova
anymore.


openstack/nova/values.yaml
```diff
@@ -72,7 +72,6 @@ pod:
     api: 6
     metadata: 2
     console: 2
-    consoleauth: 0
     conductor: 2
     scheduler: 1
     vspc: 4
```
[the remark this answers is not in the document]
joker-at-work: Nice catch. Thank you.

openstack/nova/templates/etc/_nova.conf.tpl
```diff
@@ -196,9 +196,4 @@ region_name = {{.Values.global.region}}
 [wsgi]
 default_pool_size = {{ .Values.wsgi_default_pool_size | default .Values.global.wsgi_default_pool_size | default 100 }}
 
-[workarounds]
-{{- if .Values.pod.replicas.consoleauth }}
-# This has to be removed when we also remove the deployment of nova-consoleauth
-enable_consoleauth = True
-{{- end }}
 enable_live_migration_to_old_hypervisor = True
```
grandchild, line 199: Was `enable_live_migration_to_old_hypervisor` part of the [workarounds] section and is now part of [wsgi]? Or is the empty line meaningful in this syntax (even so, I don't think there was an empty line between `enabled_consoleauth` and `enable_live_migration_to_old_hypervisor`, if I understand `{{- end }}` correctly, i.e. the leading dash makes it as if that tag were on the previous line right after the `True`.

joker-at-work: Yes, this is indeed a bug.

grandchild (approved):

PR #3961, opened 2022-08-30
[common/mariadb] Use backup-user with connection limit

fwiesel:
Instead of using the root-user, use a dedicated user with
its own connection limit in order to isolate the
backup process from the main application

### DIFF
--- a/common/mariadb/Chart.yaml
+++ b/common/mariadb/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: mariadb
-version: 0.5.5
+version: 0.6.0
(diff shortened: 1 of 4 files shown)


viktor-kovacs (approved):
Hi, It look fine to me

IvoGoman (approved):
Looks good to me.
Once this is rolled out to the services I will use this user as well for the replication to metis.

fwiesel:
I would like to create a separate user for metis. Users do not cost anything compared to the effort/risk when one application misbehaves. Do all services need a metis user?

fwiesel:
Either way, let's move forward with this solution, and then we can improve on that.

IvoGoman:
> I would like to create a separate user for metis. Users do not cost anything compared to the effort/risk when one application misbehaves. Do all services need a metis user?

Currently the replication is enabled for nova, neutron, glance, ironic, designate, keystone and soon also for nova-api.

PR #3967, opened 2022-08-31
[openstack-exporter] rename image.tag to imageVersion

leust:
For the CI to work properly, the version needs to be specified
inside a property that contains `imageVersion` in its name.

### DIFF
--- a/prometheus-exporters/openstack-exporter/values.yaml
+++ b/prometheus-exporters/openstack-exporter/values.yaml
@@ -8,12 +8,11 @@ exporter:
   log_level: "DEBUG"
 openstack:
   enabled: false
-  image:
-    name: openstack-exporter
-    tag: DEFINED-IN-REGION-SECRETS
+  imageName: openstack-exporter
+  # imageVersion:
   alerts:
     enabled: false
-    prometheus: DEFINED-IN-REGION-SECRETS
+    # prometheus: {}
   limits:
     memory: 200Mi
     cpu: 300m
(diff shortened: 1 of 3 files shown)


prometheus-exporters/openstack-exporter/values.yaml
```diff
@@ -8,9 +8,8 @@ exporter:
   log_level: "DEBUG"
 openstack:
   enabled: false
-  image:
-    name: openstack-exporter
-    tag: DEFINED-IN-REGION-SECRETS
+  imageName: openstack-exporter
+  imageVersion: DEFINED-IN-REGION-SECRETS
```
joker-at-work, line 12: Can we at the same time get rid of this anti-pattern? `DEFINED-IN-REGION-SECRETS` is not a usable default!

joker-at-work: I like it, but it makes the CI fail. We need to add a `ci/` directory with some test values now that we don't have (non-working) defaults.
```
engine.go:159: [INFO] Missing required value: .Values.global.openstack_exporter_master_password missing
engine.go:159: [INFO] Missing required value: .Values.global.openstack_exporter_master_password missing
engine.go:159: [INFO] Missing required value: .Values.global.region missing
engine.go:159: [INFO] Missing required value: .Values.openstack.alerts.prometheus missing
engine.go:159: [INFO] Missing required value: .Values.global.registry missing
engine.go:159: [INFO] Missing required value: .Values.openstack.imageVersion missing
Error: 1 chart(s) linted, 1 chart(s) failed
```

joker-at-work (approved):
Looks good for me.

hemna (approved):

PR #3968, opened 2022-09-01
[ci-lint-chart] Run linter as non-root user

fwiesel:
While in general it is a good practice to avoid running
processes as root unless necessary, the main
gain of doing so here is that it won't place
files as the root user in the local chart
directory.

The script can be run by any user,
but then the user ends up with files
owned by root in their own directory.

Running it with the callers user
gives us downloaded charts owned
by the same user


fwiesel:
Hi, short reminder.

auhlig (approved):
Thanks!

PR #3982, opened 2022-09-07
gatekeeper: add "traversal" lib module for finding pod and container specs

majewsky:
The logic mostly comes from GkImagesFromNonKeppel, and is now generalized to all policies that check pod specs or container specs.

@SuperSandro2000 I want to get this topic done this week if possible, which means I would have to merge tomorrow. If you get the chance to look at this today, I'd be grateful. If you don't get the chance, I'll merge this tomorrow at my own risk. In any case, I will perform detailed verification in QA to check that the policy behavior does not change in unintended ways.

### DIFF
--- a/system/gatekeeper/lib/traversal.rego
+++ b/system/gatekeeper/lib/traversal.rego
@@ -2,11 +2,11 @@ package lib.traversal
 
 ### find_pod(obj)
 #
-# Find a Pod object or template  within the given k8s object. For Pods, `obj`
+# Find a Pod object or template within the given k8s object. For Pods, `obj`
 # itself is returned; but e.g. for Deployments, `.spec.template` is returned.
 #
-# On success, the returned object will have {"found": true} added to it.
-# If no Pod is found, an object with just {"found": false} is returned.
+# On success, the returned object will have {"isFound": true} added to it.
+# If no Pod is found, an object with just {"isFound": false} is returned.
 #
 # Pods that are owned by a ReplicaSet etc. are ignored to avoid
 # double-reporting. (We report on the highest level possible to reduce the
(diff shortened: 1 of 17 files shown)


system/gatekeeper/templates/constrainttemplate-resource-limits.yaml
```diff
@@ -13,54 +13,34 @@ spec:
...
-        # each offending pod, and we report on the deployment/daemonset level
-        # if possible to avoid useless duplication of violations.
+        # each offending pod or pod owner.
 
         resource_types = {"cpu", "memory"}
 
-        missing_limits_on_pod[{"container": container.name, "type": resource_type}] {
-          container := iro.spec.containers[_]
+        missing_limits[{"container": container.name, "type": resource_type}] {
+          pod.found
```
SuperSandro2000, line 35: I am not super happy about this name. Maybe we find something more descriptive like ``pod.isFound`` or ``pod.found?``.

system/gatekeeper/lib/traversal.rego
```diff
@@ -0,0 +1,75 @@
+package lib.traversal
+
+### find_pod(obj)
+#
+# Find a Pod object or template  within the given k8s object. For Pods, `obj`
```
SuperSandro2000, line 5:
```suggestion
# Find a Pod object or template within the given k8s object. For Pods, `obj`
```

system/gatekeeper/lib/traversal.rego
```diff
@@ -0,0 +1,75 @@
...
+}
+find_pod(obj) = result {
+  # case 2: `obj` contains a PodSpec
+  obj.kind == __violation_owners[_]
+  result := object.union(obj.spec.template, { "found": true })
+}
+find_pod(obj) = result {
+  # case 3: neither Pod nor PodSpec
+  obj.kind != "Pod"
+  count([kind | kind := __violation_owners[_]; obj.kind == kind ]) == 0
```
SuperSandro2000, line 30: I don't fully understand this construct but could it fail and we need another fallback for find_pod?

majewsky: List comprehensions work somewhat similar to rules on the top level, except that they create lists instead of objects. `[ <expr> | <body> ]` creates a list that, for each successful evaluation of `<body>`, contains the result of `<expr>` for the variables declared within the body. So in this case, this reads as: For each `kind := __violation_owners[_]` that fulfils `obj.kind == kind`, the list will contain `kind`.

So there are two cases: Either `obj.kind` is in `__violation_owners`, then the list will have that kind name as its single member and the `count(...) == 0` will not hold (which is okay, since in this situation "case 2" matches instead). Or if we have an unknown value in `obj.kind`, the list will be empty and `count(...) == 0` will hold.

PR #3997, opened 2022-09-12
[nova] Remove per-service notification rabbitmq

fwiesel:
We have already migrated in all regions
the services to the central audit rabbitmq
So we can now undeploy the now unused one


kpawar-sap (approved):

joker-at-work (approved):
Would be good to clean up in the regional-local settings, too.

PR #4000, opened 2022-09-12
Increasing rrset cache from 4MB to 100MB

dhoeller:
Increasing the rrset and msg cache because we see in the bigger regions that the cache is evicted before the TTL runs out.


bentinney (approved):
LGTM

bentinney (approved):
LGTM

occamshatchet (approved):
LGTM

PR #4002, opened 2022-09-13
[utils] Add scalefactor for connections

fwiesel:
The max_connections are on a per-process basis,
while the proxysql config assumes a single process
per pod.
This change allows the caller of the container macro
pass a scale-factor, which specifies the number of
processes in the pod.

PR #4003, opened 2022-09-13
[nova] Scale max-connections with number of processes

fwiesel:
As sqlalchemy pool sets the limits per process,
we need to scale the number of connections with the number of
processes to have the same limit per pod.

### DIFF
--- a/openstack/nova/Chart.yaml
+++ b/openstack/nova/Chart.yaml
@@ -26,7 +26,7 @@ dependencies:
     version: 0.0.10
   - name: utils
     repository: https://charts.eu-de-2.cloud.sap
-    version: 0.5.0
+    version: 0.6.0
   - name: mariadb
     alias: mariadb_cell2
     condition: mariadb_cell2.enabled
--- a/openstack/nova/templates/api-metadata-deployment.yaml
+++ b/openstack/nova/templates/api-metadata-deployment.yaml
@@ -143,7 +143,7 @@ spec:
               readOnly: true
             {{- end }}
             {{- include "utils.proxysql.volume_mount" . | indent 12 }}
-        {{- include "utils.proxysql.container" . | indent 8 }}
+        {{- tuple . .Values.api_metadata.config_file.DEFAULT.metadata_workers | include "utils.proxysql.container" | indent 8 }}
         - name: statsd
           image: {{ required ".Values.global.dockerHubMirror is missing" .Values.global.dockerHubMirror}}/prom/statsd-exporter:v0.8.1
           imagePullPolicy: IfNotPresent
--- a/openstack/nova/templates/cell2-conductor-deployment.yaml
+++ b/openstack/nova/templates/cell2-conductor-deployment.yaml
@@ -111,7 +111,7 @@ spec:
               subPath: logging.ini
               readOnly: true
             {{- include "utils.proxysql.volume_mount" . | indent 12 }}
-        {{- include "utils.proxysql.container" . | indent 8 }}
+        {{- tuple . .Values.conductor.config_file.conductor.workers | include "utils.proxysql.container" | indent 8 }}
         {{- if .Values.cell2.conductor.config_file.DEFAULT.statsd_enabled }}
         - name: statsd
           image: {{ required ".Values.global.dockerHubMirror is missing" .Values.global.dockerHubMirror}}/prom/statsd-exporter:v0.8.1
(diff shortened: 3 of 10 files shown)


fwiesel:
Based on top of  #4002. Need to update the Chart.yaml and lock when merged.

openstack/nova/Chart.yaml
```diff
@@ -30,8 +30,8 @@ dependencies:
     repository: https://charts.eu-de-2.cloud.sap
     version: 0.4.2
   - name: utils
-    repository: https://charts.eu-de-2.cloud.sap
-    version: 0.5.0
+    repository: file://../utils
```
joker-at-work, line 33: I know this is still marked as draft, but we should not forget to update these references again after merging the PR we depend on, so I'm still marking it now.

openstack/nova/templates/cell2-conductor-deployment.yaml
```diff
@@ -111,7 +111,7 @@ spec:
               subPath: logging.ini
               readOnly: true
             {{- include "utils.proxysql.volume_mount" . | indent 12 }}
-        {{- include "utils.proxysql.container" . | indent 8 }}
+        {{- tuple . .Values.rpc_workers | include "utils.proxysql.container" | indent 8 }}
```
joker-at-work, line 114: Why is this `rpc_workers` and not like the `conductor-deployment.yaml` `conductor.config_file.conductor.workers`?

openstack/nova/Chart.yaml
fwiesel: Has been updated.

openstack/nova/templates/etc/_nova.conf.tpl
```diff
@@ -23,7 +23,7 @@ default_schedule_zone = {{.Values.global.default_availability_zone}}
 default_availability_zone = {{.Values.global.default_availability_zone}}
 
 rpc_response_timeout = {{ .Values.rpc_response_timeout | default .Values.global.rpc_response_timeout | default 60 }}
-rpc_workers = {{ .Values.rpc_workers | default .Values.global.rpc_workers | default 1 }}
+rpc_workers = {{ .Values.rpc_workers }}
```
joker-at-work, line 26: I can only find that setting for in `neutron.conf` - I don't think we need that setting, as it's unused in Nova.

joker-at-work (commented):
`api-metadata-deployment` seems to be missing

fwiesel:
> `api-metadata-deployment` seems to be missing

It looks to me, like the metadata service [goes through the conductor.](https://github.com/sapcc/nova/blob/stable/rocky-m3/nova/cmd/api_metadata.py#L45)

openstack/nova/templates/cell2-conductor-deployment.yaml
fwiesel: Ah, you are right. I'll fix it.

openstack/nova/templates/etc/_nova.conf.tpl
fwiesel: Yes, I didn't find it in the source either. I've removed it.

joker-at-work:
> > `api-metadata-deployment` seems to be missing
> 
> It looks to me, like the metadata service [goes through the conductor.](https://github.com/sapcc/nova/blob/stable/rocky-m3/nova/cmd/api_metadata.py#L45)

Looks like it. Then the api-metadata pod should not have a proxysql-sidecar and we should remove it.

joker-at-work:
Oh, but we call `nova-api` and not `nova-api-metadata` in our pods. I see connections from the Pod to port 3306

joker-at-work:
https://github.com/sapcc/helm-charts/commit/70da250a43007dbea9b0f24e05f0b472ff048b1e contains the explanation for why we use `nova-api`

openstack/nova/templates/api-metadata-deployment.yaml
```diff
@@ -143,7 +143,7 @@ spec:
               readOnly: true
             {{- end }}
             {{- include "utils.proxysql.volume_mount" . | indent 12 }}
-        {{- include "utils.proxysql.container" . | indent 8 }}
+        {{- tuple . .Values.api.config_file.DEFAULT.osapi_compute_workers | include "utils.proxysql.container" | indent 8 }}
```
joker-at-work, line 146: That's the wrong variable. Need to use this: https://github.com/sapcc/helm-charts/blob/aae1048b17c973ec708e44d1ee398dc66cb5ea35/openstack/nova/values.yaml#L225

joker-at-work (approved):

PR #4005, opened 2022-09-13
Add new cinder alerts for backend & pool states

hemna:
This patch adds new alerts for cinder backends being down as well as pools being marked as down.  This patch also alters the max volume size warning from <=3 to < 3.

NOTE:
Still need to add 2 new playbooks to support the new alerts.


hemna:
The associated playbooks with this PR

https://github.wdf.sap.corp/monsoon/convergedcloud-opsdocu-ui/pull/1361

developer-abhi (approved):
lgtm

prometheus-exporters/openstack-exporter/alerts/cinder.alerts
```diff
@@ -24,8 +24,8 @@ groups:
       service: cinder
       playbook: docs/support/playbook/cinder/cinder-low-free-space.html
     annotations:
-      description: 'Cinder backend {{ $labels.shard }}/{{ $labels.backend }} has 3 or less datastores with > 30% free space left'
-      summary: 'Cinder backend {{ $labels.shard }}/{{ $labels.backend }} has 3 or less3 datastores with > 30% free space left'
+      description: 'Cinder backend {{ $labels.shard }}/{{ $labels.backend }} has 2 or less datastores with > 30% free space left'
+      summary: 'Cinder backend {{ $labels.shard }}/{{ $labels.backend }} has 2 or less datastores with > 30% free space left'
```
joker-at-work, line 28: Doesn't the expression say "< 2" and thus it cannot be 2 while the alert fires? -> "has less than 2 datastores"

hemna: done

joker-at-work (approved):

PR #4010, opened 2022-09-15
New flavor for S4 and setting the HXM flavors as public flavors

dhoeller:


frr2018 (approved):
LGTM

openstack/sap-seeds/templates/flavor-seed.yaml
```diff
@@ -335,7 +335,17 @@ spec:
...
-    is_public: false
+    is_public: true
+    extra_specs:
+      "vmware:hv_enabled": "True"
+      "hw_video:ram_max_mb": "16"
+      "trait:CUSTOM_HANA_EXCLUSIVE_HOST": "forbidden"
+  - name: "m_c24_m192"
+    id: "142"
+    vcpus: 24
+    ram: 196596
```
joker-at-work, line 346: Why is this `192 * 1024 - 12` and not `192 * 1024 - 16`? Don't we usually subtract the `16` from `hw_video:ram_max_mb`?

dhoeller: Changed to 196592

joker-at-work (approved):

PR #4019, opened 2022-09-20
[ironic] adds kos-query to set boot_interface to ipxe

stefanhipfel:


openstack/ironic/templates/nodes-update-kq.yaml
```diff
@@ -0,0 +1,36 @@
...
+        if node.boot_interface == 'ipxe':
+            continue
+        if not node.maintenance:
+          bm.node.set_maintenance(node.uuid, 'true', maint_reason='set boot_interface to ipxe')
+          unset_maintenance = True
+
+        bm.node.update(node_id=node.uuid, patch=patch)
+
+        if unset_maintenance:
+          bm.node.set_maintenance(node.uuid, 'false', maint_reason='set boot_interface to ipxe')
```
BerndKue, line 35: An unset of the maintenance should not need a maintenance reason. Maybe this even fails because of the extra parameter?

stefanhipfel: param will just be ignored, but i can remove it.

sandzwerg (approved):

PR #4025, opened 2022-09-20
[ironic] improves script setting ipxe for nodes

stefanhipfel:


stefanhipfel:
some states like 'wait callback' cannot be updated even when in maintenance

stefanhipfel:
> Does it retry when the states change?

yes the kos-operator execs the script every x seconds

PR #4034, opened 2022-09-23
[cronus]: introduce new maxConnReleaseTimeout config option

kayrus:
new option for conn pool
cc @sapmli


sapmli (approved):
I have no clue about this `{{-` syntax, but it looks syntactically identical to the treatment of `maxConnWaitTimeout` 😄, so thumbs up 👍 .

PR #4037, opened 2022-09-23
gatekeeper: add libtest and improve coverage for "traversal" module

majewsky:
We now also report violations on ReplicaSets that do not belong to a Deployment for some reason. Also, Jobs that belong to a CronJob are now suppressed, and violations will be reported on the CronJob instead.


SuperSandro2000:
We should replace the labels with violations checked by gator.

majewsky:
> We should replace the labels with violations checked by gator.

Done.

PR #4043, opened 2022-09-26
add extra label awx as in https://github.com/sapcc/helm-charts/commit…

artherd42:
…/7afe7b7771805f7fa1fb68491b1ac433819d1d04

We have moved the metrics and alerts from vcenter-exporter to cc3test's.

### DIFF
--- a/openstack/cc3test/alerts/datastore.alerts
+++ b/openstack/cc3test/alerts/datastore.alerts
@@ -12,6 +12,7 @@ groups:
       context: '{{ $labels.type }}'
       meta: 'Datastore not accessible from host'
       playbook: 'docs/support/playbook/cinder/alerts'
+      awx: '#/templates/job_template/114'
     annotations:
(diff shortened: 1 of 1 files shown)


stanislav-zaprudskiy (commented):
I only fixed the quoting issue in that commit to let the pipelines pass through.

Unfortunately, I can't tell anything about the presence of the label or its value. Perhaps @christopherhans  or @richardtief would be more appropriate reviewers for this change.

openstack/cc3test/alerts/datastore.alerts
```diff
@@ -12,6 +12,7 @@ groups:
       context: '{{ $labels.type }}'
       meta: 'Datastore not accessible from host'
       playbook: 'docs/support/playbook/cinder/alerts'
+      awx: '#/templates/job_template/114'
```
christopherhans, line 15:
```suggestion
      awx: '#/templates/job_template/2943'
```

christopherhans (changes requested):
Please update template ID as suggested.

PR #4045, opened 2022-09-26
[Cinder] Add new SAPPoolDownFilter to default filters

hemna:
This patch moves the CapacityFilter to the last filter to run as well as adding a new SAPPoolDownFilter, which will filter out all pools that are not marked as 'up' by the vmware driver.

### DIFF
--- a/openstack/cinder/values.yaml
+++ b/openstack/cinder/values.yaml
@@ -84,7 +84,7 @@ use_tls_acme: true
 
 db_name: cinder
 
-scheduler_default_filters: "AvailabilityZoneFilter,ShardFilter,CapacityFilter,CapabilitiesFilter,SAPLargeVolumeFilter,SAPDifferentBackendFilter,SAPSameBackendFilter"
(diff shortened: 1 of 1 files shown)


joker-at-work:
> This patch moves the CapacityFilter to the last filter to run

why?

hemna:
> > This patch moves the CapacityFilter to the last filter to run
> 
> why?

to reduce the number of pools it has to parse for processing.   If a pool is already removed by any of the other filters, then the capacity filter doesn't have to do all of the code for calculating space for the pool.   In regions where we have a lot of datastores, I think this might help with the speed of getting through the scheduler.

joker-at-work:
> to reduce the number of pools it has to parse for processing. If a pool is already removed by any of the other filters, then the capacity filter doesn't have to do all of the code for calculating space for the pool. In regions where we have a lot of datastores, I think this might help with the speed of getting through the scheduler.

Why don't you write that into the commit message? We already see the move in the diff, but the important thing is the "why", so we don't just move it back next year, because nobody knows why it's at the end currently.

joker-at-work (approved):

PR #4046, opened 2022-09-27
fixup! Cargocult rate limiting from keystone

BenjaminLudwigSAP:
From here: https://github.com/sapcc/helm-charts/pull/2774/files

- _ratelimit.yaml.tpl API endpoint syntax??
- octavia/values: sapcc_rate_limit enabled, persistence disabled => What does persistence mean here?


BenjaminLudwigSAP:
rebased on master

openstack/octavia/templates/etc/_ratelimit.yaml.tpl
```diff
@@ -0,0 +1,41 @@
...
+# Group multiple CADF actions to one rate limit action.
+groups:
+  write:
+    - update
+    - delete
+
+rates:
+  # global rate limits counted across all projects
+  global:
+    lbaas/loadbalancers:
```
notandy, line 29: the target type, as injected by the watcher, is not lbaas/loadbalancers, it's just `loadbalancers`.

openstack/octavia/templates/etc/_ratelimit.yaml.tpl
```diff
@@ -0,0 +1,41 @@
...
+      - action: read
+        limit: 2000r/m
+
+  # default local rate limits applied to each project
+  default:
+    lbaas/*:
```
notandy, line 35: same here, `lbaas` is not recognized and would be a no-op, replace with `'*'`

openstack/octavia/templates/etc/_ratelimit.yaml.tpl
```diff
@@ -0,0 +1,41 @@
...
+  status: 497 Blocklisted
+  status_code: 497
+  body: "Your account has been blocklisted"
+
+# Group multiple CADF actions to one rate limit action.
+groups:
+  write:
+    - update
+    - delete
+
```
notandy, line 25: you should add a group read:
```yaml
read:
  - read
  - read/list
```
this is becase openstack-watcher emits `read/list` for listing instead of `read`: https://github.com/openstack/pycadf/blob/master/pycadf/cadftaxonomy.py#L30

notandy (changes requested):

PR #4056, opened 2022-09-28
RIP vcenter-exporter

artherd42:


system/infra-monitoring/Chart.yaml
```diff
@@ -41,11 +41,11 @@ dependencies:
...
-  - name: vcenter-exporters
-    alias: vcenter_exporters
-    repository: file://vendor/vcenter-exporters
-    version: 0.1.0
-    condition: vcenter_exporters.enabled
+#  - name: vcenter-exporters
+#    alias: vcenter_exporters
+#    repository: file://vendor/vcenter-exporters
+#    version: 0.1.0
+#    condition: vcenter_exporters.enabled
```
richardtief, line 48: Why keep that? This can go away forever so it's decluttered. Git offers version tracking if we want to look up something from the past.

richardtief (changes requested):
Finally this goes away. Just one comment.

richardtief (approved):

PR #4065, opened 2022-09-30
Add 2faproxy to keystone chart

databus23:
Add an intermedite reverse proxy deployment to the keystone chart between ingress and the keystone api that processes /v3/auth/tokens requests and enforces opt-in 2fa authentication in the form of tls client certificates.

Unless `2fa.enabled` is set to `true` no changes during deployment are expected.

### DIFF
--- a/openstack/keystone/values.yaml
+++ b/openstack/keystone/values.yaml
@@ -464,3 +464,14 @@ sapcc_rate_limit:
   enabled: false
   persistence:
     enabled: false
+
+2fa:
+  enabled: false
+  replicaCount: 3
+  pollInterval: 1m #how often do we list projects
+  image: ccloud/2faproxy
+  imageTag: latest
+  imagePullPolicy: IfNotPresent
+  openstack: {}
+  resources: {}
+  podAnnotations: {}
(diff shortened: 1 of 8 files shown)


openstack/keystone/values.yaml
```diff
@@ -464,3 +464,14 @@ sapcc_rate_limit:
   enabled: false
   persistence:
     enabled: false
+
+2fa:
+  enabled: false
+  replicaCount: 3
```
bbobrov, line 470: Why are there 3 replicas? Does it make sense to make the number of replicas equal to the number of keystone-api pods?

openstack/keystone/values.yaml
```diff
@@ -464,3 +464,14 @@ sapcc_rate_limit:
...
+  pollInterval: 1m #how often do we list projects
```
bbobrov, line 471: `pollInterval: 1m` means that the proxy will list projects ~1 time per minute. Do i understand correctly, that N replicas will make ~N requests per minute together?

databus23: yes each replica will call `/v3/projects?tag=xs:2fa` every `pollInterval` to maintain and a fresh local cache of auth scopes that need 2fa enforcement.
We can extend this to 5 minutes of you have concerns. This means changing the access policies in the dashboard may take up to `pollInterval` to take effect.

openstack/keystone/values.yaml
databus23: We can make it match, but I think the throughput, memory/cpu and overall performance behaviour of this proxy is completely different then keystone. it only parses json and make a local decision to forward the request. I think the main motivation of multiple replicase is fault tolerance for this proxy and to allow for a smooth rolling update of a new version.
If make it match then I think it would make sense to run this as sidecar of the api pod to also avoid network roundtrips inside of the cluster.

PR #4077, opened 2022-10-04
[prometheus] update prometheus-server

viennaa:
newer operator requires this prometheus-server version to be in place
maia is going to use prometheus-server-pre7 instead


richardtief (approved):
All good.

PR #4087, opened 2022-10-05
[placement] Add db migration jobs

joker-at-work:
When running Placement on its own images, Placement is responsible for migrating the DB. This is done via the placement-manage command.

We add 2 Jobs here - one for migrated the schema and one for doing any necessary data migrations.

Both Jobs only run if they find "placement-manage" in the path and thus will only do any work once we upgrade Placement to Yoga and its own images.


joker-at-work:
I've tested the Jobs in qa and had to update them, because they were unable to kill proxysql if they could not find `placement-manage`. Therefore, we now sleep 5s to be able to kill it.

Additionally, the labels were missing some  stuff.

fwiesel (approved):

PR #4090, opened 2022-10-06
[Nova][alerts] Add support group label

grandchild:


geisslet (commented):
there are discussions about the support label.
what is it about:
- replace `support` by `support_group`
- replace prefix `group` by `ccloud`

need Monday / latest Tuesday to clarify //bigSorry

joker-at-work (approved):

PR #4095, opened 2022-10-06
Do not wait for neutron network agents

sebageek:
cloud.sap/wait-for-ready will mark our network agents as something the CI does not have to wait for. The rolling rollout for the network agents can take quite long (depending on the amount up to 90 minutes) and we don't need to wait for all of them being done for a deployment to succeed.


sebageek:
Not needed at the moment, will reopen if we figure out that we do need it

PR #4097, opened 2022-10-07
[alerts] alerts for vnode and admin pool members

xsen84:


m-kratochvil:
I see a possible issue with the `NetworkF5Admin1IsOffline` alert. With your current design, one of the admin pool members is always offline, and that would constantly generate alerts, no?
FYI @defo89

defo89:
@m-kratochvil thanks for pointing this out. This is something we are of (admin2 is off by default), hence our expression for admin alert is `/Common/admin1.cc.*.cloud.sap"}`
https://github.com/sapcc/helm-charts/pull/4097/files#diff-9e7f37bd85af271803809dfae2930bc32aa0f68d6c293e13ccf364e9b597c82aR114

PR #4098, opened 2022-10-07
[MariaDB] - Enable Performance Schema

viktor-kovacs:
The Performance Schema is a feature for monitoring MariaDB/MYSQL Server execution at a low level. The Performance Schema has these characteristics:

The Performance Schema provides a way to inspect internal execution of the server at runtime. It is implemented using the [PERFORMANCE_SCHEMA](https://dev.mysql.com/doc/mysql-perfschema-excerpt/8.0/en/performance-schema.html) storage engine and the performance_schema database. The Performance Schema focuses primarily on performance data. This differs from INFORMATION_SCHEMA, which serves for inspection of metadata.

The Performance Schema monitors server events. An “event” is anything the server does that takes time and has been instrumented so that timing information can be collected. In general, an event could be a function call, a wait for the operating system, a stage of an SQL statement execution such as parsing or sorting, or an entire statement or group of statements. Event collection provides access to information about synchronization calls (such as for mutexes) file and table I/O, table locks, and so forth for the server and for several storage engines.

Performance Schema can be enabled by two settings by Helm:
1. Enable the PS only without instruments
2. Enable the instruments to collect the default PS metrics

Note: The collection has been disabled by default. It can be enabled by script or from mysql. Thus this settings do not cause performance overhead.

db_performance_schema: # Enabling performance schema only (without enabling instruments for data collections).
                       # Instrument can be enabled separately. Only Performance Schema Enablement does not cause any
                       # overhead.
  enabled: true
  
  db_performance_schema_instrument: # Enabling performance schema instruments. 
                                  # In order to enable it you need to enable the db_performance_schema.
  enabled: false
  # default : false

### DIFF
--- a/common/mariadb/values.yaml
+++ b/common/mariadb/values.yaml
@@ -25,7 +25,7 @@ db_performance_schema: # Enabling performance schema only (without enabling inst
   enabled: true
   
 db_performance_schema_instrument: # Enabling performance schema instruments. 
-                                  # In order to enable it you need to enable the enable_ps.
+                                  # Please enable the db_performance_schema in order to enable the db_performance_schema_instrument.
   enabled: false
   # default : false
 
--- a/common/mariadb/values.yaml
+++ b/common/mariadb/values.yaml
@@ -9,7 +9,6 @@ imagePullPolicy: IfNotPresent
 port_public: 3306
 max_connections: 1024
 buffer_pool_size: "1024M"
-performance_schema: "OFF"
 log_file_size: "256M"
 log_warnings: "2"
 query_cache_size: "0"
@@ -19,6 +18,17 @@ binlog_format: "MIXED"
 expire_logs_days: 10
 character_set_server: 'utf8'  # Should be utf8mb4, but we started with this
 collation_server: 'utf8_general_ci'
+
+db_performance_schema: # Enabling performance schema only (without enabling instruments for data collections).
+                       # Instrument can be enabled separately. Only Performance Schema Enablement does not cause any
+                       # overhead.
+  enabled: true
+  
+db_performance_schema_instrument: # Enabling performance schema instruments. 
+                                  # In order to enable it you need to enable the enable_ps.
+  enabled: false
+  # default : false
+
 
 root_password: null
 initdb_configmap: null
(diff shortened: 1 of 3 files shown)


bashar-alkhateeb (approved):
Looks fine!

common/mariadb/values.yaml
```diff
@@ -20,6 +19,17 @@ expire_logs_days: 10
...
 
+db_performance_schema: # Enabling performance schema only (without enabling instruments for data collections).
+                       # Instrument can be enabled separately. Only Performance Schema Enablement does not cause any
+                       # overhead.
+  enabled: true
+  
+db_performance_schema_instrument: # Enabling performance schema instruments. 
+                                  # Please enable the db_performance_schema in order to enable the db_performance_schema_instrument.
+  enabled: false
+  # default : false
```
joker-at-work, line 30: I don't think we need to write the default here, as this `values.yaml` is the default - a parent chart would have to change the value in their values file to move away from the default.

joker-at-work (approved):

common/mariadb/values.yaml
```diff
@@ -25,7 +25,7 @@ db_performance_schema: # Enabling performance schema only (without enabling inst
   enabled: true
   
 db_performance_schema_instrument: # Enabling performance schema instruments. 
-                                  # In order to enable it you need to enable the enable_ps.
+                                  # Please enable the db_performance_schema in order to enable the db_performance_schema_instrument.
   enabled: false
   # default : false
```
fwiesel, line 30: Ah, I am reading that as there is a key named "default", and it its value is set to "false", and it has be commented out. Possibly because it the same as not setting it

common/mariadb/values.yaml
joker-at-work: I don't get that. Who is the intended audience?

As said, the default is, what the setting right above defines. Everything non-default would be handled in parent charts and their values files.

viktor-kovacs: It means this value has to be false for production

fwiesel: I am reading that as there is a key `default`, which is assumed to be `false`, but since it is the same as not setting it, there is a comment to indicate the existence (and type).

I would rather go with adding a comment to `enabled: false` explaining why it is on false by default

viktor-kovacs: Thanks for the feedback! I will do the update with the next PR.

But that was the explanation: # Instrument can be enabled separately. Only Performance Schema Enablement does not cause any overhead.

PR #4099, opened 2022-10-10
[thanos] option for custom args

viennaa:


common/thanos/values.yaml
```diff
@@ -78,6 +80,23 @@ compactor:
   # Minimum age of fresh (non-compacted) blocks before they are being processed. Malformed blocks older than the maximum of consistency-delay and 30m0s will be removed.
   consistencyDelay: 30m
 
+  # Specifying custom arguments means that all compactor related settings will be overwritten. It is NOT adding the options, these are all options at glance. This is usually used, when the operator is not supporting a certain option
+  customArgs: #[]
+    - compact
```
richardtief, line 85: is this intentional?

viennaa: Yes, that's the command, that's why it has no `--`

richardtief: Just saw it in the Objector CR. Seems to be valid.

richardtief (approved):

PR #4115, opened 2022-10-11
[nova] Add console deployments back again

joker-at-work:
With the commit "[nova] Remove consoleauth support" 996518a8 we removed the consoleauth-deployment.yaml - but contained not only the deployment for consoleauth, but also the deployments of the different consoles. Since we still need the consoles, we add them back now in their own file.

### DIFF
--- a/openstack/nova/templates/console-deployment.yaml
+++ b/openstack/nova/templates/console-deployment.yaml
@@ -1,0 +1,7 @@
+{{ $envAll := . }}
+{{- range $name, $config := .Values.consoles }}
+  {{- if $config.enabled }}
+---
+{{ tuple $envAll $name $config | include "nova.console_deployment" }}
+  {{- end }}
+{{- end }}


openstack/nova/templates/console-deployment.yaml
```diff
@@ -0,0 +1,7 @@
+{{ $envAll := . }}
+{{- range $name, $config := .Values.consoles }}
+  {{- if $config.enabled }}
+---
+{{ tuple $envAll $name $config | include "nova.console_deployment" }}
```
grandchild, line 5: Would you consider this for legibility?
```suggestion
    {{- tuple $envAll $name $config | include "nova.console_deployment" }}
```

grandchild (approved):
lgtm, one optional nit

openstack/nova/templates/console-deployment.yaml
grandchild: ah, sorry that doesn't work. nevermind.

PR #4121, opened 2022-10-12
swift: roll daemonsets faster

majewsky:
This is based on my own experience driving deployments in a way that does not make them a huge time sink. The most controversial part is probably the `maxUnavailable: 2` on the swift-servers, but this is also the most important in terms of reducing the time required to complete a deployment.

### DIFF
--- a/openstack/swift/templates/servers-daemonset.yaml
+++ b/openstack/swift/templates/servers-daemonset.yaml
@@ -10,6 +10,8 @@ metadata:
 spec:
   updateStrategy:
     type: RollingUpdate
+    rollingUpdate:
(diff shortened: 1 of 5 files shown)


openstack/swift/templates/servers-daemonset.yaml
```diff
@@ -10,6 +10,8 @@ metadata:
 spec:
   updateStrategy:
     type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 2 # each object has 3 replicas, so we can afford to have 2 servers missing at a time without causing outage
```
reimannf, line 14: The only problematic situation, would be `HEAD` requests for non existent objects, which would result in a `503`. But I think we can accept this.

reimannf (approved):

PR #4130, opened 2022-10-12
[Cinder] Conditionally enable FCD driver

hemna:
This patch updates the helm-chart for cinder to conditionally enable the first class disk driver.  By default it's disabled.

This will allow me to enable it for qa for testing.

### DIFF
--- a/openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
+++ b/openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
@@ -20,7 +20,7 @@ template: |
   data:
     cinder-volume.conf: |
       [DEFAULT]
-      enabled_backends = {{ .Values.backends.enabled }}
+      enabled_backends = {{ .Values.backends.enabled }}{{ if .Values.backends.vmware_fcd.enabled | default false }},vmware_fcd {{- end }}
 
       [backend_defaults]
       vmware_host_ip = {= host =}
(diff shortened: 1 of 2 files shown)


openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
```diff
@@ -20,7 +20,7 @@ template: |
   data:
     cinder-volume.conf: |
       [DEFAULT]
-      enabled_backends = {{ .Values.backends.enabled }}
+      enabled_backends = {{ .Values.backends.enabled }}{{ if .Values.backends.vmware_fcd.enabled | default false }},vmware_fcd {{- end }}
```
joker-at-work, line 23: Why are you doing this so differently than the other backends? Could just add it to `backends.enabled` in `qa` and be done, right?

openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
```diff
@@ -55,7 +55,11 @@ template: |
...
 
-      [vstorageobject]
+      {{ if .Values.backends.vmware_fcd.enabled | default false }}
+      [vmware_fcd]
       volume_driver = cinder.volume.drivers.vmware.fcd.VMwareVStorageObjectDriver
-      volume_backend_name = vstorageobject
+      volume_backend_name = vmware_fcd
+      vmware_storage_profile = {{ .Values.backends.vmware_fcd.vmware_storage_profile }}
+      extra_capabilities = '{"vcenter-shard": "{= host.split(".")[0] =}"}'
+      {{- end }}
```
joker-at-work, line 64: Why the conditional include and why not do it on `backends.enabled` containing `vmware_fcd`? It doesn't really hurt if there are more config options, I would guess, because only an enabled FCD driver would read them.

joker-at-work (commented):
Looking at the template, you can easily add an additional backend to a region by setting `backends.additional_backends`. So if you only need it in qa-de-1, then that would be an easy way.

openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
hemna: done

openstack/cinder/templates/vcenter_datacenter_cinder_configmap.yaml
hemna: I presume you mean something like this?

https://github.wdf.sap.corp/cc/secrets/pull/10001

joker-at-work: yes

PR #4131, opened 2022-10-13
[designate] Integrate locking via pvc

fwiesel:
to restarts of the nodes.

By default, stay on memcached.


fwiesel:
Apparently, coordination is not used.

PR #4160, opened 2022-10-19
[mariadb] support_group label added to alerts template

businessbean:
- it is required for the on-call alert routing
- service owners have to define that for their MariaDB instance
- chart version bumped

### DIFF
--- a/common/mariadb/Chart.yaml
+++ b/common/mariadb/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: mariadb
-version: 0.6.1
+version: 0.7.0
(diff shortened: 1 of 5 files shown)


bashar-alkhateeb (approved):
Looks fine for me!

majewsky:
I would suggest to also put this field into the `values.yaml` for reference, like here on the Postgres chart:

https://github.com/sapcc/helm-charts/blob/9158ccb18de5b0ba893ff5a3559de8e6218b5c87/common/postgresql/values.yaml#L123-L126

businessbean:
> I would suggest to also put this field into the `values.yaml` for reference, like here on the Postgres chart:
> 
> https://github.com/sapcc/helm-charts/blob/9158ccb18de5b0ba893ff5a3559de8e6218b5c87/common/postgresql/values.yaml#L123-L126
> 
> This could also obviate the need for a separate documentation page.

I will add the info to the values.yaml, but i think a central documentation (not just for this feature) will help too.

dhoeller (approved):
LGTM

PR #4162, opened 2022-10-19
[rabbitmq] add support information for rabbitmq alerts

bashar-alkhateeb:
support information must be set by overriding
Values.alerts.support_group in the wrapper chart.
the helm chart will fail if the value is not set (default value is not set for this purpose)


fwiesel (approved):

viktor-kovacs (approved):
Code looks fine to me

PR #4165, opened 2022-10-19
[memchached] add support info for alerts

bashar-alkhateeb:
it is required for the on-call alert routing

- support information added to label
- chart version bumped
- chart must fail if the Values.alerts.support_group is not set in the parent chart

### DIFF
--- a/common/memcached/Chart.yaml
+++ b/common/memcached/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: memcached
-version: 0.0.10
+version: 0.0.11
 description: Free & open source, high-performance, distributed memory object caching system.
 home: http://memcached.org/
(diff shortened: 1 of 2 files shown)


businessbean (approved):
lgtm, but would not `0.1.0` be the better version because we added a feature?

bashar-alkhateeb:
> lgtm, but would not `0.1.0` be the better version because we added a feature?

for me it more like a patch than a real change in the functionality, or an upgrade, for that i would prefer to increase the patch number only , AFAIK we don't have strict versioning schema in team, but if team think that increasing the minor version is needed for such a change, i am more than ready to fix it ;)

viktor-kovacs (approved):

joker-at-work (approved):

PR #4166, opened 2022-10-19
[nova] Bump rabbitmq to get support_group alert label

joker-at-work:
With rabbitmq chart version 0.4.4 we get to set the support_group label on the alerts defined in the rabbitmq chart.


joker-at-work:
Superseded by #4230

PR #4168, opened 2022-10-19
[manila] upgrade mariadb chart

chuan137:
Define support_group for mariadb alerts.


Carthaca:
depends on #4160

businessbean:
@Carthaca can you add a review for #4160 ?

PR #4177, opened 2022-10-20
[openstack-exporter][cinder] Update cinder alerts

hemna:
This patch adds the support_group label.
This patch changes the CinderPoolDown alert to CinderPoolOvercommited as we only care if a pool is overcommited.

### DIFF
--- a/prometheus-exporters/openstack-exporter/alerts/cinder.alerts
+++ b/prometheus-exporters/openstack-exporter/alerts/cinder.alerts
@@ -9,6 +9,7 @@ groups:
       severity: critical
       tier: vmware
       service: cinder
+      support_group: compute-storage-api
       context: "openstack-exporter"
       meta: "Cinder backend {{ $labels.shard }}/{{ $labels.backend }} has no storage capacity."
       playbook: docs/support/playbook/cinder/cinder-storage-profile-empty.html
@@ -22,6 +23,7 @@ groups:
       severity: warning
       tier: vmware
       service: cinder
+      support_group: compute-storage-api
       playbook: docs/support/playbook/cinder/cinder-low-free-space.html
     annotations:
       description: 'Cinder backend {{ $labels.shard }}/{{ $labels.backend }} has 1 or less datastores with > 30% free space left'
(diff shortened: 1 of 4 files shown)


majewsky:
This would also be a good opportunity to add owner-info, which ensures correct labeling of Kubernetes objects belonging to the chart. See my mail from yesterday for details. The tl;dr is as follows:

1. Add a chart dependency [like this](https://github.com/sapcc/helm-charts/blob/415d299d555edc9111bb0d6fca8d9f5aff67ea82/openstack/swift/Chart.yaml#L11-L13).
2. Add metadata to values.yaml [like this](https://github.com/sapcc/helm-charts/blob/415d299d555edc9111bb0d6fca8d9f5aff67ea82/openstack/swift/values.yaml#L11-L19).
3. Run "helm dep up" and commit everything.

businessbean (approved):
lgtm

businessbean:
What about bumping the [chart version](https://github.com/sapcc/helm-charts/blob/35c413711ff3fe9f4fbbb73b2d546aba494dc876/prometheus-exporters/openstack-exporter/Chart.yaml#L3)?

prometheus-exporters/openstack-exporter/values.yaml
```diff
@@ -1,4 +1,6 @@
 owner-info:
+  support-group: compute-storage-api
+  service: cinder
```
businessbean, line 3: The [owner-info template](https://github.com/sapcc/helm-charts/blob/cinder_update_alerts/common/owner-info/templates/configmap.yaml#L7) expects a string and not a collection. This commit fixes the problem.

businessbean (approved):
lgtm

prometheus-exporters/openstack-exporter/alerts/cinder.alerts
```diff
@@ -101,6 +108,7 @@ groups:
       severity: info
       tier: vmware
       service: cinder
```
businessbean, line 110: Potential further improvement: If we define the service in the `values.yaml` we should look it up here to avoid having to maintain it in multiple areas.

prometheus-exporters/openstack-exporter/values.yaml
joker-at-work, line 3: I'd leave out the service, because it's a generic "openstack" exporter - currently only used by cinder - unless you feel like only cinder-folk should maintain it.

Also one of the maintainers is not in the appropriate support_group.

joker-at-work (approved):

prometheus-exporters/openstack-exporter/values.yaml
hemna: done

PR #4194, opened 2022-10-21
openstack/limes: add alerts for inconsistencies in quotas

SuperSandro2000:


majewsky (changes requested):
We have better descriptions of what different types of inconsistencies mean and how they arise in the API spec: https://github.com/sapcc/limes/blob/master/docs/users/api-spec-resources.md?plain=1#L639-L648 - Please include these in the alert descriptions.

majewsky (approved):

PR #4202, opened 2022-10-21
alert moved from blackbox to cc3test

developer-abhi:
@notandy : please approve this, as all neutron alert moved to https://github.com/sapcc/helm-charts/blob/master/openstack/cc3test/alerts/network.alerts


Carthaca (approved):
lgtm

PR #4219, opened 2022-10-24
[ironic] Name jobs dynamically

fwiesel:
The job can usually not be changed, so instead
we give it a new name on each image change
and as well as a change of code or configuration
as far as possible.

The pxe setup does only depend on
a pvc being created, which is done outside of this chart, so it can be turned into a helm-hook

### DIFF
--- a/openstack/ironic/templates/_helpers.tpl
+++ b/openstack/ironic/templates/_helpers.tpl
@@ -1,0 +1,44 @@
+{{ define "ironic.pxe_setup_script" }}
+cd /tftpboot
+
+if [ ! -f lpxelinux.0 ]; then
+  [ -z "$a" ] && apt-get update && a=1
+  apt-get install pxelinux
+  cp /usr/lib/PXELINUX/* .
+fi
+
+if [ ! -f ldlinux.c32 ]; then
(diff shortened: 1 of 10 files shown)


sandzwerg (commented):
I'd wait for @stefanhipfel here

fwiesel:
I see @stefanhipfel added a different attempt at the problem: https://github.com/sapcc/helm-charts/commit/a996a7c456bfdd70a82185ae8de9309a174a2c60

There is though an issue: When the chart is installed, the database will be empty, so the api pods will fail and the deployments won't get ready, so the post-install hook will never be run and the deployment will fail.
The migration can also not be a pre-install hook, as there is no database, and no config-files there yet.

The only option I see is to run the migration jobs with the chart when the chart is being installed.

stefanhipfel:
> a996a7c

`post-install` will make sure that there is a database first, before the migrate job will be executed. 
-->  post-install: Executes after all resources are loaded into Kubernetes

When the job run successfully the api deployments will run, since they are waiting:
```
        - name: DEPENDENCY_JOBS
          value: "ironic-db-migration"
```

PR #4226, opened 2022-10-25
Switch to cloud_objectstore_admin as reseller in Swift

reimannf:
Swift is using the new role with this. I checked all regions and all users have the old and the new role assignments.
Remove all seeded role assignments.

Final Step is to manually delete the role `swiftreseller`


reimannf:
yes. will create the PR if this is rolled out

PR #4237, opened 2022-10-26
[nova] Add alert for failing detach_volume RPCs

joker-at-work:
There are cases when we still have duplicate BDM entries or general BDM entries without connection_info. This is one of the cases, where volume detachment will fail.

We build this alert to become aware of such problems. This alert is also the first of its kind using oslo_messaging_events metric. We would like to see if we can build alerts based on that metrics and if this helps us.


grandchild (approved):
Super-tiny-typo-nit: last sentence, "based on that metric".
Should I even spend the effort mentioning it? Oh no now I've increased the effort again. And now twice more! 😭

grandchild (approved):

PR #4246, opened 2022-10-27
[metis] adds sync-pod, adds config for vcenter guestOS sync to metis

IvoGoman:
Changes:
- sync-pod replace replication. Pulls full backup + increments from swift instead of pulling dump from DB.
- metis pod: adds config required for guestOS collection from vCenter


richardtief (approved):
Looks good.

PR #4247, opened 2022-10-27
[cinder] Change support_group to compute

hemna:
David said the alerts needed to be changed to update the support_group to compute

### DIFF
--- a/prometheus-exporters/openstack-exporter/alerts/cinder.alerts
+++ b/prometheus-exporters/openstack-exporter/alerts/cinder.alerts
@@ -9,7 +9,7 @@ groups:
       severity: critical
       tier: vmware
       service: cinder
-      support_group: compute-storage-api
+      support_group: compute
       context: "openstack-exporter"
       meta: "Cinder backend {{ $labels.shard }}/{{ $labels.backend }} has no storage capacity."
       playbook: docs/support/playbook/cinder/cinder-storage-profile-empty.html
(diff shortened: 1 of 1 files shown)


prometheus-exporters/openstack-exporter/alerts/cinder.alerts
```diff
@@ -78,7 +78,7 @@ groups:
       severity: info
       tier: vmware
       service: cinder
-      support_group: compute-storage-api
+      support_group: compute
       context: "openstack-exporter"
       meta: "Cinder pool {{ $labels.shard }}/{{ $labels.backend }}/{{ $labels.pool }} is overcommited"
       playbook: docs/support/playbook/cinder/cinder-pool-down.html
```
joker-at-work, line 84: playbook points to pool-down for an "overcommited" alert

joker-at-work (commented):
In the docs, the compute group is called `ccloud-compute`, so I cannot verify this. Asked Tilo for help.

prometheus-exporters/openstack-exporter/alerts/cinder.alerts
hemna: David told me it had to be 'compute' in slack yesterday, so I'm quite confused.

joker-at-work: it's probably correct. I just cannot verify it myself, because the docs have contradicting information. I would assume the docs are wrong here, tbh, but need Tilo to verify this.

Without being able to verify this independently, I cannot give approval. Feel free to merge anyways :)

hemna: > playbook points to pool-down for an "overcommited" alert

yes, because the pool is down.   And on the pool down documentation there are 2 possible solutions documented that describes what to do when it's down because it's overcommited.

joker-at-work: ah. right. things go down when overcommited. forgot that. thanks.

PR #4248, opened 2022-10-28
[common][mariab] make slow queries alert less flappy

Carthaca:
e.g. having a slow query every 7 minutes (seen in manila) would make this alert flap around a lot, smoothen a bit by increasing the rate to 30 min

### DIFF
--- a/common/mariadb/Chart.yaml
+++ b/common/mariadb/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: mariadb
-version: 0.7.3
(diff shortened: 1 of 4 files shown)


viktor-kovacs (approved):
It looks fine to me
Thanks

galkindmitrii:
I'd keep 3 sec by default and set it to 5 in manila values, but either is fine...

Carthaca:
> I'd keep 3 sec by default and set it to 5 in manila values, but either is fine...

I understood Viktor, that not only manila, but also nova, designate, ironic have this longer execution times

PR #4251, opened 2022-10-28
[Ironic] Limit authentication to initial connection

fwiesel:
Prior this change, nginx would expire the url for all requests. The problem here is that the javascript app running in the browser will periodically do a POST request with a session id and possibly commands
to the url, and the validation will break that.

To make matters worse, the app won't close the session normally, keeping the shell on the server running
until it timeouts.

The session-id replaces the authenticated url as a secret, so we do not need to validate the POST requests anymore.

In similar vein, validating the url for static resources is superfluous, and we can pass those requests right through.

### DIFF
--- a/openstack/ironic/templates/console-configmap.yaml
+++ b/openstack/ironic/templates/console-configmap.yaml
@@ -17,19 +17,44 @@ data:
         listen       80;
         server_name  {{ include "ironic_console_endpoint_host_public" . }};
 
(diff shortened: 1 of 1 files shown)


sandzwerg (approved):
Isn't the health check (if there is any) expecting a HTTP 200? Apart from that, looks fine as far as I'm able to judge.

fwiesel:
At least for kubernetes, anything in the range `200 <= x < 400` is considered to be healthy.

PR #4252, opened 2022-10-30
[hermes] Alert changes, matching to maia, folder structure

notque:


notque:
The intention here is to have a consistent set of alerting setups as to the other golang services. I am trying to match them, while learning about the details of why they are doing it a particular way. 

I am also taking down each component and validating that every alert is actually firing. We've had issues with alerts not being routed to the proper channels based on labels being different.

PR #4274, opened 2022-11-02
[maia] move maia-oprom alerts

richardtief:


notque:
I am merging and deploying this as an interim solution, however as we've discussed this is not tenable long term.

PR #4277, opened 2022-11-03
[vrops-exporter] Avoid using Analytics Metrics

max-len:
The "OnlineCapacityAnalytics" metrics are not reliable for monitoring current capacity. As suggested by VMware we should solely use the "capacity" metrics.
Metric keys were tested manually in qa-de-1.


richardtief (commented):
Looks good.

richardtief (approved):

PR #4281, opened 2022-11-03
[designate] rework readiness and liveness probes

galkindmitrii:
- liveness to use /healthcheck oslo middleware
- readiness to additionally check if DB is up


occamshatchet (approved):
Looks good.

PR #4283, opened 2022-11-03
[maia] scrape jobs with single match query

viennaa:
given the amount of data returned by single matches in large regions respective queries should be as minimal as possible. This will reduce scrape times to a minimum. Only implemented for prometheus-vmware targets, as it is only needed there

Additionally: neo federate is in a dedicated job now if applicable


richardtief (approved):
All good.

PR #4285, opened 2022-11-03
[cinder] Update cinder attaching alert to 1 hour

hemna:
Since we have potentially long attachment times, we should be warned if a volume is stuck in attaching state for more than 1 hour.

### DIFF
--- a/openstack/cinder/alerts/openstack-cinder.alerts
+++ b/openstack/cinder/alerts/openstack-cinder.alerts
@@ -2,7 +2,7 @@ groups:
 - name: openstack-cinder.alerts
   rules:
   - alert: OpenstackCinderVolumeInDeletingState
-    expr: sum(openstack_stuck_volumes_max_duration_gauge{status="deleting"}) BY (id,display_name) > 15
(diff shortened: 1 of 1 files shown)


openstack/cinder/alerts/openstack-cinder.alerts
```diff
@@ -17,22 +17,22 @@ groups:
...
+      severity: warning
       tier: os
     annotations:
       description: Volume {{ $labels.display_name }} with {{ $labels.id }} in Attaching State
-      summary: Cinder Volumes taking more than 15s to attach
+      summary: Cinder Volumes taking more than 1 hour to attach
 
   - alert: OpenstackCinderVolumeInDetachingState
-    expr: sum(openstack_stuck_volumes_max_duration_gauge{status="detaching"}) BY (id,display_name) > 15
+    count(sum(openstack_stuck_volumes_max_duration_gauge{status="detaching"}) BY (id,display_name) > 15) by (id, display_name)
```
joker-at-work, line 35: You've removed the `expr:` YAML key here.

joker-at-work (changes requested):

joker-at-work (commented):
Why does the `OpenstackCinderVolumeInDeletingState` not need the same fix? From my current theory, that one would never fire either.

joker-at-work (approved):

PR #4291, opened 2022-11-07
[prometheus-server] update PrometheusMultipleTargetScrapes alert

richardtief:


common/prometheus-server/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 7.0.11
+
+* Exclude `job=prometheus-vmware` from PrometheusMultipleTargetScrapes alert
```
viennaa, line 3:
```suggestion
* Exclude `job=prometheus-vmware` from PrometheusMultipleTargetScrapes alert because this target is scraped with multiple jobs to the same target intentionally. Every scrape just replies a single metric.
```

viennaa (changes requested):
Thanks for taking this. Can you please integrate this change to the pre7 chart too?

viennaa (approved):

PR #4292, opened 2022-11-07
[kube-system-metal] add owner-label-injector

onuryilmaz:


databus23 (changes requested):
umbrella-chart needs a version bump

databus23 (approved):

PR #4298, opened 2022-11-08
[prometheus-server] Enablement to create multiple prometheus

richardtief:
Enabling the common chart to deploy multiple Prometheus. This can be done either via the list `names: []` or via a list of globally available targets to create Prometheis `global.targets: []`. Existing configurations will not break as this chart is still able to create a single Prometheus.


common/prometheus-server/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 7.1.0
+
+* Enabling the common chart to deploy multiple Prometheus. This can be done either via the list `names: []` or via a list of targets to create Prometheis `global.targets: []`. Existing configurations will not break as this chart is still able to create single Prometheus.
```
viennaa, line 3:
```suggestion
* Enabling the common chart to deploy multiple Prometheus. This can be done either via the list `names: []` or via a list of targets to create Prometheis `global.targets: []`. Existing configurations will not break as this chart is still able to create single Prometheus configured with `name:` only. 
```

common/prometheus-server/templates/_helpers.tpl
```diff
@@ -1,32 +1,61 @@
-{{/* Name of the Prometheus. */}}
+{/* Name of the Prometheus. */}}
```
viennaa, line 1:
```suggestion
{{/* Name of the Prometheus. */}}
```

As other comments are always double `{{`, this should be the case here too

common/prometheus-server/templates/_helpers.tpl
```diff
@@ -1,32 +1,61 @@
...
-{{- if and .Values.ingress.hosts .Values.ingress.hostsFQDN -}}
+{{- $name := index . 0 -}}
+{{- $root := index . 1 -}}
+{{- if and $root.Values.ingress.hosts $root.Values.ingress.hostsFQDN -}}
 {{- fail ".Values.ingress.hosts and .Values.ingress.hostsFQDN are mutually exclusive." -}}
 {{- end -}}
-{{- if .Values.ingress.hosts -}}
-{{- $firstHost := first .Values.ingress.hosts -}}
-{{- required ".Values.ingress.hosts must have at least one hostname set" $firstHost -}}.{{- required ".Values.global.region missing" .Values.global.region -}}.{{- required ".Values.global.domain missing" .Values.global.domain -}}
+prometheus-{{- $name -}}.{{- required "$root.Values.global.region missing" $root.Values.global.region -}}.{{- required "$root.Values.global.domain missing" $root.Values.global.domain -}}
```
viennaa, line 37: `.Values.ingress.hostsFQDN` or `.Values.ingress.hosts` are not being taken into consideration here. We're checking for them L34, but the names are independently set in L37.

I've read the comment in [values](https://github.com/sapcc/helm-charts/blob/0a359337be238b9b70ed5e89abe43152f0fc3b58/common/prometheus-server/values.yaml#L87) and it is a smart move, to technically disable this functionality if multiple prometheis are deployed, but L37 should use one of `.Values.ingress.hostsFQDN` or `.Values.ingress.hosts` in the default cause (1 name). If it's more, then we go like implemented. Or am I missing something here?

viennaa (changes requested):
Thanks for providing this very mature PR. Tested it against an existing Prometheus and there is no diff, apart from normal changes like versions.

I have some minor adjustments to comments and one thing to discuss, which is the ingress handling, which does not add up to me.

common/prometheus-server/templates/_helpers.tpl
```diff
@@ -1,32 +1,61 @@
-{{/* Name of the Prometheus. */}}
+{/* Name of the Prometheus. */}}
 {{- define "prometheus.name" -}}
-{{- required ".Values.name missing" .Values.name -}}
+{{- $name := index . 0 -}}
+{{- $root := index . 1 -}}
+{{- if and (not $root.Values.names) (not $root.Values.global.targets) (not $root.Values.name) -}}
+{{- fail "Connot create any Prometheus resource. Please define a name or at least one list element to names or global.targets" -}}
```
viennaa, line 6:
```suggestion
{{- fail "Can not create any Prometheus resource. Please define a name or at least one list element to names or global.targets" -}}
```

viennaa (changes requested):

common/prometheus-server/templates/_helpers.tpl
richardtief: You're right, the first host in `.Values.ingress.hosts` must be considered. `.Values.ingress.hostsFQDN` does not need an externalURL helper because the FQDN is already specified.

auhlig:
Not sure if that's what you want, but in kube-monitoring we already create multiple Prometheis, namely collector and frontend, using the common chart:
```
dependencies:
  - alias: prometheus-frontend
    name: prometheus-server
    version: 7.0.8
  - alias: prometheus-collector
    name: prometheus-server
    version: 7.0.8
...
```
Each has its own values according to its alias like
```
prometheus-collector:
  ...

prometheus-frontend:
  ...
```

richardtief:
No. Basically, we want to have common values for multiple Prometheus created for different monitoring targets. For VMware in particular, we need to scale Prometheus horizontally because vertical scaling is limited. Then we integrate that with Thanos.

viennaa (approved):

viennaa:
Alright, let's give it a spin. Thanks :)

PR #4299, opened 2022-11-08
[common thanos] Enablement to create multiple thanos

richardtief:
Enabling the common chart to deploy multiple Thanos along Prometheus. This can be done either via the list names: [] or via a list of targets to create Thanos global.targets: [].


richardtief:
Moved to https://github.com/sapcc/helm-charts/pull/4301

PR #4300, opened 2022-11-08
Introducing vmware-monitoring

richardtief:
Merge `vrops-exporter` and `prometheus-vmware` to create resources for multiple monitoring targets. 

The chart requires the newly implemented dependency prometheus-server 7.1.0 and thanos 0.2.0, which are capable of creating multiple resources. 

vrops-exporter alerts are still deployed with this chart. They will be moved to a separate chart in the future.

### DIFF
--- a/system/vmware-monitoring/values.yaml
+++ b/system/vmware-monitoring/values.yaml
@@ -1,0 +1,260 @@
+global:
+  domain: cloud.sap
+  clusterType: controlplane
+  alerts:
+    enabled: true
+    prometheus: vmware
+
+  aggregations:
+    enabled: true
+    prometheus: vmware
+
+owner-info:
+  support-group: observability
+  maintainers: 
+    - Richard Tief
+    - Tommy Sauer
+  helm-chart-url: https://github.com/sapcc/helm-charts/tree/master/system/vmware-monitoring
+
+thanos:
+  vmware: true
+  deployWholeThanos: true
+
+prometheus-server:
+  vmware: true
+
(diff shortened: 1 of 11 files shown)


richardtief:
The validation error is intentional. The dependencies must be merged first. 
`Error: can't get a valid version for repositories prometheus-server, thanos.`

system/vmware-monitoring/values.yaml
```diff
@@ -0,0 +1,260 @@
...
+        authTLSVerifyDepth: 3
+        authTLSVerifyClient: on
+
+  internalIngress:
+    enabled: false
+
+  persistence:
+    enabled: true
+    accessMode: ReadWriteOnce
+    size: 100Gi
```
viennaa, line 52:
```suggestion
    size: 10Gi
```

This is the metal variant, let's see where we go here. If we go to scaleout, we need to adjust anyway.

system/vmware-monitoring/values.yaml
```diff
@@ -0,0 +1,260 @@
...
+  helm-chart-url: https://github.com/sapcc/helm-charts/tree/master/system/vmware-monitoring
+
+thanos:
+  vmware: true
+  deployWholeThanos: true
+
+prometheus-server:
+  vmware: true
+
+  retentionTime: 4h
```
viennaa, line 26:
```suggestion
  retentionTime: 7d
```

viennaa (changes requested):
Thanks for finishing up, only some minor things.

PR #4311, opened 2022-11-09
remove prometheus-exporters/pvc-exporter chart

majewsky:
Since Kubernetes 1.20, the kubelet itself provides PVC usage metrics (`kubelet_volume_stats_...`). At the same time, the metrics provided by the pvc-exporter are deteriorating more and more. For example, in na-us-1, I'm currently seeing only 1 timeseries in the `pvc_usage` metric, compared to over 100 timeseries in the `kubelet_volume_stats_available_bytes` metric.

I have added a [new alert in prometheus-controlplane-rules](https://github.com/sapcc/helm-charts/blob/5791c9a7f5a8261b123e0cfdb5294801a133412f/prometheus-rules/prometheus-controlplane-rules/alerts/kubernetes-node.alerts#L105-L120) to supplant the one from this chart.


defo89 (approved):
lgtm but do we want to have this alert land in our queue? shouldn’t we pick right support group based on labels?

defo89:
FYI @galkindmitrii since you drove the initial deployment with Mike

PR #4312, opened 2022-11-09
[Manila] Add NODE_NAME as host for multi-scheduler deployment

kpawar-sap:


Carthaca:
Did you try what happens if you remove line https://github.com/sapcc/helm-charts/blob/2bf7b250b444bb1711ee8f4afa2d270ff3e1018a/openstack/manila/templates/scheduler-deployment.yaml#L139 - maybe that is enough?

PR #4314, opened 2022-11-10
Moving vmware alerts to a dedicated chart

richardtief:


prometheus-rules/prometheus-vmware-rules/values.yaml
```diff
@@ -0,0 +1,2 @@
+# The name of the Prometheus to which the alert- and aggregation rules will be assigned to.
+prometheusName: vmware
```
viennaa, line 2: What happens with the sharded prometheus setup here. Haven't checked the label yet, but i'd expect to have a label `prometheus: vmware-vc-a-0` or similar. or is this the in between approach, with the other PR upcoming to be merged after the sharded setup?

richardtief: This chart reflects the current state. We can easily extend it by looping through the same secrets file with the different targets.

richardtief: Yes.
> is this the in between approach, with the other PR upcoming to be merged after the sharded setup?

viennaa (approved):
Thanks for moving over the alerts. It's good from my pov, just opened a discussion in the respective team channel, if this folder is looped or not (hence every folder would need to reflect a cluster)

PR #4323, opened 2022-11-11
[nova] Update policy for system-scoped endpoints

joker-at-work:
System-scoped endpoints got their "target" dict set to an empty dict which made our previous way to write the policy break. Instead of having only "rule:owner", we now also allow "rule:projectless" for view, member and admin rules. This makes sure that these roles are also allowed to use non-project endpoints e.g. listing availability zones where the target is empty.

Previously, those endpoints did not set an explicit target and thus got
the default: the token's project_id.

### DIFF
--- a/openstack/nova/templates/etc/_nova-policy.yaml.tpl
+++ b/openstack/nova/templates/etc/_nova-policy.yaml.tpl
@@ -2,10 +2,10 @@
 #context_is_admin: role:admin
 context_is_admin: role:cloud_compute_admin
 
-owner: project_id:%(project_id)s
-member: role:member and rule:owner
-viewer: role:compute_viewer and rule:owner
-admin: (role:compute_admin or role:compute_admin_wsg) and rule:owner
+owner_or_no_project: project_id:%(project_id)s or ccloud_no_project_id_in_target:True
+member: role:member and rule:owner_or_no_project
+viewer: role:compute_viewer and rule:owner_or_no_project
+admin: (role:compute_admin or role:compute_admin_wsg) and rule:owner_or_no_project
 context_is_compute_admin: rule:context_is_admin or rule:admin
 context_is_editor: rule:context_is_compute_admin or rule:member
 context_is_viewer: rule:context_is_editor or rule:viewer


joker-at-work:
requires https://github.com/sapcc/nova/pull/373

openstack/nova/templates/etc/_nova-policy.yaml.tpl
```diff
@@ -3,9 +3,10 @@
 context_is_admin: role:cloud_compute_admin
 
 owner: project_id:%(project_id)s
```
grandchild, line 5:
```suggestion
owner_or_no_project: project_id:%(project_id)s or ccloud_no_project_id_in_target:True
```

openstack/nova/templates/etc/_nova-policy.yaml.tpl
```diff
@@ -3,9 +3,10 @@
...
-member: role:member and rule:owner
-viewer: role:compute_viewer and rule:owner
-admin: (role:compute_admin or role:compute_admin_wsg) and rule:owner
+projectless: ccloud_no_project_id_in_target:True
+member: role:member and ( rule:owner or rule:projectless )
+viewer: role:compute_viewer and ( rule:owner or rule:projectless )
+admin: (role:compute_admin or role:compute_admin_wsg) and ( rule:owner or rule:projectless )
```
grandchild, line 9:
```suggestion
member: role:member and rule:owner_or_no_project
viewer: role:compute_viewer and rule:owner_or_no_project
admin: (role:compute_admin or role:compute_admin_wsg) and rule:owner_or_no_project
```

grandchild (changes requested):
I'd merge the `projectless` thing into `owner` and rename that. `owner` is only used in the admin/member/viewer rules.

The commit message doesn't really explain it well for me:

```diff
-got their "target" dict set to an empty dict which made our previous
-way to write the policy break.
+got their "target" dict (containing user and project ID) set to an
+empty dict, i.e. missing the project_id required for the `rule:owner`
+check which made our policy fail.
```

```diff
-Instead of having only "rule:owner", we now also allow "rule:projectless"
-for view, member and admin rules.
+Instead of matching only on project_id which might now be missing, we
+also explicitly allow requests without project_id for view, member and
+admin rules.
```


```diff
-and thus got the default: the token's project_id.
+and thus got the default: the token's project_id, which made the rule pass.
```

grandchild (approved):

joker-at-work:
This PR had my approval, too :)

PR #4330, opened 2022-11-14
Adding VCClusterHasNotEnoughResourcesToDistributeNSXTControllers alert

christopherhans:
Every NSX-T Management Cluster consists out of 3 nodes. Each node must be placed on a different host. The alert triggers, if we have less than 3 hosts with enough resources.


prometheus-rules/prometheus-vmware-rules/alerts/vccluster.alerts
```diff
@@ -94,3 +94,17 @@ groups:
     annotations:
       description: "Cluster {{ $labels.vccluster }} DRS configuration is not set to fully automated."
       summary: "Cluster {{ $labels.vccluster }} DRS configuration is not set to fully automated."
+
+  - alert: VCClusterHasNotEnoughResourcesToDistributeNSXTControllers
```
richardtief, line 98:
```suggestion
  - alert: VCClusterCannotDistributeNSXTControllers
```

richardtief (approved):

PR #4336, opened 2022-11-14
[prometheus-monitors] subchart for Pod|ServiceMonitors with default relabelings

IvoGoman:
For proper alert routing we require certain labels to be attached to a metric (e.g. region, cluster_type). Many Pod|ServiceMonitors in this repository do not set these labels. The problem is that every chart maintainer must be aware of these mandatory labels for alert routing.
Previously this was not an issue, since most pods/services were scraped by one scrape job discovering targets via the prometheus scrape annotations.
This Subchart should provide a similarly convenient  option of adding a Pod|ServiceMonitor with a certain set of defaults and the option to further adjust the scrape configuration.

Example metric with default relabelings:
https://prometheus-infra-collector.qa-de-1.cloud.sap/graph?g0.expr=metis_replication_restart_counter%7Bpod%3D~%22mariadb-sync-nova-api.*%22%7D&g0.tab=1&g0.stacked=0&g0.show_exemplars=0&g0.range_input=1h

Example PR for basic usage: #4337

### DIFF
--- a/common/prometheus-monitors/templates/podmonitor.yaml
+++ b/common/prometheus-monitors/templates/podmonitor.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.podMonitor.enabled }}
+{{- if and .Values.podMonitor.enabled ( not .Values.serviceMonitor.enabled ) }}
 apiVersion: monitoring.coreos.com/v1
 kind: PodMonitor
 metadata:
--- a/common/prometheus-monitors/templates/podmonitor.yaml
+++ b/common/prometheus-monitors/templates/podmonitor.yaml
@@ -1,4 +1,7 @@
-{{- if and .Values.podMonitor.enabled ( not .Values.serviceMonitor.enabled ) }}
+{{- if .Values.podMonitor.enabled }}
+{{- if .Values.serviceMonitor.enabled }}
+{{ fail "failed to render PodMonitor. ServiceMonitor is also enabed. Only one is allowed."}}
+{{- end }}
 apiVersion: monitoring.coreos.com/v1
 kind: PodMonitor
 metadata:
@@ -7,13 +10,22 @@ metadata:
     prometheus: {{ required "required '.Values.prometheus' is not specified" .Values.prometheus }}
 
 spec:
+  jobLabel: {{ .Release.Namespace }}
   selector:
+{{- if and ( empty .Values.matchLabels ) ( empty .Values.matchExpressions ) }}
+{{ fail "required either '.Values.matchLabels' or '.Values.matchExpressions' but neither is defined" }}
+{{- end }}
+
+{{- if .Values.matchLabels }}
     matchLabels:
-    
-    {{- range ( required "required '.Values.labelSelector' is not specified" .Values.labelSelector ) }}
-      {{ .key }}: {{ .value | quote }}
-    {{- end }}
-  
+{{ .Values.matchLabels | toYaml | indent 6 }}
+{{- end }}
+
+{{- if .Values.matchExpressions }}
+    matchExpressions:
+{{ .Values.matchExpressions | toYaml | indent 6 }}
+{{- end }}
+
   namespaceSelector:
     matchNames:
 {{  default (list .Release.Namespace ) .Values.namespaces | toYaml | indent 6 }}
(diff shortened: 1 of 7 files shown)


common/prometheus-monitors/templates/podmonitor.yaml
```diff
@@ -0,0 +1,56 @@
+{{- if .Values.podMonitor.enabled }}
+{{- if .Values.serviceMonitor.enabled }}
+{{ fail "failed to render PodMonitor. ServiceMonitor is also enabed. Only one is allowed."}}
+{{- end }}
```
richardtief, line 4:
```suggestion
{{- if and .Values.podMonitor.enabled .Values.serviceMonitor.enabled }}
{{ fail "Could not render Podmonitor. '.Values.podMonitor.enabled' and '.Values.serviceMonitor.enabled' are mutually exclusive."}}
{{- end }}
```

common/prometheus-monitors/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,49 @@
+{{- if .Values.serviceMonitor.enabled }}
+{{- if .Values.podMonitor.enabled }}
+{{ fail "failed to render ServiceMonitor. PodMonitor is also enabed. Only one is allowed."}}
+{{- end }}
```
richardtief, line 4: {{- if and .Values.podMonitor.enabled .Values.serviceMonitor.enabled }}
{{ fail "Could not render Servicemonitor. '.Values.podMonitor.enabled' and '.Values.serviceMonitor.enabled' are mutually exclusive."}}
{{- end }}

common/prometheus-monitors/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+  clusterType: controlplane
+
+podMonitor:
+  enabled: false
+serviceMonitor:
+  enabled: false
+
+prometheus: ""
+
+namespaces: []
```
richardtief, line 13:
```suggestion
# List of namespace names from which the Endpoint objects are selected.
namespaces: []
```

common/prometheus-monitors/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+matchExpressions: []
```
richardtief, line 15:
```suggestion
# matchExpressions is a list of label selector requirements. The requirements are ANDed.
matchExpressions: []
```

common/prometheus-monitors/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+matchLabels: []
```
richardtief, line 16:
```suggestion

# matchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels map
# is equivalent to an element of matchExpressions, whose key field is "key", the operator
# is "In", and the values array contains only "value". The requirements are ANDed.
matchLabels: []
```

common/prometheus-monitors/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+customRelabelings: []
+customMetricRelabelings: []
```
richardtief, line 19:
```suggestion

# RelabelConfig allows dynamic rewriting of the label set, being applied to
# samples before ingestion.
customMetricRelabelings: []
```

common/prometheus-monitors/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+  enabled: false
+
+prometheus: ""
+
+namespaces: []
+
+matchExpressions: []
+matchLabels: []
+
+customRelabelings: []
```
richardtief, line 18:
```suggestion
#  RelabelConfigs to apply to samples before scraping.
customRelabelings: []
```

richardtief (changes requested):
Thank you for taking this on. I just have some suggestions like some descriptions on the values.

common/prometheus-monitors/templates/podmonitor.yaml
IvoGoman: This is intentionally split. The outer if is for determining if the template should be rendered the inner if to fail if both service and pod monitor are enabled.

common/prometheus-monitors/templates/servicemonitor.yaml
IvoGoman: This is intentionally split. The outer if is for determining if the template should be rendered the inner if to fail if both service and pod monitor are enabled.

common/prometheus-monitors/templates/podmonitor.yaml
richardtief: I see. Overlooked it at first glance.

common/prometheus-monitors/templates/servicemonitor.yaml
richardtief: All good.

richardtief (approved):
Thanks! Good to go

PR #4339, opened 2022-11-15
federate baremetal regression test status from cc3tests

artherd42:


global/prometheus-infra/templates/_prometheus.yaml.tpl
```diff
@@ -133,6 +133,7 @@
       - '{__name__=~"^node_logind_sessions", class="user", job="jumpserver", type="tty"}'
       - '{__name__=~"^prom_fluentd_.+"}'
       - '{job="netbox", __name__!~"^(up|ALERTS.*|scrape.+)"}'
+      - '{__name__=~"^cc3test_status", service="ironic", type="baremetal_and_regression"}'
```
richardtief, line 136: Looks good. But the query doesn't match any time series currently. Is this intended?

richardtief: I see. Looked in two other regions.

PR #4348, opened 2022-11-16
[mariadb] adds toggle for prechangejob

IvoGoman:
The PreChange will never succeed in case the DB was not yet deployed or is not up for some reason.
This toggle is a convenience to be used in this scenarios to avoid the deployment to fail.

### DIFF
--- a/common/mariadb/Chart.yaml
+++ b/common/mariadb/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: mariadb
-version: 0.7.5
+version: 0.7.6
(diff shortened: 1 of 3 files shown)


fwiesel:
I am not too happy with the change, I fear it will be misused:

The hook should not run on the initial installation:
```
"helm.sh/hook": pre-upgrade,pre-rollback
```

That leaves us with a helm upgrade failing due to the db being down.
My argument would be, that the upgrade needs to fail, since the settings could not be applied, so the desired state as declared could not be achieved. 

You can also run helm with `--no-hooks` to work around it.

IvoGoman:
--no-hooks does pretty much what I need. Thanks for bringing it to my attention.

PR #4349, opened 2022-11-16
[nova] Set initial allocation ratios

joker-at-work:
These settings have defaults we do not want - e.g. 16 x CPU overcommit and 1.5 x RAM overcommit. They are only used to set defaults when a new ComputeNode object is created and during an online data migration when upgrading to Xena if they are set to None or 0. The allocation ratios can then overwritten by the nova-compute service later on.

### DIFF
--- a/openstack/nova/templates/etc/_nova.conf.tpl
+++ b/openstack/nova/templates/etc/_nova.conf.tpl
@@ -203,3 +203,8 @@ default_pool_size = {{ .Values.wsgi_default_pool_size | default .Values.global.w
 
 [workarounds]
 enable_live_migration_to_old_hypervisor = True
+
+[compute]
+initial_cpu_allocation_ratio = 1.0
+initial_ram_allocation_ratio = 1.0
+initial_disk_allocation_ratio = 1.0


openstack/nova/templates/etc/_nova.conf.tpl
```diff
@@ -203,3 +203,8 @@ default_pool_size = {{ .Values.wsgi_default_pool_size | default .Values.global.w
 
 [workarounds]
 enable_live_migration_to_old_hypervisor = True
+
```
grandchild, line 206: An explanatory comment would be helpful here, since these have little-to-no effect and are merely a safeguard (IIUC).

grandchild (changes requested):
```diff
-The allocation ratios can then overwritten by the nova-compute service later on.
+The allocation ratios are then overwritten by the nova-compute service later on.
```
or maybe "can be", but they most definitely "are" being overwritten, from what you told me.

openstack/nova/templates/etc/_nova.conf.tpl
joker-at-work: I don't think so. If you want to read up on it, look into the commit adding it.

grandchild (approved):

PR #4353, opened 2022-11-18
[Alert Manager] - route clean up

bassel-z:
Content:
* remove dev-null receiver
* make elastic as a default receiver


geisslet (approved):
lgtm.
let's roll out on monday

PR #4355, opened 2022-11-18
[nova] Shellinabox & Dell KMS Proxy in Nginx

fwiesel:
This implements the shellinabox proxy in nginx
as lua scripts, and also solves proxying kvm
for dell hosts.

The existing custom solution in nova only proxies
to a single host, and cannot be changed easily
to multiple.

For the Dell KMS, it also needs to
proxy websockets, which is simply a configuration
option in nginx, but would otherwise need
to be implemented in the python proxy

### DIFF
--- a/openstack/nova/shellinabox/revproxy.lua
+++ b/openstack/nova/shellinabox/revproxy.lua
@@ -1,0 +1,257 @@
+local _M = {} -- That will be the module we return in the end
+
+local config = require "config" -- Hold the config rendered by the helm-chart
+
+local cjson = require "cjson"
+local resty_sha256 = require "resty.sha256"
+local str = require "resty.string"
+local memcached = require "resty.memcached"
+local mysql = require "resty.mysql"
+
+local guid_pattern = "%x%x%x%x%x%x%x%x%-%x%x%x%x%-%x%x%x%x%-%x%x%x%x%-%x%x%x%x%x%x%x%x%x%x%x%x"
+local url_base = "/shellinabox/" -- TODO: Get the "shellinabox" from the proxy dynamically somehow
+
+-- Little wrapper function, which avoids a
+-- potentially expensive function call "f"
+-- by caching the result in the ngx.ctx, which
+-- is preserved over all phases of a request
+local function store_in_ctx(key, f)
+    local function caching_function(...)
+        local value = ngx.ctx[key]
+        if value ~= nil then return value end
+        value = f(...)
+        if value ~= nil then ngx.ctx[key] = value end
+        return value
+    end
+
+    return caching_function
+end
+
+local function sha256_hex(token)
+    local sha256 = resty_sha256:new()
+    sha256:update(token)
+    local final  = sha256:final()
+    if final then return str.to_hex(final) end
+end
+
+-- Converts the entry to the upstream with an optional port
+local function host_from_entry(entry)
+    if entry.port and entry.port ~= 443 then
+        return "https://" .. entry.host .. ":" .. entry.port
+    else
+        return "https://" .. entry.host
+    end
+end
+
+-- set_keepalive puts a socket back in a pool,
+-- that may fail so we log it as a warning
+-- That is needed for memcached as well as mysql
+local function logged_keepalive(sock)
+    local ok, err = sock:set_keepalive(10000, 50)
+    if not ok then ngx.log(ngx.WARN, "failed to set keepalive: ", err) end
+end
+
+-- Extracts the token from either the query parameter
+-- the url, or from the referrer
+-- Caches the value, as it won't change over the request
+-- and apparently ngx.var lookups are not the cheapest
+local get_token = store_in_ctx("token", function()
+    local var_token = ngx.var.token
+    if var_token then return var_token end
+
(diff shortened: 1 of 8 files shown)


fwiesel:
This goes together with the change in nova sapcc/nova#379

openstack/nova/shellinabox/revproxy.lua
```diff
@@ -0,0 +1,226 @@
...
+    local new_location = token_base .. path .. "?" .. new_query
+    ngx.header.Location = new_location
+end
+
+function _M.process()
+    local entry = _M.lookup()
+    if not entry then return end
+    local host = host_from_entry(entry)
+    local request_uri = ngx.var.request_uri
+    local sub_uri = request_uri:sub(38)
```
joker-at-work, line 212: Can you point me to any docs for this `sub()`? I can only find `string.sub()` and would like to find out what the `38` might mean here.

openstack/nova/shellinabox/revproxy.lua
```diff
@@ -0,0 +1,226 @@
+local _M = {}
+local cjson = require "cjson"
+local config = require "config"
+local resty_sha256 = require "resty.sha256"
+local str = require "resty.string"
+local memcached = require "resty.memcached"
+local mysql = require "resty.mysql"
+local mlcache_loaded, mlcache = pcall(require, "resty.mlcache")
+local x = "%x"
+local t = { x:rep(8), x:rep(4), x:rep(4), x:rep(4), x:rep(12) }
```
joker-at-work, line 10: What that do? Is that string-formatting for numbers?

joker-at-work (commented):
Do you have some hint for a lua/openresty tutorial I could read so this review of mine could make any form of sense?

openstack/nova/shellinabox/revproxy.lua
```diff
@@ -0,0 +1,226 @@
...
+    return t1
+end
+
+local function host_from_entry(entry)
+    if entry.port and entry.port ~= 443 then
+        return "https://" .. entry.host .. ":" .. entry.port
+    else
+        return "https://" .. entry.host
+    end
+end
```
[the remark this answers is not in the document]
fwiesel: Yes, the code would be simpler. I find it though more difficult to parse when reading the logs.

openstack/nova/shellinabox/revproxy.lua
fwiesel: Well, yes and no. It is the same "code", but for matching [patterns](https://www.lua.org/manual/5.3/manual.html#6.4.1). They are similar to regular expressions, but less powerful/complex.
The alternative would be hard-coding the "%x%x%x..-%x%x"... not sure if that is easier to read. I was fearing, I might miss a `%x` somewhere.

joker-at-work: Ah. It means "repeat". Thanks.

fwiesel:
The canonical introduction would be [Programming in Lua](https://www.lua.org/pil/contents.html).
Personally, I didn't read it cover from cover, but I often end up there looking up some basics again.
Which happed when writing this too, as I also do not remember how the functions are called here again

I do not claim to have mastered the language, but I think the more difficult part is rather the nginx library.
I look most thinks up in the [nginx-api for lua](https://github.com/openresty/lua-nginx-module#nginx-api-for-lua) 
Or about  [nginx request phases](https://nginxguts.com/2011/01/phases.html)

Coming from python, the language shouldn't be to hard. Both come from an educative background as imperative with object orientation added after the fact. Especially at the scope we are looking (no inheritance, metatables...)

In python the map doing the object orientation is hidden as `__dict__` , and there are some magic methods which handle the member/attribute resolution  (`__getattr__`, `__getattribute__`).  That is handled in lua by "syntactic sugar".

A very short introduction form my side:

`x.y`  ⇔ `x["y"]`
`x:y(...)` ⇔ `x.y(x, ...)`
`function a(...)`  ⇔ `a = function(...)`, similar to python (or javascript) where the function declaration returns the function itself
So in summary: `function a.b(x, ...)`  ⇔ `a["b"] = function(x, ...)`
`[[...]]` ⇔ `"..."`
`..`  concatenate objects as string (`nil` doesn't work though)
`...` varargs in lua, which can be passed on to other functions easily.
So, this is actual code: `function a(...) print(...) end` and simply passes all arguments passed to `a` over to `print`. 

In lua there is no `@` for decorators, so I called them explicitly.

All variables should be declared `local` because like in javascript, they went with a global default.

Similar to python, when a module is required, the code is executed once the first time it is imported.
But in contrast to python there is no implicit namespace where the file lies. 
The namespaces result from the fact, that `require` returns what the module returns when it is executed, and there you create a table, which contains what you want to be exported.
By declaring objects `local` and not explicitly adding them to the returned table they become `private` in other languages.

Like in python, `"a.b"` in a module translates to `a/b` in the path.

openstack/nova/shellinabox/revproxy.lua
fwiesel: I've added some more comments, I hope that clarifies things a bit more

PR #4356, opened 2022-11-18
Helm Chart for future generation slack app

uwe-mayer:
Future generation slack app running in an alpine-Deno container interacting with internal github and `convergedcloud` slack workspace


jknipper (approved):
Looks good to me!

PR #4358, opened 2022-11-21
ttl reject added

niranba:
added for cronus reject cache ttl


dhalimi (commented):
You need to add a default value in values.yaml

PR #4364, opened 2022-11-22
[Keystone] Migrate pdb manifests to use the policy/v1 API version

rajivmucheli:
policy/v1beta1 API version of PodDisruptionBudget will no longer be served in v1.25

Ref : https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125


defo89 (approved):
lgtm

PR #4366, opened 2022-11-22
[prometheus-monitors] adds basicAuth, minor improvements

IvoGoman:


richardtief (approved):
All good. I like the omission of the default naming.

PR #4379, opened 2022-11-24
maia: add federation for Keppel metrics

majewsky:
Based on a customer request, I added the metric `keppel_manifests_with_unclean_vuln_status_count` in af5e6d6f2. Our pgmetrics end up in prometheus-openstack, so this needs to be federated into Maia.

### DIFF
--- a/openstack/maia/templates/etc/_maia_scrape_config.yaml.tpl
+++ b/openstack/maia/templates/etc/_maia_scrape_config.yaml.tpl
@@ -96,6 +96,7 @@
       - '{__name__=~"^openstack_.+",project_id!=""}'
       - '{__name__=~"^limes_(?:project|domain)_(?:quota|usage)$"}'
       - '{__name__=~"^limes_swift_.+",project_id!=""}'
+      - '{__name__=~"^keppel_.+",project_id!=""}'
 
 - job_name: 'prometheus-infra-collector'
   scrape_interval: 1m


openstack/maia/templates/etc/_maia_scrape_config.yaml.tpl
```diff
@@ -96,6 +96,7 @@
       - '{__name__=~"^openstack_.+",project_id!=""}'
       - '{__name__=~"^limes_(?:project|domain)_(?:quota|usage)$"}'
       - '{__name__=~"^limes_swift_.+",project_id!=""}'
+      - '{__name__=~"^keppel_.+",project_id!=""}'
```
richardtief, line 99: Info to @viennaa and @notque: this looks way too inclusive, but it only matches a few hundred metrics per region.

richardtief (approved):
All good.

openstack/maia/templates/etc/_maia_scrape_config.yaml.tpl
notque: Provided it's keppel metrics with a project_id, it seems reasonable that those are available to the customer.

notque (approved):
LGTM

openstack/maia/templates/etc/_maia_scrape_config.yaml.tpl
notque: I guess I should add we should look at what metrics exist, and add those to the documentation for expected metrics.

majewsky: Already done in cc/documentation-customer#355.

PR #4380, opened 2022-11-24
openstack/clair: give the update its own pod instead of running it in every matcher

SuperSandro2000:


openstack/clair/templates/etc/_config.yaml.tpl
```diff
@@ -21,7 +25,12 @@ matcher:
   max_conn_pool: 32
   indexer_addr: "http://clair-indexer:8080"
   migrations: true
-  period: {{ quote .Values.clair.update_interval }}
+  period: {{ quote $context.Values.clair.update_interval }}
+
+{{- if not (eq $type "updater") }}
```
majewsky, line 30:
```suggestion
{{- if ne $type "updater" }}
```

majewsky (approved):

PR #4384, opened 2022-11-25
[nova] Unset X-Frame-Options in shellinabox

fwiesel:
The header is set by default in the docker image,
and stops us from using the console in an iframe
as Elektra does.


fwiesel:
Sorry, that didn't seem to do the trick

PR #4404, opened 2022-11-30
openstack/keppel: split vuln_status metric into vuln_status = Pending…

SuperSandro2000:
… and vuln_status != Pending


majewsky:
We decided to go with a different approach.

PR #4411, opened 2022-12-01
[neutron] Optional Log-shipping for dnsmasq

fwiesel:
It depends on the image build from
https://github.com/sapcc/dnsmasq-log-exporter
which prints and rotates the logs from dnsmasq
and prefixes the logs with the network-id

### DIFF
--- a/openstack/neutron/templates/etc/_dhcp-agent.ini.tpl
+++ b/openstack/neutron/templates/etc/_dhcp-agent.ini.tpl
@@ -15,3 +15,6 @@ num_sync_threads = {{.Values.agent.dhcp.num_sync_threads | default 4 }}
 rpc_response_timeout = {{ .Values.rpc_response_timeout | default .Values.global.rpc_response_timeout | default 50 }}
 rpc_workers = {{ .Values.rpc_workers | default .Values.global.rpc_workers | default 5 }}
 rpc_conn_pool_size = {{ .Values.rpc_conn_pool_size | default .Values.global.rpc_conn_pool_size | default 100 }}
+{{- if .Values.dnsmasq_log_export }}
+dnsmasq_base_log_dir = /var/lib/neutron/dhcp
+{{- end }}
--- a/openstack/neutron/templates/statefulset-network-agent-apod.yaml
+++ b/openstack/neutron/templates/statefulset-network-agent-apod.yaml
@@ -71,6 +71,7 @@ spec:
           volumeMounts:
             - name: host
               mountPath: "/host"
+      shareProcessNamespace: true
       containers:
         - name: neutron-dhcp-agent
           image: {{$.Values.global.registry}}/loci-neutron:{{$.Values.imageVersionNetworkAgentDHCP | default $.Values.imageVersionNetworkAgent | default $.Values.imageVersion | required "Please set neutron.imageVersionNetworkAgentDHCP or similar"}}
@@ -116,6 +117,10 @@ spec:
               readOnly: true
             - name: network-namespace
               mountPath: /var/run/netns
+{{- if $.Values.dnsmasq_log_export }}
+            - name: dhcp
+              mountPath: /var/lib/neutron/dhcp
+{{- end }}
         - name: neutron-linuxbridge-agent
           image: {{$.Values.global.registry}}/loci-neutron:{{$.Values.imageVersionNetworkAgentLinuxBridge | default $.Values.imageVersionNetworkAgent | default $.Values.imageVersion | required "Please set neutron.imageVersionNetworkAgentLinuxBridge or similar"}}
           imagePullPolicy: IfNotPresent
@@ -183,6 +188,15 @@ spec:
             - name: etc-neutron-metadata-agent
               mountPath: /etc/neutron
               readOnly: true
+{{- if $.Values.dnsmasq_log_export }}
+        - name: dnsmasq-log-exporter
+          image: {{$.Values.global.registry}}/dnsmasq-log-exporter:{{ $.Values.imageVersionDnsMasqLogExporter | required "Please set .Values.imageVersionDnsMasqLogExporter" }}
+          imagePullPolicy: Present
+          command: ["/app/dnsmasq-log-exporter"]
+          volumeMounts:
+            - name: dhcp
+              mountPath: /var/lib/neutron/dhcp
+{{- end }}
       volumes:
         - name: metadata-proxy
           emptyDir: {}
@@ -258,6 +272,10 @@ spec:
                   path: logging.conf
                 - key: metadata-agent.ini
                   path: metadata-agent.ini
+{{- if $.Values.dnsmasq_log_export }}
+        - name: dhcp
+          emptyDir: {}
+{{- end }}
 {{- end }}
 {{- end }}
 {{- end }}
--- a/openstack/neutron/templates/statefulset-network-agent.yaml
+++ b/openstack/neutron/templates/statefulset-network-agent.yaml
@@ -80,6 +80,7 @@ spec:
           volumeMounts:
             - name: host
               mountPath: "/host"
+      shareProcessNamespace: true
       containers:
         - name: neutron-dhcp-agent
           image: {{$.Values.global.registry}}/loci-neutron:{{$.Values.imageVersionNetworkAgentDHCP | default $.Values.imageVersionNetworkAgent | default $.Values.imageVersion | required "Please set neutron.imageVersionNetworkAgentDHCP or similar"}}
@@ -154,6 +155,10 @@ spec:
               readOnly: false
             - name: network-namespace
               mountPath: /var/run/netns
+{{- if $.Values.dnsmasq_log_export }}
+            - name: dhcp
+              mountPath: /var/lib/neutron/dhcp
+{{- end }}
 {{- if $.Values.agent.neutron_l3 | default false }}
         - name: neutron-l3-agent
           image: {{$.Values.global.registry}}/loci-neutron:{{$.Values.imageVersionNetworkAgentL3 | default $.Values.imageVersionNetworkAgent | default $.Values.imageVersion | required "Please set neutron.imageVersionNetworkAgentL3 or similar"}}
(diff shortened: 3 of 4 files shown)


notandy (changes requested):
If you like this for QA/development - I am fine. But it way too blown for production - I'am already not happy with 3 containers per network pod. I neither see any tests for your software nor a build pipeline.

fwiesel:
> If you like this for QA/development - I am fine. But it way too blown for production

We currently observe in production dnsmasq becoming non-responsive on tcp ports, and since the logs are send to /dev/null, we have zero visibility.
So, you think it is better to not get any error logs in production in order to avoid a container?

>  I neither see any tests for your software nor a build pipeline.

I am not to do any test or pipeline before there is some buy-in. Maybe someone has a better proposal. But I can't find any tool doing that for us.

fwiesel:
I see following options:
1. Run dnsmasq with --no-daemon (Which has the comment "run in debug mode" in the help) and change the code in neutron to handle a non-forking process. Question there is how to get the network uuid into the log output.
2. Get the nodes logs into ELK  at least for dnsmasq. It logs to syslog, and the container has the nodes /dev/log device mounted, so it seems that was the intend, but I can't find any logs in ELK. Maybe we just to fix this.
3. Log to files, ship & rotate them as in this proposal.

joker-at-work:
Imho 1. doesn't work easily, because `dnsmasq` is started (afaics) via the `neutron-rootwrap-daemon` which also daemonizes itself and thus closes stdout.

I agree with Andy, that we don't want detailed DNS and DHCP logs as that would overwhelm our log-storage in production.

fwiesel:
> Imho 1. doesn't work easily,

Yes, I agree it is not easily done. It was the first thing I looked at. Can it be easily logged to stdout? In my opinion, the changes are much more effort than (3).  It would though address the number of containers.

> I agree with Andy, that we don't want detailed DNS and DHCP logs as that would overwhelm our log-storage in production.

Andy is arguing the number of containers, there was no mention of amount of logs.

DNS is not logged by dnsmasq, unless explicitly enabled with `log-queries` (as done in qa)
DHCP and other logs can be disabled by the options `--quiet-dhcp, --quiet-dhcp6, --quiet-ra, --quiet-tftp`.

From the manual:
> Suppress logging of the routine operation of these protocols. Errors and problems will still be logged.

I've added them as default to address your concern. We probably want to override them in qa though, should be proceed with this.

joker-at-work:
Sorry, I interpreted the

> But it way too blown for production 

as meaning logs.


> DNS is not logged by dnsmasq, unless explicitly enabled with log-queries (as done in qa)

Setting `dnsmasq_base_log_dir` will [set `--log-queries`](https://github.com/sapcc/neutron/blob/40f87abeeffca36786f82fcf3628898a6c6165bb/neutron/agent/linux/dhcp.py#L504) afaics. What am I missing here?

>  --quiet-dhcp and quiet-dhcp6 are over-ridden by --log-dhcp.

according to the manpage. `--log-dhcp` is set when `dnsmasq_base_log_dir` is set, too.

I'd prefer option 2, because then we could also get logs from the rootwrap-daemon (which supports syslog and which swalled problems for us before).

Regarding option 1, the `--no-daemon` documentation states

> to stop dnsmasq daemonising in production, use --keep-in-foreground

So there is a way to do foreground without also getting all the debugging enabled.

fwiesel:
> Setting `dnsmasq_base_log_dir` will [set --log-queries](https://github.com/sapcc/neutron/blob/40f87abeeffca36786f82fcf3628898a6c6165bb/neutron/agent/linux/dhcp.py#L504) afaics. What am I missing here?

Ah, no. I missed that. Should we go this way, that needs to be adjusted (either made configurable, or removed, since we configure it already through the config file)

> I'd prefer option 2, because then we could also get logs from the rootwrap-daemon (which supports syslog and which swalled problems for us before).

I am not against syslog/journald, quite the contrary, but I had trouble seeing how to transport the network-id through.

I looked into it,  and since the command is part of the record, and the uuid is passed in the command, it can be extracted by the equivalent of:
```
journalctl _EXE=/usr/sbin/dnsmasq -o json-pretty --since '10m ago' --output-fields="_CMDLINE,MESSAGE" \
  | jq -r  '"\(._CMDLINE[65:100]): \(.MESSAGE)"'
```

I'll check with the colleagues if they see a problem shipping those logs to ELK.

joker-at-work:
Until then, maybe we can use your way of shipping logs in a small qa region and see if we can reproduce the problem from production there.

businessbean:
> > Setting `dnsmasq_base_log_dir` will [set --log-queries](https://github.com/sapcc/neutron/blob/40f87abeeffca36786f82fcf3628898a6c6165bb/neutron/agent/linux/dhcp.py#L504) afaics. What am I missing here?
> 
> Ah, no. I missed that. Should we go this way, that needs to be adjusted (either made configurable, or removed, since we configure it already through the config file)
> 
> > I'd prefer option 2, because then we could also get logs from the rootwrap-daemon (which supports syslog and which swalled problems for us before).
> 
> I am not against syslog/journald, quite the contrary, but I had trouble seeing how to transport the network-id through.
> 
> I looked into it, and since the command is part of the record, and the uuid is passed in the command, it can be extracted by the equivalent of:
> 
> ```
> journalctl _EXE=/usr/sbin/dnsmasq -o json-pretty --since '10m ago' --output-fields="_CMDLINE,MESSAGE" \
>   | jq -r  '"\(._CMDLINE[65:100]): \(.MESSAGE)"'
> ```
> 
> I'll check with the colleagues if they see a problem shipping those logs to ELK.



> > Setting `dnsmasq_base_log_dir` will [set --log-queries](https://github.com/sapcc/neutron/blob/40f87abeeffca36786f82fcf3628898a6c6165bb/neutron/agent/linux/dhcp.py#L504) afaics. What am I missing here?
> 
> Ah, no. I missed that. Should we go this way, that needs to be adjusted (either made configurable, or removed, since we configure it already through the config file)
> 
> > I'd prefer option 2, because then we could also get logs from the rootwrap-daemon (which supports syslog and which swalled problems for us before).
> 
> I am not against syslog/journald, quite the contrary, but I had trouble seeing how to transport the network-id through.
> 
> I looked into it, and since the command is part of the record, and the uuid is passed in the command, it can be extracted by the equivalent of:
> 
> ```
> journalctl _EXE=/usr/sbin/dnsmasq -o json-pretty --since '10m ago' --output-fields="_CMDLINE,MESSAGE" \
>   | jq -r  '"\(._CMDLINE[65:100]): \(.MESSAGE)"'
> ```
> 
> I'll check with the colleagues if they see a problem shipping those logs to ELK.

[Filebeat](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-journald.html) can directly fetch journald logs. Lets check if fluentd can do the same...

businessbean:
> > > Setting `dnsmasq_base_log_dir` will [set --log-queries](https://github.com/sapcc/neutron/blob/40f87abeeffca36786f82fcf3628898a6c6165bb/neutron/agent/linux/dhcp.py#L504) afaics. What am I missing here?
> > 
> > 
> > Ah, no. I missed that. Should we go this way, that needs to be adjusted (either made configurable, or removed, since we configure it already through the config file)
> > > I'd prefer option 2, because then we could also get logs from the rootwrap-daemon (which supports syslog and which swalled problems for us before).
> > 
> > 
> > I am not against syslog/journald, quite the contrary, but I had trouble seeing how to transport the network-id through.
> > I looked into it, and since the command is part of the record, and the uuid is passed in the command, it can be extracted by the equivalent of:
> > ```
> > journalctl _EXE=/usr/sbin/dnsmasq -o json-pretty --since '10m ago' --output-fields="_CMDLINE,MESSAGE" \
> >   | jq -r  '"\(._CMDLINE[65:100]): \(.MESSAGE)"'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'll check with the colleagues if they see a problem shipping those logs to ELK.
> 
> > > Setting `dnsmasq_base_log_dir` will [set --log-queries](https://github.com/sapcc/neutron/blob/40f87abeeffca36786f82fcf3628898a6c6165bb/neutron/agent/linux/dhcp.py#L504) afaics. What am I missing here?
> > 
> > 
> > Ah, no. I missed that. Should we go this way, that needs to be adjusted (either made configurable, or removed, since we configure it already through the config file)
> > > I'd prefer option 2, because then we could also get logs from the rootwrap-daemon (which supports syslog and which swalled problems for us before).
> > 
> > 
> > I am not against syslog/journald, quite the contrary, but I had trouble seeing how to transport the network-id through.
> > I looked into it, and since the command is part of the record, and the uuid is passed in the command, it can be extracted by the equivalent of:
> > ```
> > journalctl _EXE=/usr/sbin/dnsmasq -o json-pretty --since '10m ago' --output-fields="_CMDLINE,MESSAGE" \
> >   | jq -r  '"\(._CMDLINE[65:100]): \(.MESSAGE)"'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'll check with the colleagues if they see a problem shipping those logs to ELK.
> 
> [Filebeat](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-journald.html) can directly fetch journald logs. Lets check if fluentd can do the same...

Fluentd can do the same with a [3rd party plugin](https://github.com/fluent-plugin-systemd/fluent-plugin-systemd) but not [natively](https://docs.fluentd.org/input#list-of-input-plugins)

notandy:
> So, you think it is better to not get any error logs in production in order to avoid a container?

I didn't said that - stop implying. This PR has no chance of merging in it's current form.

fwiesel:
> > So, you think it is better to not get any error logs in production in order to avoid a container?

> I didn't said that - stop implying.

I am saying it to make clear what the implication is. And the temporal aspect is here an important one:
As long as we do not have any solution in place, we practically do not have any logs to investigate the problem in production. I thought this was a quick and possibly dirty way to do that.

> This PR has no chance of merging in it's current form.

Fine by me. I have no vested interest in that particular solution. 
I made this and other proposals, if you do not see any urgency in getting a handle at fixing the problem in production then we leave it at that.

PR #4413, opened 2022-12-02
[ironic] Enable idrac-redfish and redfish

fwiesel:
Since it is working in some regions, we might as well open the option to enable it in all regions.

We still need to configure it on a host-by-host
basis, so the impact of this change is close to zero by itself.


sandzwerg (approved):

sandzwerg:
Fine with me

PR #4416, opened 2022-12-05
[Swift] alerts for quarantined objects,containers,accounts

reimannf:


majewsky (approved):

majewsky:
I added a `max()` around each metric name to prevent the alert from flapping during deployments that restart swift-health-exporter.

PR #4417, opened 2022-12-05
[nova] memcached alert becomes info-level and less triggery

joker-at-work:
This reduces the alert-level to info by updating the memcached chart and also sets the threshold for triggering the alert in Nova to a higher threshold, because looking at the current levels we didn't see any service degradation.


grandchild (approved):
LGTM

PR #4421, opened 2022-12-05
vmware-monitoring: add Maia federation service monitors

richardtief:


viennaa (approved):
It's a little odd, but it works as expected. Thanks for coming up with this solution.

PR #4432, opened 2022-12-08
[maia] customer used metric added back

viennaa:


richardtief (approved):
Thanks for re-adding it.

PR #4433, opened 2022-12-08
[Cinder] Enable independent clone by default

hemna:
This patch updates the cinder helm chart to enable the new cinder ability to allow independent cinder clones. This means that volumes created from another volume can be scheduled on a different pool (datastore) on the same backend as the source volume.

By default cinder doesn't allow this.


hemna:
related to this cinder PR:
https://github.com/sapcc/cinder/pull/159

PR #4434, opened 2022-12-09
keystone: avoid dns dependency for 2faproxy

databus23:
We used to inject http://keystone.monsoon3.svc.kubernetes.qa-de-2.cloud.sap:5000/v3 via secrets but this requires a dns lookup and a working coredns in the clusters.

By using the service ip directly we remove a dependency and make the keystone datapath more robust

### DIFF
--- a/openstack/keystone/templates/deployment-2faproxy.yaml
+++ b/openstack/keystone/templates/deployment-2faproxy.yaml
@@ -35,6 +35,8 @@ spec:
             - -exclude-domain={{ $domain }}
             {{- end}}
           env:
+            - value: OS_AUTH_URL
+              name: "http://$(KEYSTONE_SERVICE_HOST):$(KEYSTONE_SERVICE_PORT)/v3"
           {{- range $key, $value :=  index .Values "2fa" "openstack" }}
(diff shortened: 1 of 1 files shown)


openstack/keystone/templates/deployment-2faproxy.yaml
```diff
@@ -35,6 +35,8 @@ spec:
             - -exclude-domain={{ $domain }}
             {{- end}}
           env:
+            - value: OS_AUTH_URL
+              name: "http://$(KEYSTONE_SERVICE_HOST):$(KEYSTONE_SERVICE_PORT)/v3"
```
bbobrov, line 39: could you please explain, why we have http here and not http**s**? My local `OS_AUTH_URL` starts with https.

databus23: that's the backend connection from the 2faproxy to the keystone-api, its behind the ingress which does the tls termination:

```
[client] --https--> [nginx-ingress ] --http--> [2faproxy] --http--> [keystone]
```

bbobrov (approved):

PR #4435, opened 2022-12-09
maia: remove prom alert managers

richardtief:


viennaa:
Seems to be forgotten. @richardtief Can you fix the conflict and @notque review pls?

PR #4444, opened 2022-12-15
[unbound] enable infra-keep-probing reduce host ttl from 15 to 10 mins.

galkindmitrii:
by default hosts that are down, eg. they did not respond during the one probe  at  a  time period,  are  marked as down and it may take infra-host-ttl time to get probed again.

### DIFF
--- a/system/unbound/templates/config.yaml
+++ b/system/unbound/templates/config.yaml
@@ -26,7 +26,12 @@ data:
 
         cache-max-negative-ttl: 100
         cache-max-ttl: 28800
-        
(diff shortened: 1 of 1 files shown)


occamshatchet:
Since this shouldn't happen very often, but the impact of it happening is very disruptive to our customers. Wouldn't it make sense to lower this value to slightly more than the 240 seconds for the full probe?

occamshatchet:
True. Maybe we can reduce it more in QA and watch it for a while.

PR #4449, opened 2022-12-16
prometheus-vmware-rules: onboard to sharded concept

richardtief:
This is a temporary templating to accompany the progress as we roll out region by region.


prometheus-rules/prometheus-vmware-rules/values.yaml
```diff
@@ -1,2 +1,4 @@
 # The name of the Prometheus to which the alert- and aggregation rules will be assigned to.
 prometheusName: vmware
+shardedRegions:
+  - qa-de-1
```
viennaa, line 4: Shouldn't we default to non-sharded? Wouldn't bake in a static region here.

richardtief: This is our region list that we maintain as long as we roll out. After that this entry goes away again.

richardtief: In other words, we need a switch somewhere.

richardtief: And the default is non-sharded. This `else` condition
https://github.com/sapcc/helm-charts/blob/f3ca66bb5a61ffa74e21a3fff1ddeee0ad985ea2/prometheus-rules/prometheus-vmware-rules/templates/prometheus-aggregations.yaml#L19-L34

viennaa: Can you please add a comment for this variable describing this? Something like:  `static list to check if region is a sharded region, to be removed later`

viennaa (approved):

PR #4457, opened 2022-12-20
Add nova-bigvm prometheus exporter

leust:
Depends on https://github.com/sapcc/nova/pull/388

### DIFF
--- a/openstack/nova/templates/bigvm-deployment.yaml
+++ b/openstack/nova/templates/bigvm-deployment.yaml
@@ -75,6 +75,9 @@ spec:
           resources:
 {{ toYaml .Values.pod.resources.bigvm | indent 12 }}
           {{- end }}
+          ports:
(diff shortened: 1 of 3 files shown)


openstack/nova/templates/bigvm-exporter-service.yaml
joker-at-work, line 1: Do we need a service? I thought prometheus would just pick up on the pods/containers and scrape them automatically. E.g. https://github.com/sapcc/helm-charts/blob/709398f67546f1fb781af35c9d7b0bc7d7f29658/openstack/neutron/templates/deployment-cc-fabric-agent.yaml#L70-L72 doesn't have an accompanying `Service` either.

joker-at-work (changes requested):

joker-at-work (approved):

PR #4465, opened 2022-12-29
Ironic: Add Cronjob to manage BMC certs

fwiesel:
The cronjob runs a script, which checks
the certificates and replaces them, if
- the issuer is not matching an expected one
- the common name or the SAN is not matching/containing the hostname
- the expiry date is within a certain grace period

### DIFF
--- a/openstack/ironic/templates/redfish-certrobot-cronjob.yaml
+++ b/openstack/ironic/templates/redfish-certrobot-cronjob.yaml
@@ -1,0 +1,62 @@
+{{- define "_values_to_env" }}
+{{- $prefix := index . 1 }}
+    {{- with $map := index . 0 }}
+        {{- $keys := keys $map | sortAlpha }}
+        {{- range $_, $k := $keys }}
+            {{- $path := append $prefix $k }}
+            {{- $v := get $map $k }}
+            {{- if kindIs "map" $v }}
+                {{- tuple $v $path | include "_values_to_env" }}
+            {{- else }}
+- name: {{ $path | join "_" | upper }}
+{{- if kindIs "slice" $v }}
+  value: {{ $v | join "," }}
+{{- else }}
+  value: {{ $v }}
+{{- end }}
+            {{- end }}
+        {{- end }}
+    {{- end }}
+{{- end }}
+
+{{- define "values_to_env" }}
+{{- tuple . (list) | include "_values_to_env" }}
+{{- end }}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ .Release.Name }}-redfish-certrobot
+spec:
+  schedule: {{ .Values.cert_robot.schedule | quote }}
+  jobTemplate:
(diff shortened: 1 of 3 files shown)


openstack/ironic/values.yaml
```diff
@@ -31,6 +31,36 @@ debug: "True"
...
+      state:
+      city:
+      organization:
+      organizational_unit:
+    os_cloud: default  # Needs to match the name of the cloud in cloud_common and cloud_secure
+  cloud_common:
+    clouds:
+      default:
+        auth:
+          auth_url: http://keystone:5000/v3
```
sandzwerg, line 52: Does this get overwritten with a sensible value by default?

openstack/ironic/templates/redfish-certrobot-cronjob.yaml
```diff
@@ -0,0 +1,62 @@
...
+  value: {{ $v }}
+{{- end }}
+            {{- end }}
+        {{- end }}
+    {{- end }}
+{{- end }}
+
+{{- define "values_to_env" }}
+{{- tuple . (list) | include "_values_to_env" }}
+{{- end }}
```
sandzwerg, line 24: I really can't comment if this works

sandzwerg (commented):
Looks okay as far as I can tell. I'm not really good with helm. Maybe Stefan can take a look as well?

openstack/ironic/values.yaml
fwiesel: It doesn't get overwritten, and should work out of the box by relying on some conventions:
Ironic and Keystone get deployed in the same namespace, and the service for keystone has been named "keystone", and listens on port 5000.

It works together with our default dns resolution and the fact that keystone has been deployed in the same namespace as ironic.
Since it has no dots in the name, it will trigger a dns-search, and the first item in the list is `monsoon3.svc.kubernetes.<region>.cloud.sap`, so it expands to `keystone.monsoon3.svc.kubernetes.qa-de-2.cloud.sap`, which is declared [here](https://github.com/sapcc/helm-charts/blob/master/openstack/keystone/templates/service-api.yaml).

openstack/ironic/templates/redfish-certrobot-cronjob.yaml
fwiesel: Probably a bit over-engineered considering we only need to inject one or two values that way now.
When I started it, I considered to inject all values that way,
but I eventually resorted to doing it in `openstack/ironic/templates/redfish-certrobot-secret.yaml` as I need different credentials for the openstacksdk and lego.

sandzwerg: I assume you tested it, I just wanted to express that I feel not properly qualified to judge this. But it's fine if you tested it.

sandzwerg (approved):
Fine with me. Not sure if @stefanhipfel has anything to add?

PR #4466, opened 2023-01-02
[prometheus] adding remote_write alert

viennaa:


common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
```diff
@@ -214,3 +214,23 @@ groups:
       description: 'Prometheus `{{`{{ $labels.prometheus }}`}}` is backlogging on the notifications queue. The queue has not been empty for 10 minutes. Current queue length `{{`{{ $value }}`}}`.'
       summary: Prometheus is queueing notifications.
   {{- end }}
+
+  {{- if and $root.Values.remoteWriteTargets (gt (len $root.Values.remoteWriteTargets) 0) }}
```
richardtief, line 218: The second condition is very similar to the first. In my opinion, the first condition should be sufficient.
```suggestion
  {{- if $root.Values.remoteWriteTargets  }}
```

richardtief (commented):
Looks good, just one little thing.

richardtief (approved):
Thanks!

PR #4468, opened 2023-01-03
[atlas] moving atlas to dedicated chart

viennaa:
This is a prerequisite for the vmware-monitoring deployment in scaleout. Atlas needs to be present in both clusters and is moved to a dedicated chart therefore.


richardtief (approved):
Thanks. All good

PR #4470, opened 2023-01-04
Send notifications to cc-exporter slack channel

artherd42:
instead to moni channels which will be retired soon


richardtief (approved):

IvoGoman:
These alerts are raised to Critical/Warning but do not have a support_group set. This way they won't end up in our alert-observability-* channels. Is this intended? AFAIK this way there won't be any paging for them.

artherd42:
These alerts and underlying metrics were originally created to get some transparency on broken inter-dc network connectivity. CCloud network team is responsible for the network. We will discuss with network team if they want to own these alerts (as they do for apic-exporter).

IvoGoman:
Understood, I just assumed that our OnDuty should take care of these alerts since they were raised to Crit/Warn.

PR #4473, opened 2023-01-05
[Cinder] Update mariadb chart to 0.7.5

hemna:
This patch updates the mariadb chart to 0.7.5 for cinder.


joker-at-work (approved):
I'd prefer if the commit messages also stated why we update and not only that :)

PR #4477, opened 2023-01-06
Change kubeAPIServer.allowedMetrics

SchwarzM:
The kubeapiserver metrics have been renamed

### DIFF
--- a/common/prometheus-server-pre7/values.yaml
+++ b/common/prometheus-server-pre7/values.yaml
@@ -353,11 +353,12 @@ serviceDiscoveries:
     enabled: false
     # Full list of APIserver metrics: https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/metrics/metrics.go .
     allowedMetrics:
-      - apiserver_request_count
-      - apiserver_request_latencies_bucket
       - apiserver_request_total
-      - apiserver_request_latency_seconds
+      - apiserver_longrunning_requests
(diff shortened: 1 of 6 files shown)


SchwarzM:
here are the current metrics: https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/metrics/metrics.go
i did only things based on STABLE metrics

viennaa (changes requested):
Thanks for providing this update.

Can you please add the following things?
* bump the chart version and add the information about what changed into CHANGELOG
* Please update the common/prometheus-server chart too.

common/prometheus-server-pre7/values.yaml
```diff
@@ -353,11 +353,12 @@ serviceDiscoveries:
...
-      - apiserver_request_latencies_bucket
       - apiserver_request_total
-      - apiserver_request_latency_seconds
+      - apiserver_longrunning_requests
+      - apiserver_request_duration_seconds
       - apiserver_dropped_requests_total
+      - apiserver_requested_deprecated_apis
+      - apiserver_current_inflight_requests
       - etcd_request_latency_seconds
       - etcd_object_counts
```
databus23, line 363: Since 1.21
> Rename metrics etcd_object_counts to apiserver_storage_object_counts and mark it as stable. The original etcd_object_counts metrics name is marked as "Deprecated" and will be removed in the future.

PR #4481, opened 2023-01-06
maia: fix vmware metrics federation

richardtief:


notque (approved):
Looks reasonable. Please make sure to validate with it running that it’s working fine.

openstack/maia/templates/etc/_maia_scrape_config.yaml.tpl
```diff
@@ -149,58 +149,65 @@
...
   params:
     'match[]':
-      - '{{ "{" }}__name__=~"{{ $match }}", project!~"internal", vccluster!~".*management.*"{{ "}" }}'
+      - '{__name__=~"{{- include "prometheusVMwareFederationMatches" $root }}",project!~"internal",vccluster!~".*management.*"}'
 {{- end }}
 {{- end }}
-
-{{- if .Values.neo.enabled }}
-- job_name: 'prometheus-vmware-neo'
+{{- else }}
```
viennaa, line 181: Adding a little verbosity here
```suggestion
# supporting not migrated regions
# TODO: remove after global vmware-monitoring rollout
{{- else }}
```

openstack/maia/templates/maia-podmonitor.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    interval: 30s
+    path: federate
+    port: web
+    scheme: http
+    scrapeTimeout: 25s
+    params:
+      match[]:
+        - '{__name__="up"}'
+  jobLabel: prometheus-maia-oprom-2
```
viennaa, line 19: Two things: why does it have `-2`?
I can't see this reflected in the actual job name, is it supposed to do something?

viennaa (changes requested):
Really like the implementation with the smart `prometheusVMwareFederationMatches` helper.
There are only a few minor things to do before it's good to go.

openstack/maia/templates/etc/_maia_scrape_config.yaml.tpl
```diff
@@ -149,58 +149,65 @@
...
-    - targets: ['prometheus-vmware.vmware-monitoring:9090']
+    - targets: 
+        - '{{ include "prometheusVMware.fullName" (list $target $root) }}-internal.{{ $root.Values.global.region }}.cloud.sap'
   metric_relabel_configs:
     - source_labels: [__name__, project ]
       regex: '^vrops_virtualmachine_.+;(.+)'
       replacement: '$1'
       target_label: project_id
-    - regex: 'project|collector|exported_job|instance|internal_name|prometheus|prometheus_replica|resource_uuid|cluster|cluster_type|vccluster|vcenter'
+    - regex: 'project|collector|exported_job|instance|internal_name|prometheus|resource_uuid|thanos_cluster|thanos_cluster_type|vccluster|vcenter'
```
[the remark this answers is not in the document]
richardtief: You're right. Will change this.

openstack/maia/templates/maia-podmonitor.yaml
richardtief: also thought something was gonna happen with it, but it doesn't seem so. There is already a ServiceMonitor with this name that queries Prometheus metrics. It has the same name, so I gave this one a `-2` identifier. Will remove it then.

viennaa (approved):
Thanks!

PR #4486, opened 2023-01-10
elastiflow: drop unnecessary privileges on initContainer

majewsky:
This init container only changes file permissions on a mounted volume. It does not need to run in privileged mode as far as I can see.

### DIFF
--- a/system/elastiflow/values.yaml
+++ b/system/elastiflow/values.yaml
@@ -5,8 +5,10 @@ owner-info:
     - "Ivo Gosemann"
   helm-chart-url: "https://github.com/sapcc/helm-charts/tree/master/system/elastiflow"
 
+# Deploy Prometheus alerts.
 alerts:
   enabled: false
+  # Name of the Prometheus to which the alerts should be assigned to.
   prometheus: infra-frontend
 
 curator:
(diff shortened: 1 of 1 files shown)


system/elastiflow/values.yaml
```diff
@@ -351,9 +351,3 @@ kibana:
   #     memory: 4Gi
   #   limits:
   #     memory: 5Gi
-
-# Deploy Prometheus alerts.
-alerts:
-  enabled: false
-  # Name of the Prometheus to which the alerts should be assigned to.
-  prometheus: infra-frontend
```
majewsky, line 359: This part was flagged by the CI:
```
system/elastiflow/values.yaml
  354:1     error    duplication of key "alerts" in mapping  (key-duplicates)
```
I merged it with the functionally identical section at the top of the file.

notque (approved):
LGTM, curious if @IvoGoman has an argument for the privileges, I'm not aware of them being needed

majewsky:
My best guess is "copy-paste from the configure-sysctl init container". The sysctl init container does indeed need privileged access to edit kernel settings.

notque:
Yeah, that makes sense, and it's just changing permissions, so it seems fine to just merge.

IvoGoman (approved):
I do not see a reason for having privileged more here.

PR #4487, opened 2023-01-10
prometheus-server: add option to set support_group

richardtief:


viennaa (approved):
More flexible now, like it.

PR #4492, opened 2023-01-10
system/hubcopter: init

SuperSandro2000:


system/hubcopter-seeds/templates/seed.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: "openstack.stable.sap.cc/v1"
+kind: OpenstackSeed
+metadata:
+  name: hubcopter-seed
+
+spec:
+  requires:
+    - {{ .Values.global.keystoneNamespace }}/keystone-seed
```
SuperSandro2000, line 8: ``[ERROR] templates/: template: hubcopter/templates/seed.yaml:8:16: executing "hubcopter/templates/seed.yaml" at <.Values.global.keystoneNamespace>: nil pointer evaluating interface {}.keystoneNamespace``

Not sure what the problem is here

majewsky: This value usually comes from the `globals.yaml` values file. We don't have that one during `helm lint`, so you need to supply the respective values in `ci/test-values.yaml`.

system/hubcopter/templates/deployment-glas.yaml
```diff
@@ -0,0 +1,71 @@
...
+            weight: 10
+      containers:
+        - name: api
+          image: {{ include "glas_image" $ }}
+          imagePullPolicy: IfNotPresent
+          args: [ ]
+          env: {{ include "glas_environment" $ | indent 12 }}
+          volumeMounts:
+            - mountPath: /etc/glas/github_enterprise_secret
+              name: glas-secret
```
majewsky, line 48: The definition for this volume (in `.spec.template.spec.volumes`) is missing.

system/hubcopter/templates/_utils.tpl
```diff
@@ -0,0 +1,55 @@
+{{- define "glas_image" -}}
+  {{ $.Values.global.registry }}/hubcopter:{{ $.Values.glas.image_tag | required ".Values.hubcopter.image_tag is missing" }}
+{{- end -}}
+{{- define "hubcopter_image" -}}
+  {{ $.Values.global.registry }}/hubcopter:{{ $.Values.hubcopter.image_tag | required ".Values.hubcopter.image_tag is missing" }}
+{{- end -}}
+
+{{- define "glas_environment" }}
```
majewsky, line 8: Since you're only using these sets of environment variables once, you can inline this template into the respective Deployment spec. I'm usually only templating this when a set of environment variables is shared by multiple Deployments (e.g. tenso-api with tenso-worker).

majewsky (approved):

PR #4497, opened 2023-01-11
D072896 ack button support groups

geisslet:
to have the ack button working we need to separate the routes unfortunately to use the channel specific webhook

### DIFF
--- a/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
+++ b/global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
@@ -242,10 +242,66 @@ route:
     match_re:
       service: wsus
 
-  - receiver: support_group_alerts
+  - receiver: support_group_alerts_critical_compute
     continue: true
     match_re:
-      severity: warning|critical
+      severity: critical
+      cluster_type: abapcloud|admin|controlplane|customer|internet|kubernikus|metal|scaleout|virtual
+      region: global|ap-ae-1|ap-au-1|ap-cn-1|ap-jp-1|ap-jp-2|ap-sa-1|ap-sa-2|eu-de-1|eu-de-2|eu-nl-1|la-br-1|na-ca-1|na-us-1|na-us-2|na-us-3
+      support_group: compute
+
+  - receiver: support_group_alerts_critical_compute_storage_api
+    continue: true
+    match_re:
+      severity: critical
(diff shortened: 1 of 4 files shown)


global/prometheus-alertmanager-operated/values.yaml
```diff
@@ -35,6 +35,13 @@ slack:
   metal: {}
   network: {}
   vmware: {}
+  compute: {}
+  compute_storage_api: {}
+  email: {}
+  identity: {}
+  network_api: {}
+  observability: {}
+  containers: {}
```
viennaa, line 44: you have specified a dict in ci/test-values.yaml already. the idea would be, to leave this entirely blank. This leads to a fail of the pipeline, in case the values are not in place, instead of deploying it with emtpy values. Linting then would fail too, in case the values are missing. You can incorporate the above lines too.

just go with this:
```suggestion
  compute:
  compute_storage_api:
  email:
  identity:
  network_api:
  observability:
  containers:
```

global/prometheus-alertmanager-operated/templates/_alertmanager.yaml.tpl
```diff
@@ -945,6 +980,134 @@ receivers:
...
+        pretext: {{"'{{template \"slack.sapcc.pretext\" . }}'"}}
+        icon_emoji: {{"'{{template \"slack.sapcc.iconemoji\" . }}'"}}
+        callback_id: "alertmanager"
+        color: {{`'{{template "slack.sapcc.color" . }}'`}}
+        send_resolved: true
+        actions:
+          - name: {{"'{{template \"slack.sapcc.actionName\" . }}'"}}
+            type: {{"'{{template \"slack.sapcc.actionType\" . }}'"}}
+            text: {{"'{{template \"slack.sapcc.acknowledge.actionText\" . }}'"}}
+            value: {{"'{{template \"slack.sapcc.acknowledge.actionValue\" . }}'"}}
```
viennaa, line 1000: Nothing to be addressed in this PR, but as a general note. This duplication can be avoided. The only things which change here are the name and the values in the `api_url`. Both could be derived from an array as input, with a single alertconfiguration, just adding this particular information. It would be more part of a template optimisation, but in general only one occurrence of the similar looking things should be here. @bassel-z this is for you :).

viennaa (changes requested):
Thanks for providing this PR with the exact teams. I'd recommend some minor change to values, to actually leverage linting and the functionality of not having the values set. Explanation is inline.
I've added some general note to the single slack configs too, maybe something for @bassel-z to continue on.

PR #4498, opened 2023-01-11
[alertmanager] bring awx routing back

geisslet:


max-len (approved):
AWX part looks good to me

PR #4503, opened 2023-01-13
common/prometheus-server: order of coalescence corrected

richardtief:
The order can cause the targets to be preferred to the set name. For example, if a prometheus needs the targets additionally as for scrape jobs, multiple prometheus instances of the targets are created by default. What is not wanted.


viennaa (approved):
Good catch, the global targets took precedence over a single name. Thanks.

PR #4506, opened 2023-01-16
[infra-prom] removes dedicated apic-exporter scrape job

IvoGoman:
The apic-exporter is scraped by a servicemonitor. I believe this is a leftover from the previous version of the exporter


IvoGoman:
While this is a double scrape we cannot catch it with the `PrometheusMultipleTargetScrapes`. For this Job definition the instance is hard-coded to the service with `apic-exporter:9102`. While the ServiceMonitor uses the actual pod IP for the instance.

PR #4509, opened 2023-01-16
[rabbitmq] add support for rabbitmq-prometheus

bashar-alkhateeb:
added a switch to activate rabbitmq-prometheus plugin instead of the side car exporter. to activate collect metrics with plugin please set your values to:
```
    metrics:
      port: 9150
      enabled: false
      sidecar:
        enabled: false
```
RabbitMQ-prometheus plugin provides different naming schema of the metrics as the side car RabbitMQ exporter.
  metrics can be found: https://github.com/rabbitmq/rabbitmq-prometheus/blob/master/metrics.md

default values still using metrics sidecar exporter

### DIFF
--- a/common/rabbitmq/CHANGELOG.md
+++ b/common/rabbitmq/CHANGELOG.md
@@ -3,6 +3,22 @@ Rabbitmq CHANGELOG
 
 This file is used to list changes made in each version of the common chart rabbitmq.
 
+0.5.0
+-----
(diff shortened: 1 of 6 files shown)


viktor-kovacs (approved):
Looks fine to me

businessbean (approved):
lgtm

fwiesel (approved):

PR #4513, opened 2023-01-17
common/prometheus-server: forward multiple scrape alerts

richardtief:


common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
```diff
@@ -11,6 +11,9 @@ Note: The pods define the 'alert-service' label but Prometheus replaces the hyph
 {{- define "alertServiceLabelOrDefault" -}}
 "{{`{{ if $labels.label_alert_service }}`}}{{`{{ $labels.label_alert_service}}`}}{{`{{ else }}`}}{{ required "default value is missing" . }}{{`{{ end }}`}}"
 {{- end -}}
+{{- define "alertSupportGroupOrDefault" -}}
+"{{`{{ if $.Values.alerts.support_group }}`}}{{`{{ $.Values.alerts.support_group }}`}}{{`{{ else if $labels.ccloud_support_group }}`}}{{`{{ $labels.ccloud_support_group }}`}}{{`{{ else }}`}}{{ required "default value is missing" . }}{{`{{ end }}`}}"
+{{- end -}}
```
IvoGoman, line 16: I am not sure about the order of `$.Values.alerts.support_group` and `$labels.ccloud_support_group`.
Shouldn't the more specific support group of the metric take precedence over the support group from the values?

IvoGoman (changes requested):

common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
richardtief: Yes, you are right. The label takes the lead.

richardtief: The values entry can then be omitted, which would only make sense in rare cases.

IvoGoman (approved):
Thank you 👍

PR #4521, opened 2023-01-18
common/thanos: revise thanos alerts

richardtief:


common/thanos/templates/alerts/_thanos-compactor.alerts.tpl
```diff
@@ -3,83 +3,109 @@
 groups:
 - name: thanos-compactor.alerts
   rules:
-    - alert: ThanosCompactHalted
-      expr: thanos_compact_halted{app="thanos-compactor", prometheus="{{ include "thanos.name" . }}"} == 1
+    - alert: ThanosCompactMultipleRunning
+      expr: sum by (thanos) (up{job=~".*thanos.*compact.*", thanos="{{ include "thanos.name" . }}"}) > 1
```
viennaa, line 7: Why don't we sum by `job` here?

common/thanos/templates/alerts/_thanos-compactor.alerts.tpl
```diff
@@ -3,83 +3,109 @@
...
+    - alert: ThanosCompactHalted
+      expr: thanos_compact_halted{job=~".*thanos.*compact.*", thanos="{{ include "thanos.name" . }}"} == 1
+      for: 5m
       labels:
-        context: thanos
         service: {{ default "metrics" $root.Values.alerts.service }}
-        support_group: observability
+        support_group: {{ default "observability" $root.Values.alerts.support_group }}
         severity: info
         playbook: 'docs/support/playbook/prometheus/thanos_compaction.html'
```
viennaa, line 28: FWIW: the playbook does not exist

common/thanos/templates/alerts/_thanos-store.alerts.tpl
```diff
@@ -4,29 +4,82 @@ groups:
...
+      labels:
+        service: {{ default "metrics" $root.Values.alerts.service }}
+        support_group: {{ default "observability" $root.Values.alerts.support_group }}
+        severity: info
+        meta: Thanos Store `{{`{{ $labels.thanos }}`}}` has a 99th percentile latency for store series gate requests.
+      annotations:
+        description: |
+          Thanos Store `{{`{{ $labels.thanos }}`}}` has a 99th 
+          percentile latency of `{{`{{ $value }}`}}`seconds 
+          for store series gate requests.
```
viennaa, line 43: Would be interesting to know what `series gate requests` are about. Guess we have to figure this out, once it happens. Hence `info` is good 👍

viennaa (commented):
Thanks again for also bringing these alerts back on track. Why did you swap `job` with `thanos` in the queries. The job itself would give a little bit more insight, wouldn't it? I noticed this in `prometheus-server` already but didn't commented it there.

common/thanos/templates/alerts/_thanos-compactor.alerts.tpl
richardtief: sum by `job` would mean that `$labels.thanos` is not resolved in the meta.

common/thanos/templates/alerts/_thanos-compactor.alerts.tpl
richardtief: Yeah, I know. Let's create the playbooks step by step afterwards when we need to respond to one of the new alerts. Monitoring the #cc-metrics channel after deployment.
```suggestion
```

richardtief:
Since we tag the job's metrics with the Thanos name, we can use the `thanos` label to refer to the affected resource. 

https://github.com/sapcc/helm-charts/blob/master/common/thanos/templates/servicemonitor.yaml#L27-L29

viennaa (approved):
Good to go from my side.

PR #4524, opened 2023-01-20
volume-claims: remove PVCs from ap-sa-2 that did not exist there before

majewsky:
While we're at it, some handover instructions for you:
* The first version was created by adopting the left-over remains of the Helm-2-era region charts. The chart was generated to match the existing PVCs, hence why there are some rather bizarre if-statements in the templating.
* There is no pipeline for this. My process so far was to review diffs and then apply with:
  ```
  h diff upgrade volume-claims . --color --set global.region="$(u8s env | awk -F= '/U8S_CONTEXT/{print$2}')"
  h upgrade volume-claims . --set global.region="$(u8s env | awk -F= '/U8S_CONTEXT/{print$2}')"
  ```


fwiesel:
Closing in favour of #4525

PR #4526, opened 2023-01-20
prometheus-server: revised alerts

richardtief:


common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
```diff
@@ -21,145 +21,284 @@ groups:
...
       severity: info
-      playbook: 'docs/support/playbook/prometheus/wal.html'
-      meta: 'Prometheus `{{`{{ $labels.prometheus }}`}}` has `{{`{{ $value }}`}}` WAL corruptions.'
+      playbook: docs/support/playbook/prometheus/wal.html
+      meta: Prometheus `{{`{{ $labels.prometheus }}`}}` has WAL corruptions.
     annotations:
-      description: 'Prometheus `{{`{{ $labels.prometheus }}`}}` has `{{`{{ $value }}`}}` WAL corruptions.'
-      summary: Prometheus has WAL corruptions
+      description: |
+        Prometheus `{{`{{ $labels.prometheus }}`}}` has `{{`{{ $value }}`}}` WAL corruptions.
```
viennaa, line 155: The description does not really fit the alert expression. If the corruption increases, `$value` will be at a certain value. Once increasing stops, it will be back to normal. However, the corrupt WAL blocks are staying in place.

Maybe we should split this in two. This would rather be a `PrometheusWALCorruptionIncreasing` alert. Maybe we should have one for total too, not taking the increase in consideration?

common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
```diff
@@ -21,145 +21,284 @@ groups:
...
+      summary: Prometheus is reaching its maximum capacity serving concurrent requests.
 
   {{- if $root.Values.alerts.multipleTargetScrapes.enabled }}
   - alert: PrometheusMultipleTargetScrapes
-    # we exclude the following:
-    # * cadvisor metrics because it has the same instance as the kubelet but a different path
-    # e.g. 10.246.204.80:10250/metrics vs. 10.246.204.80:10250/metrics/cadvisor
-    # * pod service discovery job, we have a dedicated alert for that
-    expr: sum by (job, ccloud_support_group) (up * on(instance, cluster) group_left() (sum by(instance, cluster) (up{job !~ "kubernetes-cadvisors|kubernetes-kubelet|.*-pod-sd|{{ $root.Values.alerts.multipleTargetScrapes.exceptions | join "|" }}"}) > 1))
+    expr: sum by (job, ccloud_support_group) (up * on(instance, cluster) group_left() (sum by(instance, cluster) (up{job!~"{{ $root.Values.alerts.multipleTargetScrapes.exceptions | join "|" }}"}) > 1))
```
viennaa, line 287: What about the previously excluded jobs? Aren't they around anymore?

common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
```diff
@@ -171,69 +310,140 @@ groups:
...
   - alert: PrometheusRemoteWriteDown
-    expr: sum by (prometheus, url) (rate(prometheus_remote_storage_samples_failed_total[5m])) > 0 or sum by (prometheus, url) (rate(prometheus_remote_storage_bytes_total[5m])) <= 0
+    expr: |
+      (
+        (rate(prometheus_remote_storage_failed_samples_total{prometheus="{{ include "prometheus.name" . }}"}[5m]) or rate(prometheus_remote_storage_samples_failed_total{prometheus="{{ include "prometheus.name" . }}"}[5m]))
+      /
+        (
+          (rate(prometheus_remote_storage_failed_samples_total{prometheus="{{ include "prometheus.name" . }}"}[5m]) or rate(prometheus_remote_storage_samples_failed_total{prometheus="{{ include "prometheus.name" . }}"}[5m]))
+        +
+          (rate(prometheus_remote_storage_succeeded_samples_total{prometheus="{{ include "prometheus.name" . }}"}[5m]) or rate(prometheus_remote_storage_samples_total{prometheus="{{ include "prometheus.name" . }}"}[5m]))
```
viennaa, line 406: Why is `prometheus_remote_storage_failed_samples_total` and `prometheus_remote_storage_succeeded_samples_total` in here? Is this an older version? At least I can't find the metrics in our Prometheus.

Also I don't get the logic behind this:
Shouldn't it be

```failed_samples_total/samples_total``` ?

Then we would get a percentage on how many are failed. Maybe you need to elaborate, what the idea is.

**Review by viennaa — COMMENTED** [review]
Thanks for updating the alerts and bringing in the community ideas. Got some remarks to talk about.

File: common/prometheus-server/templates/alerts/_prometheus.alerts.tpl (hunk 1 above)
Comment (richardtief) [reply common/prometheus-server/templates/alerts/_prometheus.alerts.tpl:155]: Fair. To my mind, there should only be one WAL corruption alert without `increase`. 
Like `prometheus_tsdb_wal_corruptions_total > 0` to indicate corrupt WAL blocks.

File: common/prometheus-server/templates/alerts/_prometheus.alerts.tpl (hunk 2 above)
Comment (richardtief) [reply common/prometheus-server/templates/alerts/_prometheus.alerts.tpl:287]: This is only needed for kube-state-metrics. Therefore, this must be excluded in the specific Prometheus, not in the common chart. https://github.com/sapcc/helm-charts/blob/master/common/prometheus-server/values.yaml#L395-L398

Comment (richardtief) [reply common/prometheus-server/templates/alerts/_prometheus.alerts.tpl:287]: I will create a pull request on the affected Prometheus charts to regain the exceptions.

File: common/prometheus-server/templates/alerts/_prometheus.alerts.tpl (hunk 3 above)
Comment (richardtief) [reply common/prometheus-server/templates/alerts/_prometheus.alerts.tpl:406]: yes, the older metric can go. But otherwise, it's as it should be: `failed / total `(where `total` is the addition of `successful` and `failed`).

File: common/prometheus-server/templates/alerts/_prometheus.alerts.tpl [hunk 4]
```diff
@@ -21,145 +21,284 @@ groups:
...
+      description: |
+        Prometheus `{{`{{ $labels.prometheus }}`}}` as failed
+        `{{`{{ printf "%.0f" $value }}`}}` scrapes in the last 5m because
+        some targets exceeded the configured sample_limit.
+      summary: Prometheus has failed scrapes that have exceeded the configured sample limit.
 
   - alert: PrometheusWALCorruption
-    expr: increase(prometheus_tsdb_wal_corruptions_total{prometheus="{{ include "prometheus.name" . }}"}[5m]) > 0
+    expr: |
+      increase(prometheus_tsdb_wal_corruptions_total{prometheus="{{ include "prometheus.name" . }}"}[5m]) > 0
```
viennaa, line 146:
```suggestion
      round(increase(prometheus_tsdb_wal_corruptions_total{prometheus="{{ include "prometheus.name" . }}"}[2h1m]) > 0)
```

viennaa (approved):
Let's give it a go!

PR #4527, opened 2023-01-20
prometheus-server-pre7: added PodMonitor for Thanos sidecar and revised alerts

richardtief:


viennaa (approved):
Thanks for taking care of the older chart too. Changes were discussed in #4526, hence approving same changes here too.

PR #4539, opened 2023-01-24
lyra: use new common/pgbackup chart

majewsky:

This new setup is already deployed and tested in my own services (Castellum, Keppel, Limes and Tenso).

Since this will restart the postgres container anyway, I took the liberty of bumping Postgres from 12.11 to 12.13. This is a bugfix release with no incompatibilities, my own services are already running on 12.13 without any issues.

Merge this together with the respective companion PR in cc/secrets.


databus23:
LGTM @ArtieReus can you take care to merge and rollout, Thanks

databus23 (approved):

PR #4541, opened 2023-01-25
[logs][controlplane] Set time_key to timestamp

fwiesel:
The grok-filter defaults to the key `time`,
but our index defaults to `@timestamp`

See: https://github.com/fluent/fluent-plugin-grok-parser/issues/58


fwiesel:
I am not sure, that is correct, it just follows the example given in the fluentd issue.

There are a couple of timestamps, which are not iso, and a couple of grok expressions, which have timestamp twice.

PR #4542, opened 2023-01-25
[neutron] Remove livenessProbes and report via exporter instead

swagner-de:


openstack/neutron/templates/statefulset-network-agent-apod.yaml
```diff
@@ -141,10 +165,32 @@ spec:
...
+                PROMETHEUS = "http://prometheus-openstack.prometheus-openstack:9090"
+                QUERY = """
+                  sum(openstack_subnets_with_dhcp_ip_allocation{network_host="%s"}) by (bridge, network_host)
+                    unless on(bridge, network_host)
+                  label_replace(neutron_agent_checks_linux_bridge_present, "network_host", "$1", "hostname", "(.*)")
+                """
+                response =requests.get(PROMETHEUS + "/api/v1/query", params={"query": QUERY % os.uname()[1]}) 
+                if  response.status_code != 200:
+                    print(f"Error while running prometheus query: {response.json()}")
+                    exit(-1)
```
notandy, line 184: there is no negative program exit code

Also, maybe you don't want to fail  here:
If prometheus is down and neutron is being rolled out, kubernetes would wait indefinitely to bring up new neutron agent pods since they would stay unready. Jusy maybe taking it into consideration.

notandy (changes requested):
couldn't test it so far, but if it works, thats cool!

openstack/neutron/templates/statefulset-network-agent-apod.yaml
swagner-de: ACK on the exit code even though it just get 2**16 iirc.
To the other topic: if we'd succeed here, kubernetes would just roll all agent pods quickly in sequence thus creating an outage. So I was thinking that being stuck is better than rolling in quickly and creating an outage.

notandy (dismissed):

joker-at-work (commented):
Could you maybe refactor the liveness check into a (parameterized) include?

They all have the same missing space in `response =requests`, so they look copy-pasted anyways.

Do I read it correctly, that a pod asks prometheus for the metrics it exports itself in another container?

Would a pod show as "ready" after startup until prometheus scraped the metrics again from the pod, because it queries the stale data from before the restart?

swagner-de:
Hi @joker-at-work

> Could you maybe refactor the liveness check into a (parameterized) include?
> They all have the same missing space in response =requests, so they look copy-pasted anyways.

[Done here](https://github.com/sapcc/helm-charts/compare/fa085cbe0013d3143f578bf6a5af51715e639255..ad15ae7b5dc314f35b24356d760b8ed52c20684f)

> Do I read it correctly, that a pod asks prometheus for the metrics it exports itself in another container?

Yes that is correct. On top of the data that the pod itself generates we also need the bridges/netns that shall be on this network agent. The pod only exports what is configured on it. That data is held by neutron or replicated into prometheus. As I want to reduce the queries towards neutron, I ask prometheus.

> Would a pod show as "ready" after startup until prometheus scraped the metrics again from the pod, because it queries the stale data from before the restart?

No. Prometheus [marks time series as stale](https://prometheus.io/docs/prometheus/latest/querying/basics/#staleness):
> If a target scrape or rule evaluation no longer returns a sample for a time series that was previously present, that time series will be marked as stale. If a target is removed, its previously returned time series will be marked as stale soon afterwards.

Our scrape interval is [30s](https://github.com/sapcc/helm-charts/blob/master/common/prometheus-server/values.yaml#L304). All time series of a target that did not deliver new values after these 30s will be marked stale and will not be evaluated in the instant query we run.

For the livenessCheck we use [`initial_delay = 20s`](https://github.com/sapcc/helm-charts/pull/4542/files#diff-40286c32692138502b0f19265eed58ad8c975b27d3fdc691108ca8a087b06a3bR109) and `successThreshold 
 = 2, periodSeconds = 15s`, meaning we need a minimum of 20s + 15s to become ready. While the first probe might get a outdated result back, the second would get the data from the new pod.

sven-rosenzweig:
can be closed due to https://github.com/sapcc/helm-charts/pull/12003

PR #4545, opened 2023-01-25
[Manila][tempest] Change branch to ccloud-python3

kpawar-sap:
Under testing


kpawar-sap:
Merging after successful run on argo 
https://argo.cc.qa-de-1.cloud.sap/workflows/argo/manila-tempest-h8nnq?tab=workflow

PR #4548, opened 2023-01-27
[neutron] Move defaults for workers to values &  scale proxysql accordingly

fwiesel:
Please, have a look at the individual commits for more details.


notandy:
api_workers are not used since we moved to uwsgi in general.

fwiesel:
Changed it to `.Values.api.processes` for the uwsgi case.

PR #4549, opened 2023-01-27
Adjusted service labels for compute alerts

christopherhans:
For better analyses and accountability I would like to adjust the service labels of some alerts from the compute support group according to their service group (compute, storage, network).

@geisslet I hope the service label is free to choose and won't affect routing/paging.


max-len (approved):
lgtm

geisslet:
lgtm
note: support_group is the criteria for routing (fixed set - check ops-docu). the service is free to define - to bind an alert to a specific service/topic

PR #4550, opened 2023-01-27
[vmware-monitoring] deploy inventory per vcenter

viennaa:
* To get ATLAS out of the critical path for providing targets the idea is to start multiple inventories.
* Every inventory should be scoped to a single target and provide the resources to all pods for the given target (vrops)
* exporters are not querying vrops-inventory anymore, instead they would query vrops-inventory-$vrops

foo


system/vmware-monitoring/templates/vrops-inventory-vpa.yaml
```diff
@@ -17,13 +20,14 @@ spec:
   targetRef:
     apiVersion: v1
     kind: Deployment
-    name: vrops-inventory
+    name: {{ include "vropsInventory.fullName" (list $target $root) }}
   updatePolicy:
-    updateMode: "Auto"
```
richardtief, line 22: Why changing this?

richardtief (changes requested):
Thanks for providing this. Could you please resolve the conflict?

system/vmware-monitoring/templates/vrops-inventory-vpa.yaml
viennaa: I like to try this in QA first before setting it for real. Also, it would currently have no effect to the prod regions, but if we decide to rollout the updater everywhere, I rather want to have the rollout of the VPA controlled, instead of having the VPA waiting to fire at soon as possible

richardtief (approved):
Thanks!

PR #4554, opened 2023-01-31
[glance] Bump proxysql and scale by workers

fwiesel:
Sqlalchemy has a pool per process, and for each worker a process is started. To maintain the same number of maximum connections in proxysql, we need to multiply the number by the number of workers

Also, update proxysql by a minor version


rajivmucheli (approved):
lgtm

PR #4557, opened 2023-01-31
sentry: use new common/pgbackup chart

majewsky:

The new pgbackup chart is already deployed and tested across all regions in my own services (Castellum, Limes, Keppel, Tenso).

This also fixes the lack of backup monitoring and alerts for Sentry.

Please merge this together with the respective companion PR in cc/secrets.


rajivmucheli (approved):
lgtm, thanks for taking of sentry!

PR #4559, opened 2023-01-31
[cc3test] add alert for cc3test_status metric missing completely

thgrs:


artherd42 (approved):
looks like a good starting point. Maybe we can reduce the `for` period later.

PR #4561, opened 2023-01-31
adding changes to postfix statefulset

corey-aloia:


sapmli (approved):
LGTM

openstack/cronus/values.yaml
```diff
@@ -70,7 +70,7 @@ cronus:
     enabled: true
     replicas: 1
   statefulSet:
-    enabled: false
```
dhalimi, line 73: Set it back to false, the enablement is not from here

PR #4566, opened 2023-02-01
infra-monitoring: enable pod-sd

richardtief:


richardtief:
I think not, since we need the default settings here. 

```yaml
  # SD for Kubernetes Pods. See https://github.com/coreos/prometheus-operator/issues/38.
  pods:
    enabled: false
    # Only scrape pods with annotation prometheus.io/targets: $prometheusName.
    # If set to false, all targets are scraped.
    limitToPrometheusTargets: true
    # Scrape all metrics by default.
    forbiddenMetrics: []
```

viennaa (approved):
As the other job has no targets it's ok to remove it. Thanks for clarifying.

PR #4567, opened 2023-02-01
Ironic: Add setting to controll disk erase

sandzwerg:
This setting allows us to enable or disable the erasure of the disk
during the cleaning of a node. For now, we want to disable this. Later we
can then enable it again. The other cleaning steps shouldn't cause major
issues but by enabling automated cleaning for some regions blocks we can
test if we find any other issues.

@stefanhipfel please make sure that makes sense. I'm still not very familiar with helm.


stefanhipfel (approved):

stefanhipfel:
so in general we disable it for all blocks for now.

sandzwerg:
Yes, but that shouldn't matter much as automated cleaning is still disabled, and it only affects automated cleaning.

PR #4573, opened 2023-02-02
[cc3test] keep last logs and avoid parallel jobs

thgrs:


artherd42:
Maybe better to increase TTL rather than removing completely. With this, resources jobs, pods, ... will pile up.

thgrs:
i think nothing should pile up because of "successfulJobsHistoryLimit: 1" and "failedJobsHistoryLimit: 1"

PR #4575, opened 2023-02-03
[ironic] Handle kos-template and -query the same

fwiesel:
Remove the template also when you remove the query


sandzwerg (commented):
If you tested it, it's fine with me, but maybe let @stefanhipfel also take a look.

PR #4576, opened 2023-02-06
[cc3test] split purgers into their own pods for all services

ashifnihalb:
split purgers into their own pods for all services


thgrs (approved):
lgtm

thgrs (approved):
lgtm

PR #4582, opened 2023-02-06
grafanasix: use new common/pgbackup chart

majewsky:

This change is already rolled out to prod for Arc, Castellum, Elektra, Keppel, Limes, Lyra, Sentry and Tenso.

Note that this only generates a diff in grafanasix-author; the other Grafanas do not have a database backup.

Merge this together with the respective companion PR in cc/secrets.


thgrs (approved):
lgtm

PR #4585, opened 2023-02-07
[glance] Update utils with proxysql startup fixes

fwiesel:

for various bugfixes

### DIFF
--- a/openstack/glance/Chart.lock
+++ b/openstack/glance/Chart.lock
@@ -10,12 +10,12 @@ dependencies:
   version: 0.2.7
 - name: utils
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.7.0
+  version: 0.7.2
 - name: redis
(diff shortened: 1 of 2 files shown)


rajivmucheli (changes requested):
https://github.com/sapcc/helm-charts/pull/4585/files#diff-d8b9787e9f1642b8e345ae0641139508c38498f273f2734deab0c19dd4c4fe9bR21

is this a typo ?

fwiesel:
No, it is intentional:

> tilde ranges (patch version changes allowed), where ~1.2.3 is equivalent to >= 1.2.3 < 1.3.0.

So, the Chart.yaml only specifies only a bound. The exact version is still specified in the Chart.lock.

Next time, we update the utils, we can do that via `helm dep up` without having to edit the Chart.yaml

rajivmucheli (approved):
lgtm

PR #4586, opened 2023-02-07
[nova] Update utils with proxysql startup fixes

fwiesel:

for various bugfixes

### DIFF
--- a/openstack/nova/Chart.yaml
+++ b/openstack/nova/Chart.yaml
@@ -26,7 +26,7 @@ dependencies:
     version: 0.0.12
   - name: utils
     repository: https://charts.eu-de-2.cloud.sap
-    version: 0.6.0
+    version: ~0.7.2
   - name: mariadb
     alias: mariadb_cell2
     condition: mariadb_cell2.enabled
--- a/openstack/nova/Chart.lock
+++ b/openstack/nova/Chart.lock
@@ -16,7 +16,7 @@ dependencies:
   version: 0.0.12
 - name: utils
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.6.0
+  version: 0.7.2
 - name: mariadb
   repository: https://charts.eu-de-2.cloud.sap
   version: 0.7.5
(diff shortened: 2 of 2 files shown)


openstack/nova/Chart.yaml
```diff
@@ -26,7 +26,7 @@ dependencies:
     version: 0.0.12
   - name: utils
     repository: https://charts.eu-de-2.cloud.sap
-    version: 0.6.0
+    version: ~0.7.2
```
joker-at-work, line 29: What does this `~` mean?

fwiesel: > tilde ranges (patch version changes allowed), where ~1.2.3 is equivalent to >= 1.2.3 < 1.3.0

That way, we avoid having to specify the exact version twice:
- Once in Chart.yaml
- Second time in the Chart.lock

The next time we can pull in the new version with `helm dep up`

joker-at-work: So when-ever I render the chart locally, which includes a `helm dep up` to download the dependencies, it will switch away from the specified version to a newer version released by now? I don't really like that.

For upgrading to a new chart version, the dev should (imho) look at what changed in the versions e.g. to be able to write a good commit message for the other devs. Setting that version explicitly in `Chart.yaml` should be a negligible overhead compared to that, so I would stay with exact versions.

fwiesel: > So when-ever I render the chart locally, which includes a `helm dep up` to download the dependencies, it will switch away from the specified version to a newer version released by now? I don't really like that.

I would suggest to  run `helm dependency build` instead, which does what you want. It just downloads the dependencies b on the `Chart.lock`.

`helm dep up`  is `helm dependency update`, and ignores the lock-file, and it shouldn't come as a surprise that it updates the dependencies.

joker-at-work: Makes sense. Didn't know about that command.

joker-at-work (approved):

PR #4587, opened 2023-02-07
[ironic] Update utils with proxysql startup fixes

fwiesel:

for various bugfixes


sandzwerg (approved):

fwiesel:
Already fixed in commit ed20426b59b3cfcffe6e4b93f1d4637405eedfee

PR #4590, opened 2023-02-08
[glance] Generate job-name from proxysql settings

fwiesel:
Jobs cannot be changed, so all settings need
result in a new job-name.


rajivmucheli (approved):
lgtm

PR #4594, opened 2023-02-08
[cc3test] lower grafana alert to warning and make it less sensitive

thgrs:


artherd42 (approved):
agree. leftover from the past where ALL datapath checks were considered as critical

PR #4613, opened 2023-02-09
Change some alerts to warnings

hemna:
The info alerts are basically being ignored and some of these alerts are pretty critical for visibility.  Changing these alerts to warning so they get into the right slack channel with visibility.


chuan137 (approved):
lgtm

PR #4617, opened 2023-02-10
[common/rabbitmq] Shorter hostnames

fwiesel:
Use short hostnames instead of the fqdn,
and rely on the DNS for Services discovery
from kubernetes to resolve it

For normal pods, that should be just a cosmetic change but for the KVM use-case, which runs on baremetal, it allows us to resolve and use external ips
with the same configuration

### DIFF
--- a/common/rabbitmq/CHANGELOG.md
+++ b/common/rabbitmq/CHANGELOG.md
@@ -3,6 +3,11 @@ Rabbitmq CHANGELOG
 
 This file is used to list changes made in each version of the common chart rabbitmq.
 
+0.5.1
+_____
+fabian.wiesel@sap.common
+- Use just the hostname without domain for the transport-url
+
 0.5.0
 -----
 b.alkhateeb@sap.com
--- a/common/rabbitmq/Chart.yaml
+++ b/common/rabbitmq/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rabbitmq
-version: 0.5.0
+version: 0.5.1
 description: A Helm chart for RabbitMQ
 sources:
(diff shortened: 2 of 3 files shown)


bashar-alkhateeb (commented):
If i understood it correctly
it is for the use case of KVM hypervisors and the search domain will be configured on the host itself 
the question is does that apply as well for the Namespace part of the FQDN ? are we going to have only one ?  
AFAIK standard resolver libraries are allowing only 6 search domains to be configured, that's why i am asking

fwiesel:
For a kubernetes pod, this change is practically no change thanks to how [DNS behaviour](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#namespaces-of-services) is specified in kubernetes 
```
search {{.Release.Namespace}}.svc.kubernetes.{{.Values.global.region}}.{{.Values.global.tld}} svc.kubernetes.{{.Values.global.region}}.{{.Values.global.tld}}
```

So, the own namespace is always in the search path.

For a baremetal host, I will configure the search path to

```
search cc.{{.Values.global.region}}.{{.Values.global.tld}}
```
So, for example, `nova-rabbitmq` would trigger a search resolution in both cases
* In a pod would first be searched with `nova-rabbitmq.{{.Release.Namespace}}.svc.kubernetes.{{.Values.global.region}}.{{.Values.global.tld}}`, which is exactly what the chart was saying before.
* In a baremetal host, I would map to `nova-rabbitmq.{{.Values.global.region}}.{{.Values.global.tld}}`, and I have already setup the dns records (in qa) to map that to the publicly exposed url of the rabbitmq service.

The PR doesn't add or requires to add more search domains, it would make use of the first entry already there.

bashar-alkhateeb (approved):

PR #4642, opened 2023-02-13
[PX] Deploy podmonitor with each domain, per domain

swagner-de:
Carve the podmonitor again out of the bird.yaml file and deploy it for every domain found usning loops to find out which domains exists. Previous chart did not deploy the podmonitor at all, since it was in a conditional that requires pxmon to be present.

### DIFF
--- a/px/bird/templates/podmonitor.yaml
+++ b/px/bird/templates/podmonitor.yaml
@@ -1,35 +1,46 @@
-{{- define "podmonitor" -}}
+{{- $domains := list -}} 
+{{- range $service, $service_config := .Values.config -}}
+{{- range $domain, $domain_config := $service_config -}}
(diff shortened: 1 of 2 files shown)


rhalina (approved):

px/bird/templates/podmonitor.yaml
```diff
@@ -1,35 +1,49 @@
...
+---
 apiVersion: monitoring.coreos.com/v1
 kind: PodMonitor
 metadata: 
   labels:
     prometheus: infra-collector
-  name: px
-  namespace: px
+  name: px-domain_{{ $domain_number }}
+  namespace: infra-collector
```
richardtief, line 19: namespace`infra-collector` doesn't exist. You can leave it at `px`.

richardtief (changes requested):

richardtief (approved):

PR #4645, opened 2023-02-14
[dns][bind] health check

mutax:
There is not neccessarily a bash in the image, and the script gets called via sh externally anyhow.
Additionally hostname -i is deprecated and to work it requires working name resolution but could return a different IP than the local one. 
Anyhow, because rndc uses localhost by default, the parameter can be omitted completely.


occamshatchet (approved):
Looks good!

PR #4647, opened 2023-02-14
[dns][unbound] updating root.hints to new version

mutax:


occamshatchet (approved):
Wow, rare to see an ipv4 root server change.

PR #4648, opened 2023-02-14
[dns][unbound] adding serverid with region

mutax:
add server name and region to config, so we can query them using
`dig CH TXT id.server`


occamshatchet (approved):
LGTM

PR #4653, opened 2023-02-15
[nova] Re-enable heartbeat in pthread for uwsgi

fwiesel:
Heartbeat in pthread has been disabled due to
https://bugs.launchpad.net/oslo.messaging/+bug/1961402

But uwsgi needs that setting, so we override in that case via environment according to
https://docs.openstack.org/oslo.config/xena/reference/drivers.html#module-oslo_config.sources._environment

Similarly like already done in neutron 58520f0

### DIFF
--- a/openstack/nova/templates/api-deployment.yaml
+++ b/openstack/nova/templates/api-deployment.yaml
@@ -62,6 +62,10 @@ spec:
               value: "localhost"
             - name: STATSD_PORT
               value: "9125"
+            {{- if .Values.api.use_uwsgi }}
+            - name: OS_OSLO_MESSAGING_RABBIT__HEARTBEAT_IN_PTHREAD
+              value: "true"
+            {{- end }}
             {{- if .Values.sentry.enabled }}
             - name: SENTRY_DSN
               valueFrom:


joker-at-work:
I don't think it's a good idea to do this via environment setting, because the usual way we set settings is via the config and thus people (at least me) are used to searching there. Would it make much of a difference to move that `if` into the configmap instead?

Additionally, `heartbeat_in_pthread` is the default from some point in time but at least in Xena. So the upgrade happening any day now should make this setting mood. https://github.com/sapcc/oslo.messaging/blob/66a77c535c089272accb87e871ade45474f314cf/oslo_messaging/_drivers/impl_rabbit.py#L90-L100

I also found this related bug-report https://bugs.launchpad.net/nova/+bug/1825584 where they implemented ignoring the missed heartbeat error messages.

fwiesel:
>Additionally, heartbeat_in_pthread is the default from some point in time but at least in Xena. So the upgrade happening any day now should make this setting mood.  [...]
> Would it make much of a difference to move that if into the configmap instead?

I think, I was missing to give the necessary context:

`heartbeat_in_pthread=True` is indeed the default in Xena, and the upgrade is the reason we might be in need of this patch, should be choose to enable `uwsgi`.

`heartbeat_in_pthread=True`  triggers a deadlock since Wallaby, which has only been fixed in Yoga.
That is why we set it to `heartbeat_in_pthread=false`  it in the `<project>.conf` [here](https://github.com/sapcc/helm-charts/blob/master/openstack/utils/templates/_ini_sections.tpl#L5) (i.e. `neutron.conf`, `manila.conf`, [PR](https://github.com/sapcc/helm-charts/pull/4654) for `cinder.conf`) for all processes of the service.

I am not sure about the details, @Carthaca can probably give more information, but I know that isn't working for `uwsgi`, and our helm-chart for nova has the option to run it in uwsgi.
So, we need to override the override in some way, and I see two options

- Environment variable, as in all the other applications
- Create a new `nova-api.conf` and place it there.

My main issue with the `nova-api.conf` would be that nova would be the only service going that way. Personally, I am already quite unhappy how divergent the charts are implemented, but especially with changes which apply across services, I would say it is especially important to keep them consistent, so they can also be adapted when needed consistently as well (i.e. removing this override for yoga or later releases)

Carthaca:
I don't have first hand experience with uwsgi and this setting, only more reading material: https://bugs.launchpad.net/oslo.messaging/+bug/1934937

I just observed the hanging services on non-wsgi with manila, which was odd because it did not always hit all my agents.
After a rabbitmq restart some agents were not completely stopping their work, only the parts that relied on rabbitmq were 'hanging'.

----

> My main issue with the nova-api.conf would be that nova would be the only service going that way.

That means providing manila-api.conf and neutron-server.conf could make everyone happy? 😃 
(except for the extra work to implement this, probably)

joker-at-work:
There's on small error I'd like to point out: We already have a `nova-api.conf`: https://github.com/sapcc/helm-charts/blob/f0fca8a9884ece31ba4c20a1b7575f8fb5bdc431/openstack/nova/templates/etc-configmap.yaml#L13-L14

I see the consistency argument.

Before merging this, I'd like to see if it makes a difference i.e. wait for qa-de-2 to be upgraded to Xena (so we actually have that option available) and see how this environment variable changes the options. If there is no change, e.g. because the option set via utils chart in nova.conf already sets this, I would not merge this as it should not have any other effect.

joker-at-work (approved):
I didn't get it before, now I think I do. We only need this in uWSGI-enabled regions and there, only for nova-api. Everywhere else, we use and need the override from the utils chart.

I think this PR makes sense.

PR #4657, opened 2023-02-16
common/prometheus-server: VPA only for prometheus pod

richardtief:


viennaa (approved):
Thanks, matching all is not really fitting the use case anymore.

PR #4665, opened 2023-02-20
[prometheus-infra-global] alert not longer needed

viennaa:
old thanos version which produced this error is not around anymore.


richardtief (approved):
Thanks.

PR #4671, opened 2023-02-21
Adding parent object properties

christopherhans:
Sometimes we get false positive alerts, because the information from the vrops-inventory is outdated.

Would it make sense to compare i.e. the hostsystem label with the parent host from the parentHost property?


richardtief:
Not necessary from my side. The inventory pod must be reliable (on us) and is the source of truth.

PR #4673, opened 2023-02-21
arc, lyra: bump pgbackup to 0.1.7

majewsky:
This updates backup-tools from 0.7.0 to 0.8.1, with the following changes:

- All usage of python-swiftclient is replaced with native Go code, thereby reducing CPU consumption and memory usage spikes.
- Backups are streamed directly into Swift instead of going through the local filesystem first.

Furthermore, the Helm chart has been extended with a configurable memory limit that is raised in the big regions (see companion PR in cc/secrets). This should permanently fix the current backup pod instability in eu-de-1/2.


ArtieReus (approved):
lgtm @databus23 do you want also to check??

PR #4675, opened 2023-02-21
Federate kube_pod_container_resource_requests

Nuckal777:
It replaces kube_pod_container_resource_requests_cpu_cores and kube_pod_container_resource_requests_memory_bytes in version 2 of kube-state-metrics.


viennaa (changes requested):
Thanks for noticing this detail. Please bump the chart version, then we're good to go.

viennaa (approved):

PR #4678, opened 2023-02-22
Update DatastoreNotAccessibleFromHost Alert

christopherhans:
Update DatastoreNotAccessibleFromHost alert. Alert does not fire anymore, if the whole datastore is inaccessible. There are dedicated alerts for this case.


artherd42 (approved):
lgtm

PR #4691, opened 2023-02-23
[cinder] Redo non-vcenter-volume agents

fwiesel:
Most options are simply passed through,
so the logic within the charts is fairly limited.

We need to work around NFS shares though,
and mount them with kubernetes.


fwiesel:
Mounting is completely delegated to kubernetes for now, until we have an image with sapcc/loci#16

hemna:
I am not sure the etc/_volume.conf.tpl file is even used.  I can't see the output from that template in the deployed cinder pods.   I think the vcenter operator generates the cinder-volume.conf file from the 
templates/vcenter_datacenter_cinder_configmap.yaml

fwiesel:
Exactly, the  vsphere volumes from where you describe. 
The changes are related to the "static" configuration via the values files, which we haven't used for ages.

In the past, it was triggered by the values in
```
global:
  volumes_netapp: []
  volumes: []
```
The `global.volumes` was my first attempt some time ago to generalise the `global.volumes_netapp`.
Since they are not used, I propose to remove them and replace with something new.

To trigger those templates, you need to define a volume in a value file. The corresponding PR is in our internal repo for qa-de-1 (and the deployed code as well).

openstack/cinder/values.yaml
```diff
@@ -72,6 +69,22 @@ cinderApiPortPublic: 443
 cinderApiPortInternal: 8776
 cinderApiPortAdmin: 8776
 
+defaults:
+  backends:
+    common: {}
+    cinder.volume.drivers.netapp.common.NetAppDriver:
```
joker-at-work, line 75: Should we consolidate the key `backend_defaults` from above into this structure (as a follow-up maybe)?

joker-at-work (approved):
In general, it looks like we're only supporting static NFS volumes here. Not sure, but maybe that should be reflected in the name.

Overall, no blockers from my side.

openstack/cinder/values.yaml
fwiesel: Yes, I think, that would make sense. I can create a PR for that.

fwiesel: I had another look, and I am not sure, how to integrate it.
The defaults in this section are the defaults for the settings for an individual backend.
The `backend_defaults` are the `backend_defaults` section, which apply to all backends of a host, but only vmware.
It definitely needs some rework, but I am unclear a bit on how to do that.

joker-at-work: I thought we could maybe do something like `defaults.backends."cinder.volume.drivers.vmware.vmdk.VMwareVcVmdkDriver".backend_defaults` or something - but not sure either. If we don't have a good way, then we leave it until we figure it out $somewhen.

PR #4696, opened 2023-02-24
prometheus-server: option to disable object storage

richardtief:


viennaa (approved):
Really cool it can work without the storage backend.

PR #4698, opened 2023-02-27
Alert Manager: add health check alert

bassel-z:


bassel-z:
related: [https://github.wdf.sap.corp/cc/documentation-operation/pull/1642](url)

richardtief (approved):
Looks good. We need to think about improving this, since the alert manager cannot alert to its own existence.

bassel-z:
> Looks good. We need to think about improving this, since the alert manager cannot alert to its own existence.

Agree. I talked with Martin about this. We can follow up on the topic in the next iterations.

PR #4701, opened 2023-02-27
system/hubcopter: add support for multiple deployments

SuperSandro2000:


majewsky (changes requested):
I find it surprising that specific configuration for the different instances is hardcoded here. Usually we put that in the `values.yaml` (i.e. in the secrets repo). Having all the configuration relating to the specific instances in `values.yaml` would also reduce duplication of effort, e.g. the list in `index.html` could be generated by a template.

SuperSandro2000:
Moved the values into the secrets repo.

majewsky (approved):

PR #4705, opened 2023-02-28
[manila] Enable oslo.messaging's ping endpoint

joker-at-work:
Since the Victoria release of OpenStack, oslo.messaging supports an RPC endpoint called "oslo_rpc_server_ping" [0]. This endpoint can be used to implement a generic RPC check to see if a server is still receiving messages/rpc calls via rabbitmq.

[0] https://github.com/sapcc/oslo.messaging/commit/82492442f3387a0e4f19623ccfda64f8b84d59c3


Carthaca (approved):
thx

PR #4706, opened 2023-02-28
Swift: enable write restriction middleware

talal:
Do not merge before https://github.com/sapcc/swift-addons/pull/3 .


majewsky (approved):

majewsky:
I'm going to merge this with the intent of having it in QA. It will not automatically go to prod by itself because of the manual promote gate. Please align with Falk on when it's safe to promote this to prod.

PR #4726, opened 2023-03-02
teardown backup-cleanup deployment

majewsky:
The new backup-tools 0.8.x series uses the Schwift library for uploading to Swift. Schwift knows how to set X-Delete-At on the segments of SLOs, not only the SLOs themselves, so we do not need a separate cleanup machinery for the segments anymore.


reimannf (approved):
Very nice

PR #4743, opened 2023-03-08
[tarslite] Support Basic Auth and Securid Redirect

darpan92:
This change will allow for alertmanager to communicate with basic auth and everyone else with Securid MFA. This was achieved by splitting the urls for both auth types.


max-len:
We need to add the corresponding secret for basic auth. See https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#authentication

darpan92:
The secret has been added to the tars namespace. Its called basic-auth. Both mfa and basic auth are working currently in the production deployment with the secret.

PR #4745, opened 2023-03-09
[nova] Set image default architecture for scheduler

fwiesel:
The scheduler decides according to the image properties on which hypervisor to put the vm.
That requires hw_architecture to be set, but for convenience we can set a default architecture in the schedulers config


joker-at-work:
I don't see in the code, why it requires us to set `hw_architecture`. https://github.com/sapcc/nova/blob/stable/xena-m3/nova/scheduler/filters/image_props_filter.py#L77-L81 looks to me, like it would just ignore unset values.

fwiesel:
You are right, it isn't required. It is just that if that isn't set and we would have an arm64 node, we would deploy on that,
unless this property is set, or all images have  the `hw_architecture` property set.

joker-at-work (approved):
commit-message type:
> In order to avoid trouble should be ever introduce another

should *we*

Otherwise looks good. Thanks for updating the commit message.

PR #4748, opened 2023-03-10
Pyroscope ironic

fwiesel:

PR #4749, opened 2023-03-10
Add env-vars snippet for pyroscope

fwiesel:
The snippet should take care of setting all the environment variables,
taking the defaults from utils, allowing an chart level override, but
also one on an application-level.

It requires a python user-customization file merge in sapcc/loci#15,
and additionally, the process cannot be run as root,
as a root process in a container is more restricted than a non-root user
one.

All settings from pyroscope can be set via the `PYRO_*` environment
variables.

Additionally, there is the `PYRO_APP_SUBSET` variable,
which is a regular expression, which needs to match the hostname,
for the agent to be come active.
The idea is, that you can thereby select a subset of the application
instances to be sampled for a potential A/B split.

### DIFF
--- a/openstack/utils/Chart.yaml
+++ b/openstack/utils/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: utils
-version: 0.7.4
+version: 0.8.0
--- a/openstack/utils/templates/_env.tpl
+++ b/openstack/utils/templates/_env.tpl
@@ -1,0 +1,15 @@
+{{- define "utils.env.pyroscope" }}
+    {{- $envAll := index . 0 -}}
+    {{- $application := index . 1 -}}
+    {{- with $envAll }}
+        {{- $settings := merge (get .Values.pyroscope $application) (dict "app_name" $application) .Values.pyroscope.defaults .Values.utils.pyroscope_defaults }}
+        {{- if $settings.enabled }}
+            {{- range $k, $v := $settings }}
+                {{- if not (eq $k "enabled") }}
+- name: PYRO_{{ $k | upper}}
+  value: "{{ $v }}"
+                {{- end }}
+            {{- end }}
+        {{- end }}
+    {{- end }}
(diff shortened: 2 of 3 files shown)


fwiesel:
See #4748 for a usage-example.

fwiesel:
> What I noticed is that in the ironic example you're setting a SecurityContext with file ownership.
> * Why is this needed?

Not sure where to begin.

Loci containers do not set any user in the Dockerfile, so the user used when starting those images is root.
I was assuming that it is clear, that running things as root is questionable in itself and doesn't need further explanation

But regarding this change it has a funny implication: Root is not allowed to access some things in `/proc` which normal users are, which stops the profiler from working, as it cannot attach itself to the processes.

The file ownership (fsGroup) is more part of the same practice, technically only necessary when using a persistentVolume, which is not NFS. Since we do use currently NFS for PVs, `fsGroup` is rather unnecessary, but that is rather a detail of the implementation and not the spec.

> * How can we document this in this change, that this is needed, why it's needed, what to look for & how to apply it properly?

From my perspective, setting up a correct security context is sufficiently documented in the [standard documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/), and not having any is in my view simply an oversight when switching away from kolla-containers to loci. 

It is simply best practice not to run as root, and I am fairly sure we need rather need to come up with a rationale for why we do, if we do not change it with the next audit.

grandchild:
The need to not run things as root is obvious, it was just not obvious that this is the only thing that's happening. So this is basically a bugfix and not something that Pyroscope _needs_ for some reason I couldn't see. Because the commit message in #4748 just said "Pyroscope integration" I assumed it was something Pyroscope specific.

As I understand it now this is not something that the service author has to change for Pyroscope specifically.

grandchild:
Oh, I guess with the `/proc` thing, it does _need_ it. Then just a remark along the lines of "make sure you don't run as root" would probably be helpful.

fwiesel:
I've updated the commit message with more explanations hopefully answering those questions also
for future users.

PR #4750, opened 2023-03-10
keppel: expose health-monitor result on GET /healthcheck

majewsky:
This can be used by the F5 for keppel.global to monitor the health of the Keppel API in an E2E fashion (including basic health of Keystone and Swift).


defo89 (approved):
lgtm

reimannf (commented):
How are the results exposed? A simple `200` if everything is fine?

reimannf (approved):

PR #4753, opened 2023-03-13
common/thanos: thanos ruler functionality added

richardtief:
This PR enables the common/thanos chart to specify a Thanos Ruler. This is intended to sit alongside a Thanos Query deployment and evaluate alerting and aggregations rules across prometheus.

### DIFF
--- a/common/thanos/templates/alerts/thanos-alerts.yaml
+++ b/common/thanos/templates/alerts/thanos-alerts.yaml
@@ -1,3 +1,6 @@
+{{- /*
+Alerts are deployed dependent upon the components chosen. It could either be Compactor|Query|Store, Query|Ruler or Query.
+*/}}
 {{- if .Values.enabled }}
 {{- $root := . }}
 {{- range $name := coalesce .Values.names .Values.global.targets (list .Values.name) }}
@@ -5,6 +8,7 @@
 {{- range $component := tuple "compactor" "query" "store" }}
 {{- $path := (printf "%s/alerts/_thanos-%s.alerts.tpl" $root.Template.BasePath $component ) }}
 ---
+# This PrometheusRules cover Compactor|Query|Store alerts
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 
@@ -21,6 +25,7 @@ spec:
 {{- range $component := tuple "query" "ruler" }}
 {{- $path := (printf "%s/alerts/_thanos-%s.alerts.tpl" $root.Template.BasePath $component ) }}
 ---
+# This PrometheusRules cover Query|Ruler alerts
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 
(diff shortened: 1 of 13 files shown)


common/thanos/templates/alerts/thanos-alerts.yaml
```diff
@@ -16,19 +16,39 @@ metadata:
 spec:
 {{ include $path (list $name $root) | indent 2 }}
 
+{{- end }}
+{{- else if and $.Values.ruler.enabled (eq $.Values.deployWholeThanos false) }}
+{{- range $component := tuple "query" "ruler" }}
```
viennaa, line 21: it's a little bit unfortunate, that this file has the same code three times. It reads ok but it takes a while to digest. Maybe we can add a comment in the beginning which would help my further self to faster parse this file.

> Alerts are deployed dependent upon the components chosen.  It could either be `Compactor|Query|Store`, `Query|Ruler` or `Query`.

common/thanos/values.yaml
```diff
@@ -136,6 +153,51 @@ query:
...
+  # match the regular expression `[0-9]+(ms|s|m|h|d|w|y)` (milliseconds seconds
+  # minutes hours days weeks years).
+  # retention: <string>
+
+  # Interval between consecutive evaluations. Defaults to 15s
+  # evaluationInterval: <string>
+
+  # The external URL the Thanos Ruler instances will be available under. This
+  # is necessary to generate correct URLs. This is necessary because Thanos Ruler is
+  # served from a subpath of the Thanos Query.
```
viennaa, line 174:
```suggestion
  # The external URL the Thanos Ruler instances will be available under. This
  # is necessary to generate correct URLs to serve Thanos Ruler from a subpath of Thanos Query.
```

common/thanos/values.yaml
```diff
@@ -136,6 +153,51 @@ query:
     #   namespace: default
     #   port: 10901
 
+ruler:
+  enabled: false
+  # When a ThanosRuler deployment is paused, no actions except for deletion
+  # will be performed on the underlying objects.
+  paused: false
```
viennaa, line 160: Interesting, haven't looked into this yet and the explanation is straight from k8s. Not sure how it works though.

common/thanos/templates/ruler/ruler.yaml
```diff
@@ -0,0 +1,58 @@
...
+  name: {{ include "thanos.name" (list $name $root) }}
+spec:
+  image: {{ include "thanosimage" $root }}
+  paused: {{ $.Values.ruler.paused }}
+  logLevel: {{ default "info" $.Values.ruler.logLevel }}
+{{- if $.Values.ruler.retention }}
+  retention: {{ $.Values.ruler.retention | quote }}
+{{- end }}
+{{- if $.Values.rbac.create }}
+  serviceAccountName: {{ include "serviceAccount.name" (list $name $root) }}
```
viennaa, line 16: Is this ruler specific? It's mapped in the overall values but we haven't used it before.

viennaa (commented):
Thanks for providing this well prepared PR and also incorporating the monitoring.mixins straight.

Very little remarks from my side.

common/thanos/templates/ruler/ruler.yaml
viennaa: Ok I've seen the label added to the ServiceAccount. Still would move this to the `ruler` section as it is very specific judging its label.

PR #4766, opened 2023-03-16
[nova] No db in nova.conf / merge confs for wsgi apps

fwiesel:
The compute nodes should not know about the databases, as they are communicating via the conductor with the respective one So, we need to move that information in to the process specific configs.
The wsgi apps though are only loading a nova.conf, so we need to merge the configs to a single file


joker-at-work (commented):
Neutron seems to solve having multiple config files with uWSGI likes this: https://github.com/sapcc/helm-charts/blob/803ae493a49d2399786fa8bbf1b028cb63297e62/openstack/neutron/templates/etc/_uwsgi.ini.tpl#L4

Do you think that would be an option for Nova, too?

I'd prefer to have the same config layout on all pods if possible.

openstack/nova/templates/etc/_nova-api-metadata.conf.tpl
```diff
@@ -0,0 +1,9 @@
+{{ include (print .Template.BasePath "/etc/_nova.conf.tpl") . }}{{- /*
+Included twice, as the the api apps only look at the nova.conf file, so we merge the options manually
+*/}}
+
+{{- include "nova.helpers.ini_sections.api_database" . }}
+
+{{- include "ini_sections.database" . }}
```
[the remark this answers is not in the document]
fwiesel: Unfortunately, it is either no database or "both":
See: https://github.com/sapcc/nova/blob/stable/xena-m3/nova/config.py#L104-L105

fwiesel:
You are right, we can do that without pasting the config-files together.
I've moved the database now to the application specific  config file, and added the arguments to the uwsgi.ini.

openstack/nova/templates/api-metadata-deployment.yaml
```diff
@@ -47,7 +47,7 @@ spec:
             - kubernetes-entrypoint
           env:
             - name: COMMAND
-              value: "nova-api"
+              value: "nova-api-metadata"
```
joker-at-work, line 50: We can't do that. From the [docs](https://docs.openstack.org/nova/latest/admin/cells.html#cells-v2-layout-metadata-api):
```
Global

If you have networks that span cells, you might need to run Nova metadata API globally. When running globally, it should be configured as an API-level service with access to the [api_database.connection](https://docs.openstack.org/nova/latest/configuration/config.html#api_database.connection) information. The nova metadata API service must not be run as a standalone service, using the nova-api-metadata service, in this case.
```

joker-at-work (changes requested):
Please see the inline comment for the requested change.

Seeing all this, I wonder if it would make sense to have an `/etc/nova/db.conf` and specify the config files or just the config dir on the command line for every container.

openstack/nova/templates/api-metadata-deployment.yaml
fwiesel: Changed it back

fwiesel:
I would leave that change for a different PR.
Personally, I have a strong preference of making use of defaults. "Convention over configuration".

But I think, you wish can be addressed with the default of [config_dir](https://docs.openstack.org/oslo.config/latest/configuration/options.html#default)
So,  `/etc/nova/nova.conf.d/db.conf` should do the trick.

openstack/nova/templates/etc/_api_uwsgi.ini.tpl
```diff
@@ -16,3 +16,4 @@ master = true
 memory-report = true
 processes = {{ .Values.api.wsgi_processes }}
 wsgi-file = /var/lib/openstack/bin/nova-api-wsgi
+pyargv = --config-file /etc/nova/nova.conf --config-file /etc/nova/nova-api-os-compute.conf
```
joker-at-work, line 19: This needs changing to `nova-api.conf`

joker-at-work (changes requested):

openstack/nova/templates/etc/_api_uwsgi.ini.tpl
fwiesel: Fixed it.

joker-at-work (approved):

PR #4772, opened 2023-03-17
[nova] Fixup PR #4766: DB access for Jobs with projected volumes

fwiesel:
This should just fix the jobs, the other specs will get updated accordingly.

- Instead of pasting the credentials into serveral files, rely on the default include path pattern.

- Instead of having the credentials for the dbs in a configmap, put them in a secret.

- Instead of mounting all files individually, create a projected volume in the spec which contains them.

### DIFF
--- a/openstack/nova/templates/db-migrate-job.yaml
+++ b/openstack/nova/templates/db-migrate-job.yaml
@@ -37,50 +37,41 @@ spec:
           value: {{ .Values.python_warnings | quote }}
         volumeMounts:
         - mountPath: /etc/nova
-          name: etcnova
-        - mountPath: /etc/nova/nova.conf
           name: nova-etc
-          subPath: nova.conf
-          readOnly: true
-        - mountPath: /etc/nova/nova-api.conf
-          name: nova-etc
-          subPath: nova-api.conf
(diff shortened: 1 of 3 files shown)


openstack/nova/templates/db-migrate-job.yaml
```diff
@@ -37,50 +37,41 @@ spec:
...
-          subPath: nova.conf
-          readOnly: true
-        - mountPath: /etc/nova/nova-api.conf
-          name: nova-etc
-          subPath: nova-api.conf
-          readOnly: true
-        - mountPath: /etc/nova/api-paste.ini
-          name: nova-etc
-          subPath: api-paste.ini
-          readOnly: true
```
joker-at-work, line 52: Any idea why we would need api-config and policy for the migration job?

joker-at-work (approved):

openstack/nova/templates/db-migrate-job.yaml
fwiesel: No idea. I suspect it is a simply the result of copy-and-paste, but I didn't want delay my upgrade of kvm to xena to test which of the files are neccessary.

PR #4774, opened 2023-03-20
Deploy multiple statsd-exporter

artherd42:


prometheus-exporters/statsd-exporter/templates/deployment.yaml
```diff
@@ -1,82 +1,76 @@
...
-      {{- include "statsd-exporter.selectorLabels" . | nindent 6 }}
+      {{- include "statsd-exporter.labels" $exporter.name | nindent 6 }}
   template:
     metadata:
       labels:
-        {{- include "statsd-exporter.selectorLabels" . | nindent 8 }}
+       {{- include "statsd-exporter.labels" $exporter.name | nindent 8 }}
       annotations:
-        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") $ | sha256sum }}
+        checksum/config: {{ include (print $.Template.BasePath "/config.yaml") $ | sha256sum }}
```
thgrs, line 24: shoudn't this remain configmap.yaml?

prometheus-exporters/statsd-exporter/values.yaml
```diff
@@ -1,98 +1,78 @@
...
+      # Prometheus
+      metricsPort: 9102 # The address on which to expose the web interface and generated Prometheus metrics.
+      metricsPath: /metrics # Path under which to expose metrics.
+      prometheus: infra-frontend
+      scrapeInterval: 60s
+      scrapeTimeout: 55s
+
+    - name: 'statsd-exporter-inf'
+      enabled: true
+      udpPort: 9125 # The UDP port on which to receive statsd metric lines.
```
thgrs, line 49: just a thought: would it maybe make sense to use a different port here? but maybe it does not make too much sense ...

thgrs (changes requested):

prometheus-exporters/statsd-exporter/templates/deployment.yaml
artherd42: I have deleted the wrong config file. Just fixed.

prometheus-exporters/statsd-exporter/values.yaml
artherd42: Doesn't matter since each has it's own ingress,

PR #4784, opened 2023-03-20
arc/lyra: bump pgbackup chart from 0.1.7 to 0.1.10

majewsky:
This upgrades https://github.com/sapcc/backup-tools from 0.8.0 to 0.9.1, including a cleanup of all code towards current best practices. Starting with 0.9.x, backup-tools releases are validated by E2E tests in Concourse.

*Update:* This now also bumps the pgmetrics chart to update prometheus-exporter to 0.11.1 and set `support_group` on all alerts.

### DIFF
--- a/openstack/lyra/values.yaml
+++ b/openstack/lyra/values.yaml
@@ -84,13 +84,3 @@ pgmetrics:
   db_name: monsoon-automation_production
   alerts:
     support_group: containers
-  customMetrics:
-    pg_database:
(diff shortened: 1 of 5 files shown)


openstack/lyra/values.yaml
```diff
@@ -84,13 +84,3 @@ pgmetrics:
   db_name: monsoon-automation_production
   alerts:
     support_group: containers
-  customMetrics:
```
ArtieReus, line 87: custom metrics is gone now or are included in the new version?

PR #4803, opened 2023-03-24
[sap-seeds] Extract common flags to snippets, set hypervisor to vmware

fwiesel:
We want to be able to allow users to target another hypervisor (libvirt) in the future, so for compatability reasons, the existing flavors need to target vsphere.

We chose flavors over image attributes as it is more explicit, and we only want allow boot-from-volume on KVM, while the existing flavors have root-disks.

As a side-effect, it "breaks" forcing those non-baremetal flavor to a baremetal host.

### DIFF
--- a/openstack/sap-seeds/templates/_flavors_hana_no_exclusive.tpl
+++ b/openstack/sap-seeds/templates/_flavors_hana_no_exclusive.tpl
@@ -4,8 +4,7 @@
   ram: 373352
   disk: 64
   extra_specs:
-    "vmware:hv_enabled": "True"
-    "hw_video:ram_max_mb": "16"
+    {{- tuple . "vmware_common" | include "sap_seeds.helpers.extra_specs" | indent 4 }}
     "resources:CUSTOM_MEMORY_RESERVABLE_MB": "373352"
     "trait:CUSTOM_NUMASIZE_C48_M729": "required"
     "hw:cpu_cores": "24"  # used in nova-vmware as cores-per-socket (12pCPU = 24vCPU)
@@ -15,8 +14,7 @@
   ram: 746720
   disk: 64
   extra_specs:
-    "vmware:hv_enabled": "True"
-    "hw_video:ram_max_mb": "16"
+    {{- tuple . "vmware_common" | include "sap_seeds.helpers.extra_specs" | indent 4 }}
     "resources:CUSTOM_MEMORY_RESERVABLE_MB": "746720"
     "trait:CUSTOM_NUMASIZE_C48_M729": "required"
     "hw:cpu_cores": "48"  # used in nova-vmware as cores-per-socket (24pCPU = 48vCPU)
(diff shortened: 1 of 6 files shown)


joker-at-work:
> As a side-effect, it "breaks" forcing those non-baremetal flavor to a baremetal host.

People will be unhappy about that. We need to bring that up more prominently. @B1Sandmann

fwiesel:
Definitely, I would propose to create a special private flavor for that use-case, and give it out to the projects that need it.

openstack/sap-seeds/templates/flavor-seed.yaml
```diff
@@ -13,477 +13,385 @@ spec:
     disk: 1
     is_public: true
     extra_specs:
-      "vmware:hv_enabled": "True"
-      "hw_video:ram_max_mb": "4"
```
joker-at-work, line 17: idk why, but this flavor has only 4 MB of video RAM and would now get 16 MB. With 16 MB it would be larger than 512 MB and we usually set `ram` to "$fullsize - $videoram"

openstack/sap-seeds/templates/_flavors_hana_no_exclusive.tpl
```diff
@@ -93,20 +86,17 @@
   disk: 64
   is_public: true
   extra_specs:
-    "vmware:hv_enabled": "True"
+    {{- tuple . "vmware" | include "sap_seeds.helpers.extra_specs" | indent 4 }}
     "hw_video:ram_max_mb": "16"
```
joker-at-work, line 90: `hw_video:ram_max_mb` is included in the tuple and can be removed

joker-at-work (changes requested):
When are we going to add the new private flavor? With rolling these changes out, we would need to have the alternative available at roughly the same time, I would say.

openstack/sap-seeds/templates/flavor-seed.yaml
fwiesel: Fixed it, now one can pass the macro a dictionary with values for a flavor which override the defaults.

openstack/sap-seeds/templates/_flavors_hana_no_exclusive.tpl
fwiesel: Removed it

sandzwerg:
FYI this is my current github.com user

openstack/sap-seeds/templates/flavor-seed.yaml
```diff
@@ -6,484 +6,400 @@ spec:
...
   - monsoon3/nova-flavor-seed
   flavors:
+  - name: "forced"
+    id: "1"
+    vcpus: 1
+    ram: 1024
+    disk: 0
+    is_public: false
+    extra_specs:
+      {{- tuple . "forced" | include "sap_seeds.helpers.extra_specs" | indent 6 }}
```
joker-at-work, line 16: This renders as `extra_specs:` - are you sure this will work with the seeder? Shouldn't it maybe be `extra_specs: {}` or something to have the right type?

fwiesel: It looks it doesn't matter. The parsed results are looking like this:
```
{'name': 'forced', 'id': '1', 'ram': 1024, 'vcpus': 1, 'is_public': False}
```
It may be due to the `omitempty` annotation for the [field](https://github.com/sapcc/openstack-seeder/blob/2b1a09bd9676575277640fecdfb512f874e66687/api/v2/openstackseed_spec.go#L229).

joker-at-work (approved):

PR #4816, opened 2023-03-30
[Cinder] Upgrade memcached chart to 0.1.0

hemna:
This patch updates the memcached dependency to 0.1.0.


joker-at-work:
I think #4810 is the better one, because the commit-message contains more information about what changes with the dependency update.

PR #4817, opened 2023-03-30
srs: add `runAsNonRoot: true` or `runAsUser: 0` to most containers

majewsky:
This is a first pass that mostly just documents the state of things and marks where work can be done to move more stuff from root to non-root.

Most of our own container images (using the Dockerfile template from go-makefile-maker) are already running as `USER 4200:4200`, so we can immediately add `runAsNonRoot: true` here and be done.

Same for statsd-exporter: It has `USER 65534` (that UID is `nobody` by convention), so we can also just set `runAsNonRoot: true`.

In `keep-image-pulled`, several external images default to user `root`, but we can force the `sleep` to execute as user `nobody` with explicit `runAsUser/runAsGroup: 65534` directives.

In Swift, several images are legitimately running as root, so I documented this by putting in an explicit `runAsUser: 0` directive. We should work on reducing the number of images that need this (esp. by checking if shared mount propagation can be used to obsolete the unmount-helper), but this is outside the scope of this PR.

In those images which currently run with root permissions, but which don't look like they need it, I put a TODO in to move the image and/or container towards a `runAsNonRoot` setup. Final note: I did not consider the containers in the `common` charts yet.

This change should be fairly safe to apply. Even if I made a mistake, it will only cause the containers to immediately fail and go into state `RunContainerError`, so the CI in QA will likely catch all mistakes.

### DIFF
--- a/openstack/backup-replication/templates/replication-deployment.yaml
+++ b/openstack/backup-replication/templates/replication-deployment.yaml
@@ -59,6 +59,8 @@ spec:
           value: 'true'
         - name: 'DEBUG'
           value: {{ $.Values.swift_http_import.debug | quote }}
+        securityContext:
+          runAsNonRoot: true
         volumeMounts:
         - name: bin
           mountPath: /bin-backup-replication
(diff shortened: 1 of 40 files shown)


reimannf (approved):
Nice prep work. Thank you

PR #4820, opened 2023-03-31
[netapp-exporter] using single prometheus directive

viennaa:
services are created only if explicitly set in netapp-perf-exporter


chuan137 (approved):
lgtm
thanks for taking extra care for maintaining the change logs

PR #4822, opened 2023-04-03
Cbshelf

corey-aloia:
Adding in cbshelf chart


openstack/cbshelf/Chart.yaml
```diff
@@ -0,0 +1,13 @@
...
+description: A Helm chart for Shelf Application
+type: application
+version: 0.1.0
+appVersion: 1.16.0
+dependencies:
+  - name: postgresql
+    repository: file://../../common/postgresql
+    version: 0.3.0
+  - name: pgbackup
+    repository: file://../../common/pgbackup
```
majewsky, line 12: Please pull this chart from Chartmuseum.
```suggestion
    repository: https://charts.eu-de-2.cloud.sap
```
Otherwise this reference will break when we update pgbackup the next time. (This does not apply to the postgresql chart which we serve as path dependency for historical reasons.)

openstack/cbshelf/templates/cronjob.yaml
```diff
@@ -0,0 +1,66 @@
...
+                      values:
+                      - {{ tpl .Values.volume.backup.app $ }}
+                  topologyKey: "kubernetes.io/hostname"
+          volumes:
+          - name: backup
+            persistentVolumeClaim:
+                claimName: {{ .Values.volume.backup.pvcName }}
+          containers:
+            - name: backup
+              image: 'keppel.eu-de-1.cloud.sap/ccloud-dockerhub-mirror/restic/restic:0.15.1'
```
majewsky, line 28: Please use templating here, in order to automatically pick up the change if we ever need to move this account.
```suggestion
              image: '{{ .Values.global.dockerHubMirror }}/restic/restic:0.15.1'
```

openstack/cbshelf/templates/backend-configmap.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: {{ .Release.Namespace }}
```
majewsky, line 4: As a general remark that applies to all template files, you do not need to set `metadata.namespace = {{.Release.Namespace}}`. That's automatically implied by Helm.

PR #4824, opened 2023-04-03
[Cinder] Created critical alert for CinderShardMaxVolumeSize

hemna:
This patch renames CinderShardMaxVolumeSizeInfo to CinderShardMaxVolumeSize as well creates a warning where < 2 datastores match and a critical alert where there are no datastores that can create a max size volume.


kevin-fischer (changes requested):
Thanks a lot Walter, that looks good

But could you @hemna please change the alert name so that CinderShardMaxVolumeSizeInfo then becomes CinderShardMaxVolumeSize for both warning and critical

kevin-fischer (approved):

PR #4825, opened 2023-04-03
[Cinder] Enable stacking of cinder volumes on pools

hemna:
This patch changes the default scheduler behavior to stack volumes on pools (datastores) again.  We had disabled this because we couldn't create independent clones or snapshots, so when a datastore got full a customer couldn't snap or clone their volume.  Cinder now supports creating independent snaps and clones, so it makes sense to start stacking volumes again.

We want to stack volumes to ensure that we always have storage available to house the max volume size creation on the new datastores.


joker-at-work (approved):

hemna:
no need to do this now.

PR #4826, opened 2023-04-03
[Cinder] Modify the CinderFreeSpace alerts

hemna:
for warning it will fire when <= 2 datastores are low on space critical it will fire when <= 1 datastores are low on space


kevin-fischer (approved):
Looks good
Thanks @hemna!

PR #4831, opened 2023-04-05
summarize resource utilization by support-group and service

majewsky:
I'm almost certain that these queries work as intended. :)


Nuckal777 (approved):
lgtm

PR #4838, opened 2023-04-12
Pull keystone-statsd from dockerHubMirrorAlternateRegion

bbobrov:
Pulling keystone-statsd from the region where keystone runs might lead to circular dependency: keystone image cannot be obtained, because image service cannot run, because keystone is not available.

Pull keystone-statsd components from another region to avoid the dependency.


rajivmucheli (approved):
lgtm

PR #4846, opened 2023-04-18
Updating HostNotResponding alerts

christopherhans:
1. changing power state for HostNotResponding alerts to not "Powered Off". Not reponding hosts usually respond an unknown powerstate. However just filtering for "Unknown" will not cover "Powered On" hosts, in case the powerstate gets reported correctly.
2. Removing not responding hosts from HADetectedAPossibleHostFailure to prevent alert duplcation.


max-len (approved):
lgtm

PR #4847, opened 2023-04-18
[openstack_performance] Helm chart to deploy locust with openstack_performance tests

misamoylov:
[openstack_performance] Helm chart to deploy locust with openstack_performance repository.

### DIFF
--- a/openstack/openstack_performance/templates/deployment.yaml
+++ b/openstack/openstack_performance/templates/deployment.yaml
@@ -30,7 +30,7 @@ spec:
             defaultMode: 0400
       containers:
         - name: {{ .Chart.Name }}
-          image: "{{ .Values.image.repository }}:latest"
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command:
             - sh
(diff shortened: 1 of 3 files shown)


openstack/openstack_performance/values.yaml
```diff
@@ -0,0 +1,13 @@
+# Default values for locust.
+# This is a YAML-formatted file.
+locustCmd: "locust -f /data/src/openstack_performance/openstack.py OpenstackUser"
```
galkindmitrii, line 3: what is the point for duplicating this in both entrypoint and in args of the deployment?

openstack/openstack_performance/values.yaml
```diff
@@ -0,0 +1,13 @@
...
+pip_packages:
+  - keystoneauth1
+  - urllib3
+
+command:
```
galkindmitrii, line 8: why do we need this value? it is not used.

openstack/openstack_performance/templates/deployment.yaml
```diff
@@ -0,0 +1,104 @@
...
+        - name: config
+          configMap:
+            name: openstack-performance-config
+        - name: git-ssh-key
+          secret:
+            secretName: git-ssh-key
+            defaultMode: 0400
+      containers:
+        - name: {{ .Chart.Name }}
+          image: "{{ .Values.image.repository }}:latest"
```
galkindmitrii, line 33: for development latest is fine, but we should use actual tags later on.

galkindmitrii (changes requested):

galkindmitrii (approved):

PR #4848, opened 2023-04-19
Fix ACI endpoint learning metrics

swagner-de:
Prometheus labels should be used to _differentiate the characteristics of the thing that is being measured_.
(https://prometheus.io/docs/practices/naming/#labels) In this case the label contained the actual value that was being measured. This should hence be 3 individual metrics, where their value carries the measured state. Also simplified the regex a bit.

ACI leafs currently carry the platform NXOS in our inventory which is technically incorrect. However this is what the SD will feed into prometheus, so I deleted the aci-leaf device type. `terminal length 0` will be executed as init command and will return an error, yet that will be ignored.

### DIFF
--- a/prometheus-exporters/network-generic-ssh-exporter/templates/_config.yaml.tpl
+++ b/prometheus-exporters/network-generic-ssh-exporter/templates/_config.yaml.tpl
@@ -2,9 +2,9 @@ credentials:
   default:
     username: {{ .Values.network_generic_ssh_exporter.user }}
     password: {{ .Values.network_generic_ssh_exporter.password }}
-#  apic:
-#    username: {{ .Values.network_generic_ssh_exporter.apic_user }}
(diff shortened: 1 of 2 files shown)


IvoGoman (commented):
Do we need the credentials section for `apic` in line 5-7? Or should this just be deleted?

swagner-de:
I suppose they are needed, as ACI uses a different syntax for the username than all the other equipment. Do you know how which get-params prometheus uses to query the exporter? Where do we  have that defined?

IvoGoman:
This will be configured in Atlas: https://github.com/sapcc/helm-charts/blob/7c894cfc7985e3ac3628c92a177880d45cde42af/system/atlas/templates/configmap.yaml#L48-L59

swagner-de:
Alright, I matched atlas and exporter config now.

IvoGoman (approved):
Looks good 👍

PR #4850, opened 2023-04-19
Change VMFS DS Capacity Alert tier and Severity

himanip94:
Change VMFS Datastore Capacity Alert tier from OS to VMware and reduce severity from Warning to Info


kevin-fischer (approved):
Looks good @himanip94
Thx Himani

PR #4852, opened 2023-04-20
[cronus] add k8s startup probe

sapmli:


openstack/cronus/templates/cronus/statefulset.yaml
```diff
@@ -68,6 +68,16 @@ spec:
...
 {{- end }}
+          startupProbe:
+            httpGet:
+{{- if .Values.cronus.tls }}
+              scheme: HTTPS
+{{- end }}
+              path: /healthz
+              port: http
+            initialDelaySeconds: 3
+            failureThreshold: 3
```
corey-aloia, line 79: I guess this means that worst case startup time is "15 seconds". i.e. threshold * period seconds. Is this correct? From https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes

sapmli: yes, exactly.

corey-aloia: Just to add: I believe what would happen in this case is that if the startup probe does not succeed in this time, then the pod is restarted. So if 15 seconds is not long enough, then it could end up never starting

sapmli: or better worst case non-startup time.

corey-aloia (approved):

openstack/cronus/templates/cronus/statefulset.yaml
sapmli: The current readiness is defined for having an initial 3 seconds delay. I guess this is the time we can expect the pod to come up. So I used 5 seconds as a boundary and let it happen 3 times in a row. In my eyes, this should definitely be enough to decide whether this pod is ever starting up or not.

PR #4854, opened 2023-04-20
remove remaining references to k8s.gcr.io

majewsky:
I want to start tearing down the `ccloud-k8sgcr-mirror`. I plan to merge this at the end of next week latest, until someone objects.


richardtief (approved):
All good.

defo89 (approved):
lgtm, but maybe a side note - should we consider pointing to `keppel.global.cloud.sap` everywhere instead?

majewsky:
@defo89 Yes, I forgot about that. I adjusted the places that I was touching anyway.

PR #4861, opened 2023-04-24
[nannies] manila: switch prom host

Carthaca:
follow move of netapp metrics from infra-collector to prometheus-storage


chuan137 (approved):
lgtm

PR #4866, opened 2023-04-25
openstack/keppel-trivy: init

SuperSandro2000:


openstack/keppel-trivy/templates/deployment.yaml
```diff
@@ -0,0 +1,45 @@
...
+        alert-tier: os
+        alert-service: keppel
+    spec:
+      containers:
+        - name: trivy-server
+          image: {{ $.Values.global.ghcrIoMirror }}/aquasecurity/trivy:{{ $.Values.trivy.image_version | required ".Values.clair.image_version is not set!" }}
+          imagePullPolicy: IfNotPresent
+          command:
+            - trivy
+            - server
```
majewsky, line 31: The `--token` argument should be added to avoid unauthorized usage of the trivy-server instance.

majewsky (changes requested):

openstack/keppel-trivy/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - Muhammad Talal Anwar
+  helm-chart-url: https://github.com/sapcc/helm-charts/tree/master/openstack/keppel-trivy
+
+trivy:
+  image_version: "0.40.0"
+
+redis:
+  persistence:
+    enabled: true
+    accessMode: ReadWriteMany
```
majewsky, line 22: We don't support ReadWriteMany on our Cinder storage class, so this gets downgraded to ReadWriteOnce by the CSI.
```suggestion
    accessMode: ReadWriteOnce
```

openstack/keppel-trivy/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+    maxmemory-policy: allkeys-lru
+    maxmemory-samples: 5  
+
+  resources:
+    limits:
+      memory: 1Gi
+      cpu: 100m
+    requests:
+      memory: 100Mi
+      cpu: 10m
```
majewsky, line 38: We are expecting around 3-5 GiB initially and maybe more in the future, so we need to give a substantially higher limit here. Let's start with `5Gi` for both memory request and limit and then we'll see where it goes from there.

majewsky: Also, going by https://redis.io/docs/reference/eviction/, it appears that we should prefer `maxmemory-policy: allkeys-random`.

openstack/keppel-trivy/README.md
```diff
@@ -0,0 +1,5 @@
+## Trivy for Keppel
+
+This is the Trivy deployment that is bundled with Keppel.
```
majewsky, line 3:
```suggestion
This is the [Trivy](https://github.com/aquasecurity/trivy) deployment that is bundled with Keppel.
```

openstack/keppel-trivy/values.yaml
SuperSandro2000: like we discussed I went with volatile-ttls

majewsky (approved):

PR #4870, opened 2023-04-27
[tempest] Use oslo.config env source for "dynamic" variables

fwiesel:
Instead of editing the config-file, we can simply set the environment variable according to the following schema:

https://github.com/openstack/oslo.config/blob/master/oslo_config/sources/_environment.py


fwiesel:
Doesn't seem to work

PR #4873, opened 2023-04-28
[openstack/utils] Snippet for kubernetes-entrypoint in initcontainer

fwiesel:
Moving the logic to an initcontainer means we do not have to build in the executable in the main image and can share that image among services.

Also, a missing dependency won't get mixed up with "normal" program failures.


joker-at-work (approved):
Having a usage-example would be nice e.g. in the commit-message, so it's easier for people to adopt it and to know how the parameters should look like.

fwiesel:
Added an example to the commit message

PR #4874, opened 2023-04-28
[ironic] Move kubernetes-entrypoint to init containers

fwiesel:
services.

failures.


sandzwerg (approved):
I admit I don't get all the details, but that was true for the current state as well. IIRC you've tested this already? If so, we can try it in QA and roll out later.

fwiesel:
I manually deployed this to qa, so I am fairly confident it works.
I wouldn't move it though to production, simply because there is no change for better or worse in the actual application.

PR #4877, opened 2023-04-28
[glance] Move kubernetes-entrypoint to init containers

fwiesel:
services.

failures.


rajivmucheli (approved):
lgtm, thanks for taking care.

PR #4879, opened 2023-05-02
[cinder] Move kubernetes-entrypoint to init containers 

fwiesel:
Split into smaller commits
Some minor cleanup plus the actual change:

services.

failures.


hemna:
Need to resolve the conflicts

fwiesel:
Rebased it

fwiesel:
Closing in favour of  #5805

PR #4884, opened 2023-05-02
[mariadb] explicitly add some namespace metadata

Carthaca:
it is already implicitly included as shown in the helm diff steps


viktor-kovacs (approved):
Hi, If the changed required the code looks good to me.
Service Owner should be notified about the change.

PR #4886, opened 2023-05-03
[glance] Change the job-name when the init-container changes

fwiesel:
The job needs to be immutable, so any change to it should be a different job(-name).


rajivmucheli (approved):
lgtm

PR #4888, opened 2023-05-04
[manila] adding init to manila-api for nfs permissions

Carthaca:
42424 is the default uid and gid coming from loci

directly going for the workaround provided by
https://github.com/kubernetes/examples/issues/260


chuan137 (approved):
lgtm

PR #4889, opened 2023-05-04
[Cinder] Downgrade CinderShardMaxVolumeSize alerts

hemna:
This patch temporarily downgrades the  CinderShardMaxVolumeSize alerts to info.  We are going to test out stacking for 2 weeks in ap-cn-1 and see how it goes.  So for the duration we decided in the storage meeting to downgrade the max volume size alerts to info.


kevin-fischer (approved):
LGTM

PR #4895, opened 2023-05-05
[sap-seeds] Add m_c2_m16 & m_c64_m512 for Hana Cloud

grandchild:


openstack/sap-seeds/templates/flavor-seed.yaml
```diff
@@ -267,6 +267,15 @@ spec:
...
     extra_specs:
       {{- tuple . "vmware" | include "sap_seeds.helpers.extra_specs" | indent 6 }}
+  - name: "m_c2_m16"
+    id: "138"
+    vcpus: 2
+    ram: 16368
+    disk: 64
+    is_public: true
+    extra_specs:
+      "catalog:alias": "m2.4xlarge"
```
joker-at-work, line 277: This seems to be an error during copy and paste as we now have 2 flavors with the same alias.

joker-at-work (changes requested):

openstack/sap-seeds/templates/flavor-seed.yaml
grandchild: oops, thanks!

dhoeller (approved):
LGTM

joker-at-work (approved):

PR #4902, opened 2023-05-09
[ironic] Make random job-name lowercase

fwiesel:
Only lower-case characters are allowed


sandzwerg (approved):
of course they are. I'd really like that linting :/

PR #4908, opened 2023-05-10
Adding capacity metrics

christopherhans:
Capacity metrics needed for "Percent Cluster Resource" admission control policy.


richardtief (approved):
Looks good.

PR #4914, opened 2023-05-11
common/thanos: add support for regional thanos

richardtief:


viennaa (approved):
Everything clarified via call, good to go.

PR #4919, opened 2023-05-12
thanos-metal: thanos-regional spec added

richardtief:


viennaa (approved):
Thanks for introducing this guy 👍

PR #4945, opened 2023-05-17
[compute-alerting] Fix DatastoreDisconnected* alerts

max-len:
As discussed, this will use only the string with the first dash in it for lable matching to make it work for e.g. `node042-ap133-nvme-7`


christopherhans (approved):
ty

PR #4949, opened 2023-05-22
prombump, bringing back hw metrics

viennaa:
* metrics were previously dropped to ensure a stable environment (high cardinality, low resources available)
* technically reverts #a88e363 #062a6e5, additionally to #3cd790f


richardtief (approved):
Thanks. All good.

PR #4951, opened 2023-05-22
[infra-monitoring] Add job for linux kvm machines

fwiesel:
Only in qa as a PoC, so subject to change before going to production.


viennaa (changes requested):
Thanks for providing the KVM scrape job. Two things to before we can continue: 

* rebase on latest master to fix the failing test
* update the minor chart version

viennaa (approved):

PR #4955, opened 2023-05-22
ipmi-exporter: move metal-esxi alerts to thanos-ruler regional

richardtief:
The newly introduced Thanos Rule component is intended to allow the evaluation of Prometheus Rules across multiple Prometheis. In this case, metrics from the infra-collector and the numerous vmware Prometheis are needed to evaluate the rules. Therefore, they are labeled on the regional Thanos Rule.


BerndKue (approved):
lgtm

PR #4963, opened 2023-05-24
system/elastic: add sts metric to cluster yellow alert

IvoGoman:
the alert is firing often in the last weeks due to node
maintenance in the region. Yellow status of the cluster
itself is not so problematic, as the cluster is able to
resync. If one or more nodes is also facing volume attach
issues then this can become a problem.
This additional condition should improve this alert, as
the OnDuty only needs to check once the cluster is actually
missing elastic cluster nodes.


Kuckkuck:
👍

PR #4964, opened 2023-05-24
[prober-static] add cp filer alert

defo89:


databus23 (approved):

databus23:
Out of curiosity, where can I see the probe definition?
Its not clear to me from which region the prober is called.

defo89:
> Out of curiosity, where can I see the probe definition? Its not clear to me from which region the prober is called.

I agree this was missing and not quite clear, so I added an option to supply extra labels per each probe. Now the alert summary and description should have it.

PR #4965, opened 2023-05-24
[openstack/utils] Templates to spread pods across AZs

fwiesel:
topologySpreadConstraints lets us define, how the pods should
be spread across different topologies, like nodes, racks, or in our case
(availability-)zones.

The topology key is by default the deprecated
"failure-domain.beta.kubernetes.io/zone", but can be switched via the
global value `.global.topology_key`.

By spreading the pods out, we reduce the risk of a downtime
due to all pods being affected by a problem in one zone.

The scheduler should handle it with best effort.

This also a preparation for [Topology Aware Routing](https://kubernetes.io/docs/concepts/services-networking/topology-aware-routing/),
when configured, would allow us prioritize communication within a zone.

But for that to make even sense, the (usually api) nodes
need to be spread out a bit.

### DIFF
--- a/openstack/utils/Chart.yaml
+++ b/openstack/utils/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: utils
-version: 0.9.2
+version: 0.9.3
--- a/openstack/utils/templates/_affinity.tpl
+++ b/openstack/utils/templates/_affinity.tpl
@@ -32,10 +32,10 @@ affinity:
       - weight: 1
         preference:
           matchExpressions:
-            - key: "failure-domain.beta.kubernetes.io/zone"
+            - key: "{{ .Values.global.topology_key }}"
               operator: In
               values:
-                - {{$availability_zone}}
+                - {{ $availability_zone }}
 {{- end }}
 
 {{- define "kubernetes_maintenance_affinity" }}
--- a/openstack/utils/templates/_topology.tpl
+++ b/openstack/utils/templates/_topology.tpl
@@ -1,0 +1,57 @@
+{{/*
+  Topology constraints to spread nodes across azs to ensure the
+  service being up even in the face of an az being down
+*/}}
+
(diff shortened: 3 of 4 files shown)


galkindmitrii:
jfyi: the zone label is not yet set in all regions.

Carthaca:
I asked myself: Why not simply use `podAntiAffinity`? 
[stackoverflow tells me this](https://stackoverflow.com/questions/73157345/kubernetes-spread-pods-across-nodes-using-podantiaffinity-vs-topologyspreadconst)
I'm thinking in this case, it could as well be mapped the other way (I believe availability zones are a linear topology)

What were your thoughts, Fabian?

fwiesel:
My understanding is, they serve different purposes. `podAntiAffinity` is for clear affinity / anti-affinity rules, where it is either more or less a yes or no decision pod-by-pod decision (i.e. place all those pods together, or do not place them together), while for `topologySpreadConstraints` the constraint is something to be optimized "globally" for, and unlikely to be fulfilled perfectly.

The official documentation explicitly points for the use-case we are looking at to [topologySpreadConstraints](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#pod-topology-spread-constraints).

I am a bit unclear about how the "failure" mode for `podAntiAffinity` looks like. The documentation says:
> You can indicate that a rule is soft or preferred, so that the scheduler still schedules the Pod even if it can't find a matching node.

It doesn't say where to. The `topologySpreadConstraints` makes clear it does a best effort to fulfill the request:
> if you select `whenUnsatisfiable: ScheduleAnyway`, the scheduler gives higher precedence to topologies that would help reduce the skew.

fwiesel:
@galkindmitrii now that you say it, I reconsidered and make the value configurable and default to `failure-domain.beta.kubernetes.io/zone`, we can then change it centrally if the region has the `topology.kubernetes.io/zone` and it will be used by both the `kubernetes_pod_az_affinity` macro, as well as the new one here.

Carthaca (approved):

PR #4973, opened 2023-05-26
[ironic] switches console nginx to ssl

stefanhipfel:


BerndKue (approved):
lgtm

openstack/ironic/templates/console-configmap.yaml
```diff
@@ -7,15 +7,36 @@ metadata:
     type: config
     component: ironic
 data:
+  dhparam.pem: |-
+{{ .Values.console.ssl_dhparam | nindent 4 | trim  }}
+
```
sandzwerg, line 12: The linter complains about this, but I don't get why:
```
==> Linting openstack/ironic

[ERROR] templates/: template: ironic/templates/console-configmap.yaml:11:41: executing "ironic/templates/console-configmap.yaml" at <4>: invalid value; expected string
```

stefanhipfel: me too. it expects a string, which will be added via templating from vault

stefanhipfel: maybe we need those test values setup:
Linting chart with values file 'openstack/ironic/ci/test-values.yaml'...

PR #4975, opened 2023-05-26
Feature/certificate management 61

corey-aloia:
This PR adds logic to mount the cert and key instead of reading if from the config map. It also changes the way we set ldap so that we don't need to adjust the secret in every env


schardosin (approved):
LGTM

PR #4977, opened 2023-05-27
Update nsxt.alerts

atcyuth:
changing severity to info until we replace the certs with digi certs on PROD NSXT


christopherhans (approved):
lgtm

PR #4980, opened 2023-05-30
[infra-monitoring] Fix metrics path for KVM

fwiesel:
Copy-and-paste error: It is a standard node-exporter, which expects the path /metrics


viennaa (changes requested):
Please bump the chart patch version.

viennaa (approved):
Thanks and good to go

PR #4983, opened 2023-05-30
[maia] aggregate Manila share used bytes with logical used bytes

chuan137:
The metrics should report sizes that are consistent with `df` on the client side. For volumes/shares that have logical space reporting/enforcement enabled, the metrics openstack_manila_share_used_bytes should report the logical used bytes.


Carthaca (approved):
I approve the intention: openstack_manila_share_used_bytes should report different values depending on the netapp_volume_is_space_reporting_logical metric.

But I cannot judge the correctness of applying this to the scrape config

PR #4986, opened 2023-05-30
[infra-monitoring]enables ipmi-exporter by default

stefanhipfel:


viennaa (changes requested):
Please bump the chart accordingly, then we can merge this.

richardtief:
It's already enabled in every region but QA: https://github.wdf.sap.corp/cc/secrets/blob/master/qa-de-1/values/infra-monitoring.yaml#L76. Please enable it regional secrets.

PR #4989, opened 2023-05-30
[openstack/utils] Topology refinement

fwiesel:
- Use default instead of value as the default value from global is not properly merged
- Add service annotations for topology aware routing
- Add the annotations only in regions with more than one az

### DIFF
--- a/openstack/utils/templates/_affinity.tpl
+++ b/openstack/utils/templates/_affinity.tpl
@@ -24,15 +24,16 @@ affinity:
             - {{$component}}
 {{- end }}
 
-{{ define "kubernetes_pod_az_affinity" -}}
-{{- $availability_zone := index . 0 -}}
+{{ define "utils.kubernetes_pod_az_affinity" -}}
+{{- $envAll := index . 0 -}}
+{{- $availability_zone := index . 1 -}}
 affinity:
   nodeAffinity:
     preferredDuringSchedulingIgnoredDuringExecution:
       - weight: 1
         preference:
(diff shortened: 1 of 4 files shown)


fwiesel:
Follow up to #4965

Carthaca:
With removing the `.global.topology_key`value, can you please also rework `kubernetes_pod_az_affinity`? 
I noticed, that the `.Values` are anyhow not working in https://github.com/sapcc/helm-charts/pull/4965/files#diff-a8000139f34581224d5ac4276723eb108ae9879555d815bbe05577f0076d4534R35, the input to the function is only a availability zone name

fwiesel:
Done: I've renamed it to `utils.kubernetes_pod_az_affinity`, and pass it `envAll`. I'll need then to update the charts using it when making use of it.

openstack/utils/templates/_affinity.tpl
```diff
@@ -47,3 +48,4 @@ affinity:
                   values:
                   - operational
 {{- end }}
+s
```
Carthaca, line 51: This `s` comes probably unintended

Carthaca (changes requested):
Thanks, a tiny extra letter sneaked in :)

openstack/utils/templates/_affinity.tpl
fwiesel: Sorry, indeed. I've removed it.

Carthaca (approved):
Thanks!

PR #4995, opened 2023-05-31
[maia] bump version for changes in scrape config

chuan137:


openstack/maia/templates/etc/_maia_scrape_config.yaml.tpl
```diff
@@ -150,9 +150,6 @@
   metric_relabel_configs:
     - regex: "cluster|cluster_type|instance|job|kubernetes_namespace|kubernetes_pod_name|kubernetes_name|pod_template_hash|exported_instance|exported_job|type|name|component|app|system|thanos_cluster|thanos_cluster_type|thanos_region|alert_tier|alert_service"
       action: labeldrop
-    - action: drop
```
viennaa, line 153: why aren't they dropped any more?

viennaa:
for reference, chart bump is also for 
#4983

openstack/maia/templates/etc/_maia_scrape_config.yaml.tpl
chuan137: because the federate rules are changed, only metrics with ":maia" at the end is fetched. I am changing the fetching way from dropping the metrics to explicitly defining them in the exporters. It's much easier to read and maintain. the exporter changes are [here](https://github.com/sapcc/helm-charts/commit/99d9b8fbadce57e4cf8cd75a11d9150c022f0d64) and [here](https://github.com/sapcc/helm-charts/commit/ba43011221aab166e5a00dc39b9ef1c9b0cbfd0a)

PR #5002, opened 2023-06-01
[nova] Topology aware scheduling and spreading + AZ update

fwiesel:
Use topologySpreadConstraints to spread the api pods over the zones in a best effort manner to improve the situation in the situation of a zone is down (kubernetes may not rebalance, or cannot due to inbalance)

Use topology aware routing service-annotation to make use of the distributed placement and prefer communicating to nodes in the same AZ

In regions with a single AZ, the annotations are not set.

### DIFF
--- a/openstack/nova/templates/cell2-conductor-deployment.yaml
+++ b/openstack/nova/templates/cell2-conductor-deployment.yaml
@@ -27,7 +27,8 @@ spec:
         name: nova-conductor
         alert-tier: os
         alert-service: nova
-{{ tuple . "nova" "conductor" | include "helm-toolkit.snippets.kubernetes_metadata_labels" | indent 8 }}
+        {{- tuple . "nova" "conductor" | include "helm-toolkit.snippets.kubernetes_metadata_labels" | nindent 8 }}
+        {{- include "utils.topology.pod_label" . | indent 8 }}
       annotations:
         {{- if or .Values.cell2.conductor.config_file.DEFAULT.statsd_enabled .Values.proxysql.mode }}
         prometheus.io/scrape: "true"
@@ -35,8 +36,9 @@ spec:
         {{- end }}
         configmap-etc-hash: {{ include (print $.Template.BasePath "/etc-configmap.yaml") . | sha256sum }}
     spec:
-{{ tuple . "nova" "conductor" | include "kubernetes_pod_anti_affinity" | indent 6 }}
-{{ include "utils.proxysql.pod_settings" . | indent 6 }}
+      {{- tuple . "nova" "conductor" | include "kubernetes_pod_anti_affinity" | nindent 6 }}
+      {{- include "utils.proxysql.pod_settings" . | nindent 6 }}
+      {{- tuple . (dict "name" "nova-conductor") | include "utils.topology.constraints" | indent 6 }}
       terminationGracePeriodSeconds: {{ .Values.defaults.default.graceful_shutdown_timeout | add 5 }}
       hostname: nova-conductor
       initContainers:
--- a/openstack/nova/templates/console-shellinabox-deployment.yaml
+++ b/openstack/nova/templates/console-shellinabox-deployment.yaml
@@ -24,7 +24,8 @@ spec:
     metadata:
       labels:
         name: nova-console-cell1-shellinabox
-{{ tuple . "nova" "console-cell1-shellinabox" | include "helm-toolkit.snippets.kubernetes_metadata_labels" | indent 8 }}
+        {{- tuple . "nova" "console-cell1-shellinabox" | include "helm-toolkit.snippets.kubernetes_metadata_labels" | nindent 8 }}
+        {{- include "utils.topology.pod_label" . | indent 8 }}
       annotations:
         {{- if .Values.proxysql.mode }}
         prometheus.io/scrape: "true"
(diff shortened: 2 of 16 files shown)


openstack/nova/templates/console-shellinabox-deployment.yaml
```diff
@@ -25,6 +25,7 @@ spec:
       labels:
         name: nova-console-cell1-shellinabox
 {{ tuple . "nova" "console-cell1-shellinabox" | include "helm-toolkit.snippets.kubernetes_metadata_labels" | indent 8 }}
```
joker-at-work, line 27: Optional: Move this to `nindent`, too?

openstack/nova/templates/console-shellinabox-deployment.yaml
```diff
@@ -37,7 +38,8 @@ spec:
         runAsGroup: 33
         fsGroup: 33
 {{ tuple . "nova" "console-cell1-shellinabox" | include "kubernetes_pod_anti_affinity" | indent 6 }}
```
joker-at-work, line 40: Optional: Move this to `nindent`, too?

openstack/nova/templates/scheduler-deployment.yaml
```diff
@@ -35,8 +35,8 @@ spec:
         configmap-etc-hash: {{ include (print $.Template.BasePath "/etc-configmap.yaml") . | sha256sum }}
         configmap-scheduler-etc-hash: {{ include (print $.Template.BasePath "/scheduler-etc-configmap.yaml") . | sha256sum }}
     spec:
-{{ tuple . "nova" "scheduler" | include "kubernetes_pod_anti_affinity" | indent 6 }}
-{{ include "utils.proxysql.pod_settings" . | indent 6 }}
+      {{- tuple . "nova" "scheduler" | include "kubernetes_pod_anti_affinity" | nindent 6 }}
+      {{- include "utils.proxysql.pod_settings" . | nindent 6 }}
```
joker-at-work, line 39: Why does `nova-scheduler` not need `utils.topology.constraints`?

openstack/nova/templates/console-shellinabox-deployment.yaml
```diff
@@ -37,7 +38,8 @@ spec:
...
-{{ include "utils.proxysql.pod_settings" . | indent 6 }}
+      {{- include "utils.proxysql.pod_settings" . | nindent 6 }}
+      {{- tuple . (dict "name" "nova-console-shellinabox") | include "utils.topology.constraints" | indent 6 }}
```
joker-at-work, line 42: I think we need to rename this to `nova-console-cell1-shellinabox`.

openstack/nova/templates/cell2-conductor-deployment.yaml
```diff
@@ -28,15 +28,17 @@ spec:
         alert-tier: os
         alert-service: nova
 {{ tuple . "nova" "conductor" | include "helm-toolkit.snippets.kubernetes_metadata_labels" | indent 8 }}
```
joker-at-work, line 30: Optional: Move to `nindent`?

openstack/nova/templates/cell2-conductor-deployment.yaml
```diff
@@ -28,15 +28,17 @@ spec:
...
         {{- if or .Values.cell2.conductor.config_file.DEFAULT.statsd_enabled .Values.proxysql.mode }}
         prometheus.io/scrape: "true"
         prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
         {{- end }}
         configmap-etc-hash: {{ include (print $.Template.BasePath "/etc-configmap.yaml") . | sha256sum }}
     spec:
-{{ tuple . "nova" "conductor" | include "kubernetes_pod_anti_affinity" | indent 6 }}
-{{ include "utils.proxysql.pod_settings" . | indent 6 }}
+      {{ tuple . "nova" "conductor" | include "kubernetes_pod_anti_affinity" | nindent 6 }}
+      {{ include "utils.proxysql.pod_settings" . | nindent 6 }}
```
joker-at-work, line 40: Shouldn't they both start with `{{-`? Doesn't break anything, I think, just adds an additional newline.

joker-at-work (changes requested):

openstack/nova/templates/console-shellinabox-deployment.yaml
fwiesel: Done

openstack/nova/templates/console-shellinabox-deployment.yaml
fwiesel: Done

openstack/nova/templates/scheduler-deployment.yaml
fwiesel: We only run one instance of it, and if we run more than one instance, we might get into trouble. AFAIK.

openstack/nova/templates/console-shellinabox-deployment.yaml
fwiesel: done

openstack/nova/templates/cell2-conductor-deployment.yaml
fwiesel: Changed it

fwiesel:
The last update pulls in a newer `utils` which uses the non-deprecated topology key.

openstack/nova/templates/scheduler-deployment.yaml
joker-at-work: hm ... correct, we currently do, but we also define `kubernetes_pod_anti_affinity` anyways, so I would have expected differently. anyways, no need to change currently, I guess. thanks.

joker-at-work (approved):

PR #5007, opened 2023-06-01
[manila] Topology aware scheduling and spreading

fwiesel:
Use topologySpreadConstraints to spread the api pods over the zones in a best effort manner to improve the situation in the situation of a zone is down (kubernetes may not rebalance, or cannot due to inbalance)

Use topology aware routing service-annotation to make use of the distributed placement and prefer communicating to nodes in the same AZ

In regions with a single AZ, the annotations are not set.


Carthaca (approved):
Thanks, Fabian!

PR #5010, opened 2023-06-01
reducing alert severity to warning because dhcp and metadata service …

artherd42:
…is not global but per project


sebageek (approved):
After Discussion with @artherd42 I concur with this assessment. Until this specific test is better understandable/actionable for the on-shift people and has a lower false-positive rate it makes sense to keep it as warning. We should remember to revert this though, once the check has been improved.

PR #5023, opened 2023-06-05
maia: disable multiple scrape alerts

richardtief:
The multiple scrape alerts from maia prometheus are evaluated by prometheus-openstack. This led to duplicate alert notifications several times in the past. Since we know what we are doing, these alerts can be disabled for maia.


notque (approved):
You've made several of the changes to make this work. I'm disconnected from what you're doing. I assume this is appropriate from your other changes, so approved.

richardtief:
The changes were made in the common chart and we use this functionality for Maia as well. Therefore, we can also turn it off specifically.

PR #5032, opened 2023-06-06
Fix openstack_neutron_monitor_agents_heartbeat_seconds/load

sven-rosenzweig:
If the last heartbeat timestamp of an agent is older than 90 seconds, the query will not return the agent due to the where condition containing  a ">". Only agents sending heartbeat_timestamp in the intervall [now-90 seconds, now] are shown. Completly removing the intervall conditions (all agent heartbeats will be fetched) allows us to customize the alert time intervall in the prometheus query. Currently this value is set to 75 seconds.


swagner-de (approved):
Looks good. Let's merge it after dangling agents have been cleaned up.

PR #5034, opened 2023-06-06
[neutron] Drop per-service rabbitmq-notifications

fwiesel:
We have now a central one for all services, and neutron is already sending to the new one, so we can drop the old one: It is only wasting resources.


fwiesel:
You can cross check it with the grafana dashboard. There are no messages queued.

sebageek (approved):

PR #5046, opened 2023-06-08
[logstash-audit] Remove host specific fields from event_data

max-len:
To enable "job_events" log aggregator on AWX we need to control the number of fields generated to avoid overloading the receiving ELK stack. Fields observed to contain host or user specific data therefore need to be excluded.

### DIFF
--- a/system/audit-logs/vendor/logstash-audit-external/templates/_logstash.conf.tpl
+++ b/system/audit-logs/vendor/logstash-audit-external/templates/_logstash.conf.tpl
@@ -181,10 +181,13 @@ filter {
         {{ end -}}
       }
 
-      if "awx" in [cluster_host_id] {
+      if "awx" in [logger_name] {
         mutate {
           add_field => { "[sap][cc][audit][source]"  => "awx" }
-          remove_field => [ "event_data" ]
+          remove_field => [ "[event_data][artifact_data]", "[event_data][changed]", "[event_data][dark]",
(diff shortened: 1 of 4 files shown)


system/audit-logs/vendor/logstash-audit-external/templates/_logstash.conf.tpl
```diff
@@ -181,10 +181,19 @@ filter {
...
+          remove_field => [ "[event_data][artifact_data]" ]
+          remove_field => [ "[event_data][changed]" ]
+          remove_field => [ "[event_data][dark]" ]
+          remove_field => [ "[event_data][failures]" ]
+          remove_field => [ "[event_data][ignored]" ]
+          remove_field => [ "[event_data][ok]" ]
+          remove_field => [ "[event_data][processed]" ]
+          remove_field => [ "[event_data][res]" ]
+          remove_field => [ "[event_data][rescued]" ]
+          remove_field => [ "[event_data][skipped]" ]
```
IvoGoman, line 196: _remove_field_ takes an array as argument. So it would be possible to call it once for all fields.

IvoGoman:
Chart version needs to be bumped in https://github.com/sapcc/helm-charts/blob/master/system/audit-logs/Chart.yaml
and https://github.com/sapcc/helm-charts/blob/master/system/audit-logs/vendor/logstash-audit-external/Chart.yaml

IvoGoman (approved):

PR #5050, opened 2023-06-09
Support web consoles for cell2

joker-at-work:

### DIFF
--- a/openstack/nova/shellinabox/10_default_server.conf
+++ b/openstack/nova/shellinabox/10_default_server.conf
@@ -31,17 +31,7 @@ server {
   add_header "X-Frame-Options"  "";     # Undo what is set in nginx.conf
   proxy_hide_header "X-Frame-Options";  # Remove it from the upstream server
 
-  location = / {
-    if ($http_user_agent ~ "^http-keepalive-monitor/") {
-      return 301 /status;
-    }
-    rewrite_by_lua_block {
-      local revproxy = require "revproxy"
-      return revproxy.root()
-    }
-  }
-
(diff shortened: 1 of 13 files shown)


fwiesel:
I think, we need eventually revisit how the cells are structured in the helm-chart. But we best leave it for another time...
The reason why I bring it up, I find [is_cell2](https://github.com/sapcc/helm-charts/pull/5050/files#diff-755fc98bf07913488d3b5f50ef3fb2234f40f897668e3dc3a0dfa1c192bf364aR56-R59) or the [various conditions](https://github.com/sapcc/helm-charts/pull/5050/files#diff-b637ac6d4a1fcbd78f1fe419dc52a4b9586a1c5e009ce76f05facea4f3582547R134-R137) in the vcenter-template not particular appealing. It is showing that the logic behind the cells is not terribly constrained or contained, which results in code needing catering to the various options taken.

The only thing I possibly would change (if possible) is that the path in the urls is of the form `/<cell-name>-<console-type>`.
Personally, I would prefer either `/<cell-name>/<console-type>`, because they form logically a hierarchy. I assume you wanted to avoid it, because you cannot actually access `/<cell-name>` by itself. I would argue though, that going "up" is not a feature of the http protocol.

I think, it depends on what you want to put the emphasis on: That the consoles belong to a cell (hierarchy), or that you cannot browse "up". More a stylistic thing, so if you disagree we keep it the way you have it.

joker-at-work:
I agree with "we need to re-visit" cells, but shouldn't do that right now.

I don't mind having a hierarchy for the cells' consoles. Let me play that through.

fwiesel (approved):

fwiesel (approved):

PR #5051, opened 2023-06-09
prometheus-kubernetes|kubernikus-global: alert routing updated

richardtief:
Updated the support group and service to forward alerts. None of the alerts are critical, so I reduced them to warning.


viennaa (approved):
Thanks for taking care.

PR #5058, opened 2023-06-13
nova/flavors: Rename & alias x1.* to m_* names

grandchild:
- x1.memory -> m_c8_m128
 - x1.2xmemory -> m_c16_m256
 - x1.4xmemory -> m_c32_m512


dhoeller (approved):
LGTM

PR #5064, opened 2023-06-14
[octavia] Fix alert to include agent name

swagner-de:
Due to the `max()` function we aggregate all metrics and only show if an
agent's heartbeat is older than 120 secs, not which one. I guess this is
not intended as even the alert description uses the `octavia_host`
label.


BenjaminLudwigSAP (approved):
lgtm, will observe as I roll out

PR #5069, opened 2023-06-15
prometheus-kubernetes: init

richardtief:


system/prometheus-kubernetes/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+description: Common Kubernetes Prometheus
+name: prometheus-kubernetes
+version: 0.0.1
+home: https://github.com/sapcc/helm-charts/tree/master/system/prometheus-kubernetes
+dependencies:
+  - name: prometheus-server
+    repository: https://charts.eu-de-2.cloud.sap
+    version: 7.2.7
```
viennaa, line 9:
```suggestion
    version: 7.2.8
```

+ new Chart.lock pls.

viennaa (changes requested):
Thanks for providing this shell.

We should go with latest prometheus-server here for the cluster-type seed adjustment.

viennaa (approved):
Alright, good to go for a simple test without deps.

PR #5070, opened 2023-06-15
[keystone] add projects metric

Carthaca:
provide a simple way to get project info in prom.
E.g. I have a project_id and I want to get the project name and domain name.
Before we abused other metrics like 'limes_project_quota' with a higher performance impact since it has more than one datapoint per project.

### DIFF
--- a/openstack/keystone/values.yaml
+++ b/openstack/keystone/values.yaml
@@ -349,6 +349,31 @@ mysql_metrics:
         WHERE domain_id != '<<keystone.domain.root>>'
       values:
         - "projects"
+    - name: openstack_projects_count_gauge
+      help: Keystone projects
+      query: |
+        SELECT
+          project.id as project_id,
+          project.name as project_name,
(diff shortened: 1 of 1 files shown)


bbobrov (approved):

rajivmucheli:
maybe we remove openstack_projects_total metric ? better info can be achieved from openstack_projects_count_gauge metric ?

Carthaca:
> maybe we remove openstack_projects_total metric ? better info can be achieved from openstack_projects_count_gauge metric ?

This could be done in a follow-up task I think, currently it is in use in the Keystone Grafana Dashboard - this would have to be changed to use the new metric before removing the old one.

Can you take care of merging and rollout, please?

rajivmucheli:
sure, i had a similar idea if we decided to remove openstack_projects_total metric.

Is there a timelines to rollout openstack_projects_count_gauge metric?

PR #5103, opened 2023-06-20
[nova] Automatically confirm resize after three days

fwiesel:
Users often forget to confirm a resize after the vm has been successfully resized.

Automatically confirming brings the VMs back to a standard state and clean up the copy of the root disk.

The time-frame was chosen to be a weekend plus one day to be hopefully as short as possible without messing up anyany manual process going on, which might get delayed over a weekend.


joker-at-work (changes requested):
Blocking this merge, because we need a blog-entry to inform our customers about this before rolling out. Please create one.

typo: double `any` in the commit-message

fwiesel:
Updated the commit message

joker-at-work (approved):

PR #5104, opened 2023-06-20
[common][mariadb] use non-deprecated topology labels

rajivmucheli:


defo89:
Isn't it the same as https://github.com/sapcc/helm-charts/pull/5087/files ?

rajivmucheli:
couldnt find this PR.

PR #5106, opened 2023-06-21
[cronus] add chart for maillog

sapmli:


openstack/cronus/templates/maillog/deployment.yaml
```diff
@@ -0,0 +1,76 @@
+{{- if .Values.maillog.enabled }}
+{{- if .Values.maillog.deployment.enabled }}
```
corey-aloia, line 2: Do you need a deployment and a sts?

openstack/cronus/values.yaml
```diff
@@ -529,6 +529,85 @@ rhea_rabbitmq:
...
+  godebug: netdns=cgo
+  service:
+    type: LoadBalancer
+    annotations:
+      disco: "true"
+      disco/record: my-ingress.region.cloud.com.
+  deployment:
+    enabled: true
+    replicas: 1
+  statefulSet:
```
corey-aloia, line 592: Can we removed the stateful set if it is not used? Might make things a bit simpler

openstack/cronus/values.yaml
```diff
@@ -529,6 +529,85 @@ rhea_rabbitmq:
...
+      labels: {}
+      servicePort: 80
+  updateStrategy: RollingUpdate
+  terminationGracePeriod: 60
+  # all below can be used only with a deployment style
+  maxSurge: 1
+  maxUnavailable: 0
+  ingressAnnotations:
+    # don't handle this resource by ingress controllers
+    kubernetes.io/ingress.class: rhea
```
corey-aloia, line 609: This looks wrong, and something I dont think we need to add. Its been a while since I worked with nginx, but I believe this is referring to which ingress to use, which should be nginx AFAIK

openstack/cronus/values.yaml
```diff
@@ -529,6 +529,85 @@ rhea_rabbitmq:
...
+    type: "elasticsearch"
+    uri: "http://localhost:9200"
+    indexMgmt: false
+    indexPrefix: "mailstate"
+    indexDateFormat: "2006_02_01_15_04_05"
+    cutOffTime: 30d
+    granularity: 1d
+    lookBacks: 5
+  tls:
+    serverTlsName: tls-secret-maillog
```
corey-aloia, line 547: The server tls certs should be handled automagically by disco I believe. So your application shouldn't need to worry about server certs to outside world AFAIK. I.E. the following ingress annotations should take care of updating dns and using server certs in the ingress:
```
    ingress.kubernetes.io/auth-tls-secret: kube-system/ingress-cacrt
    ingress.kubernetes.io/auth-tls-verify-client: "optional"
    ingress.kubernetes.io/auth-tls-verify-depth: "3"
    kubernetes.io/ingress.class: "nginx"
    disco: "true"
    ```
    This is an example and my not be 100% correct.

File: openstack/cronus/ci/test-values.yaml [hunk 5]
```diff
@@ -8,3 +8,9 @@ rhea:
     name: rhea-repository
     pullPolicy: IfNotPresent
     tag: latest
+
+maillog:
+  image:
+    name: maillog-repository
+    pullPolicy: IfNotPresent
+    tag: latest
```
corey-aloia, line 16: maillog is not enabled here meaning that you don't get the CI pipeline error checking as your maillog resources won't be included

openstack/cronus/templates/maillog/deployment.yaml
```diff
@@ -0,0 +1,76 @@
...
+  strategy:
+{{- if eq "RollingUpdate" .Values.maillog.updateStrategy }}
+    rollingUpdate:
+      maxSurge: {{ .Values.maillog.maxSurge }}
+      maxUnavailable: {{ .Values.maillog.maxUnavailable }}
+{{- end }}
+    type: {{ .Values.maillog.updateStrategy }}
+  selector:
+    matchLabels:
+      name: maillog
```
corey-aloia, line 26: maybe best that the match labels line up through a variable, this way there is not chance for typos. I.E. using `{{.Chart.Name}}` to match. And just curios: why have the label `instance: maillog`?

openstack/cronus/templates/maillog/deployment.yaml
```diff
@@ -0,0 +1,76 @@
...
+        name: maillog
+        instance: maillog
+    spec:
+      serviceAccountName: maillog
+{{- if .Values.maillog.affinity }}
+      affinity: {{ toYaml .Values.maillog.affinity | nindent 8 }}
+{{- end }}
+      containers:
+        - name: maillog
+          image: "{{ .Values.maillog.image.name }}:{{ .Values.maillog.image.tag }}"
```
corey-aloia, line 42: You should probably have the `.Values.global.registry` here. So:
`"{{ .Values.global.registry }}/{{ .Values.image.name }}:{{ .Values.image.tag }}"`

openstack/cronus/templates/maillog/role.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if .Values.maillog.enabled -}}
+apiVersion: rbac.authorization.k8s.io/v1
```
corey-aloia, line 2: Are you sure this is needed? I think by default it should be able to get all secrets in its own namespace

openstack/cronus/templates/maillog/secrets.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: maillog
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: maillog
+    chart: maillog
+    release: maillog
+    heritage: {{ .Release.Service }}
+type: Opaque
+data:
+  config.yaml: {{ include "cronus/templates/maillog/_config.yaml.tpl" . | b64enc }}
```
corey-aloia, line 14: Ive always this confusing in the other charts.... that we take the config.yaml.tpl, then stich it in the config.yaml of a secret. Why not just use the secret right away? Or stick with the configmap? I.E. why a config that goes to a secret then back to a config.... you could just take the config map and mount it

openstack/cronus/values.yaml
```diff
@@ -529,6 +529,85 @@ rhea_rabbitmq:
...
+      4SuP99G7DuItG008T8AYFUHAM2h/yskVyvoZ8+gZx54TC9aY9gPIKyX++4bHv5BC
+      qbEdU46N05R+AIBW2KvWozQkjhSQCbzcp6DHXLoZINI6y0WOImzXrvLUSIm4CBaj
+      6MTXInIkmitdURnmpxTxLva5Kbng/u20u5ylIQKqpcD8HWX97lLVbmbnPkbpKxo+
+      LvHPhNDM3rMsLu06agF4JTbO8ANYtWQTx0PVrZKJu+8fcIaUp7MVBIVZ
+      -----END CERTIFICATE-----
+  debug: true
+  godebug: netdns=cgo
+  service:
+    type: LoadBalancer
+    annotations:
```
corey-aloia, line 586: Can these annotations be directly on the service instead of in the values? then at the time they differ between envs, we can start adding them to the values. But I think an annotation like `disco: "true"` will never change, and if it does, then it would change for all envs

openstack/cronus/templates/maillog/deployment.yaml
sapmli: STS removed

openstack/cronus/values.yaml
sapmli: removed

openstack/cronus/values.yaml
sapmli: I fixed it at least to be "maillog".
That still means that - according to the cronus version - there will probably no nginx in front of the service, but I might be wrong.

openstack/cronus/templates/maillog/deployment.yaml
sapmli: No, the registry is usually part of the image spec in the secrets.

corey-aloia: It should be a variable though... for example, in the past the registry url has changed. So by using the global.registry, it is aligning with where the build step in the pipeline pushes to, which can and has changed

openstack/cronus/templates/maillog/secrets.yaml
corey-aloia: Will resolve this conversation at it is not that important really. Ideally IMHO config files would come from configmaps, but it really does not matter that much

openstack/cronus/values.yaml
corey-aloia: The nginx would be used by the ingress, so if there is external traffic from outside the cluster, it would go through the nginx and then to the service.

I was able to verify that the ingress class cronus has is invalid, and is effecttively "none":
image

Looking deeper, we only have 1 ingress class of `nginx` and this is the default. So effectively, cronus having this invalid one, it ends up using `nginx`

corey-aloia: THinking more.... we might want to test this out. That <none> then uses the default, as potentially it should show the default. This definitely feels hacky though as we set it to a class that doesn't exist. No idea why. Here are the available classes:

corey-aloia: I made a test and the default only kicks in if you leave the annotation off. So in this class, we are setting it to an invalid name in cronus in order to no use nginx. I am confused. But I found some chat in slack suggesting this was done because cronus only needs to cert creation. Still... I find this really confusing. Most likely you don't want this for maillog.

image

PR #5120, opened 2023-06-23
[bind] allow more SOA queries, but limit notifies

mutax:
We hit the limit of parallel SOA queries to mdns -- the default in bind is 20.
These settings allow more queries in parallel, allowing for faster zone syncs.
Details on the values can be found in https://kb.isc.org/docs/aa-01313


dhoeller (approved):
LGTM

galkindmitrii (approved):

PR #5131, opened 2023-06-27
[memcached] bump imageTag versions of library/memcached and prom/memcached-exporter

rajivmucheli:
The updated versions are clean as per Trivy


viktor-kovacs (approved):
Lgtm

PR #5134, opened 2023-06-28
[Glance] remove configuration-snippet: 

rajivmucheli:
disabled due to CVE-2021-25742

### DIFF
--- a/openstack/glance/templates/ingress.yaml
+++ b/openstack/glance/templates/ingress.yaml
@@ -9,7 +9,6 @@ metadata:
     component: glance
   annotations:
     ingress.kubernetes.io/proxy-request-buffering: "off"
-    ingress.kubernetes.io/rewrite-target: "/$1"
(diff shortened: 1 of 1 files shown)


fwiesel:
I think, you can simply delete the snippet. 
My understanding is, the equivalent needs to be added [here](https://github.com/sapcc/helm-charts/blob/3dd8327218c91ec747adf4dd9671d648d079e9f6/system/kube-system-global/values.yaml)  somewhere...
But I can't find it

fwiesel:
It is rolled somehow out as a location-snippet in `ConfigMap/kube-system-metal-ingress-nginx-controller` ... I can't find the source right now though

auhlig:
Yes, the request id is done centrally via ingress configmap and the snippet annotation can be removed. The source is in kube-secrets.git.

fwiesel (approved):

PR #5136, opened 2023-06-28
2faproxy: Add OpenstackSeed for service user creation

databus23:
So far I used application credentials tied to my user which is not suitable for production.

### DIFF
--- a/openstack/keystone/templates/seed-2faproxy.yaml
+++ b/openstack/keystone/templates/seed-2faproxy.yaml
@@ -1,0 +1,26 @@
+{{- if and (index .Values "2fa" "enabled") (index .Values "2fa" "seed_user") }}
+apiVersion: "openstack.stable.sap.cc/v1"
+kind: OpenstackSeed
+metadata:
+  name: {{ .Release.Name }}-2faproxy-seed
+
+spec:
+  requires:
+    - {{ .Values.global.keystoneNamespace }}/keystone-seed
+    - monsoon3/domain-default-seed
+    - monsoon3/domain-ccadmin-seed
+
+  domains:
+    - name: {{ index .Values "2fa" "openstack" "OS_DOMAIN_NAME" | required ".Values.2faproxy.openstack.DOMAIN_NAME is missing" }}
+      users:
+        - name: {{ index .Values "2fa" "openstack" "OS_USERNAME" | required ".Values.2faproxy.openstack.OS_USERNAME is missing" }}
+          description: Keystone 2faproxy service user
+          password: '{{ index .Values "2fa" "openstack" "OS_PASSWORD" | required ".Values.2faproxy.openstack.OS_PASSWORD is missing" }}'
+
+    - name: ccadmin
+      projects:
+        - name: cloud_admin
+          role_assignments:
+            - user: {{ index .Values "2fa" "openstack" "OS_USERNAME"}}@{{ index .Values "2fa" "openstack" "OS_DOMAIN_NAME" }}
+              role: cloud_identity_viewer
+{{- end }}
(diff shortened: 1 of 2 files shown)


openstack/keystone/templates/seed-2faproxy.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: {{ .Release.Name }}-2faproxy-seed
+
+spec:
+  requires:
+    - {{ .Values.global.keystoneNamespace }}/keystone-seed
+    - monsoon3/domain-default-seed
+    - monsoon3/domain-ccadmin-seed
+
+  domains:
+    - name: {{ index .Values "2fa" "openstack" "OS_DOMAIN_NAME" | required ".Values.2faproxy.openstack.DOMAIN_NAME is missing" }}
```
bbobrov, line 14: it seems that the error message is not leading us to the right path. Shouldn't it be `.Values.2fa....`?

bbobrov, line 14: where does this value come from? I don't see anything in the values.yaml about it. "Default" seems like a reasonable default value.
Also, why do we even have it configurable? This user looks very like other service users, like `nova` or `neutron` (or other users in domain Default). Should not they simply always live in Default?

bbobrov (changes requested):

openstack/keystone/templates/seed-2faproxy.yaml
databus23: Thanks for spotting this!

databus23: Good point. The idea is, that the OS_ parameters are provided as values via (cc secrets).
The reason I did not specify `.2fa.openstack.OS_DOMAIN_NAME` as a default in the charts yaml is that I'm currently using application credentials in qa-de-1 in that case a default environment variable might interfere with the OS_APPLICATION_CREDENTIAL_* variables I specified in the qa-de-1 secrets.
So I opted for this backward compatible chart change.
That way I can test the rollout with seeding in qa-de-2 where the 2faproxy is currently disabled before I change the qa-de-1 deployment.
As I needed to manually specific the OS_DOMAIN_NAME variable in the secrets anyways I opted for using this variable in all places instead of manually specifying "Default" in the various places in the seed resource.
That being said I don't plan to use a different value then "Default".

I hope my reasoning makes some sense.

databus23: @bbobrov Any thought. Can we merge this as is or do you want it changed?

openstack/keystone/templates/seed-2faproxy.yaml
```diff
@@ -0,0 +1,26 @@
...
+spec:
+  requires:
+    - {{ .Values.global.keystoneNamespace }}/keystone-seed
+    - monsoon3/domain-default-seed
+    - monsoon3/domain-ccadmin-seed
+
+  domains:
+    - name: {{ index .Values "2fa" "openstack" "OS_DOMAIN_NAME" | required ".Values.2fa.openstack.OS_DOMAIN_NAME is missing" }}
+      users:
+        - name: {{ index .Values "2fa" "openstack" "OS_USERNAME" | required ".Values.2faproxy.openstack.OS_USERNAME is missing" }}
```
bbobrov, line 16: the error message is still wrong here

openstack/keystone/templates/seed-2faproxy.yaml
```diff
@@ -0,0 +1,26 @@
...
+          description: Keystone 2faproxy service user
+          password: '{{ index .Values "2fa" "openstack" "OS_PASSWORD" | required ".Values.2faproxy.openstack.OS_PASSWORD is missing" }}'
```
bbobrov, line 18: and here

bbobrov (changes requested):

openstack/keystone/templates/seed-2faproxy.yaml
databus23: oh man. I am really sloppy. thanks again for spotting this. 👍

PR #5144, opened 2023-07-03
[alertmanager] increase silence retention time

uwe-mayer:


uwe-mayer:
> Before that goes in, please bump the chart.

Done

bassel-z (approved):
LGTM

PR #5148, opened 2023-07-03
[PX] fix syntax errors in chart

swagner-de:


richardtief (approved):
Thanks for providing a fix.

PR #5149, opened 2023-07-03
[vrops-exporter] Extract tag from summary|tagJson

max-len:
This was tested for one existent tag with the label_replace function and therefore needs further testing as relabel config, with multiple tags of different categories etc.
It might be possible to generalize it even more to create new labels for all objects in the tag json, although it could be difficult for multiple tags of the same category.

### DIFF
--- a/system/vmware-monitoring/Chart.yaml
+++ b/system/vmware-monitoring/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vmware-monitoring
-version: 1.3.1
+version: 1.3.2
 description: VMware Monitoring and Metrics Collection
 dependencies:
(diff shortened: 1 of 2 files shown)


kevin-fischer (approved):
+1

richardtief (commented):
Looks good. But I'm not sure if you want to leave out the original label then.

https://github.com/sapcc/helm-charts/blob/3ac69c601b7b476eb852ebd47517ab638df794a3/system/vmware-monitoring/templates/vrops-exporter-deployment.yaml#L124-L125

richardtief:
@max-len this pr is still open.

richardtief (approved):
Thanks, all good

PR #5151, opened 2023-07-04
[cinder] Move CapabilitiesFilter second for volume type selection

fwiesel:
The CapabilitiesFilter narrows down the host list to the ones for the volume type. When using only vmware volume types, the order should not make a difference.
But when we have different hypervisors, filtering out host with the wrong capabilities (i.e. wrong storage protocol / vendor / etc...) allows us to identify the volume type indirectly by looking at the host list in a filter, without having to implement excactly the CapabilitiesFilter logic in the filter themselves to derive the targetted hosts from the capabilities.


hemna:
I would prefer to have the SAPPoolDownFilter before this one as it filters out any pools for consideration that the volume driver or volume manager should never see.

fwiesel:
Done, I've moved it after `AvailabilityZoneFilter` and `SAPPoolDownFilter`

hemna (approved):

PR #5153, opened 2023-07-04
[alerts]prometheus-kubernetes-rules version bumped

businessbean:
- to activate the pod resources alerts from
- https://github.com/sapcc/helm-charts/pull/5130


defo89 (changes requested):
please run `helm dep up` to update the lock files with new chart versions

defo89 (approved):

PR #5170, opened 2023-07-10
openstack/cerebro: update owner-info to v0.2.0

majewsky:
Using older owner-info versions generates a Gatekeeper violation because of the missing support group reference. Since this release contains only seeds, we should be fine falling back to @viennaa's support group association.


viennaa (approved):
Thanks for bumping.

PR #5172, opened 2023-07-11
[alerts] prometheus-kubernetes-rules version bumped

businessbean:
- to use the version that contains the rule syntax fixes


IvoGoman (approved):
👍

PR #5178, opened 2023-07-12
keystone: controls for delayed assignment cleanup

bbobrov:
Immediately deleting role assignments is sometimes not desirable. People might forget to renew their ldap group membership, be on vacation or have other temporary short-term leave. Another case is a temporary unavalability of CAM resulting in users not being visible in keystone.

If the role assignment is gone, projects are left unattended, and it is not desirable.

### DIFF
--- a/openstack/keystone/Chart.yaml
+++ b/openstack/keystone/Chart.yaml
@@ -34,6 +34,10 @@ dependencies:
     name: redis
     repository: https://charts.eu-de-2.cloud.sap
     version: 1.2.23
+  - alias: keystone-cron-redis
+    name: redis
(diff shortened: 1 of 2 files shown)


rajivmucheli:
is the intent to run this over another redis pod ? why not extend the cron job schedule ?

lint failed due to :

[ERROR] templates/: parse error at (keystone/templates/etc/_keystone.conf.tpl:239): function "keystone_cron_redis" not defined

rajivmucheli:
FYI : Keystone chart is bump since a new feature is introduced.

rajivmucheli:
I presume we dont need this PR anymore ?

PR #5181, opened 2023-07-13
[rabbitmq] improve RabbitMQNotReady alert

Carthaca:
The critical alert should not fire if any rabbitmq is not ready, it should fire if no rabbitmq pods are ready.
Which is is difference, because there may be a NotReady pod still lingering around next to a working Ready pod.
We observed this with a pod stuck in Terminating due a node not being safely drained.

### DIFF
--- a/common/rabbitmq/Chart.yaml
+++ b/common/rabbitmq/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rabbitmq
-version: 0.6.0
+version: 0.6.1
 description: A Helm chart for RabbitMQ
 sources:
 - https://github.com/sapcc/helm-charts/common/rabbitmq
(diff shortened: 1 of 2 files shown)


Carthaca:
Not sure, if the same makes sense for the Cluster setup? 🤔

notque:
All of the pods for a statefulset need to be up, or it should alert. At this says notifications on the alert, most of them are using the statefulset. 

I would prefer that for the statefulset and notifications that it alerts if there are any pods down over a period of time, if a pod is struck in terminating, it should be fixed.

Carthaca:
Thanks for having a look.

I added an if-else condition, so that the alert stays like it is if the chart is deployed as statefulset.

bashar-alkhateeb:
i think we don't have any other alert which will alert for unready pods in a stuck status ?
i think we need to add another alert for any un ready pod but with a warning severity (but can be done in different PR Issue created for it [CCC-34467](https://jira.tools.sap/browse/CCC-34467) )
nit: typos in commit message `"Which is is difference" -> "which it is"`

Carthaca:
I agree, a warning for any unready pod can be added in a separate PR.

And I corrected the nit ;)

bashar-alkhateeb (approved):
and i approved :)

PR #5182, opened 2023-07-13
[mariadb] improve MariaDBNotReady alert

Carthaca:
The critical alert should not fire if any mariadb is not ready, it should fire if no mariadb pods are ready.
Which is is difference, because there may be a NotReady pod still lingering around next to a working Ready pod.

We observed this on rabbitq with a pod stuck in Terminating due a node not being safely drained.
see similar commit
fc6402dfca1d98fdab0872d4269ee51f36f29f10

### DIFF
--- a/common/mariadb/Chart.yaml
+++ b/common/mariadb/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: mariadb
-version: 0.7.11
+version: 0.7.12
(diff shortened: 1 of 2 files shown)


Carthaca:
related to https://github.com/sapcc/helm-charts/pull/5181

businessbean (approved):
Maurice and i have tested and clarified query. It works for all services that should only have one MariaDB pod ready at a time. That is currently the case for all services.

PR #5185, opened 2023-07-14
[cinder] update dependency on rabbitmq to version 0.5.2

bashar-alkhateeb:
update the dependencies of chart cinder to use rabbitmq version 0.5.2:
          adding vendor's feature enhancements and bug fixes: [ChangeLog](https://www.rabbitmq.com/changelog.html)


fwiesel:
It would be good if the PR also had the version of rabbitmq in it.
It looks like we are upgrading to 3.11.15. Given the changelog, I wonder why we do not upgrade either to 3.12.1 or at least to 3.11.17. I am asking, because it will be a disruptive change, and at least I would like to make that as seldom as possible.

bashar-alkhateeb:
> It would be good if the PR also had the version of rabbitmq in it.
> It looks like we are upgrading to 3.11.15. Given the changelog, I wonder why we do not upgrade either to 3.12.1 or at least to 3.11.17. I am asking, because it will be a disruptive change, and at least I would like to make that as seldom as possible.

Done!

fwiesel (commented):
Fine by me, but the final say is with @hemna

hemna (approved):

PR #4620, opened 2023-02-10
[nova] update dependency of rabbitmq to 0.5.1

bashar-alkhateeb:
included releases:
rabbitmq 0.5.0 which adds support for prometheus-rabbitmq plugin
rabbitmq 0.5.1 which use short hostname instead of FQDN for rabbitmq

PR #5192, opened 2023-07-14
[nova] update dependency on rabbitmq to version 0.5.2

bashar-alkhateeb:
adding vendor's feature enhancements and bug fixes: rabbitmq-3.11.15-management
[ChangeLog](https://www.rabbitmq.com/changelog.html)

### DIFF
--- a/openstack/nova/Chart.lock
+++ b/openstack/nova/Chart.lock
@@ -10,7 +10,7 @@ dependencies:
   version: 0.2.7
 - name: rabbitmq
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.5.1
+  version: 0.5.2
 - name: memcached
   repository: https://charts.eu-de-2.cloud.sap
   version: 0.1.1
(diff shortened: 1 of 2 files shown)


fwiesel:
See: https://github.com/sapcc/helm-charts/pull/5185#issuecomment-1635437392

fwiesel (commented):
Fine by me. What's your take @grandchild ?

grandchild:
Still confused.

The defaults from the rabbitmq chart are:

https://github.com/sapcc/helm-charts/blob/master/common/rabbitmq/values.yaml#L71-L74

```yaml
metrics:
  enabled: &metrics true
  sidecar:
    enabled: *metrics
```

and in the nova chart we have:

https://github.com/sapcc/helm-charts/blob/11aaecdef5cf68dc14be22742711d3ed3c0aaa8b/openstack/nova/values.yaml#L802-L811

```yaml
rabbitmq:
  [...]
  metrics:
    password: null
```

which sounds to me like we will have the explicit sidecar and not the plugin.

but in the meeting i understood we don't want the sidecar anymore? or do we, @fwiesel?

fwiesel:
No, since the metrics come from rabbitmq plugin, the sidecar is just dead weight.

bashar-alkhateeb:
first you need to merge this PR :)
https://github.com/sapcc/helm-charts/pull/4620/files

bashar-alkhateeb:
we didnt change the default config of rabbitMQ chart to be sure that all services moved ''in a planned way'' to the new config and after then we will change the default config

grandchild:
so we should merge #4620, then roll out nova, and then merge this, right?

grandchild:
or can we do it in one deployment?

bashar-alkhateeb:
yes you can

grandchild (approved):

grandchild:
oops

PR #5196, opened 2023-07-14
[octavia] Drop unused notification rabbitmq

fwiesel:
All notification messages are send to the hermes rabbitmq, and the consumer has been removed from the notification rabbitmq from octavia.


BenjaminLudwigSAP (approved):
lgtm

PR #5197, opened 2023-07-14
Cinder: Allow inexact volume/backup search by desc

hemna:
This patch updates the Cinder resource filters to allow users to do an inexact search on volume and backup for the description.

### DIFF
--- a/openstack/cinder/templates/etc/_resource_filters.json.tpl
+++ b/openstack/cinder/templates/etc/_resource_filters.json.tpl
@@ -1,8 +1,8 @@
 {
     "volume": ["name", "name~", "status", "metadata",
                "bootable", "migration_status", "availability_zone",
-               "group_id", "user_id"],
-    "backup": ["name", "status", "volume_id"],
+               "group_id", "user_id", "description~"],
+    "backup": ["name", "status", "volume_id", "description~"],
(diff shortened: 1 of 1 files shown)


hemna:
A customer request from 

https://convergedcloud.slack.com/archives/C01SYP4CCTD/p1688377486503939

joker-at-work (commented):
Do you know how costly that is if customers start using it more often? Sounds like a LIKE filter on a possibly long string.

hemna:
Even in our biggest db, the explain plan on it is still listed as simple with a possible rows of 66k.   

```
MariaDB root@127.0.0.1:cinder> explain select * from volumes where deleted=0 and display_description like '%foo%';
+----+-------------+---------+------+---------------+--------+---------+--------+-------+-------------+
| id | select_type | table   | type | possible_keys | key    | key_len | ref    | rows  | Extra       |
+----+-------------+---------+------+---------------+--------+---------+--------+-------+-------------+
| 1  | SIMPLE      | volumes | ALL  | <null>        | <null> | <null>  | <null> | 66669 | Using where |
+----+-------------+---------+------+---------------+--------+---------+--------+-------+-------------+
```

joker-at-work (approved):
Yeah, "simply" scanning 66.000 strings for a substring might still take a long time. Let's see how it behaves and if we get problems.

PR #5199, opened 2023-07-18
Update ALM alerts

artherd42:


common/prometheus-alertmanager-base/templates/alerts/_alertmanager.alerts.tpl
```diff
@@ -29,19 +29,33 @@ groups:
...
+      service: alertmanager
+      severity: warning
+      playbook: 'docs/support/playbook/prometheus/alertmanager_cluster_failures'
+      meta: 'Alertmanager {{`{{ $labels.alertmanager }}`}} is queing cluster messages.'
       support_group: {{ include "alerts.support_group" . }}
     annotations:
-      description: 'Alertmanager {{`{{ $labels.alertmanager }}`}} fails to synchronize with other Alertmanagers of the HA cluster. This can cause duplicate notifications.'
-      summary: Alertmanager in HA cluster fail to synchronize.
+      description: 'Alertmanager {{`{{ $labels.alertmanager }}`}} is queing cluster messages. This can cause dropping messages because too many are queued.'
+      summary: Alertmanager in HA cluster fails to synchronize
```
richardtief, line 58: Should we really add another warning for messages in the queue? Or is it rather an info warning for the #cc-alertmanager channel?

richardtief (commented):
Thanks for taking this in, because it clutter our warning channel. One small thing to discuss.

common/prometheus-alertmanager-base/templates/alerts/_alertmanager.alerts.tpl
artherd42: yes, good idea!

richardtief (approved):

PR #5207, opened 2023-07-20
remove old -6 from index names

notque:
It used to stand for version 6. We are on version 7 now, and it's an unneccesary relic.


notque:
Thinking about it however, we may not be able to merge this right away without balooning the number of indexes which is already an issue. This would be ideal to merge Monday the 31st of July. I will set a calendar entry for myself.

PR #5221, opened 2023-07-28
plutono: fix postgres host

richardtief:
After removing the `nameOvveride` for the postgres deployment, the plutono deployment could no longer find the postgres service. Changing this to the actual name even causes errors again, as Kubernetes truncates the name for longer release names such as `plutono-author-postgresq` without the `l` at the end. The `nameOverride` was there for a reason.


richardtief:
Solved with https://github.com/sapcc/helm-charts/pull/5225

PR #5222, opened 2023-07-28
plutono: scrape internal metrics

richardtief:
1. Deploys a PodMonitor to scrape metrics from plutono. 
2. Removes metrics port on `gitsync` container and renames port 3000 to metrics on `plutono` container.


system/plutono/values.yaml
```diff
@@ -111,6 +111,16 @@ authentication:
   # ssoKey:
   # caCert:
 
+prometheus-monitors:
+  podMonitor:
+    enabled: true
+  prometheus: "openstack"
```
richardtief, line 117: We can discuss if prometheus-openstack is the right on to ingest the metrics.

thgrs (approved):
lgtm

PR #5224, opened 2023-07-28
[Octavia] Add ESD for stateless UDP

BenjaminLudwigSAP:
That is, UDP traffic without using a connection table. Usually UDP traffic (on a Service_UDP class listener) uses a connection table that associates a source IP/port combination with a pool member. Entries from this table aren't cleared (except after a timeout of 60s), even after the pool member is deleted, which causes a lot of UDP traffic to be sinkholed on load balancers that often switch/delete pool members, especially Gardener LBs.


openstack/octavia/templates/etc/_esd.json.tpl
```diff
@@ -50,5 +50,8 @@
   "hcm_rmk_restrict_internal": {
     "lbaas_ctcp": "cc_tcp_profile",
     "lbaas_irule": ["cc_hcm_rmk_restrict_internal"]
+  },
+  "stateless_udp": {
```
notandy, line 54: I would prefer something more cryptic to stay more consistent with the other ESD names and ensure nobody accidentally tags them like this.

notandy (changes requested):

openstack/octavia/templates/etc/_esd.json.tpl
m-kratochvil: I understand what you're trying to avoid, but I also remember back in the day we once said that these weird letters/numbers in the ESD names, invented by Torsten should go, cause nobody knows what they mean. But if I follow that convention, the name could be something like:

```
udp_noconn_c788a_v1_0
```

BenjaminLudwigSAP: Not sure whether I proposed this before, but I had the idea in the past of giving every future ESD a CCloud-specific prefix. Maybe that would already suffice to prevent it from being accidentally used. The simplest form would be a `cc_` prefix, but I'm open to other ideas in that direction (`cc_esd_`, `cc_special_`, your idea here).

m-kratochvil:
```suggestion
  "cc_esd_udp_stateless": {
      "lbaas_cudp": "cc_udp_datagram_profile"
  }
```
Something like that should be fine, no random characters or arbitrary code versions and I like the `esd` in the name.

PR #5225, opened 2023-07-31
plutono: remove superfluous templates that break the postgres subchart

majewsky:
As discussed on Slack with Richard.


thgrs (approved):
lgtm

PR #5229, opened 2023-07-31
openstack/keppel: set trivy ratelimit

SuperSandro2000:
For https://github.com/sapcc/keppel/pull/267

I basically just copied manifest pulls, we should think about the numbers and if they make sense before merging


majewsky:
Correction, you copied the number for manifest *pushes*. I think the general limit is fine, but we should be a bit more generous with the burst budget (e.g. 50) to support usecases where a monthly vulnerability report is compiled by script (like the SRC team does for our own images).

openstack/keppel/templates/_utils.tpl
```diff
@@ -48,6 +48,8 @@
   value: '300'  # per account
 - name:  KEPPEL_BURST_MANIFEST_PUSHES
   value: '15'   # per account
+- name:  KEPPEL_BURST_TRIVY_REPORT_RETRIEVES
+  value: '15'   # per account
```
SuperSandro2000, line 52:
```suggestion
  value: '50'   # per account
```

majewsky (approved):

majewsky:
Values look good. As a nitpick, "retrieves" should be "retrievals".

SuperSandro2000:
fixed this also in keppel itself

PR #5231, opened 2023-08-01
openstack/content-repo: convert to cronjob

SuperSandro2000:


openstack/content-repo/templates/cronjob.yaml
```diff
@@ -0,0 +1,85 @@
+{{- range $repo, $config := .Values.repos }}
+{{- if not $config.disabled }}
+apiVersion: batch/v1
+kind: CronJob
+
+metadata:
+  name: content-repo-{{ $repo }}
+  namespace: content-repo
```
majewsky, line 8: From the "let's fix this while we're at it" department: Namespace declarations on Helm templates are superfluous since Helm will override them with the target namespace anyway.

openstack/content-repo/templates/cronjob.yaml
```diff
@@ -0,0 +1,85 @@
...
+              secretName: swift-http-import-certs
+          containers:
+          - name: swift-http-import
+            image: {{ $.Values.global.registry }}/swift-http-import:{{ include "image_version" $.Values }}
+            command:
+              - /sbin/tini
+            args:
+              - /bin-content-repo/exponential-backoff
+              - /usr/bin/swift-http-import
+              - /etc/http-import/config/{{ $repo }}.yaml
```
majewsky, line 54: If I'm not mistaken (and please double-check me on this), Kubernetes should do retries with exponential backoff on failure anyway, so we can also get rid of our custom exponential-backoff script. And with that removed, we can also get rid of tini since swift-http-import does not fork. Since the swift-http-import binary is the default entrypoint, that would all lead to just this:

```suggestion
            args:
              - /etc/http-import/config/{{ $repo }}.yaml
```

majewsky (changes requested):

majewsky (approved):
@SuperSandro2000 I have added a commit with further cleanups. Please check before merging.

PR #5251, opened 2023-08-08
[neutron] degrade NSXTExporterNoMetrics alert to info

notandy:


sven-rosenzweig:
Why don't we assign the warning to `support_group: vmware` as with `NSXTExporterScrapeOlderTwoHours?`

PR #5252, opened 2023-08-09
[neutron] Add core 2.1 related config options

swagner-de:


openstack/neutron/templates/etc/_ml2-conf-cc-fabric.ini.tpl
```diff
@@ -1,5 +1,3 @@
 [ml2_cc_fabric]
 driver_config_path = /etc/neutron/plugins/ml2/cc-fabric-driver-config.yaml
-{{- if .Values.cc_fabric.handle_all_l3_gateways }}
 handle_all_l3_gateways = {{ .Values.cc_fabric.handle_all_l3_gateways }}
```
sebageek, line 3: So this is a required value now with no default set?

swagner-de:
Okay, so I tried a few different possibilities now and I think I understand why this was as it is. helm evaluates an empty field to `false`, just as the actual value `false`. Even a `empty false` will return true.

sebageek (approved):

swagner-de:
I have removed the commit changing the behavior.

PR #5254, opened 2023-08-09
keystone: upgrade to xena

bbobrov:


rajivmucheli (changes requested):
1. strange, this value is still train:  https://github.com/sapcc/helm-charts/blob/keystone-xena/openstack/keystone/values.yaml#L32
2. Below have to be updated :
https://github.com/sapcc/helm-charts/blob/keystone-xena/openstack/keystone/templates/bin/_db-sync.tpl#L4
3. point 1 has to be updated since  `release: "{{ .Release.Name }}"` is used in ingress, job-migration, seeds, etc

rajivmucheli (approved):
2nd upgrade for the year :)

PR #5255, opened 2023-08-10
keystone: remove more references to train

bbobrov:
This reverts commit https://github.com/sapcc/helm-charts/commit/d5cd90e52eff03a12a0975a1e3009819a34ec28a. It also removes mentions of train and train-specific bits, because, turns out, they are really train-specific and not required today.


rajivmucheli (approved):
lgtm

PR #5257, opened 2023-08-10
[maia] two new snmp metrics for maia federation

Kuckkuck:


richardtief (approved):
FYI @notque - requested from: https://github.wdf.sap.corp/PlusOne/enhancements/issues/91.

PR #5261, opened 2023-08-11
prometheus-kubernetes: thanos enabled

richardtief:
This PR enables Thanos for Kubernetes Prometheis. 
- added thanos dependency
- adjusted thanos-seed naming to distinguish seed names from different clusters

### DIFF
--- a/system/thanos-seeds/templates/seed.yaml
+++ b/system/thanos-seeds/templates/seed.yaml
@@ -4,7 +4,7 @@
 apiVersion: "openstack.stable.sap.cc/v1"
 kind: OpenstackSeed
 metadata:
-  name: {{ include "prometheus.name" (list $name $root) }}-thanos
+  name: {{ include "swift.userName" (list $name $root) | trimPrefix "prometheus-" }}
   labels:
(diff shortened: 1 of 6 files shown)


system/thanos-seeds/templates/seed.yaml
```diff
@@ -4,7 +4,7 @@
 apiVersion: "openstack.stable.sap.cc/v1"
 kind: OpenstackSeed
 metadata:
-  name: {{ include "prometheus.name" (list $name $root) }}-thanos
+  name: {{ include "swift.userName" (list $name $root) | trimPrefix "prometheus-" }}
```
viennaa, line 7: It looks like you want to drag in clusterType here with this outcome:

kubernetes-clusterType-thanos

We should add the clusterType directive here then too:
https://github.com/sapcc/helm-charts/blob/master/system/thanos-seeds/values.yaml#L16-L25C3

viennaa (changes requested):
Thanks for bringing this up and providing the PR. Two little remarks from my side to talk about.

viennaa (approved):
Great 👍

PR #5272, opened 2023-08-16
thanos-virtual|admin-k3s|kubernikus: remove release

richardtief:
The cluster-wide querier is not needed for these cluster types since there will probably always be a one-to-one relationship between the querier and the kubernetes prometheus. Queriers on the prometheus will then be queried directly.


viennaa (approved):
let's remove them.

PR #5273, opened 2023-08-16
add new thanos endpoints for regional query

viennaa:
all but metal and scaleout are queried directly since it would be a 1:1 relationship with a query in between.

### DIFF
--- a/system/thanos-metal/values.yaml
+++ b/system/thanos-metal/values.yaml
@@ -57,14 +57,14 @@ regional_thanos:
     - metal
     - scaleout
     - kubernetes-virtual
-    - kubernetes-admin-k3s
+    - kubernetes-admin
     - kubernetes-kubernikus
     - kubernetes-k-master
     - kubernetes-customer
     - kubernetes-internet
-    - kubernetes-concourse
-    - kubernetes-concourse2
-    - kubernetes-concourse3
+    - kubernetes-ci
+    - kubernetes-ci2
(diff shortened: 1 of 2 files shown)


system/thanos-metal/values.yaml
```diff
@@ -55,10 +55,16 @@ regional_thanos:
...
     - scaleout
+    - kubernetes-virtual
+    - kubernetes-admin-k3s
+    - kubernetes-kubernikus
+    - kubernetes-k-master
+    - kubernetes-customer
+    - kubernetes-internet
+    - kubernetes-concourse
+    - kubernetes-concourse2
+    - kubernetes-concourse3
```
richardtief, line 67:
```suggestion
    - kubernetes-ci
    - kubernetes-ci2
    - kubernetes-ci3
```

system/thanos-metal/values.yaml
```diff
@@ -55,10 +55,16 @@ regional_thanos:
 
   clusterTypes:
     - metal
-    - virtual
-    - admin
-    - kubernikus
     - scaleout
+    - kubernetes-virtual
+    - kubernetes-admin-k3s
```
richardtief, line 60:
```suggestion
    - kubernetes-admin
```

richardtief (changes requested):
We will stick to the naming already in place.

richardtief (approved):
Thanks

PR #5274, opened 2023-08-16
Redis increase CPU limit and request

sven-rosenzweig:
The RedisDB for services using openstack-rate-limiting is heavily throttled in all regions: Average ratio of throttled CPU seconds/used CPU seconds is above > 80%. Total used cpu (if there is no limit) is above 200% regions in the stages qa, bronze and silver. For eu-de-1 this value is higher than 400%.

### DIFF
--- a/common/redis/values.yaml
+++ b/common/redis/values.yaml
@@ -28,10 +28,10 @@ persistence:
 resources:
   limits:
     memory: 128Mi
-    cpu: 100m
+    cpu: 350m
   requests:
     memory: 128Mi
-    cpu: 10m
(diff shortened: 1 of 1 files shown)


auhlig (approved):
Maybe also consider a [VPA](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler#example-vpa-configuration) to automatically have resources configured.
Initially, I'd recommend `updateMode: "Off"` and `controlledValues: "RequestsOnly"` to observe the actual usage and resource recommendations.

PR #5275, opened 2023-08-16
traefik logic adapted to multiple scheme

viennaa:
traefik part of the chart was not able to handle the newer helpers made for the multiple deployment. adopted it and made serviceName configurable. Since we are only using it in one cluster, it should default to our used name.


richardtief (approved):
Looks good.

PR #5278, opened 2023-08-16
[hermes] increase es request to 20g

notque:
Increase global request to 20g to match a realistic view of the memory limits based on java opts


IvoGoman (approved):
LGTM 👍

PR #5285, opened 2023-08-18
setting ingressClassName and adding traefik documentation

viennaa:
the admin and customer cluster both need different special handling, which needs some inline explanation


richardtief (approved):
Looks good.

PR #5287, opened 2023-08-21
[util] Created function kubernetes_pod_AZ_spread in utils/templates/_…

dusandordevicsap:
Created function kubernetes_pod_AZ_spread in utils/templates/_affinity.tpl that makes sure pods are spread between all availability zones. Function should replace kubernetes_pod_anti_affinity for scheduling pods in all AZs.

### DIFF
--- a/openstack/utils/Chart.yaml
+++ b/openstack/utils/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: utils
-version: 0.10.1
+version: 0.10.2
--- a/openstack/utils/templates/_affinity.tpl
+++ b/openstack/utils/templates/_affinity.tpl
@@ -48,3 +48,40 @@ affinity:
                   values:
                   - operational
 {{- end }}
+
+{{ define "kubernetes_pod_AZ_spread" -}}
+{{- $envAll := index . 0 -}}
+{{- $application := index . 1 -}}
+{{- $component := index . 2 -}}
+affinity:
+  podAntiAffinity:
+  preferredDuringSchedulingIgnoredDuringExecution:
+  - weight: 1
+    podAffinityTerm:
+      labelSelector:
+        matchExpressions:
+          - key: "app"
+            operator: In
+            values:
+              - {{$envAll.Release.Name}}
+          - key: "component"
+            operator: In
+            values:
+              - {{$component}}
+      topologyKey: "topology.kubernetes.io/zone"
+  - weight: 2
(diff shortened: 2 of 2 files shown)


galkindmitrii:
I'd check topology spread constraints instead of affinity for AZ spreading:
https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/

galkindmitrii:
> > I'd check topology spread constraints instead of affinity for AZ spreading:
> > https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/
> 
> We did check it but left it for future for few reasons:
> 
> * It is supported from Kubernetes version 1.27, whereas we do have previous versions in some environments
> * Some of features there are still in beta
> 
It is supported since k8s 1.16: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md
New versions just add more features to it and some of those are beta.

dusandordevicsap:
> > > I'd check topology spread constraints instead of affinity for AZ spreading:
> > > https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/
> > 
> > 
> > We did check it but left it for future for few reasons:
> > 
> > * It is supported from Kubernetes version 1.27, whereas we do have previous versions in some environments
> > * Some of features there are still in beta
> 
> It is supported since k8s 1.16: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md New versions just add more features to it and some of those are beta.

I agree with you and documentation, but after checking with few colleagues that are more involved with this I got recommendation to stick with proposed solution until we are 100% confident new one is production ready.

businessbean (approved):
This is based on my [code for MariaDB Galera](https://github.com/sapcc/helm-charts/blob/069f35db78587b39e2c68ea14446ec0fff0c1842/common/mariadb-galera/helm/templates/statefulset-mariadb.yaml#L41). So I'm biased, but it is only a function. Every colleague can decide which features should be used for a certain Helm chart. Currently not all [Pod Topology Spread Constraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#topologyspreadconstraints-field) features are GA and they also mention that they will [not replace the old (anti)affinity rules](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/895-pod-topology-spread#impact-to-other-features). 

> The motivation of this KEP is to resolve limitations of existing features, but it won't replace them

PR #5294, opened 2023-08-24
[neutron] Alerts for rate limit hits

sven-rosenzweig:
Introducing two alerts based on the number of global (project independent) and local (project-based) rate limit hits. Whenever a limit exceeds **10%** of the overall requests for 5 minutes the alert triggers. The project independent based alert compares the tuple of action (create, update, etc.) and URI hitting the limit in relation to the overall number of requests of that type. Comparing the local vs global based alerts reveals that the first additionally groups by the project_id.

Both alerts are based on metrics provided by the
`openstack-watcher-middleware` and `openstack-rate-limit-middleware`. The first provides the information how many requests are being send. The latter provides the information of limit hits.

### DIFF
--- a/openstack/neutron/alerts/openstack/api.alerts
+++ b/openstack/neutron/alerts/openstack/api.alerts
@@ -16,3 +16,39 @@ groups:
     annotations:
       description: 'Neutron API Pod `{{ $labels.pod }}` request queue > 100 for 5 min'
       summary: 'Neutron API request queue is not clearing up, increase neutron-server deployment replicas for mitigation.'
+
+- name: neutron-globalrate-limit.alerts
+  rules:
+  - alert: GlobalRateLimitExceeded
+    expr: sum by (target_type_uri, action) (rate(openstack_requests_ratelimit_total{service="network",container="statsd", level="global"}[1m])) / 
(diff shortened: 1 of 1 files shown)


openstack/neutron/alerts/openstack/api.alerts
```diff
@@ -16,3 +16,41 @@ groups:
     annotations:
       description: 'Neutron API Pod `{{ $labels.pod }}` request queue > 100 for 5 min'
       summary: 'Neutron API request queue is not clearing up, increase neutron-server deployment replicas for mitigation.'
+
+- name: neutron-globalrate-limit.alerts
+  rules:
+  - alert: GlobalRateLimitExceeded
+    expr: sum by (target_type_uri, action) (rate(openstack_requests_ratelimit_total{service="network",container="statsd", level="global"}[15m]) * 60 ) / 
+            on (target_type_uri, action) 
+            (sum by (target_type_uri, action) (rate(openstack_watcher_api_requests_total{service="network", container="statsd"}[15m]) * 60)) > 0.1
```
swagner-de, line 25: Also, I see no reason why you multiply both values by 60.

swagner-de (changes requested):

PR #5296, opened 2023-08-25
nova/flavors: Reintroduce disabled m5.192xlarge

grandchild:
We need this flavor (and possibly more in the future) with the proper
extra-specs to update old VMs via script. Hence the new extra file
"_flavors_deleted.tpl" to contain them with information on when to
remove them again.

### DIFF
--- a/openstack/sap-seeds/templates/_flavors_deleted.tpl
+++ b/openstack/sap-seeds/templates/_flavors_deleted.tpl
@@ -1,0 +1,41 @@
+# Removed flavors are kept _private_ ("is_public: false") in here instead of
+# actually deleting them to allow management of old instances with these
+# flavors.
+
+# To query if one or more flavors are still in use run the following query in
+# the "nova" DB (replace $FLAVOR with the flavor name(s)) of every region:
+#
+#   SELECT
+#       i2.project_id,
+#       i2.created_at,
+#       i2.uuid,
+#       i2.host,
+#       i2.hostname AS name,
+#       j1.flavor
+#   FROM instances AS i2
+#   JOIN (
+#       SELECT
+#           instance_uuid,
+#           json_value(flavor, '$.cur."nova_object.data".name') AS flavor
+#       FROM instance_extra AS ie
+#       JOIN (SELECT uuid FROM instances WHERE deleted = 0) AS i
+#       ON i.uuid = ie.instance_uuid
+#   ) AS j1
+#   ON j1.instance_uuid = i2.uuid
+#   WHERE flavor in ('$FLAVOR');
+
+- name: "m5.192xlarge"
+  id: "271"
+  vcpus: 106
+  ram: 2979840
+  disk: 64
+  is_public: false
+  extra_specs:
+{{- if .Values.use_hana_exclusive }}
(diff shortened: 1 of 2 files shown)


openstack/sap-seeds/templates/_flavors_deleted.tpl
```diff
@@ -0,0 +1,41 @@
...
+- name: "m5.192xlarge"
+  id: "271"
+  vcpus: 106
+  ram: 2979840
+  disk: 64
+  is_public: false
+  extra_specs:
+{{- if .Values.use_hana_exclusive }}
+    {{- tuple . "vmware" | include "sap_seeds.helpers.flavor_extra_specs" | indent 4 }}
+    "trait:CUSTOM_HANA_EXCLUSIVE_HOST": "required"
```
joker-at-work, line 36: `vmware` has the same trait in is as forbidden. Please use `vmware_hana_exclusive` instead or `vmware_common` like below if you want to explicitly set the trait here.

Additionally, shouldn't we set a NUMA trait, too?
Should we set `hw:cpu_cores`?

grandchild: re `vmware_hana_exclusive`: oops, typo, fixed, thanks 🙂

re `hw:cpu_cores`: since the flavor is not really numa-aligned, and there's no guarantee of that, we don't have to set that constraint. i would argue for more flexibility for the vmware scheduler, plus i don't know what happens if `vpcus` is not divisible by `hw:cpu_cores`. probably nothing, but i don't know. i could test later, but i didn't invest the effort so far.

openstack/sap-seeds/templates/_flavors_deleted.tpl
```diff
@@ -0,0 +1,41 @@
...
+  id: "271"
+  vcpus: 106
+  ram: 2979840
+  disk: 64
+  is_public: false
+  extra_specs:
+{{- if .Values.use_hana_exclusive }}
+    {{- tuple . "vmware_hana_exclusive" | include "sap_seeds.helpers.flavor_extra_specs" | indent 4 }}
+    "trait:CUSTOM_HANA_EXCLUSIVE_HOST": "required"
+{{- else }}
```
joker-at-work, line 37: `vmware_hana_exclusive` already contains the `"trait:CUSTOM_HANA_EXCLUSIVE_HOST": "required"`

joker-at-work (changes requested):

openstack/sap-seeds/templates/_flavors_deleted.tpl
joker-at-work: What about the `NUMA` trait? Does the memory not align with it roughly?

grandchild: ah, the NUMA trait is the trait i wanted to set instead of HANA_EXCLUSIVE_HOST!

the NUMA trait is used for proper(tm) hana flavors to make sure the host and vm numa size matches. nowadays we only have the one numa size anymore (except for a few remaining clusters here and there, such as the 12TiB clusters), so it's less of a discriminating factor. for `m5.192xlarge` we don't really care whether it fits neatly, we only want the few remaining vms to be allowed on our hana clusters.

joker-at-work (approved):

PR #5297, opened 2023-08-29
[Keystone] introduce project creation in policy

rajivmucheli:
project creation should be possible via a new project project_creator apart the default policy rule/role


rajivmucheli:
if this idea is acceptable, i will then proceed adding this role in Elektra

bbobrov:
Please explain the reason behind the change. Why is it needed? Who is the user of the feature? What is the usecase?
________________________________

PR #5299, opened 2023-08-29
keystone: upgrade to zed

bbobrov:
This is a part of the effort to get rid of vulnerabilities in dependencies. Keystone is getting upgraded to zed


rajivmucheli (approved):
lgtm

PR #5303, opened 2023-08-30
[cc3test] disable CC3TestCanaryTestsNotRunning for now

thgrs:
it creates a lot of false alerts due to decomissioned bb's and their stale metrics etc - so lets disable it for now until we have found a better way of handling this


ashifnihalb (approved):
lgtm

PR #5315, opened 2023-08-31
nova/metrics: Add nova_events

grandchild:
This is a follow-up to https://github.com/sapcc/nova/pull/449 to make
the metrics visible to prometheus.


grandchild:
fixed name & labels TBD and open for discussion

PR #5319, opened 2023-09-01
ingress should always be nginx

viennaa:
if this isn't set, it will pick up traefik in certain clusters which won't work

### DIFF
--- a/common/prometheus-server/templates/ingress-internal.yaml
+++ b/common/prometheus-server/templates/ingress-internal.yaml
@@ -26,6 +26,11 @@ metadata:
     {{ end }}
 
 spec:
+  {{- if $.Values.traefik.enabled }}
+  ingressClassName: traefik
+  {{- else }}
+  ingressClassName: nginx
+  {{- fi }}
   rules:
     {{- range $host := coalesce $.Values.internalIngress.hosts (list $name) }}
     - host: {{ include "internalFQDNHelper" (list $host $root) }}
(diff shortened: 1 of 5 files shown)


richardtief (approved):
Thanks for making this standard.

common/prometheus-server/templates/ingress-internal.yaml
```diff
@@ -26,6 +26,11 @@ metadata:
     {{ end }}
 
 spec:
+  {{- if $.Values.traefik.enabled }}
+  ingressClassName: traefik
+  {{- else }}
+  ingressClassName: nginx
+  {{- fi }}
```
richardtief, line 33:
```suggestion
  {{- if not $.Values.traefik.enabled }}
  ingressClassName: nginx
  {{- end }}
```

common/prometheus-server/templates/ingress.yaml
```diff
@@ -27,6 +27,11 @@ metadata:
     {{ end }}
 
 spec:
+  {{- if $.Values.traefik.enabled }}
+  ingressClassName: traefik
+  {{- else }}
+  ingressClassName: nginx
+  {{- fi }}
```
richardtief, line 34:
```suggestion
  {{- if not $.Values.traefik.enabled }}
  ingressClassName: nginx
  {{- end }}
```

richardtief (changes requested):
`fi` is shell language not helm template.

PR #5327, opened 2023-09-05
nova/flavor: Increase "forced" disk size to 64GiB

grandchild:
The forced flavor, used for spawning tests on baremetal and elsewhere, still needs a proper disk size because images request a "min_disk" size in their attributes for the VM.

64GiB is the most common size for regular OS images, with most other sizes being smaller.


joker-at-work (approved):

grandchild:
argh, didn't wait for the tests, sorry! 🙁

PR #5328, opened 2023-09-06
[cronus] fix default poller interval

sapmli:


corey-aloia (approved):

openstack/cronus/values.yaml
```diff
@@ -526,7 +526,7 @@ poller: # service that handles received emails
   statefulSet:
     enabled: false
   action: forwardEmail # supported values: forwardEmail, activateEmail, simulator
-  pollInterval: 5 # should be less than a default cronus ResponseHeaderTimeout (9 seconds)
+  pollInterval: 5s # should be less than a default cronus ResponseHeaderTimeout (9 seconds)
```
corey-aloia, line 529: quick question: as it is "enabled: false" shouldn't nothing be deployed?

corey-aloia (approved):

openstack/cronus/values.yaml
corey-aloia: Ah, potentially it is enabled in the other env but as the wrong default

dhalimi: Yes, the reason is to control the deployment from cc/secrets per region

corey-aloia: Hi David, yes I know about the per region :) but this still shouldn't have the wrong default in the root values file. And, I believe shouldn't even be in this values file, but in the poller chart values file (see comment on the PR for cronus-qa pipeline)

PR #5335, opened 2023-09-06
common/prometheus-server: scrapeConfigSelector added

richardtief:


viennaa (approved):
Thanks, let's go with some configs.

PR #5356, opened 2023-09-15
nova/flavor: Allow old bigVM flavors on HANA BBs

grandchild:
For decommissioning old BBs we need to move these VMs away. This patch allows to migrate VMs with these flavors onto HANA BBs. (After the old VMs are updated with the new flavor info.)


grandchild:
As discussed in the VMware meeting, made the flavors private too.

fwiesel (approved):

PR #5357, opened 2023-09-15
exclude kubelet and cadvisor by default

viennaa:
this is needed in every prometheus-kubernetes based cluser, otherwise alerts will fire

### DIFF
--- a/system/prometheus-kubernetes/values.yaml
+++ b/system/prometheus-kubernetes/values.yaml
@@ -15,6 +15,12 @@ prometheus-server:
     enabled: true
     size: 100Gi
 
+  alerts:
+    multipleTargetScrapes:
+      exceptions:
+        - kubernetes-cadvisors
+        - kubernetes-kublet
+
 thanos:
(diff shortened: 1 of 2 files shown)


system/prometheus-kubernetes/values.yaml
```diff
@@ -15,6 +15,12 @@ prometheus-server:
     enabled: true
     size: 100Gi
 
+  alerts:
+    multipleTargetScrapes:
+      exceptions:
+        - kubernetes-cadvisors
+        - kubernetes-kublet
```
richardtief, line 22: Let's bring the comment
```suggestion
        # we exclude cadvisor metrics because it has the same instance as the kubelet but a different path
        # e.g. 10.246.204.80:10250/metrics vs. 10.246.204.80:10250/metrics/cadvisor
        - kubernetes-cadvisors
        - kubernetes-kublet
```

richardtief (approved):
All good. Just a detail

PR #5360, opened 2023-09-15
update net alerts to support group

geisslet:


prometheus-exporters/snmp-ntp-exporter/alerts/snmp-ntp-cisco-n7k.alerts
```diff
@@ -1,16 +1,16 @@
 groups:
```
viennaa, line 1: Empty groups don't make sense. Either comment this one or remove the file entirely.

viennaa (changes requested):
Thanks for changing the related alerts to the new group.

Wouldn't it make sense, to remove `tier` in the alerts touched? We had this discussion earlier today with @bassel-z and @richardtief and since the `support_group` is changing here, it is like the perfect opportunity.

Apart from this, please bump all `Chart.yaml` patch or minor versions to reflect a new change. 

This should be done for:
* cc3test
* neutron
* network-generic-ssh-exporter
* snmp-exporter
* snmp-ntp-exporter
* infra-monitoring

prometheus-exporters/snmp-ntp-exporter/alerts/snmp-ntp-cisco-n7k.alerts
geisslet: I would leave the tier for now, since it does no harm. 
true at chart .. I forgot .. thanks.
will do.

viennaa:
Usually you need to bump and merge dependent chart version before you use them and then run a `helm dep up` to get the latest dependencies. Merging this now and adjusting it after the subcharts have been built.

PR #5361, opened 2023-09-18
Aggregated rules consumed by Castellum

chuan137:
We move the queries from Castellum to here. Also monitor the snapshot usage to avoid snapshot spill to active share.

### DIFF
--- a/prometheus-exporters/netapp-exporter/charts/netapp-cap-exporter/aggregations/netapp-castellum.rules
+++ b/prometheus-exporters/netapp-exporter/charts/netapp-cap-exporter/aggregations/netapp-castellum.rules
@@ -4,7 +4,7 @@ groups:
 - name: castellum_manila
   rules:
 
-  # Case one: old provision style, where snapshot reserve is allocated AS PART OF the target share size and snapshot reserver percentage is 5.
+  # Case one: old provision style, where snapshot reserve is allocated AS PART OF the target share size and snapshot reserve percentage is 5.
   # The logical space reporting/enforcement is always disabled for them.
   # Share size = netapp_volume_total_bytes + netapp_volume_snapshot_reserved_bytes
   # Share usage = netapp_volume_used_bytes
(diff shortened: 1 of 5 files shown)


prometheus-exporters/netapp-exporter/charts/netapp-cap-exporter/aggregations/netapp-castellum.rules
```diff
@@ -0,0 +1,58 @@
...
+    expr: netapp_volume_provision_case_one * (netapp_volume_total_bytes + netapp_volume_snapshot_reserved_bytes)
+
+  # Case two: new provision style and logical space is NOT enabled
+  # New provision style means snapshot reserve is allocated IN ADDTION TO share size, and sanpshot reserve percentage is 50
+  # To avoid snapshot spill, share's minimal size must be larger than the snapshot used size.
+  # Share size = netapp_volume_total_bytes
+  # Share usage = netapp_volume_used_bytes
+  # Share minimal size = max(netapp_volume_total_bytes, netapp_volume_snapshot_used_bytes)
+  - record: netapp_volume_provision_case_two
+    expr: (netapp_volume_percentage_snapshot_reserve{share_id!="", volume=~"share.*"} / 50 == 1) * (1 + netapp_volume_is_space_enforcement_logical == 1)
```
majewsky, line 31: Am I correct in assuming that the `netapp_volume_provision_case_{one,two,three}` metrics would always be 1-valued? (That would be worth clarifying in the comment.)

If yes, then I'm confused how the second factor works: `1 + metric == 1` looks like it would evaluate to 2. I recommend adding an additional set of parentheses to clarify operator order.

chuan137: yes, the metrics `netapp_volume_provision_case_{one,two,three}` are 1-valued, so that they can be easily used in other metrics. 

The `+` operator takes higher precedence, so it's equivelant to `((1 + netapp_volume_is_space_enforcement_logical) == 1)`, which evaluates to 1. If I took `netapp_volume_is_space_enforcement_logical == 0` to filter the shares with no logical space enforcement, the expression would evaluate to 0.

I will add proper parentheses to improve readability.

majewsky (approved):

PR #5365, opened 2023-09-19
[cc3test] reduce the scope of the test jobs to subdirs

thgrs:
before we were always running all the different tests with the scope set to the complete tests directory which resulted in a lot of unneeded code execution (fixtures and parametrizations for all tests, even the ones not run by a certain job) and with this change we reduce the scope for each job to only the subdirectory (or subdirectories) the corresponding tests are defined in ... this is possible now as we have properly separated all the code


ashifnihalb (approved):
lgtm

PR #5370, opened 2023-09-20
cleaning up federation

viennaa:
bastion is no longer needed
netapp is switched to global thanos setup


richardtief (approved):
Thanks for cleaning this up.

PR #5376, opened 2023-09-21
removing cronus pushgateway from prometheus infra

dhalimi:


system/prometheus-infra/values.yaml
```diff
@@ -87,11 +87,6 @@ collector:
 interconnect_sre:
   enabled: false
 
-cronus:
-  enabled: false
-
-pushgateway_infra:
-  enabled: false
```
viennaa, line 90: please remove one more blank line here, otherwise we have 2 blank lines

viennaa (changes requested):
Thanks for providing the cleanup PR.

Please increase the patch version of the Chart.yaml by one.

PR #5379, opened 2023-09-22
chore(prom-infra): remove image-usage-exporter job from config

IvoGoman:
the exporter has been deleted and this clears it from the federate config


viennaa (approved):
Thanks  🧹

PR #5385, opened 2023-09-22
common/prometheus-server: no limits

richardtief:


richardtief:
They are never triggered on our Prometheus because we have not set a threshold that can be hit.

richardtief:
default is always `0` (no limit): 

```
 # An uncompressed response body larger than this many bytes will cause the
  # scrape to fail. 0 means no limit. Example: 100MB.
  # This is an experimental feature, this behaviour could
  # change or be removed in the future.
  [ body_size_limit: [<size>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#size) | default = 0 ]

  # Per-scrape limit on number of scraped samples that will be accepted.
  # If more than this number of samples are present after metric relabeling
  # the entire scrape will be treated as failed. 0 means no limit.
  [ sample_limit: [<int>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#int) | default = 0 ]

  # Per-scrape limit on number of labels that will be accepted for a sample. If
  # more than this number of labels are present post metric-relabeling, the
  # entire scrape will be treated as failed. 0 means no limit.
  [ label_limit: [<int>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#int) | default = 0 ]

  # Per-scrape limit on length of labels name that will be accepted for a sample.
  # If a label name is longer than this number post metric-relabeling, the entire
  # scrape will be treated as failed. 0 means no limit.
  [ label_name_length_limit: [<int>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#int) | default = 0 ]

  # Per-scrape config limit on number of unique targets that will be
  # accepted. If more than this number of targets are present after target
  # relabeling, Prometheus will mark the targets as failed without scraping them.
  # 0 means no limit. This is an experimental feature, this behaviour could
  # change in the future.
  [ target_limit: [<int>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#int) | default = 0 ]
```

viennaa (approved):
Ok fair point. Can go until we need limits here.

PR #5388, opened 2023-09-25
use kubernetes_pod_AZ_spread utils function to schedule API pods

dusandordevicsap:
use kubernetes_pod_AZ_spread utils function to schedule API pods, that should make sure pods are spread across all available AZs


rajivmucheli (approved):
lgtm

PR #5389, opened 2023-09-25
 [mariadb-galera] add unit testing for statefulset

bashar-alkhateeb:
add the first unit tests example to test mariadb-galera cluster
statefulset


businessbean (approved):
good start

PR #5397, opened 2023-09-26
[ipmi-exporter] moves to scrapeconfig

stefanhipfel:


BerndKue (changes requested):
You will have to change quiet a lot the options, I figgured yesterday when change the redfish-exporter scrape config.
See https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1alpha1.ScrapeConfig
For instance relabel_configs has to be changed to relabelings.

PR #5400, opened 2023-09-26
nova/flavors: Fix Unlist m5.96xlarge & x1.32xlarge

grandchild:
- Use correct vmware extra-specs template
- Move the flavors to _flavors_deleted.tpl

Also, remove the usage of hana_exclusive_contains_legacy_bigvm_flavors config value for hana-exclusive regions, because we don't want to delete flavors anymore if old VMs still refer to them.

### DIFF
--- a/openstack/sap-seeds/templates/_flavors_deleted.tpl
+++ b/openstack/sap-seeds/templates/_flavors_deleted.tpl
@@ -24,6 +24,25 @@
 #   ON j1.instance_uuid = i2.uuid
 #   WHERE flavor in ('$FLAVOR');
 
+- name: "m5.96xlarge"
+  id: "270"
+  vcpus: 90
+  ram: 1468416
+  disk: 64
+  is_public: false
+  extra_specs:
+{{- if .Values.use_hana_exclusive }}
+    {{- tuple . "vmware_hana_exclusive" | include "sap_seeds.helpers.extra_specs" | indent 4 }}
+{{- else }}
+    {{- tuple . "vmware_common" | include "sap_seeds.helpers.extra_specs" | indent 4 }}
+    "host_fraction": "1/4,3/4,1/2,1"
+    "resources:CUSTOM_BIGVM": "2"
+    "resources:CUSTOM_MEMORY_RESERVABLE_MB": "1468416"
+{{- end }}
+    "trait:CUSTOM_NUMASIZE_C48_M729": "required"
+    "hw:cpu_cores": "48"  # used in nova-vmware as cores-per-socket (24pCPU = 48vCPU)
+    "vmware:hw_version": "vmx-18"
+
 - name: "m5.192xlarge"
   id: "271"
   vcpus: 106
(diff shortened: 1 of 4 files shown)


grandchild:
This needs a corresponding secrets change, to remove `hana_exclusive_contains_legacy_bigvm_flavors`. Will be done once this is merged.

openstack/sap-seeds/templates/_flavors_deleted.tpl
```diff
@@ -33,9 +52,30 @@
...
+{{- if .Values.use_hana_exclusive }}
+    {{- tuple . "vmware_hana_exclusive" | include "sap_seeds.helpers.extra_specs" | indent 4 }}
+{{- else }}
+    {{- tuple . "vmware_common" | include "sap_seeds.helpers.extra_specs" | indent 4 }}
+    "host_fraction": "1,0.67,0.34"
+    "resources:CUSTOM_BIGVM": "2"
+    "resources:CUSTOM_MEMORY_RESERVABLE_MB": "1991680"
+{{- end }}
+    "trait:CUSTOM_NUMASIZE_C48_M729": "required"
+    "hw:cpu_cores": "48"  # used in nova-vmware as cores-per-socket (24pCPU = 48vCPU)
```
fwiesel, line 80: I noticed that maybe a bit late, but it says here `hw:cpu_cores=48` vs `vcpus=128` in line 66. Is that intended?

openstack/sap-seeds/templates/_flavors_deleted.tpl
```diff
@@ -24,6 +24,25 @@
...
+{{- if .Values.use_hana_exclusive }}
+    {{- tuple . "vmware_hana_exclusive" | include "sap_seeds.helpers.extra_specs" | indent 4 }}
+{{- else }}
+    {{- tuple . "vmware_common" | include "sap_seeds.helpers.extra_specs" | indent 4 }}
+    "host_fraction": "1/4,3/4,1/2,1"
+    "resources:CUSTOM_BIGVM": "2"
+    "resources:CUSTOM_MEMORY_RESERVABLE_MB": "1468416"
+{{- end }}
+    "trait:CUSTOM_NUMASIZE_C48_M729": "required"
+    "hw:cpu_cores": "48"  # used in nova-vmware as cores-per-socket (24pCPU = 48vCPU)
```
fwiesel, line 43: It says here `"hw:cpu_cores": "48"` vs `vcpus: 90`  in line 29, is that intended?

openstack/sap-seeds/templates/_flavors_deleted.tpl
grandchild: As the comment implies, this number is the cores-per-socket, a.k.a. the vCPU NUMA size for the flavor. So it's always <= the number of vCPUs of the flavor.

openstack/sap-seeds/templates/_flavors_deleted.tpl
grandchild: see above.

openstack/sap-seeds/templates/_flavors_deleted.tpl
grandchild: this is highly confusing in nova -- i thought the comment would clear it up, but since other people have been confused by it too already, even with the comment, i should probably reword it to make it clearer. -> separate PR

fwiesel: So it is intended. I still surprised that it is a non-integer factor.

fwiesel (approved):

PR #5401, opened 2023-09-26
Changing postfix alert trigger #433

dhalimi:


rschardosin (approved):

prometheus-exporters/cronus-exporter/templates/postfix-alerts.yaml
```diff
@@ -1,5 +1,5 @@
 {{- $values := .Values }}
-{{- if $values.postfix.postfixEnabled }}
```
corey-aloia, line 2: Nice! Much clearer :)

PR #5405, opened 2023-09-27
[neutron] Parametrized mounting of /development volume that is used for debugging

dusandordevicsap:
At the moment /development volume is mounted in any case. That one is used for development, but is not needed for production. We already do have debug option in values yaml file, so this changes add if statement related to that option (for example if debug.asr_agent = true then mount volume). This change should remove mounting of /development when debug is set to false


dusandordevicsap:
Will create new one with /development completely removed

PR #5410, opened 2023-09-28
nova/metrics: Fix empty instances gauge

grandchild:
The query used the disused and now mostly-empty `instance_types`
table, which made the metric report no values (or only a very few
ancient instances in some regions).

Use instance_extra instead and parse the flavor name out of the JSON.
The query is a bit more expensive this way (it doubled in time from
the query without any join).

### DIFF
--- a/openstack/nova/values.yaml
+++ b/openstack/nova/values.yaml
@@ -348,10 +348,10 @@ mysql_metrics:
           instances.project_id,
           COUNT(*) AS gauge,
           instances.vm_state,
-          coalesce(instance_types.name,'N/A') AS flavor_name
+          coalesce(JSON_VALUE(instance_extra.flavor, '$.cur."nova_object.data".name'),'N/A') AS flavor_name
         FROM instances
-        JOIN instance_types ON instances.instance_type_id=instance_types.id
-        GROUP BY instances.vm_state, instances.host, instances.project_id, instance_types.name;
+        LEFT JOIN instance_extra ON instances.uuid=instance_extra.instance_uuid
+        GROUP BY instances.vm_state, instances.host, instances.project_id, flavor_name;
       values:
       - "gauge"
     - help: Total Nova VM Stuck count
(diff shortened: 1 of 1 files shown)


Carthaca:
Since you are reactivating a non-working metric, may I suggest to think about another change: exclude soft-deleted entries.

Otherwise our customers would always have to do the extra work of filtering to get the same instance list like they get via API. And I'm not aware of use-cases for having deleted instances.

openstack/nova/values.yaml
```diff
@@ -530,11 +530,11 @@ mysql_metrics:
           instances.project_id,
           instances.hostname,
           instances.vm_state,
-          coalesce(instance_types.name,'N/A') AS flavor_name,
+          coalesce(JSON_VALUE(instance_extra.flavor, '$.cur."nova_object.data".name'),'N/A') AS flavor_name
```
Carthaca, line 533: missing a comma at the end

Carthaca (changes requested):

openstack/nova/values.yaml
grandchild: ooh, well spotted!

Carthaca (approved):

grandchild:
sorry, I added the non-deleted instances clause in a separate commit. would you have a look again?

openstack/nova/values.yaml
```diff
@@ -530,11 +531,12 @@ mysql_metrics:
...
           instances.hostname,
           instances.vm_state,
-          coalesce(instance_types.name,'N/A') AS flavor_name,
+          coalesce(JSON_VALUE(instance_extra.flavor, '$.cur."nova_object.data".name'),'N/A') AS flavor_name,
           COUNT(*) AS gauge
         FROM instances
-        JOIN instance_types ON instances.instance_type_id=instance_types.id where host LIKE 'nova-compute-ironic%'
-        GROUP BY instances.vm_state, instances.project_id, instance_types.name, instances.uuid, instances.hostname, instances.node;
+        LEFT JOIN instance_extra ON instances.uuid=instance_extra.instance_uuid WHERE host LIKE 'nova-compute-ironic%'
+        WHERE instances.deleted = 0
```
Carthaca, line 538:
```suggestion
        AND instances.deleted = 0
```

Carthaca: There is already a `WHERE` in the line above

Carthaca (changes requested):

openstack/nova/values.yaml
grandchild: O.O thanks again!

Carthaca (approved):

PR #5411, opened 2023-09-29
rm tier filter for octobus routing

geisslet:
tier is removed from some alerts already so the previous routing filter to octobus caused gaps


bassel-z (approved):
Looks good to me. With this PR more alerts as before will be routed to octobus as an example tier: storage

geisslet:
yes . more alerts is the intention

PR #5420, opened 2023-10-06
Enable py-tracebacker for Neutron UWSGI

sebageek:
This option will create a unix domain socket for each worker that will print out tracebacks if a worker runs into harakiri. With harakiri-verbose we'll also see the tracebacks in our standard logs, which hopefully will allow us to better debug them.

### DIFF
--- a/openstack/neutron/templates/etc/_uwsgi.ini.tpl
+++ b/openstack/neutron/templates/etc/_uwsgi.ini.tpl
@@ -45,6 +45,7 @@ harakiri = 120
 harakiri-verbose = true
 post-buffering = 4096
 backlog-status = true
+py-tracebacker = /var/lib/neutron/uwsgi_pytracebacker.
 
 {{ if gt (.Values.api.cheaper | int64) 0 -}}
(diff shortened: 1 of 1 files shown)


openstack/neutron/templates/etc/_uwsgi.ini.tpl
```diff
@@ -45,6 +45,7 @@ harakiri = 120
 harakiri-verbose = true
 post-buffering = 4096
 backlog-status = true
+py-tracebacker = /var/lib/neutron/uwsgi_pytracebacker.
```
grandchild, line 48: A little comment explaining that this is the file prefix could be helpful to future readers being confused, like I was, about the period at the end.

grandchild (changes requested):

openstack/neutron/templates/etc/_uwsgi.ini.tpl
sebageek: I added an explanation to the commit message.

grandchild (approved):

PR #5429, opened 2023-10-11
prometheus-server: get PrometheusMultiplePodScrape alerts evaluated

richardtief:
This is due to the removal of `kube-state-metrics` from the non-Kubernetes prometheus.


common/prometheus-server/values.yaml
```diff
@@ -429,6 +433,7 @@ alerts:
   multiplePodScrapes:
     enabled: true
 
+
```
viennaa, line 436:
```suggestion
```

viennaa (commented):
Looks good, just removing a blank line.

viennaa (approved):

PR #5431, opened 2023-10-11
audit-logs: move alerts partially to Thanos Rule

richardtief:


system/audit-logs/vendor/fluent-audit-container/templates/prometheus-alerts.yaml
```diff
@@ -12,11 +13,17 @@ metadata:
     app: fluent-audit-container-prometheus
     tier: infra
     type: alerting-rules
-    prometheus: {{ required ".Values.alerts.prometheus missing" $values.alerts.prometheus | quote }}
+    {{- if contains "thanos" $target }}
+    # alerts, which combine metrics from infra-collector and kubernetes, are available from the thanos metal query.
+    {{ $target }}: {{ index $values.alerts.prometheus $target | required (printf "$values.alerts.prometheus.%s missing" $target) }}
```
IvoGoman, line 18: Can this be missing? The loop is only entered if there is any entry in `$values.alerts.prometheus`

system/audit-logs/values.yaml
```diff
@@ -56,7 +56,10 @@ fluent_audit_container:
       host: none
   alerts:
     enabled: false
-    prometheus: infra-collector
+    # Keys = directory names in alerts/ and aggregations/
+    prometheus:
+      infra-collector: infra-collector
+      thanos-ruler: metal
```
IvoGoman, line 62: Maybe we can restructure this to not having to rely on parsing the name
```suggestion
    prometheus:
      - name: infra-collector
        type: prometheus
      - name: metal
        type: thanos-ruler
```

system/audit-logs/vendor/fluent-audit-container/templates/prometheus-alerts.yaml
```diff
@@ -1,7 +1,8 @@
 {{- if eq .Values.global.clusterType "metal" }}
 {{- $values := .Values }}
 {{- if $values.alerts.enabled }}
-{{- range $path, $bytes := .Files.Glob "alerts/*.alerts" }}
+{{- range $target, $unused := $values.alerts.prometheus }}
+{{- range $path, $bytes := $.Files.Glob (printf "alerts/%s/*.alerts" $target) }}
```
IvoGoman, line 5:
```suggestion
{{- range $path, $bytes := $.Files.Glob (printf "alerts/%s/*.alerts" $target.name) }}
```

system/audit-logs/vendor/fluent-audit-container/templates/prometheus-alerts.yaml
```diff
@@ -12,11 +13,17 @@ metadata:
     app: fluent-audit-container-prometheus
     tier: infra
     type: alerting-rules
-    prometheus: {{ required ".Values.alerts.prometheus missing" $values.alerts.prometheus | quote }}
+    {{- if contains "thanos" $target }}
+    # alerts, which combine metrics from infra-collector and kubernetes, are available from the thanos metal query.
+    {{ $target }}: {{ index $values.alerts.prometheus $target | required (printf "$values.alerts.prometheus.%s missing" $target) }}
+    {{- else }}
+    prometheus: {{ index $values.alerts.prometheus $target | required (printf "$values.alerts.prometheus.%s missing" $target) }}
+    {{- end }}
```
IvoGoman, line 21:
```suggestion
    {{ $target.type | required (printf "$values.alerts.prometheus.[%v].type missing" $i) }}: {{ $target.name | required (printf "$values.alerts.prometheus.[%v].name missing" $i) }}
```

system/audit-logs/vendor/fluent-audit-container/templates/prometheus-alerts.yaml
```diff
@@ -1,7 +1,8 @@
 {{- if eq .Values.global.clusterType "metal" }}
 {{- $values := .Values }}
 {{- if $values.alerts.enabled }}
-{{- range $path, $bytes := .Files.Glob "alerts/*.alerts" }}
+{{- range $target, $unused := $values.alerts.prometheus }}
```
IvoGoman, line 4:
```suggestion
{{- range $i, $target := $values.alerts.prometheus }}
```

IvoGoman (commented):
I have added some comments w.r.t. using a list of dicts instead. I feel this makes it easier to understand, as all target are handled the same and there is no need for searching in strings. 
What do you think about it? (Have not rendered this locally yet)

system/audit-logs/vendor/fluent-audit-container/templates/prometheus-alerts.yaml
richardtief: This is just one solution that you can find on several charts in this repo. But yes - this is not failsafe, so we need to make sure it's there.

richardtief:
Just rendered it locally and the thanos-ruler alert is not coming up. Only the prometheus: infra-collector one.

PR #5432, opened 2023-10-11
infra-monitoring: kube-state-metrics removed

richardtief:


richardtief:
https://github.com/sapcc/helm-charts/pull/5431

richardtief:
px-exporter: https://github.com/sapcc/helm-charts/pull/5433

viennaa (approved):
Looks good, would merge this before #5433 to avoid px conflicts.

PR #5433, opened 2023-10-11
infra-monitoring|px-exporter: move alerts to kubernetes

richardtief:


viennaa (commented):
Thanks. 

I'd bump the chart after #5432 has been merged to avoid any clashes.

viennaa (commented):
Can you bump infra-monitoring to `1.8.1` here?

swagner-de (approved):

PR #5443, opened 2023-10-13
[mariadb][common] Enable VPA for mariadb container

rajivmucheli:
Service owner can set enabled: True to enable, by default its false

### DIFF
--- a/common/mariadb/templates/deployment.yaml
+++ b/common/mariadb/templates/deployment.yaml
@@ -26,9 +26,7 @@ spec:
         prometheus.io/scrape: "true"
         prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
 {{- end }}
-{{- if .Values.vpa.set_main_container }}
         vpa-butler.cloud.sap/main-container: mariadb
-{{- end }}
     spec:
       affinity:
         nodeAffinity:
--- a/common/mariadb/templates/deployment.yaml
+++ b/common/mariadb/templates/deployment.yaml
@@ -26,7 +26,9 @@ spec:
         prometheus.io/scrape: "true"
         prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
 {{- end }}
+{{- if .Values.vpa.set_main_container }}
         vpa-butler.cloud.sap/main-container: mariadb
+{{- end }}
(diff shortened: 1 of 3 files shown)


Nuckal777 (approved):
lgtm

rajivmucheli:
Hi @viktor-kovacs, 

Shall we rollback the minor version mariadb upgrade to allow this commit to be merged ?

rajivmucheli:
FYI @joker-at-work @sebageek, this PR introduces VPA for nova, neutron mariadb deployments.

common/mariadb/templates/deployment.yaml
```diff
@@ -26,7 +26,9 @@ spec:
         prometheus.io/scrape: "true"
         prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
 {{- end }}
-
+{{- if .Values.vpa.enabled }}
+        vpa-butler.cloud.sap/main-container: mariadb
+{{- end }}
```
joker-at-work, line 31: Why do have to enable it explicitly? Does it hurt if it's active by default?

I don't think `vpa.enabled` catches the meaning very good. I'd prefer something like `vpa.set_main_container`.

rajivmucheli: only nova, neutron, keystone were part of the list. Other services were not.

asseco-int-mgmt:
> Hi @viktor-kovacs,
> 
> Shall we rollback the minor version mariadb upgrade to allow this commit to be merged ?

Hi Rajiv!

Yes, We are going to rolling back today.

Regards,Viktor

common/mariadb/templates/deployment.yaml
joker-at-work: Doesn't answer the second question, but with the rename of the value, I'm fine.

Nuckal777: Should work, maybe make a toggle to remove the annotation instead?

viktor-kovacs: An Enable/Disable (toggle) option is more useful even if the default value is enabled or disable (true/false) than and option to remove if you don't want to enable it. In that case nobody won't forget to specify if they want to enable/disable it. Also it helps during templating.

businessbean (approved):
lgtm with the recommended changes and the added documentation snippet

PR #5445, opened 2023-10-13
[Thanos] request ephemeral storage by default

viennaa:
Compaction is done on ephemeral storage and it tends to use gigabytes. If it is more than 25Gi a pvc needs to be added. In case of bigger node an overwrite exists.


richardtief (approved):
Thanks. Looks good.

PR #5446, opened 2023-10-13
[manila] run ensure as service

chuan137:


kpawar-sap (approved):

Carthaca:
can only be rolled out after https://github.com/sapcc/manila/pull/143 is in

PR #5447, opened 2023-10-16
replaced metric to match

viennaa:
lots of missing alerts due to a non existent metric name
  replaced with latest mixin:
  https://monitoring.mixins.dev/thanos/#thanosstoreseriesgatelatencyhigh


richardtief (approved):
Cool, thanks.

PR #5448, opened 2023-10-16
[alertmanager] rm all pd routes && rm ack buttons

geisslet:
PD sunset
rm routes to PD before to avoid delivery errors.
Ack buttons also disappear since Pulsar can't access new engagement tool


geisslet:
@bassel-z as discussed

bassel-z (approved):
LGTM

PR #5451, opened 2023-10-17
common/prometheus-alertmanager: slack links grafana replaced with plutono

richardtief:


bassel-z:
Please update the ALM-base-dependency version in alertmanager-operated chart so we can pull the new version in alertmanager deployment.
https://github.com/sapcc/helm-charts/blob/master/global/prometheus-alertmanager-operated/Chart.yaml#L11

PR #5459, opened 2023-10-17
prometheus-infra.scaleout: kube-state-metrics removed

richardtief:
With Thanos, it is no longer necessary to have `kube-state-metrics` available in multiple Prometheus. Kubernetes Prometheus keeps kube_ metrics available so that they only need to be maintained in one place. Dependent services have been adjusted accordingly so that no absent metric alerts are generated.


richardtief:
https://github.com/sapcc/helm-charts/pull/5460

viennaa (approved):
Yeah! Leets angle grind this out 👍

PR #5464, opened 2023-10-18
delete old alert without metrics

artherd42:


richardtief (approved):
Looks good. Thanks for the cleanup.

PR #5468, opened 2023-10-18
[keystone] MariaDB Galera support added to templates

businessbean:
- hostname for the DB service
- option to disable the seed CRD (needed for integration tests)
- related values.yaml entries will be added later
- chart version bumped


bbobrov (changes requested):
I am putting this on hold while we are working on VPA and upgrade to Zed; i will lift my -1 when we are done with the tasks

bbobrov (approved):

openstack/keystone/Chart.lock
```diff
@@ -1,10 +1,10 @@
 dependencies:
 - name: mariadb
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.7.11
+  version: 0.8.0
```
bbobrov, line 4: Will this upgrade mariadb? I am concerned, because we never figured out why mariadb failed last time, and decided to keep it at the old version until we manage to investigate.

bbobrov: disregard this

businessbean: I have used the value from the master branch. Why is it shown as a diff?
- https://github.com/sapcc/helm-charts/blob/master/openstack/keystone/Chart.yaml#L17
- https://github.com/sapcc/helm-charts/blob/master/openstack/keystone/Chart.lock#L4

PR #5469, opened 2023-10-18
prepare to switch the ingress-nginx annotation prefix

majewsky:
We want to migrate from the old `ingress.kubernetes.io/`-prefixed annotations to the standard format `nginx.ingress.kubernetes.io/`. During the migration, we need to have all Ingress objects annotated in both ways to ensure forward-compatibility. After the move, the old annotations can be removed.

This change is mostly autogenerated through
```
sed -i 's+^\(\s*\)\(ingress.kubernetes.io/.*\)+\1\2\n\1nginx.\2+g' $(rg ingress.kubernetes.io -l)
```

I am currently removing the Gatekeeper check that complains about `nginx.ingress.kubernetes.io/`. Once we start rolling this out, I can add a new check that complains about `ingress.kubernetes.io/foo` without `nginx.ingress.kubernetes.io/foo`.

### DIFF
--- a/common/linkerd-support/README.md
+++ b/common/linkerd-support/README.md
@@ -86,6 +86,7 @@ metadata:
   annotations:
     {{- if and $.Values.global.linkerd_enabled $.Values.global.linkerd_requested }}
     ingress.kubernetes.io/service-upstream: "true"
+    nginx.ingress.kubernetes.io/service-upstream: "true"
     {{- end }}
 ```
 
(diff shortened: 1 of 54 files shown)


databus23 (approved):

majewsky:
> Once we start rolling this out, I can add a new check that complains about `ingress.kubernetes.io/foo` without `nginx.ingress.kubernetes.io/foo`.

Added in https://github.com/sapcc/helm-charts/commit/2d0c4bae1ee3b30fc6f9b19f45ca6bd7fb1adf2d

PR #5470, opened 2023-10-18
remove collector federation

viennaa:
dashboards and alerts have been migrated and are available via Thanos now.


richardtief (approved):
Cool. Let's remove this finally

PR #5339, opened 2023-09-07
fix(mariadb): metric was renamed

IvoGoman:
The metric `mysql_info_schema_threads_seconds` has been renamed to `mysql_info_schema_processlist_seconds` with exporter v0.14.0 see https://github.com/prometheus/mysqld_exporter/pull/603

PR #5313, opened 2023-08-31
[MariaDB] - Security Patches upgrade to 10.5.21

viktor-kovacs:
Patch includes:
     - Mariadb upgrade from 10.5.17 to 10.5.21
     - mysqld-exporter upgrade from v0.12.1 to v0.14.0

PR #5472, opened 2023-10-18
[MariaDB] - Rolling back Security Patches 10.5.21 to 10.5.17

viktor-kovacs:
Reason: The 10.5.21 release blocking connection on various system

Reverting back the following PRs:
[MariaDB] - Security Patches upgrade to 10.5.21: https://github.com/sapcc/helm-charts/pull/5313 
rename metric in alert: https://github.com/sapcc/helm-charts/pull/5339

### DIFF
--- a/common/mariadb/Chart.yaml
+++ b/common/mariadb/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: mariadb
-version: 0.7.15
+version: 0.7.16
--- a/common/mariadb/templates/alerts/_mysql.alerts.tpl
+++ b/common/mariadb/templates/alerts/_mysql.alerts.tpl
@@ -28,7 +28,7 @@
       summary: {{ include "fullName" . }} reports slow queries.
 
   - alert: {{ include "alerts.service" . | title }}MariaDBWaitingForLock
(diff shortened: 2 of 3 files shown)


IvoGoman:
Do we need to downgrade the mysqld exporter as well? v0.12.1 is 2 years old and the GH repo states that the exporter currently supports MariaDB >=10.3. If this is just a precaution, then I am fine with it.

businessbean:
> Do we need to downgrade the mysqld exporter as well? v0.12.1 is 2 years old and the GH repo states that the exporter currently supports MariaDB >=10.3. If this is just a precaution, then I am fine with it.

- `0.12.0` has no official MariaDB support https://github.com/prometheus/mysqld_exporter/tree/release-0.12#mysql-server-exporter-
- `0.13.0` supports `10.2` and newer https://github.com/prometheus/mysqld_exporter/tree/release-0.13#mysql-server-exporter-
- `0.14.0` supports `10.2` and newer https://github.com/prometheus/mysqld_exporter/tree/v0.14.0#mysql-server-exporter-

`0.14.0` would still be the preferred version. It is just not the complete rollback that Viktor wants to achieve.

viktor-kovacs:
The roll back is only temporary and all the changes should be rolled back.

businessbean (approved):
lgtm. Let's roll back to a working version combination

PR #5473, opened 2023-10-18
[memcached] priorityClassName made optional

businessbean:
- to support K8s setups without preconfigured PriorityClass objects
- will be used for integration tests
- chart version bumped

### DIFF
--- a/common/memcached/templates/deployment.yaml
+++ b/common/memcached/templates/deployment.yaml
@@ -93,6 +93,10 @@ spec:
             memory: {{ .Values.metrics.resources.requests.memory | quote }}
         {{- end }}
       {{- end }}
-      {{- if not (and (hasKey $.Values "priorityClass") (hasKey $.Values.priorityClass "disabled") ($.Values.priorityClass.disabled)) }}
-      priorityClassName: {{ default "openstack-service-critical" .Values.priority_class | quote }}
+      {{- if .Values.priority_class }}
+        {{- if and (kindIs "bool" .Values.priority_class) (eq .Values.priority_class true) }}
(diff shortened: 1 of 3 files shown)


common/memcached/templates/deployment.yaml
```diff
@@ -93,4 +93,6 @@ spec:
             memory: {{ .Values.metrics.resources.requests.memory | quote }}
         {{- end }}
       {{- end }}
+      {{- if not (and (hasKey $.Values "priorityClass") (hasKey $.Values.priorityClass "disabled") ($.Values.priorityClass.disabled)) }}
       priorityClassName: {{ default "openstack-service-critical" .Values.priority_class | quote }}
+      {{- end }}
```
majewsky, line 98: This interface looks unnecessarily confusing. Why have `.Values.priority_class` be a string and then use the different key `.Values.priorityClass` to hold `{"disabled": true}`? I would suggest using a sentinel value:

```suggestion
      {{- $priority_class := .Values.priority_class | default "openstack-service-critical" }}
      {{- if ne $priority_class "none" }}
      priorityClassName: {{ quote $priority_class }}
      {{- end }}
```

Or even simpler, declare `priority_class: openstack-service-critical` in this chart's values.yaml and allow using charts to disable it by overriding that to null. Then you only need:

```suggestion
      {{- if .Values.priority_class }}
      priorityClassName: {{ quote .Values.priority_class }}
      {{- end }}
```

majewsky (changes requested):

majewsky (approved):
I still find it overly surprising to allow boolean values instead of just strings and nulls, but I'm assuming you have some very specific reason to need to do that.

PR #5479, opened 2023-10-20
feat(audit-logs): add probes and alerts for all endpoints

IvoGoman:
This will allow adding more probe targets and respective alerts via entries in the secrets.
Once this is deployed, we can remove the redundant probes from the `probe-exporter`


artherd42 (approved):
lgtm

PR #5485, opened 2023-10-20
[manila] prep for antelope chart

Carthaca:
make more options configurable

Most of the things are separated by the chart name
and will change when used in an dependent chart with different name.

### DIFF
--- a/openstack/manila/Chart.lock
+++ b/openstack/manila/Chart.lock
@@ -1,10 +1,10 @@
 dependencies:
 - name: mariadb
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.7.12
+  version: 0.7.16
 - name: memcached
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.1.2
+  version: 0.1.4
 - name: mysql_metrics
   repository: https://charts.eu-de-2.cloud.sap
   version: 0.2.7
@@ -16,9 +16,9 @@ dependencies:
   version: 0.5.2
 - name: utils
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.10.1
+  version: 0.10.4
 - name: redis
   repository: https://charts.eu-de-2.cloud.sap
   version: 1.2.1
-digest: sha256:c011baadcd527910c8ff1b12f6cabc639a5051395336be95e48fdc25461f1bbc
-generated: "2023-08-08T10:15:30.903346+02:00"
+digest: sha256:0b9f025d6c1e574828335e08628c0305ce1921583c7578f6c87a42e059d3ea17
+generated: "2023-10-23T10:58:00.307139+02:00"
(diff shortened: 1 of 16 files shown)


openstack/manila-antelope/values.yaml
```diff
@@ -0,0 +1,8 @@
+# Default values for manila-antelope.
+alerts:
+  enabled: false
+coordinationBackendMountPath: manila
+seeds:
+  enabled: false
```
Carthaca, line 6: needs a better name - since the endpoints are still seeded, but nothing else

openstack/manila-antelope/Chart.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - name: chuan137
+name: manila-antelope
+sources:
+  - https://github.com/sapcc/manila
+version: 0.0.1
+dependencies:
+  - name: manila
+    alias: manila-antelope
+    # repository: https://charts.eu-de-2.cloud.sap
+    repository: file://../manila
```
Carthaca, line 17: will be adjusted

openstack/manila/Chart.lock
```diff
@@ -15,10 +15,10 @@ dependencies:
   repository: https://charts.eu-de-2.cloud.sap
   version: 0.5.2
 - name: utils
-  repository: https://charts.eu-de-2.cloud.sap
-  version: 0.10.1
+  repository: file://../utils
```
Carthaca, line 18: will be adjusted, waiting for https://github.com/sapcc/helm-charts/pull/5484

chuan137 (changes requested):
Thanks for working on this. Only one thing, the coordinate pvc name is defined using Chart name. So is the mount file name. Maybe changing the util function can quickly fix it

Carthaca:
I already implemented https://github.com/sapcc/helm-charts/commit/a4fa85952f77cd9c48386af4b867718e510c63ce to fix the mount path.

And I'm okay with having a new pvc for coordination - were you thinking of re-using the existing one for antelope then?

chuan137:
reuse the pvc is fine for me, but then it should not be deployed twice

Carthaca:
No, here nothing will change and in the antelope installation it will be set https://github.com/sapcc/helm-charts/pull/5498/files#diff-b4405cb045726a33d9d8d6321aaadec7de545581a2a559d79e8bdc7ee72aedc3R4

Carthaca:
I think re-using is more complex - and since we plan to do this in qa only not worth the effort

Carthaca:
No, defined here https://github.com/sapcc/helm-charts/blob/27f1ae247eff35093e32b917f557166eea526118/openstack/utils/templates/_coordination.tpl#L2

it contains the Chart name which will be different

chuan137:
hmm, does it work as expected? the resource is defined in manila chart, what is the chart name  when the yaml file is parsed? manila or manila-antelope. I am not very sure

chuan137 (approved):

Carthaca:
according to helm template it works 🙂

PR #5490, opened 2023-10-23
[prometheus] userName and containerName configurable

viennaa:
the directive previously had no impact. setting these values will now result in actually configuring swift the way it is specified in the config.
as thanos-seeds are leveraging the same helper class, they need to follow the pattern. went for a simple copy approach to not overfill the template with too many if-else patterns to differentiate and still provide the old behaviour


richardtief (approved):
This looks good.

PR #5493, opened 2023-10-23
[Cinder] Update mariadb chart to 0.7.12

hemna:
This patch updates the mariadb chart to 0.7.12 to help prevent false positives with CinderMariaDBNotReady alerts.


Carthaca (approved):
lgtm

PR #5494, opened 2023-10-23
elastiflow: move alerts partially to Thanos Rule

richardtief:
With Thanos, it is no longer necessary to have kube-state metrics available in multiple Prometheus. Kubernetes Prometheus keeps these metrics available so that they only need to be maintained in one place. Dependent services have been adjusted accordingly so that no absent metric alerts are generated.


IvoGoman (approved):
Thank you 👍

PR #5495, opened 2023-10-23
poller chart: fixing volume mount

corey-aloia:
It looks like the volume name was changed but not the volume mount for `poller-mx-records`


rick-sanchez-in-c137 (approved):
LGTM

PR #5496, opened 2023-10-23
adjustable storage config name

viennaa:
when having a name crunch with thanos and prometheus and different containeres are used, the auto generated storage-config-secret is not matching.


richardtief (approved):
Looks good

PR #5498, opened 2023-10-24
[manila] add chart for antelope release

Carthaca:
to be run in parallel, modelled as depedent chart


Carthaca:
waiting for https://github.com/sapcc/helm-charts/pull/5485 to be able to update Chart.lock

PR #5520, opened 2023-10-26
removing invalid defaults

corey-aloia:
The chart had invalid values as defaults. I believe these values are also optional. IMO, For optional values, no default should be supplied, for required values that do not have a valid default, we should rather use the required function.

### DIFF
--- a/openstack/poller/values.yaml
+++ b/openstack/poller/values.yaml
@@ -18,7 +18,7 @@ poller:
     mtaName: dns; cronus-poller.example.com
     smtpPorts:
       - 25000
-    localName: 
+    localName:
   sesCredentials:
   activateEmail:
   dumpContainer:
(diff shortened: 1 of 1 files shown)


openstack/poller/values.yaml
```diff
@@ -31,7 +31,7 @@ poller:
   printMessage: false
   rebuildDSN: false
   rhea:
-  queueName: 
+  queueName:
```
sapmli, line 34: Sure this shouldn't be nested inside the `rhea:` key?

sapmli (changes requested):

openstack/poller/values.yaml
sapmli: see https://github.com/sapcc/helm-charts/blob/master/openstack/poller/templates/_config.yaml.tpl#L72

corey-aloia: You are right.... I did not actually change this value, and my IDE just removed an unneeded space at this line.
I will go ahead and nest and then merge.... but looking at this values file.... there are also other things wrong. Someone likely needs to go through this chart and improve

PR #5523, opened 2023-10-26
global/alertmanager: deploy new keppel inhibit in prod regions

SuperSandro2000:


viennaa (approved):
To be rolled out on Wednesday next week.

PR #5545, opened 2023-11-02
alertmanager: move Swift alerts from #cc-os-swift to #cc-swift

majewsky:
Same as for my own services that were already moved. This was discussed with the Swift service owners (cc @sumitarora2786).


richardtief (approved):
@bassel-z fyi I'll push this change in your absence.

PR #5548, opened 2023-11-03
Ironic: update chart dependencies

sandzwerg:
Lots of updates piled up


fwiesel (commented):
You also need to run a `helm dep up` to update the `Chart.lock` file.

sandzwerg:
That makes totally sense. Please check again @fwiesel

fwiesel (approved):

PR #5556, opened 2023-11-06
[linkerd-support] Add helper snippets

joker-at-work:
Instead of having to add 3 lines into every template, one can use these snippets and only add a single line. This also makes it easier to upgrade (most of) the annotations if necessary.

### DIFF
--- a/common/linkerd-support/Chart.yaml
+++ b/common/linkerd-support/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v2
 name: linkerd-support
-version: 0.1.3
+version: 0.1.4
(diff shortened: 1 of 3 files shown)


majewsky:
I tried having something like this, but it does not work in all cases. Library charts like `common/...` cannot access this if the `linkerd-support` chart is only added as a dependency on the top-level chart. And adding this one as dependency on a library chart carries the risk of duplicate objects being rendered. In the end, I preferred having consistent instructions.

joker-at-work:
Ok, then I'l have to put it in `openstack/utils` instead. Too bad.

PR #5560, opened 2023-11-07
[Nova] Add basic linkerd support

joker-at-work:
This adds the `linkerd-support` chart as dependency and also adds the annotations to all `Deployment`, `Service` and `Ingress` objects. `Job` objects are out of the picture for now, because they need more than just annotations as we need to shut down linkerd for the pod to shut down. There will be a follow-up to support these.

We depend on the `utils` chart for the annotation snippets and the conditions so our templates stay more compact and thus better readable.

### DIFF
--- a/openstack/nova/templates/_console-service.yaml.tpl
+++ b/openstack/nova/templates/_console-service.yaml.tpl
@@ -9,6 +9,7 @@ metadata:
   name: nova-console-{{ $name }}
   annotations:
   {{- include "utils.topology.service_topology_mode" . | indent 2 }}
+  {{- include "utils.linkerd.pod_and_service_annotation" . | indent 4 }}
   labels:
     system: openstack
     type: backend
--- a/openstack/nova/Chart.lock
+++ b/openstack/nova/Chart.lock
@@ -16,7 +16,7 @@ dependencies:
   version: 0.1.1
 - name: utils
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.10.4
+  version: 0.11.1
 - name: mariadb
   repository: https://charts.eu-de-2.cloud.sap
   version: 0.7.17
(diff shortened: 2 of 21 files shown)


openstack/nova/templates/_console-service.yaml.tpl
```diff
@@ -9,6 +9,7 @@ metadata:
   name: nova-console-{{ $name }}
   annotations:
   {{- include "utils.topology.service_topology_mode" . | indent 2 }}
+  {{- include "utils.linkerd.pod_and_service_annotation" . | indent 4 }}
```
fwiesel, line 12: There is a mismatch... Is that really correct?

fwiesel (commented):
I start to think, we should have simply a `utils.snippets.pod_annotations`, and `utils.snippets.ingress_annotations`, and then we update them in one place instead. But that is probably a different discussion.

openstack/nova/templates/_console-service.yaml.tpl
joker-at-work: Yes, this come from the fact that `utils.linkerd.pod_and_service_annotation` doesn't contain any indentation while `utils.topology.service_topology_mode` [contains 2 spaces](https://github.com/sapcc/helm-charts/blob/5904aaefdd8bcdd7ab04babe8a2e4a60adf1a069/openstack/utils/templates/_topology.tpl#L69).

joker-at-work:
> I start to think, we should have simply a `utils.snippets.pod_annotations`, and `utils.snippets.ingress_annotations`, and then we update them in one place instead. But that is probably a different discussion.

You mean for all the possible annotations we can have? This would make sense if we have the same annotations everywhere. Not sure if that's the case. Especially across services.

openstack/nova/templates/_console-service.yaml.tpl
fwiesel: Ah, my bad.

fwiesel:
> You mean for all the possible annotations we can have? 

Let's take it elsewhere. I was just mentioning in the hope. you would see it immediately the same. But we best discuss such a broader change also in a broader context then one for nova only.

PR #5532, opened 2023-10-30
[rabbitmq][mariadb][memcached] linkerd annotation support added

businessbean:
- parameter added to enable the annotation per pod type independent from the `global.linkerd_enabled` & `global.linkerd_requested` parameters
- chart version bumped

PR #5565, opened 2023-11-08
[rabbitmq] add linkerd annotation

bashar-alkhateeb:
adding support for linkerd annotation to rabbitmq helm chart the annotation can be activated by setting up the following global values to true :
```
.Values.global.linkerd_enabled = true
.Values.global.linkerd_requested = true
```

if the previous values are set globally in the top level Chart, you can deactivate the annotation for this particular chart by overriding:` .Values.linkerd.enabled` to `false` (default true)

for more information about linkerd lease find:
https://github.com/sapcc/helm-charts/tree/master/common/linkerd-support

splitting PR: https://github.com/sapcc/helm-charts/pull/5532 into multiple PRs per helm chart

### DIFF
--- a/common/rabbitmq/values.yaml
+++ b/common/rabbitmq/values.yaml
@@ -112,3 +112,7 @@ upgrades:
 #  rollingUpdate:
 #    maxUnavailable: 1
 #    maxSurge: 3
+
+linkerd:
+  # linkerd annotation for the RabbitMQ pod (true/false)
+  enabled: true
--- a/common/rabbitmq/Chart.yaml
+++ b/common/rabbitmq/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rabbitmq
-version: 0.5.2
+version: 0.6.0
 description: A Helm chart for RabbitMQ
 sources:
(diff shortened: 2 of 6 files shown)


common/rabbitmq/values.yaml
```diff
@@ -112,3 +112,7 @@ upgrades:
 #  rollingUpdate:
 #    maxUnavailable: 1
 #    maxSurge: 3
+
+linkerd:
+  #linkerd annotation for the RabbitMQ pod (true/false)
```
joker-at-work, line 117: nit: missing a space after the `#` and annmotations also happen for Service objects not only for pods

fwiesel (approved):

common/rabbitmq/values.yaml
bashar-alkhateeb: space added after `#`

businessbean:
I would opt for more complexity in the template like [here](https://github.com/sapcc/helm-charts/pull/5532/files#diff-90340c00f939e26fb7f4f8b1f2fed98569011391c334c96db3ea15424ed5dd3cR29) to keep the `values.yaml` as simple as possible. But anyhow, these are only details...

joker-at-work (approved):

PR #5573, opened 2023-11-09
[Cinder] Update chart to include linkerd

hemna:
This adds the `linkerd-support` chart as dependency and also adds the annotations to all `Deployment`, `Service` and `Ingress` objects.

We depend on the `utils` chart for the annotation snippets and the conditions so our templates stay more compact and thus better readable.

### DIFF
--- a/openstack/cinder/Chart.lock
+++ b/openstack/cinder/Chart.lock
@@ -1,24 +1,27 @@
 dependencies:
 - name: utils
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.10.4
+  version: 0.11.1
 - name: mariadb
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.7.12
+  version: 0.8.0
 - name: mysql_metrics
   repository: https://charts.eu-de-2.cloud.sap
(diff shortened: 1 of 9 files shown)


joker-at-work (commented):
[This one](https://github.com/sapcc/helm-charts/blob/03f865fd54afb9e51c37a3d68932cf4156bd5156/openstack/cinder/templates/volumes/_volume-deployment.yaml.tpl#L31-L38) seems to be missing.

Additionally, the `api-ratelimit-redis` chart contains `Service` and `Deployment` objects, but I'm not sure if it's prepared for linkerd, yet. Cinder might need another version bump for that as a follow-up. Please contact the owner of the chart.

joker-at-work:
Regarding the `api-ratelimit-redis`: I failed to see that this is just the `redis` chart. It looks to me like you need at least version 1.3.5 for that to support linkerd. Might be best to bump to the latest version to also get any (security) fixes.

joker-at-work (approved):
I haven't looked at the rendered output, but I don't see any obvious errors with the change.

joker-at-work (approved):

PR #5581, opened 2023-11-13
[openstack/utils] Common snippets for a trust-bundle

fwiesel:
The trust-manager[^1] offers the capability to centrally manage root certificates. This allows us to inject a root of trust with kubernetes means during runtime instead of patching the images to contain them during build-time.

We can revoke root-certificates in the same fashion.

The small side-effect is not having to patch images to contain our company CA.

The bigger degree of flexiblity also allows us to easily add a custom CA for doing mutual tls authentication on the service level, adding one option for removing credentials from the charts.

Defaulting to enabled, as the charts need to be adapted on a per service-level anyway, and is more an option to disable it, in case there is a problem discovered after deployment.


fwiesel:
I would take care of the chart changes in the services, they would look like this: https://github.com/sapcc/helm-charts/pull/5582/files

I've validated with curl that certificates from our company CA certificates are actually accepted.

openstack/utils/templates/_proxysql.tpl
```diff
@@ -63,6 +63,7 @@
     {{- include "utils.proxysql.volume_mount" . | indent 2 }}
       {{- end }}
     {{- end }}
+  {{- include "utils.trust_bundle.volume_mount" . | indent 2 }}
```
joker-at-work, line 66: > Defaulting to enabled, as the charts need to be adapted on a per service-level anyway, and is more an option to disable it, in case there is a problem discovered after deployment.

With this line here, every chart updating the utils chart will automatically get one trust-manager volume mount in (basically) every pod, while it's not clear that the chart is including the volume definition. I suspect this will lead to problems. Can we include the `utils.trust_bundle.volumes` somehow (with another name) to make it work or do you think every chart will be adapted before upgrading to utils `0.12.0`?

fwiesel: My idea was update all the charts before upgrading to `utils>=0.12.0`.
Yes, it will break things, but it would fail quite early in the pipeline when trying to apply the chart, not somewhere down the pipeline.

I think, we should avoid keeping too many variants of how things are done around, we should rather focus our effort on making things more equal more quickly.

But you are right, that might block changes on utils and requires syncing with people on that.
I'll default to disable it, and enable it in the values files with the changes to the charts using it in the values.

fwiesel: Updated the commit message accordingly as well.

joker-at-work (approved):
There's a `_debug.tpl` snippet containing some container, but I'm not sure what it's used for. That one might maybe eventually also need the certificates.

The kubernetes-entrypoint snippet defines a container, but imho doesn't need the certificate, because it only interacts with k8s itself and that should be fine, I guess.

PR #5583, opened 2023-11-13
disable pod scrape alerts by default

viennaa:
since kube_state_metrics are available in the respective kubernetes proms only, they are usually not needed

### DIFF
--- a/common/prometheus-server/CHANGELOG.md
+++ b/common/prometheus-server/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 7.4.4
+
+* default MultiplePodScrape-alert to false 
+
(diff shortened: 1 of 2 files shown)


richardtief (changes requested):
This is **opt out** implemented in every chart. There is no need to change it. If a Prometheus does not have `kube_` metrics, it can specify the respective Thanos 
Rule instance instead. https://github.com/sapcc/helm-charts/pull/5429

richardtief:
e.g. https://github.com/sapcc/helm-charts/blob/3f85354e213d12ea83a68182916587da43a643c5/system/infra-monitoring/values.yaml#L91-L96

PR #5587, opened 2023-11-13
[hermes] keyword to custom analyzer for audit index to enable searching

notque:


notque:
This should not be merged until later this week when all other pending changes have gone through.

PR #5591, opened 2023-11-14
[openstack/nova] Make use of trust-bundle snippets

fwiesel:
The snippets replace the system ca trust-bundle with a centrally managed one by trust-manager in k8s. This eliminates the need for patching the images to contain the company internal CA.

### DIFF
--- a/openstack/nova/Chart.lock
+++ b/openstack/nova/Chart.lock
@@ -16,7 +16,7 @@ dependencies:
   version: 0.1.5
 - name: utils
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.11.1
+  version: 0.12.1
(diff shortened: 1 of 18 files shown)


joker-at-work (commented):
Do our `statsd` exporter containers not need the new trust bundle? They get scraped from the outside and don't do any requests on their own, I guess.

fwiesel:
Exactly, I thought it would be not necessary. It would be more consistent, but then I would have to check if it also works with go applications. I would rather revisit that at a later stage. First get it out of the door for the loci images, and then we'll see. Maybe it would be easier to do all that simply in web mutation hook.

joker-at-work (approved):

PR #5601, opened 2023-11-14
[openstack/manila] Make use of trust-bundle snippets

fwiesel:
The snippets replace the system ca trust-bundle with a centrally managed one by trust-manager in k8s. This eliminates the need for patching the images to contain the company internal CA.


chuan137 (approved):
lgtm

PR #5606, opened 2023-11-15
[Neutron] Add basic linkerd support, bump chart and supporting charts…

dusandordevicsap:
… versions. This adds the `linkerd-support` chart as dependency and also adds the
annotations to all `Deployment`, `Service`, `Daemonset`  and `Ingress` objects. 

We depend on the `utils` chart for the annotation snippets and the

Chart version bumped to 0.2.2.

### DIFF
--- a/openstack/neutron/Chart.yaml
+++ b/openstack/neutron/Chart.yaml
@@ -1,26 +1,26 @@
 apiVersion: v2
 description: A Helm chart for Openstack Neutron
 name: neutron
-version: 0.2.0
+version: 0.2.2
 appVersion: "yoga"
 dependencies:
   - condition: mariadb.enabled
     name: mariadb
     repository: https://charts.eu-de-2.cloud.sap
-    version: 0.7.5
+    version: 0.8.0
   - name: memcached
     repository: https://charts.eu-de-2.cloud.sap
-    version: 0.1.1
+    version: 0.1.5
   - condition: mariadb.enabled
     name: mysql_metrics
     repository: https://charts.eu-de-2.cloud.sap
     version: 0.2.7
   - name: rabbitmq
     repository: https://charts.eu-de-2.cloud.sap
-    version: 0.5.1
+    version: 0.6.0
   - name: utils
     repository: https://charts.eu-de-2.cloud.sap
-    version: ~0.7.2
+    version: 0.11.1
   - name: owner-info
     repository: https://charts.eu-de-2.cloud.sap
     version: 0.2.0
@@ -29,3 +29,6 @@ dependencies:
     repository: https://charts.eu-de-2.cloud.sap
     version: 1.2.29
     condition: rate_limit.enabled
+  - name: linkerd-support
+    repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
+    version: 0.1.3
--- a/openstack/neutron/Chart.yaml
+++ b/openstack/neutron/Chart.yaml
@@ -27,7 +27,7 @@ dependencies:
   - name: redis
     alias: api-ratelimit-redis
     repository: https://charts.eu-de-2.cloud.sap
-    version: 1.2.29
+    version: 1.3.5
     condition: rate_limit.enabled
   - name: linkerd-support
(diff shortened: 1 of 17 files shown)


openstack/neutron/Chart.yaml
```diff
@@ -29,3 +29,6 @@ dependencies:
     repository: https://charts.eu-de-2.cloud.sap
     version: 1.2.29
```
joker-at-work, line 30: We need to bump redis to at least `1.3.5` to make it support linkerd.

openstack/neutron/templates/deployment-rpc-server.yaml
```diff
@@ -37,6 +37,7 @@ spec:
         {{- if .Values.proxysql.mode }}
         prometheus.io/scrape: "true"
         prometheus.io/targets: {{ required "$.Values.metrics.prometheus missing" $.Values.metrics.prometheus }}
+        {{- include "utils.linkerd.pod_and_service_annotation" $ | indent 8 }}
```
joker-at-work, line 40: Should this really be inside the `if`?

openstack/neutron/templates/ingress-server.yaml
```diff
@@ -10,6 +10,7 @@ metadata:
   annotations:
     kubernetes.io/tls-acme: "true"
     disco: "true"
+    {{- include "utils.linkerd.pod_and_service_annotation" $ | indent 4 }}
```
joker-at-work, line 13: This needs to be the ingress and not he pod_and_service annotation.

openstack/neutron/templates/vct-nsxv3-agent-deployment.yaml
```diff
@@ -43,6 +43,7 @@ template: |
           configmap-etc-hash: {{ include (print $.Template.BasePath "/configmap-etc.yaml") . | sha256sum }}
           prometheus.io/scrape: "true"
           prometheus.io/targets: {{ required ".Values.metrics.prometheus missing" .Values.metrics.prometheus | quote }}
+          {{- include "utils.linkerd.pod_and_service_annotation" $ | indent 10 }}
```
joker-at-work, line 46: Just wondering: did it work properly with the `$` here? The other includes use `.` in this file.

joker-at-work (commented):
It looks like you missed to update `templates/deployment-ironic-agent.yaml`.

openstack/neutron/templates/vct-nsxv3-agent-deployment.yaml
dusandordevicsap: It was failing when I tried . so I used $

openstack/neutron/Chart.yaml
dusandordevicsap: Updated it now,  thanks !

openstack/neutron/templates/deployment-rpc-server.yaml
dusandordevicsap: Mistake, moved it out

openstack/neutron/templates/vct-nsxv3-agent-deployment.yaml
dusandordevicsap: I tested it and it works with . but I left $ for consistency, since on some other places . is failing. Do you think it would be better to use . or leave $ everywhere ?

joker-at-work: I'd try to be consistent with the usage of `.` and `$` within the same file.

dusandordevicsap: OK makes sense, updated it as well

joker-at-work: I just marked this one, but there are lots of others, too, that use `$` even though the rest of the file uses `.`. Since you decided to be consistent in a file, could you please do the same in the other files, too?

joker-at-work (commented):
I don't see any obvious problems anymore but also haven't rendered the chart to check in detail.

PR #5619, opened 2023-11-16
common/postgresql-ng: add user provisioning

SuperSandro2000:


common/postgresql-ng/templates/configmap.yaml
```diff
@@ -0,0 +1,69 @@
...
+data:
+{{- $users := .Values.users }}
+{{- if .Values.persistence.createBackupUser }}
+{{- $backup = nil }}
+{{- $users = $backup }}
+{{- end }}
+{{- if .Values.createMetricsUser }}
+{{- $metrics = nil }}
+{{- $users = $metrics }}
+{{- end }}
```
majewsky, line 33: What is the purpose of this structure? It looks broken to me. For example, `if .Values.createMetricsUser`, then `$metrics` is set to nil and then therefore `$users` is also set to nil, which would make the entire following `range` ineffective.

common/postgresql-ng/templates/secrets.yaml
```diff
@@ -16,20 +16,26 @@ spec:
       - name: channel-config
         # at the time of writing, kubectl is only available in edge
         image: "{{ $.Values.global.dockerHubMirror }}/library/alpine:edge"
+        {{- $users := "" }}
+        {{- range $user, $settings := .Values.users }}
+        {{- $users = printf "%s postgres-user-password-%s" $users $user}}
```
majewsky, line 21: These secret names should contain `.Release.Name` to avoid conflicts between multiple Postgres DBs within the same namespace.

common/postgresql-ng/templates/secrets.yaml
```diff
@@ -16,20 +16,26 @@ spec:
...
+            for password in postgres-password {{ $users }}; do
+              # if we already have a secret, we can stop here
+              if [[ "$(kubectl get secrets "$password" --ignore-not-found)" != "" ]]; then
+                exit 0
+              fi
 
-            head -c 30 </dev/urandom | base64 > postgres-password
-            kubectl create secret generic postgres-password --from-file=postgres-password
+              head -c 30 </dev/urandom | base64 > "$password"
+              kubectl create secret generic "$password" --from-file="$password"
```
majewsky, line 37: This is not something that needs to be addressed in this PR, but we should think about setting `.metadata.ownerReferences` on these secrets to ensure that they are cleaned up when the Helm release is uninstalled (and also to ensure proper owner-info attribution).

common/postgresql-ng/values.yaml
```diff
@@ -19,11 +19,13 @@ postgresUser: postgres
...
-# dbInit: name of a configmap that provides the db-init.sql script
+# custom init db sql lines
+# initDb: |
+#   ...
 
-# custom db-maintain.sql
-# dbMaintain: name of a configmap that provides the db-maintain.sql script
+# custom maintain db sql lines
+# maintainDb: |
+#   ...
```
majewsky, line 28: The fields can also be set to empty-string instead of being commented out like this.

```suggestion
# A custom SQL script that will run when the database is first created.
initDb: ""
# A custom SQL script that will run everytime the database starts up.
maintainDb: ""
```

Also, we can consider giving these more plausible names, e.g. `sql.onCreate` and `sql.onStartup`.

common/postgresql-ng/values.yaml
```diff
@@ -37,9 +39,25 @@ config:
...
+    # Ensure the user has those grant lines run on each start.
+    # TO $USER with the username is added per line.
+    # %PGDATABASE% is replaced with the actual database name.
+    #grant: GRANT CONNECT,SELECT ON DATABASE "%PGDATABASE%"
+    # give out a general set of default permissions to let an app manage its own db
+    #manageAppDb: true
+    # Alter the table owner to this user. Make sure to only set this once!
+    #setOwnerOnTable: true
+
+createMetricsUser: true
```
majewsky, line 55: I don't see a reason to not create a pgmetrics user, so we could put this user as an example here to document how this works.
```suggestion
# users to create; by default a metrics user for postgres-exporter is created
users:
  # NOTE: backup and postgres are reserved user names and adding them here might cause unexpected issues
  metrics:
    # Ensure the user has those grant lines run on each start.
    # TO $USER with the username is added per line.
    # %PGDATABASE% is replaced with the actual database name.
    # For example, a read-only application user would need `GRANT CONNECT,SELECT ON DATABASE "%PGDATABASE%"`
    grant: "GRANT pg_monitor"
    # give out a general set of default permissions to let an app manage its own db
    manageAppDb: false
    # Alter the table owner to this user. Make sure to only set this once!
    setOwnerOnTable: false
```

common/postgresql-ng/values.yaml
```diff
@@ -37,9 +39,25 @@ config:
...
+# users to create
+users:
+  # NOTE: backup, metrics and postgres are reserved user names and adding them here might cause unexpected issues
+  # acme-user:
+    # Ensure the user has those grant lines run on each start.
+    # TO $USER with the username is added per line.
+    # %PGDATABASE% is replaced with the actual database name.
+    #grant: GRANT CONNECT,SELECT ON DATABASE "%PGDATABASE%"
+    # give out a general set of default permissions to let an app manage its own db
+    #manageAppDb: true
```
majewsky, line 51: This does not appear to be used.

common/postgresql-ng/templates/configmap.yaml
SuperSandro2000: I was searching for the dict function and didn't know how to do it....

common/postgresql-ng/values.yaml
SuperSandro2000: The metrics user needs quite special permissions and I don't want to proxy them through the values if not necessary.

common/postgresql-ng/templates/secrets.yaml
SuperSandro2000: I added a todo for that

PR #5623, opened 2023-11-17
[Octavia] Add basic linkerd support, bump chart and supporting charts…

dusandordevicsap:
… versions. This adds the linkerd-support chart as dependency and also adds the

Chart version bumped to 10.0.1

### DIFF
--- a/openstack/octavia/templates/octavia-api-ingress.yaml
+++ b/openstack/octavia/templates/octavia-api-ingress.yaml
@@ -12,8 +12,7 @@ metadata:
     type: api
     component: octavia
   annotations:
-  {{- include "utils.linkerd.ingress_annotation" $ | indent 4 }}
-
+  {{- include "utils.linkerd.ingress_annotation" . | indent 4 }}
 {{- with .Values.ingress.annotations }}  
 {{ toYaml . | indent 4 }}
 {{- end }}
--- a/openstack/octavia/templates/octavia-api-ingress.yaml
+++ b/openstack/octavia/templates/octavia-api-ingress.yaml
@@ -13,6 +13,7 @@ metadata:
     component: octavia
 {{- with .Values.ingress.annotations }}
   annotations:
+  {{- include "utils.linkerd.pod_and_service_annotation" $ | indent 4 }}
 {{ toYaml . | indent 4 }}
 {{- end }}
 spec:
--- a/openstack/octavia/templates/octavia-api-ingress.yaml
+++ b/openstack/octavia/templates/octavia-api-ingress.yaml
@@ -12,10 +12,10 @@ metadata:
     type: api
     component: octavia
   annotations:
+    # linkerd annotations
     {{- include "utils.linkerd.ingress_annotation" . | indent 4 }}
-    {{- with .Values.ingress.annotations }}  
-{{ toYaml . | indent 4 }}
-    {{- end }}
+    # other annotations
+    {{- .Values.ingress.annotations | toYaml | nindent 4 }}
 spec:
   tls:
      - secretName: tls-{{ include "octavia_api_endpoint_host_public" . | replace "." "-" }}
--- a/openstack/octavia/templates/octavia-api-ingress.yaml
+++ b/openstack/octavia/templates/octavia-api-ingress.yaml
@@ -11,9 +11,10 @@ metadata:
     system: openstack
     type: api
     component: octavia
-{{- with .Values.ingress.annotations }}
   annotations:
   {{- include "utils.linkerd.pod_and_service_annotation" $ | indent 4 }}
+
+{{- with .Values.ingress.annotations }}  
 {{ toYaml . | indent 4 }}
 {{- end }}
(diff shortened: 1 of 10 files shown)


openstack/octavia/templates/octavia-api-service.yaml
```diff
@@ -3,7 +3,7 @@ kind: Service
 metadata:
   name: octavia-api
   annotations:
-  {{- include "utils.topology.service_topology_mode" . | indent 2 }}
+    {{- include "utils.topology.service_topology_mode" . | indent 2 }}
```
joker-at-work, line 6: Looks like you missed adding the linkerd annotation here.

openstack/octavia/templates/octavia-api-ingress.yaml
```diff
@@ -13,6 +13,7 @@ metadata:
     component: octavia
 {{- with .Values.ingress.annotations }}
   annotations:
+  {{- include "utils.linkerd.pod_and_service_annotation" $ | indent 4 }}
 {{ toYaml . | indent 4 }}
 {{- end }}
```
joker-at-work, line 18: This needs to change, because when `.Values.ingress.annotations` is empty, we will not add the `annotations` key and the linkerd annotation. While `.Values.ingress.annotations` is currently not empty by default, the logic is still flawed.

Additionally, this needs to be `utils.linkerd.ingress_annotation`.

openstack/octavia/templates/octavia-api-service.yaml
dusandordevicsap: Fixed, thanks !

openstack/octavia/templates/octavia-api-ingress.yaml
dusandordevicsap: I fixed wrong annotation and changed order of things, annotation key is not any more subject to `.Values.ingress.annotations`

openstack/octavia/templates/octavia-api-ingress.yaml
```diff
@@ -11,9 +11,10 @@ metadata:
     system: openstack
     type: api
     component: octavia
-{{- with .Values.ingress.annotations }}
   annotations:
   {{- include "utils.linkerd.pod_and_service_annotation" $ | indent 4 }}
+
+{{- with .Values.ingress.annotations }}  
 {{ toYaml . | indent 4 }}
```
joker-at-work: I'd rather not add an empty line here so it's easier to see that these belong to annotations. It would imho also make sense to indent the `with` and `end` block (and maybe the `toYaml`, too).

Additionally, with moving the `include` out of the `with`, you don't need to use `$` anymore.

dusandordevicsap: Fixed empty line and $/.
Re indentation, it is not rendering properly if indented, so I left it

BenjaminLudwigSAP: I pushed commit 573099c06a1cbe7136fb6fceff3ac54de8939491 to fix this.

BenjaminLudwigSAP (changes requested):

openstack/octavia/templates/octavia-api-service.yaml
```diff
@@ -3,7 +3,8 @@ kind: Service
...
```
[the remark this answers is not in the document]
joker-at-work: The `service_topology_mode` has 2 indents in addition to the ones given via `indent 2` in its template: https://github.com/sapcc/helm-charts/blob/73820c446cdb136d4dacc5d0dde7b646a201709d/openstack/utils/templates/_topology.tpl#L69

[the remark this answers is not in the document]
BenjaminLudwigSAP: Ugh okay. Even though [on other lines it's correct](https://github.com/sapcc/helm-charts/blob/73820c446cdb136d4dacc5d0dde7b646a201709d/openstack/utils/templates/_topology.tpl#L33)... Whatever, a PR for that can be opened at another point in the future. I'm going to approve this here PR.

BenjaminLudwigSAP (approved):

openstack/octavia/templates/octavia-api-ingress.yaml
```diff
@@ -11,10 +11,11 @@ metadata:
...
     type: api
     component: octavia
-{{- with .Values.ingress.annotations }}
   annotations:
-{{ toYaml . | indent 4 }}
-{{- end }}
+    # linkerd annotations
+    {{- include "utils.linkerd.ingress_annotation" . | indent 4 }}
+    # other annotations
+    {{- .Values.ingress.annotations | toYaml | nindent 4 }}
```
joker-at-work, line 18: With this change, you cannot unset `.Values.ingress.annotations` anymore. This is a change to the previous behavior. Could be fine, because we have defaults for that. The chart renders unless linkerd is also enabled afaics.

PR #5603, opened 2023-11-14
[openstack/octavia] Make use of trust-bundle snippets

fwiesel:
The snippets replace the system ca trust-bundle with a centrally managed one by trust-manager in k8s. This eliminates the need for patching the images to contain the company internal CA.

Also update the job-name to include more fields to monitor for a change.

PR #5624, opened 2023-11-17
octavia: fix jobs being left over after redeployment

majewsky:
The explanation in the deleted code comment that jobs cannot be recreated was correct at the time, in commit 24c84622a6, because the job was created as a regular manifest object instead of a hook object.

As of now, the job has a `hook-delete-policy: before-hook-creation`, so job spec immutability is not a concern.

We found this issue while I was explaining the GkVulnerableImages check to Benjamin Ludwig, and we saw that dozens of old images are kept from being GC'd because these jobs are still referencing them.

#### Todos after this deployment:
- [ ] clean up existing job objects, e.g. with `k delete $(k get job -o name | grep octavia-migration-yoga-202)`.

### DIFF
--- a/openstack/octavia/templates/_helpers.tpl
+++ b/openstack/octavia/templates/_helpers.tpl
@@ -30,10 +30,3 @@ Create chart name and version as used by the chart label.
 {{- define "octavia.chart" -}}
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
-
-{{/*
-DB migration job name
-*/}}
-{{- define "octavia.migration_job_name" -}}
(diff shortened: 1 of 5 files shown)


BenjaminLudwigSAP (approved):
thx, lgtm

BenjaminLudwigSAP:
Rebased to master

BenjaminLudwigSAP:
I just noticed that #5603 also [changes the migration job name](https://github.com/sapcc/helm-charts/pull/5603/files#diff-cb9b2ab36d93ee1a2e1de468a7913093ab95981a3e91bff30c6b50a448b1d536L38-R43), in order "to include more fields to monitor for a change".
But I suppose that part of #5603 can be removed after this here PR is merged, right @majewsky?

majewsky:
yes

PR #5625, opened 2023-11-17
manila: add share_type_id to snapshot size/count metrics

majewsky:
I need these to collect the respective usage data in Limes.

Otherwise this would require extensive shenanigans to join in the respective labels from the share metrics, which would put unreasonable strain on the Prometheus server.

Note that I have not tested this; I'm just going by what is already working in the corresponding share metrics queries, since the `share_instances` table is already joined into this query anyway.


Carthaca (approved):
I directly commited a small fix

PR #5628, opened 2023-11-20
[Barbican]Add basic linkerd support

rajivmucheli:
This adds the `linkerd-support` chart as dependency and also adds the annotations to all `Deployment`, `Service` and `Ingress` objects. `Job` objects are out of the picture for now

### DIFF
--- a/openstack/barbican/templates/ingress.yaml
+++ b/openstack/barbican/templates/ingress.yaml
@@ -17,6 +17,7 @@ metadata:
     kubernetes.io/tls-acme: "true"
     disco: "true"
   {{- end }}
+  {{- include "utils.linkerd.ingress_annotation" . | indent 4 }}
 spec:
   tls:
      - secretName: tls-{{ include "barbican_api_endpoint_host_public" . | replace "." "-" }}
--- a/openstack/barbican/templates/ingress.yaml
+++ b/openstack/barbican/templates/ingress.yaml
@@ -12,8 +12,8 @@ metadata:
     system: openstack
     type: api
     component: barbican
-  {{- if .Values.tlsacme }}
   annotations:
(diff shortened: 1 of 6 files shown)


openstack/barbican/templates/ingress.yaml
```diff
@@ -17,6 +17,7 @@ metadata:
     kubernetes.io/tls-acme: "true"
     disco: "true"
   {{- end }}
+  {{- include "utils.linkerd.ingress_annotation" . | indent 4 }}
```
joker-at-work, line 20: This will create invalid YAML if `.Values.tlsacme` is not set. I think the template needs restructuring to always add the `annotations` key, i.e. move the `if` from line 15 down a line.

openstack/barbican/templates/service.yaml
```diff
@@ -16,6 +16,7 @@ metadata:
     prometheus.io/scrape: "true"
     prometheus.io/port: {{.Values.statsd.port | quote }}
     prometheus.io/targets: {{ .Values.alerts.prometheus | quote }}
+    {{- include "utils.linkerd.pod_and_service_annotation" . | indent 8 }}
```
joker-at-work, line 19: Does that indentation level look correct in the rendered chart? I'd assume it should be something like `4` instead.

joker-at-work (changes requested):

openstack/barbican/templates/service.yaml
rajivmucheli: helm-diff worked with 8, however reducing it to 4 instead.

joker-at-work (approved):
Cannot find any obvious problems, but haven't looked at the rendered chart.

PR #5629, opened 2023-11-20
[Keystone] Fix White space, update ingress annotation

rajivmucheli:


openstack/keystone/templates/service-api.yaml
```diff
@@ -16,6 +16,7 @@ metadata:
     prometheus.io/port: "{{.Values.api.metrics.port}}"
     prometheus.io/targets: "{{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus }}"
 {{- end }}
+    {{- include "utils.linkerd.pod_and_service_annotation" . | indent 4 }}
```
joker-at-work, line 19: This template needs restructuring, because there will be no `annotations` key if `.Values.api.metrics.enabled` is false. Should be fine to move the `if` in line 13 down one line.

joker-at-work (changes requested):
You also have to bump the `redis` chart to at least 1.3.5 to make it support `linkerd`.

Looks like you forgot to add the `linkerd_requested` key in the `values.yaml`.

There are 2 more templates (one `Service`, one `Deployment`) for 2fa-proxy. I'm not sure if they're talking to Keystone itself or some other service where they'd need `linkerd`-enablement.

joker-at-work (approved):
I don't see anything obvious, but I also haven't looked at the rendered chart.

PR #5634, opened 2023-11-20
[designate] Add basic linkerd support

dusandordevicsap:
This adds the linkerd-support chart as dependency and also adds the

Chart version bumped to 0.2.1


fwiesel (approved):

galkindmitrii:
please don't merge until other changes are rolled out.

PR #5637, opened 2023-11-20
[octavia] Do not enable proxysql pre-upgrade

fwiesel:
We can only do the pre-upgrade job with proxysql enabled, if proxysql has already been installed, as the configmap hasn't been created yet.


fwiesel:
Copied from [manila](https://github.com/sapcc/helm-charts/blob/master/openstack/manila/templates/migration-job.yaml)

BenjaminLudwigSAP (approved):

PR #5643, opened 2023-11-23
common/thanos: don't discover global thanos in eu-de-1

richardtief:


viennaa (approved):
Thanks, I knew it was hidden somewhere here but did not have a closer look at Tuesday 👍

PR #5646, opened 2023-11-23
split controlplane api availability metrics to prometheus-kubernetes …

artherd42:
…and prometheus-openstack

### DIFF
--- a/prometheus-rules/prometheus-openstack-rules/aggregations/haproxy.rules
+++ b/prometheus-rules/prometheus-openstack-rules/aggregations/haproxy.rules
@@ -1,0 +1,26 @@
+groups:
+- name: haproxy.rules
+  rules:
+  - record: global:api_errors_per_request_sli:ratio_rate5m
+    expr: |2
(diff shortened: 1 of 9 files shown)


prometheus-rules/prometheus-openstack-rules/aggregations/haproxy.rules
```diff
@@ -0,0 +1,26 @@
+groups:
+- name: haproxy.rules
+  rules:
+  - record: global:api_errors_per_request_sli:ratio_rate5m
```
viennaa, line 4: Why don't we take the configurable time here from the previous chart? Even though we might not need the flexibility, it would avoid the duplication and future errors in case of changes needed. (same for ingress rules above)

https://github.com/sapcc/helm-charts/blob/49dc3518b8cc0b501bb35f03b45de47302f5b48d/openstack/sre/templates/prometheus-aggregations.yaml#L33-L40

viennaa (changes requested):
Thanks for taking this item and getting the duplicated metrics resolved here.

I'd vow for having the same configurable time interval like we had before, see my inline comment.

viennaa (approved):
Thanks for updating it to the flexible time interval. Ready to ship.

PR #5649, opened 2023-11-23
adding global ruler

viennaa:


global/thanos-global/values.yaml
```diff
@@ -71,3 +71,12 @@ thanos:
   
   alerts:
     prometheus: kubernetes
+
+  ruler:
+    enabled: true
+    loglevel: warn
+    evaluationInterval: 30s
```
richardtief, line 78: I'm not sure if we want to do this that often? I would start with `60s`. 
```suggestion
    evaluationInterval: 60s
```

richardtief (changes requested):

richardtief (approved):
Thanks. Let's rule the cloud

PR #5650, opened 2023-11-23
adding global stores alert

viennaa:
Global should be able to reach 9 stores, not checking all vmware stores and ignoring special clusters. This is a sanity check and does not show, which store is missing exactly


richardtief (approved):
Looks good. Thanks

PR #5656, opened 2023-11-24
adding vmware stores and playbook links

viennaa:


richardtief (approved):
Thanks. Please bump the chart and then it's ready to go.

PR #5669, opened 2023-11-28
Add linkerd support, bump chart and supporting charts versions

dusandordevicsap:
This adds the linkerd-support chart as dependency and also adds the

Chart version bumped to 0.3.11

### DIFF
--- a/openstack/manila/Chart.yaml
+++ b/openstack/manila/Chart.yaml
@@ -35,7 +35,7 @@ dependencies:
   - name: redis
     alias: api-ratelimit-redis
     repository: https://charts.eu-de-2.cloud.sap
-    version: 1.2.1
+    version: 1.3.5
     condition: api_rate_limit.enabled
   - name: linkerd-support
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
--- a/openstack/manila/Chart.yaml
+++ b/openstack/manila/Chart.yaml
@@ -18,7 +18,7 @@ dependencies:
   - condition: memcached.enabled
     name: memcached
     repository: https://charts.eu-de-2.cloud.sap
-    version: 0.1.5
+    version: 0.2.0
   - condition: mariadb.enabled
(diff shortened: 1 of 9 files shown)


openstack/manila/Chart.yaml
```diff
@@ -9,16 +9,16 @@ maintainers:
...
   - condition: mariadb.enabled
     name: mariadb
     repository: https://charts.eu-de-2.cloud.sap
-    version: ~0.7.6
+    version: 0.8.0
   - condition: memcached.enabled
     name: memcached
     repository: https://charts.eu-de-2.cloud.sap
-    version: ~0.1.0
+    version: 0.1.5
```
joker-at-work, line 21: This needs to be `2.0.0` at least.

openstack/manila/templates/api-service.yaml
```diff
@@ -10,7 +10,9 @@ metadata:
   annotations:
     prometheus.io/scrape: "true"
     prometheus.io/targets: {{ required ".Values.alerts.prometheus.openstack" .Values.alerts.prometheus.openstack |  quote }}
+    {{- include "utils.linkerd.pod_and_service_annotation" . | indent 4 }}
   {{- include "utils.topology.service_topology_mode" . | indent 2 }}
+
```
joker-at-work, line 15: Did you add this newline on purpose?

joker-at-work (changes requested):
It looks like you missed 2 deployments:
```
templates/shares/_netapp-deployment.yaml.tpl:kind: Deployment
templates/shares/_netapp-ensure-deployment.yaml.tpl:kind: Deployment
```

openstack/manila/templates/api-service.yaml
dusandordevicsap: May be typo, I am not sure, removed it

joker-at-work (approved):
I cannot see any obvious problems, but also haven't rendered the chart.

PR #5671, opened 2023-11-28
[octavia] Add project/AZ metric labels for AZ-aware LB usage reporting

BenjaminLudwigSAP:
This PR is for AZ-aware LB usage reporting with Limes. Limes will probably want to `sum by (loadbalancer_availabilityzone)` or something and set label `loadbalancer_status!="DELETED"`.

Neutron added a dedicated metric for this in #5374, but the already existing `octavia_loadbalancers_count_gauge` should work fine, since our Grafana/Plutono dashboards all use this metric with `sum()`. The only place that has to be changed is the `OpenstackOctaviaLoadbalancerErrored` alert, see below.

---

Also:

- Fix `OpenstackOctaviaLoadbalancerErrored` alert: Since `octavia_loadbalancers_count_gauge` is labelled by status and host (and now also project and AZ), `rate()` will calculate the rate of `ERROR` LB count per status and host (and project and AZ), but we are interested in the *overall* `ERROR` LB count.

- `values.yaml`: Remove trailing whitespace

---

Belongs to CCC-34464

### DIFF
--- a/openstack/octavia/alerts/octavia.alerts
+++ b/openstack/octavia/alerts/octavia.alerts
@@ -60,7 +60,7 @@ groups:
       description: 'Failover detected for agent {{ $labels.app_kubernetes_io_name }}, AS3 target will be changed to active device.'
       summary: Openstack Octavia Metric to monitor F5 Worker Agent failover events
   - alert: OpenstackOctaviaLoadbalancerErrored
-    expr: rate(octavia_loadbalancers_count_gauge{loadbalancer_status="ERROR"}[5m]) > 0
+    expr: sum(rate(octavia_loadbalancers_count_gauge{loadbalancer_status="ERROR"}[5m])) > 0
     for: 10m
     labels:
       context: Loadbalancer go to ERROR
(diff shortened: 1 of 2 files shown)


majewsky (approved):

majewsky (changes requested):
I just noticed that usage is only half of what we need. We also need to have the capacity broken down by AZ.

We are currently computing capacity by counting octavia-worker pods using this query:

```promql
(1300 * (count(max by (label_app_kubernetes_io_name) (kube_pod_labels{label_f5_device_model="i5800"})) or vector(0)) + 500 * (count(max by (label_app_kubernetes_io_name) (kube_pod_labels{label_f5_device_model="5250v"})) or vector(0)))
```

Would it be possible to add an AZ label to these pods?

BenjaminLudwigSAP:
> Would it be possible to add an AZ label to these pods?
Yes, that should be possible. Have a look in the secrets repo in `*/values/octavia.yaml` at the list under the `workers` key. They have an `availability_zone` sub-key. It just has to be added as a pod label so that Limes can then do filter on it.
I opened #5752 for this.

majewsky (approved):
This part is good to go. Capacity is discussed in the other PR.

BenjaminLudwigSAP:
Will merge next year due to current change freeze.

PR #5675, opened 2023-11-28
[cronus] Add CORS configuration for nebula

sapmli:


openstack/cronus/templates/nebula/_config.yaml.tpl
```diff
@@ -180,6 +180,16 @@ nebula:
   {{- end }}
 {{- end }}
 {{- end }}
-
+{{- if .Values.nebula.cors.enabled }}
+  cors:
+    enabled: {{ .Values.nebula.cors.enabled }}
+    allowedOrigins:
+{{- range $k, $v := .Values.nebula.cors.allowedOrigins }}
```
corey-aloia, line 187: Just a side note: I noticed that we do this in a few places, but it would potentially be easier to just do:
```
{{- .Values.nebula.cors.allowedOrigins | toYaml | indent 6 }}
```

openstack/cronus/templates/nebula/_config.yaml.tpl
```diff
@@ -180,6 +180,16 @@ nebula:
   {{- end }}
 {{- end }}
 {{- end }}
-
+{{- if .Values.nebula.cors }}
```
corey-aloia, line 183: I see you just changed :) another option would have been to update the ci/test-values.yaml. Not sure why the CI doesn't use the charts values files 🤔

corey-aloia (approved):

openstack/cronus/templates/nebula/_config.yaml.tpl
sapmli: Ah yea. I wondered why my values entries didn't work. :-)

openstack/cronus/templates/nebula/_config.yaml.tpl
corey-aloia: made a mistake in my comment... it should be nindent and not indent

PR #5676, opened 2023-11-28
nova: Rename config vars for CPU reservation

grandchild:
The upcoming CPU reservation change in nova renames these two config vars. The old name should be removed after the rollout.


grandchild:
See also: https://github.com/sapcc/nova/pull/460

PR #5702, opened 2023-12-01
[RabbitMQ]add custom configuration file

bashar-alkhateeb:
Adding a mechanism to add custom configuration of RabbitMQ, the configuration will be set under `/etc/rabbitmq/conf.d/20-custom.conf` in RabbitMQ pod, that if any key is set under `.Values.customConfig` .

Configuration can be set as [key, value] pairs under `.Values.customConfig`

```
Example:
customConfig:
  disk_free_limit.absolute: 50MB
```

Any key under` .Values.customConfig` with null value will be ignored in the config file.

For more information please find:
https://www.rabbitmq.com/configure.html#advanced-config-file


common/rabbitmq/CHANGELOG.md
```diff
@@ -3,13 +3,40 @@ Rabbitmq CHANGELOG
 
 This file is used to list changes made in each version of the common chart rabbitmq.
 
+
+0.6.2
+-----
+b.alkhateeb@sap.com
+- new mechanism to add customs configuration file to RabbitMQ pods, under /etc/rabbitmq/conf.d/20-custom.conf
```
joker-at-work, line 10: typo: "customs" -> "custom"

common/rabbitmq/CHANGELOG.md
```diff
@@ -3,13 +3,40 @@ Rabbitmq CHANGELOG
...
+  custom configuration can be added as `key, value` pairs under .Values.customConfig
+  ```
+  customConfig:
+    key1: value1
+    key2: value2
+  ```
```
joker-at-work, line 16: Imho missing the important "keys with a null value are ignored", because that means I cannot unset a value via config, right?

common/rabbitmq/templates/deployment.yaml
```diff
@@ -133,4 +138,9 @@ spec:
         - name: container-init
           configMap:
             name: {{ template "fullname" . }}-bin
+      {{- if $.Values.customConfig }}
```
joker-at-work, line 141: We don't seem to use `$.Values` but only `.Values` in the other parts of the file - any reason to change that here in the `if`?

common/rabbitmq/templates/deployment.yaml
```diff
@@ -133,4 +138,9 @@ spec:
...
+        - name: rabbitmq-custom-config
```
joker-at-work, line 142: Why is it named `rabbitmq-custom-config` and not `rabbitmq-etc` like the configmap? It might (at some point) contain more than the custom-config, right? And if not, why not name the configmap `…-custom-config`?

common/rabbitmq/templates/etc/_rabbitmq-custom-config.tpl
```diff
@@ -0,0 +1,15 @@
...
+## RabbitMQ custom config
+## Generated by helm template etc/_rabbitmq-custom-config.tpl
+## For more information please find helm chart
+##     https://github.com/sapcc/helm-charts/tree/master/common/rabbitmq
+
+## For more information about RabbitMQ configuration please find
+##     https://www.rabbitmq.com/configure.html
+
+{{- range $k, $v := .Values.customConfig }}
+{{ if $v }}
```
joker-at-work, line 11: This also disallows me from setting anything to `false` or 0, right? From reading the provided link on how to configure rabbitmq and it does not look like the config can handle a string `"false"`, too.

common/rabbitmq/CHANGELOG.md
bashar-alkhateeb: Added!

common/rabbitmq/templates/deployment.yaml
bashar-alkhateeb: you are right actually in this context is doesn't make any difference to use .`Value` or` $.Value`
changed it to `.Value` to keep consistent

common/rabbitmq/templates/deployment.yaml
bashar-alkhateeb: it was because i thought it is butter to mount the hole conf.d but then changed it to only a mounted file instead, that to not override default config files 
naming schema set to ! rabbitmq-custom-config

common/rabbitmq/templates/etc/_rabbitmq-custom-config.tpl
bashar-alkhateeb: Correct fixed only `nil` values will be ignored 
fixed!

common/rabbitmq/templates/etc/_rabbitmq-custom-config.tpl
```diff
@@ -0,0 +1,15 @@
...
+## RabbitMQ custom config
+## Generated by helm template etc/_rabbitmq-custom-config.tpl
+## For more information please find helm chart
+##     https://github.com/sapcc/helm-charts/tree/master/common/rabbitmq
+
+## For more information about RabbitMQ configuration please find
+##     https://www.rabbitmq.com/configure.html
+
+{{- range $k, $v := .Values.customConfig }}
+{{ if not (eq $v nil) }}
```
fwiesel, line 11: I understand to test for that is [described here](https://github.com/Masterminds/sprig/issues/53#issuecomment-483414063)
```gotpl
{{ if kindIs "invalid" $value }}
```

fwiesel, line 11: Personally, I am more a fan of using everywhere 
``{{- `` and only in the places you really want a newline to leave the `-` away.
That serves two purposes:
1. It is consistent and easily to remember
2. You can indent the loops and conditions logically.
I.e.
```
{{- range $k, $v := .Values.customConfig }}
  {{- if kindIs "invalid" $value }}
{{ $k }} = {{ $v }}
  {{- end }}
{{- end }}

No a requirement from me to change, but just a suggestion.

```

joker-at-work: Th `kindIs` needs a "not", right?

bashar-alkhateeb: Done!

fwiesel (approved):

joker-at-work (commented):
The name of the template-file of the configmap does not match its name (`etc-configmap.yaml` vs. `{{ template "fullname" . }}-custom-conf`).

sapmli (approved):
Thanks, Bashar!

bashar-alkhateeb:
> The name of the template-file of the configmap does not match its name (`etc-configmap.yaml` vs. `{{ template "fullname" . }}-custom-conf`).

file name changed as requested 
`etc-configmap.yaml ->  custom-conf-configmap.yaml`

joker-at-work (approved):

PR #5705, opened 2023-12-01
adding filter option

viennaa:
Specific urls can no explicitly be ignored and will not be added as targets

### DIFF
--- a/common/thanos/Chart.yaml
+++ b/common/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: Deploy Thanos via operator
 type: application
-version: 0.6.7
+version: 0.6.8
 appVersion: v0.31.0
 maintainers:
   - name: Tommy Sauer (viennaa)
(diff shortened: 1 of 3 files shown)


richardtief:
I don't see why this is necessary, as we define the query stores manually anyway. e.g. https://github.com/sapcc/helm-charts/blob/ed1f5d9dc8226113b55a6e65936d996c22266987/global/thanos-global/values.yaml#L46-L47

viennaa:
If we exclude the whole region, we would miss out on the stores which are deployed there.

qa-de-2 example:
available:
* thanos-infra-collector-grpc.qa-de-2
* thanos-kubernetes-metal-grpc.qa-de-2.cloud.sap
* thanos-kubernetes-virtual-grpc.qa-de-2.cloud.sap
* thanos-openstack-grpc.qa-de-2.cloud.sap

not available:
* thanos-storage-grpc.qa-de-2.cloud.sap
* thanos-kubernetes-kubernikus-grpc.qa-de-2.cloud.sap
* thanos-kubernetes-admin-grpc.qa-de-2.cloud.sap
* thanos-scaleout-grpc.qa-de-2.cloud.sap

This is raising the GRPC error rate, since they can never be contacted.

viennaa:
Good idea, #5709

PR #5707, opened 2023-12-01
[vmware-collector] add hostsystem metric cpu|corecount_provisioned

ninePebbles:


viennaa (changes requested):
Thanks for proactively providing this addition. Next time consider adding one of the [maintainers](https://github.com/sapcc/helm-charts/blob/master/system/vmware-monitoring/values.yaml#L15-L17) to speed up the approve/merge process.

Please increase the patch version of the [Chart.yaml](https://github.com/sapcc/helm-charts/blob/master/system/vmware-monitoring/Chart.yaml#L3) and we're good to go.

PR #5718, opened 2023-12-04
Neutron default rate limits

sven-rosenzweig:
Limits are calculated based on the maximal number of requests per URI (e.g. ports, ports/{some-id}) and action (e.g. create, delete) in the time window of 4 weeks.
Additionally a 50% suprlus is added to the limit as buffer.
    
 Corresponding Prometheus query
 ```
max_over_time(sum by (target_type_uri, action) (rate(openstack_watcher_api_requests_total{service="network",container="statsd",service="network", target_type_uri=~".*",status=~".*", action=~".*", container="statsd"}[5m]) * 60 )[4w:1m])
```
    
 To keep the limits administrable, limits of smaller regions (bronze and silver) are grouped together as defaults since the number of neutron-server replicas is more or less the same. For the gold regions it is necessary to set custom limits since the request volume is much higher.

### DIFF
--- a/openstack/neutron/templates/etc/_ratelimit.yaml.tpl
+++ b/openstack/neutron/templates/etc/_ratelimit.yaml.tpl
@@ -13,36 +13,21 @@ whitelist_users:
 {{- if .Values.rate_limit.blacklist }}
 #Blocked Projects List
 blacklist:
-{{ .Values.rate_limit.blacklist | toYaml | indent 3 }}
+{{ .Values.rate_limit.blacklist | toYaml | indent 2 }}
 {{- end }}
 
 {{- if .Values.rate_limit.blacklist_users }}
 #Blocked Users List
 blacklist_users:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: openstack/neutron/templates/etc/_ratelimit.yaml.tpl [hunk 1]
```diff
@@ -37,12 +37,6 @@ groups:
...
-rates:
-{{- range $level_k, $level := .Values.rate_limit.rates }}
-  {{ $level_k }}:
-        {{- range $target_uri_type, $target_uri_type_limits := $level }}
-    {{ $target_uri_type }}:
-{{ $target_uri_type_limits | toYaml | indent 6 }}
-        {{- end }}
-{{- end }}
+rates: 
+{{ .Values.rate_limit.rates | toYaml | indent 3 }}
```
swagner-de, line 41: Why do you use indent 3 here? This is a key on the second indention level. so I think `indent 2` would be fine.

swagner-de (approved):

PR #5727, opened 2023-12-04
[octavia] Bump utils dependency for proxysql update

fwiesel:


BenjaminLudwigSAP (approved):

BenjaminLudwigSAP:
Thanks, I will roll it out.

PR #5728, opened 2023-12-04
updating prometheus operator and crds

viennaa:
updating prometheus-operator
updating kube-prometheus-stack
updating crd dependend charts


richardtief (approved):
Thanks

PR #5730, opened 2023-12-04
[nova] Bump utils dependency for proxysql update

fwiesel:


joker-at-work (changes requested):
There are conflicts against the `master` branch, because this contains Placement in addition to Nova and we already did this for Nova in #5719

fwiesel:
Sorry, I thought I missed that one.

PR #5732, opened 2023-12-05
adding cronus toleration for logging

corey-aloia:
It looks like the cronus nodes in the scaleout clusters have a taint `dedicated=cronus`, causing no fluent bit on the nodes and not logging being shipped


Kuckkuck (approved):

corey-aloia:
Hi @Kuckkuck , thanks for approving, but I actually had an error with how I wrote it :) For some reason I put "cronus" as the operator..... I just fixed and pushed

PR #5752, opened 2023-12-08
[Octavia] Add availability zone to worker pod labels

BenjaminLudwigSAP:
When the worker does not have an availability zone set, the label will be empty.
@majewsky please keep this in mind and check whether this behavior is okay.

### DIFF
--- a/openstack/octavia/templates/octavia-worker-deployment.yaml
+++ b/openstack/octavia/templates/octavia-worker-deployment.yaml
@@ -21,7 +21,9 @@ spec:
       labels:
         app.kubernetes.io/name: octavia-worker-{{ $name }}
         app.kubernetes.io/instance: {{ .Release.Name }}
-        f5-device-model: {{ $v.model }} # used by Limes for capacity discovery (via Prometheus metric "kube_pod_labels")
+        # Labels used by Limes for capacity discovery (via Prometheus metric "kube_pod_labels")
+        f5-device-model: {{ $v.model }}
+        availability-zone: {{ $v.availability_zone }}
       annotations:
         configmap-etc-hash: {{ include (print $.Template.BasePath "/octavia-etc-configmap.yaml") . | sha256sum }}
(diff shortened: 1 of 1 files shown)


openstack/octavia/templates/octavia-worker-deployment.yaml
```diff
@@ -21,7 +21,9 @@ spec:
       labels:
         app.kubernetes.io/name: octavia-worker-{{ $name }}
         app.kubernetes.io/instance: {{ .Release.Name }}
-        f5-device-model: {{ $v.model }} # used by Limes for capacity discovery (via Prometheus metric "kube_pod_labels")
+        # Labels used by Limes for capacity discovery (via Prometheus metric "kube_pod_labels")
+        f5-device-model: {{ $v.model }}
+        availability-zone: {{ $v.availability_zone }}
```
majewsky, line 26: I would like for this label to always have a value. What does it mean for a worker to not be attached to a specific AZ? If it means that we actually do not know which AZ it belongs to, I would like to have:
```go
        availability-zone: {{ $v.availability_zone | default "unknown" }}
```
If it means that it is used only for non-AZ-aware loadbalancers, I would like to have:
```go
        availability-zone: {{ $v.availability_zone | default "any" }}
```
Ref: Enum in <https://pkg.go.dev/github.com/sapcc/go-api-declarations/limes#AvailabilityZone>

majewsky (changes requested):

openstack/octavia/templates/octavia-worker-deployment.yaml
BenjaminLudwigSAP: Every worker is assigned to one and only one F5 pair. Every F5 pair is either assigned to an AZ or cross-AZ. Therefore I will go with the "any" option.

majewsky (approved):

BenjaminLudwigSAP:
Will merge next year due to current change freeze.

PR #5790, opened 2023-12-13
[unbound] -- restructured the image/image_tag values

vssldmtrv:
use the "latest" tag as a placeholder
use proper production tags in your secrets

set the image pull policy to "Always"
"IfNotPresent" does not work reliably with the "latest" tag


mutax (approved):
lgtm

PR #5799, opened 2023-12-14
chore(elastiflow): removing all artifacts & references

IvoGoman:
- remove `system/elastiflow`
- remove `prometheus-exporters/elastiflow-query-exporter`
- remove Gatekeeper exception (#5801)
- remove Prometheus Global federation
- remove settings for ELK/OpenSearch Manager


majewsky (approved):
Gatekeeper exception removal was moved to #5801.

viennaa (approved):
Global prometheus-infra is turned off, we can remove the reference to elastiflow, but it will not be deployed :)

richardtief (approved):

PR #5801, opened 2023-12-14
chore(gatekeeper): remove elastiflow exceptions

IvoGoman:


majewsky (approved):
Confirmed that Elastiflow has been undeployed in all prod clusters.

PR #5804, opened 2023-12-14
[cinder] Bump utils dependency for proxysql update

fwiesel:


fwiesel:
This becomes obsolete on merging #5594

joker-at-work (approved):

PR #5805, opened 2023-12-14
Fix migration job name, move kubernetes-entrypoint logic to init-container, plus cleanup

fwiesel:
Please look at the individual commits for detail.

### DIFF
--- a/openstack/cinder/templates/vcenter_datacenter_cinder_deployment.yaml
+++ b/openstack/cinder/templates/vcenter_datacenter_cinder_deployment.yaml
@@ -49,6 +49,8 @@ template: |
         hostname: cinder-volume-vmware-{= name =}
         {{- tuple . "{= availability_zone =}" | include "utils.kubernetes_pod_az_affinity" | nindent 8 }}
         {{- include "utils.proxysql.pod_settings" . | nindent 8 }}
+        initContainers:
+        {{- tuple . (dict "service" (print .Release.Name "-mariadb," .Release.Name "-rabbitmq")) | include "utils.snippets.kubernetes_entrypoint_init_container" | indent 8 }}
         containers:
         - name: cinder-volume-vmware-{= name =}
           image: {{required ".Values.global.registry is missing" .Values.global.registry}}/loci-cinder:{{.Values.imageVersionCinderScheduler | default .Values.imageVersion | required "Please set cinder.imageVersion or similar" }}
@@ -58,12 +60,8 @@ template: |
               add: ["SYS_ADMIN"]
           command:
           - dumb-init
-          - kubernetes-entrypoint
+          - cinder-volume
           env:
-          - name: COMMAND
-            value: 'cinder-volume'
-          - name: NAMESPACE
-            value: {= namespace =}
           {{- if .Values.sentry.enabled }}
           - name: SENTRY_DSN
             valueFrom:
--- a/openstack/cinder/templates/vcenter_datacenter_cinder_deployment.yaml
+++ b/openstack/cinder/templates/vcenter_datacenter_cinder_deployment.yaml
@@ -73,8 +73,6 @@ template: |
           {{- end }}
           - name: PYTHONWARNINGS
             value: 'ignore:Unverified HTTPS request'
-          - name: PGAPPNAME
-            value: cinder-volume-vmware-{= name =}
           livenessProbe:
             exec:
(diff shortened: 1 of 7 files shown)


openstack/cinder/templates/vcenter_datacenter_cinder_deployment.yaml
```diff
@@ -49,6 +49,8 @@ template: |
         hostname: cinder-volume-vmware-{= name =}
         {{- tuple . "{= availability_zone =}" | include "utils.kubernetes_pod_az_affinity" | nindent 8 }}
         {{- include "utils.proxysql.pod_settings" . | nindent 8 }}
+        initContainers:
+        {{- tuple . (dict "service" (print "cinder-api," .Release.Name "-rabbitmq")) | include "utils.snippets.kubernetes_entrypoint_init_container" | indent 8 }}
```
joker-at-work, line 53: I'm not sure about this one. I looked at the `nova-compute` template and we don't have any dependencies defined. It was the same here before (no env to configure `kubernetes-entrypoint`, so no dependencies).

Why would `cinder-volume` depend on `cinder-api`? I think this should be `cinder-mariadb` instead?

Do we need to wait for the migration job to finish before starting up?

joker-at-work (commented):
Should remove the `postgresql` key from the `values.yaml`, too? It's also unused.

openstack/cinder/templates/vcenter_datacenter_cinder_deployment.yaml
fwiesel: > Why would cinder-volume depend on cinder-api? I think this should be cinder-mariadb instead?

I think, my idea was that `cinder-volume` talks to `cinder-api`  (via rabbitmq). It does not access the db directly, AFAIK.

> Do we need to wait for the migration job to finish before starting up?

Probably, I'll add it.

fwiesel: Okay, it gets complicated... We probably do want to wait on a major release upgrade, but not just because some random attribute has changed.

Say, we change add a newline in the script, that triggers the creation of a new job, because that is an immutable attribute. But I don't think we want to replace the job right then.

I would say, it would make sense to revisit the migration jobs, but I would like to consider that a different problem then the one I am addressing here.

joker-at-work: `cinder-volume` does have a `proxysql` container in it and has the DB configure in `cinder.conf`. I also found none of the `remotable` decorators Nova has in the code.

fwiesel: Ah, then we definitely want to wait on the db. I'll add it.

joker-at-work: > I would say, it would make sense to revisit the migration jobs, but I would like to consider that a different problem then the one I am addressing here.

follow-up somewhen later sounds fine. we didn't make things worse here, as there was no waiting done before anyways.

fwiesel:
> Should remove the postgresql key from the values.yaml, too? It's also unused.

I've removed it.

openstack/cinder/templates/vcenter_datacenter_cinder_deployment.yaml
```diff
@@ -49,6 +49,8 @@ template: |
         hostname: cinder-volume-vmware-{= name =}
         {{- tuple . "{= availability_zone =}" | include "utils.kubernetes_pod_az_affinity" | nindent 8 }}
         {{- include "utils.proxysql.pod_settings" . | nindent 8 }}
+        initContainers:
+        {{- tuple . (dict "service" (print "cinder-api," .Release.Name "-mariadb," .Release.Name "-rabbitmq")) | include "utils.snippets.kubernetes_entrypoint_init_container" | indent 8 }}
```
joker-at-work, line 53: Now that we have established, that `cinder-volume` talks directly to the DB, why do we still wait for `cinder-api`? This makes us effectively wait for the migration job, doesn't it?

fwiesel: Yes, probably it doesn't make sense that way around, since `cinder-api` calls the `cinder-volume` and not the other way around.

> This makes us effectively wait for the migration job, doesn't it?

Yes, but not a specific migration job with a name that changes on every attribute change, but just a migration job. Either way, it is gone now.

joker-at-work (approved):

PR #5806, opened 2023-12-14
[nova] Adding VPA support

dusandordevicsap:
This adds optional VPA support to Nova. VPA support can be enabled/disabled through values.yaml file.

More details about VPA here https://operations.global.cloud.sap/docs/devops/kubernetes/resource-mgmt/


joker-at-work (commented):
Could you change the commit-message to contain some information on what we do here and why, please?

Imho, "Adding VPA support" sounds like it doesn't work without this, ,but that's not true. I'd rather have the commit-message say "Optionally set main-container for vpa-butler" or similar.

joker-at-work (approved):

PR #5815, opened 2023-12-18
alert for misbehaving Thanos compactor

viennaa:


common/thanos/templates/alerts/_thanos-compactor.alerts.tpl
```diff
@@ -109,3 +109,17 @@ groups:
           Thanos Compact `{{`{{ $labels.thanos }}`}}` has disappeared.
           Prometheus target for the component cannot be discovered.
         summary: Thanos component has disappeared.
+
+    - alert: ThanosCompactSpawningManyPods 
+      expr: count by (region, namespace) (kube_pod_info{pod=~"thanos-.*-compactor-*.+"}) > 100
```
richardtief, line 114: We wait until there are over 100 pods? That seems like a high threshold to me

viennaa: It is piling up quiet fast, every 5mins a new pod per compactor. We need to keep the vmware-compactors in mind. 

When the pods are cycled, we're going from 33 to 66 in the biggest region. Giving it one additional iteration of slack should avoid accidentally triggering and provide headroom for growth.

It is also just to be aware, that we are running into problems, usually we see way later than with 100 pods :).

richardtief: Okay, I agree. 

Since it is the common chart, the alert will fire multiple times because it is deployed with more than one release. Hence, the condition still needs to be bound to the release.

viennaa: Good remark, pushing

richardtief (approved):
Thanks!

PR #5818, opened 2023-12-19
refining new compactor alert to be more precise

viennaa:


richardtief (approved):
Nice. Thanks for improving.

PR #5828, opened 2023-12-20
[designate-nanny] moving user seeds to designate

mutax:
To enable seeding in all namespaces necessary, user seeding has moved to the designate helm chart, to continue having a single source of truth the nanny charts now read the designate values as well.

Enabling the nanny and the users used is now configured in value files of designate centrally.


openstack/designate/templates/seeds.yaml
```diff
@@ -938,3 +938,42 @@ spec:
       role: cloud_dns_admin
       inherited: true
 {{- end }}
+
+{{- if .Values.nanny_enabled }}
+---
+apiVersion: "openstack.stable.sap.cc/v1"
+kind: "OpenstackSeed"
+metadata:
+  name: designate-nanny-seed
```
galkindmitrii, line 947: would you like to replace designate here with {{ .Release.Name }} ?
In that case seeds for global case will be called different, automatically.

openstack/designate/values.yaml
```diff
@@ -7,6 +7,7 @@ debug: "True"
 
 global_setup: false
 tempest_enabled: false
+nanny_enabled: false
```
galkindmitrii, line 10: since this is coming to all region, I think it is fine to leave it to true per default.

galkindmitrii (approved):

openstack/designate/values.yaml
mutax: I just aligned it with tempest and had the idea to have a "safe" default in the helm charts to be overwritten where we need it. in global for example we don't have it in all regions and when its true by default there will be missing configuration.

openstack/designate/templates/seeds.yaml
mutax: oh, makes sense. the name then will be designate-global-nanny which is not in sync with the general naming scheme of designate-nanny-global.  But will change it, makes sense!

galkindmitrii (approved):

PR #5830, opened 2023-12-20
[designate-nanny] using makro for keystone url at proper location

mutax:
cannot use the makro in values.yaml, reverting for now and setting explicitly in secrets.


galkindmitrii (approved):

mutax:
this version now compiles locally and should get the auth url automatically.

galkindmitrii (approved):

PR #5836, opened 2023-12-22
[openstack/utils] Set SSL_CERT_FILE specifically

fwiesel:
The defaults seem to differ between implementations, so we rather want to set it explicitly to make sure it is picked up as expected.


rajivmucheli (approved):
lgtm

PR #5838, opened 2023-12-22
[glance] Fix trust-bundle

fwiesel:
- Enable trust-bundle explicitly (as the default is false).
- Set env variable to ensure the cert bundle is used.


rajivmucheli (approved):
thanks a lot!

PR #5846, opened 2024-01-03
common/prometheus: reduce memory request

richardtief:
Since the new version with Go tag `stringlabels` is deployed everywhere, we can reduce the default memory request to 4Gi.


viennaa:
Superseeded by #5847

PR #5847, opened 2024-01-03
setting vpa defaults

viennaa:
big regions need to be adapted, if they need 10GB memory or more for their Prometheus


richardtief (approved):
Looks good. Thanks!

PR #5851, opened 2024-01-04
moving alert to regional and version bump

viennaa:
alert needs to move to respective regional thanos. since it is pushed to every thanos it is missing the kube_pod_info. We want it centrally only

Using the opportunity to upgrade.


richardtief (approved):
Looks all right. Thanks

PR #5853, opened 2024-01-04
update latest common

viennaa:


openstack/maia/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Expose Prometheus as multi-tenant OpenStack service
 name: maia
-version: 1.4.14
+version: 1.4.15
```
richardtief, line 4: The script is too ambitious.

richardtief (approved):
The charts have been bumped up twice. Looks like the script does it once each for Thanos and Prometheus.

system/vmware-monitoring/Chart.yaml
```diff
@@ -1,14 +1,14 @@
 apiVersion: v2
 name: vmware-monitoring
-version: 1.5.6
+version: 1.5.8
```
richardtief, line 3: Double bump.

viennaa: all addressed

richardtief (approved):

PR #5855, opened 2024-01-04
common/thanos: set default user to 0 (root)

richardtief:
The upstream Thanos image has changed the default user in the container image to 1001 instead of 0 (root).

https://github.com/thanos-io/thanos/pull/6107

As we are dependent on user 0 in our environment, we explicitly set it to 0.


viennaa (approved):
Thanks, security wise not great, but the existing files are root owned, so we go for it.

PR #5864, opened 2024-01-08
[plutono] Add locust timescaledb password variable

misamoylov:


system/plutono/templates/_plutono-start.tpl
```diff
@@ -73,6 +73,7 @@ function start_application {
   sed -i 's,__OPENSEARCH_PASSWORD__,{{.Values.authentication.opensearch_password}},g' /var/lib/plutono/provisioning/datasources/*
   sed -i 's,__ALERTMANAGER_PASSWORD__,{{.Values.alertmanager.password}},g' /var/lib/plutono/provisioning/datasources/*
   sed -i 's,__PROMETHEUS_REGION__,{{.Values.global.region}},g' /var/lib/plutono/provisioning/datasources/*
+  sed -i 's,__LOCUST_TIMESCALEDB_PASSWORD__,{{.Values.global.region}},g' /var/lib/plutono/provisioning/datasources/*
```
richardtief, line 76: copy paste error
```suggestion
  sed -i 's,__LOCUST_TIMESCALEDB_PASSWORD__,{{.Values.authentication.locust_timescaledb_password}},g' /var/lib/plutono/provisioning/datasources/*
```

richardtief (changes requested):
Looks good. Just a small issue. And could you please bump the chart version?

system/plutono/templates/plutono-secrets.yaml
```diff
@@ -12,4 +12,5 @@ data:
   elk_password: {{ required ".Values.authentication.elk_password missing" .Values.authentication.elk_password | b64enc | quote }}
   opensearch_password: {{ required ".Values.authentication.opensearch_password missing" .Values.authentication.opensearch_password | b64enc | quote }}
   metisdb_password: {{ required ".Values.authentication.metisdb_password missing" .Values.authentication.metisdb_password | b64enc | quote }}
+  locust_timescaledb_password: {{ required ".Values.authentication.locust_timescaledb_password missing" .Values.authentication.locust_timescaledb_password | b64enc | quote }}
```
richardtief, line 15: Required means the value is set in all Plutono regions? I guess that won't work since you only need it in `eu-nl-1`.

richardtief (changes requested):

system/plutono/templates/plutono-secrets.yaml
richardtief: We can consider default empty string.
```suggestion
  locust_timescaledb_password: {{ default "" .Values.authentication.locust_timescaledb_password | b64enc | quote }}
```

richardtief (approved):
Thanks!

PR #5867, opened 2024-01-08
vpa recommendation exceeded alert

viennaa:


common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
```diff
@@ -376,3 +376,20 @@ groups:
         Prometheus `{{`{{ $labels.prometheus }}`}}` remote write is behind for `{{`{{ $labels.url }}`}}`.
       summary: Prometheus remote write is behind.
   {{- end }}
+  {{- if eq $root.Values.vpaUpdateMode "Auto" }}
+  - alert: PrometheusVPARecommendsVPAMaxAllowedExceeded
+    expr: vpa_butler_vpa_container_recommendation_excess{verticalpodautoscaler=~"prometheus.+"} / 1024 / 1024 / 1024 > 0
```
richardtief, line 381: As this is a common chart, we need to tie the alert to the release in question to avoid receiving multiple unrelated alerts.

common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
```diff
@@ -376,3 +376,20 @@ groups:
...
+    for: 15m
+    labels:
+      service: {{ default "metrics" $root.Values.alerts.service }}
+      support_group: {{ default "observability" $root.Values.alerts.support_group }}
+      severity: info
+      meta: Prometheus VPA for `{{`{{ $labels.prometheus }}`}}` in `{{`{{ $labels.prometheus }}`}}` is recommending more resources.
+      playbook: docs/support/playbook/prometheus/remote_write
```
richardtief, line 388: wrong playbook.

common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
```diff
@@ -376,3 +376,20 @@ groups:
...
+    annotations:
+      description: |
+        Prometheus `{{`{{ $labels.prometheus }}`}}` in `{{`{{ $labels.cluster }}/{{ $labels.namespace }}`}}`  needs more `{{`{{ $labels.resource}}`}}`.
+        It is hitting the VPA maxAllowed boundry and is not ensured to run properly at its current place. Consider upgrading the VPA maxAllowed value if
+        the cluster permits.
```
richardtief, line 393:
```suggestion
        Prometheus `{{`{{ $labels.prometheus }}`}}` in `{{`{{ $labels.cluster }}/{{ $labels.namespace }}`}}` needs more `{{`{{ $labels.resource}}`}}`.
        It is hitting the VPA maxAllowed boundary and is not ensured to run properly at its current place. Consider upgrading the VPA maxAllowed value if
        the cluster permits.
```

richardtief (changes requested):
Thanks! Some small issues.

common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
viennaa: Can you elaborate a little more? We are only configuring the alert if it is set to `Auto`, which is not the case yet. When we're enabling respective Prometheis the set to `Auto` will appear at the same time as the version bump.

richardtief: I see. 

Three things, imagine we roll this out with e.g. prometheus-openstack
1. the metric `vpa_butler_vpa_container_recommendation_excess` does not exist. Leads to an absent-metric alert
2. if this is depoyed with the corresponding Thanos Rule, multiple prometheus are in scope - not just prometheus-openstack. In other words: We get an alert deployed with prometheus-openstack also for other prometheus.
3. the metric also has `resource="cpu", unit="core"` labels - so this does not only apply to memory

viennaa: 1. and 2.: 
You're right about the thanos rule, since prometheus-kubernetes was the testing horse I haven't had this in mind.
We got to move this accordingly.

3. Yes, the calculation makes no sense in this case. It was just for better readability. Initially I thought of handling this in one alert, but the GiB conversion makes no sense then. We can split it in two, and have better readable Memory and CPU values.

viennaa:
@richardtief moved to thanos-regional

richardtief:
The alerts should not be linked to the regional Thanos. This topic should be part of the common chart. Since we already have that, we simply use the logic of multiple pod scrape alerts to propagate the vpa alerts to Thanos Rule when the metric lives in prometheus-kubernetes. In addition, we can use all the templating we already have to equip the alerts. Changes pushed. Now we need to turn the review around.

common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
```diff
@@ -249,6 +249,40 @@ groups:
...
+      severity: info
+      meta: Prometheus VPA for `{{`{{ $labels.verticalpodautoscaler }}`}}` in `{{`{{ $labels.namespace }}`}}` is recommending more CPUs.
+    annotations:
+      description: |
+        `{{`{{ $labels.verticalpodautoscaler }}`}}` in `{{`{{ $labels.cluster }}/{{ $labels.namespace }}`}}` needs more `{{`{{ $labels.resource }}`}}`. It is overutilized by `{{`{{ $value }}`}}` cores.
+        It is hitting the VPA maxAllowed boundary and is not ensured to run properly at its current place. Consider upgrading the VPA maxAllowed
+        CPU core value if the host memory size permits.
+      summary: Prometheus needs more CPU.
+  {{- end }}
+
```
viennaa: Why is this needed? It would only work in the kubernetes case.
Since we leverage the prometheus scoped thanos rule, it can be removed, no?

richardtief: The entire common chart is built on the assumption that prometheus-kubernetes is the default. If this is not the case, we need to forward the alerts to the Thanos rule. This switch is here:
```
{{- if and (eq $root.Values.vpaUpdateMode "Auto") (not $root.Values.alerts.thanos.enabled) }}
```
ans the values:
https://github.com/sapcc/helm-charts/blob/ed1f5d9dc8226113b55a6e65936d996c22266987/common/prometheus-server/values.yaml#L417-L420

richardtief: To label them with the specified `thanos-ruler`, there is another PrometheusRule in that [file](https://github.com/sapcc/helm-charts/blob/ed1f5d9dc8226113b55a6e65936d996c22266987/common/prometheus-server/templates/alerts/prometheus-alerts.yaml).

common/prometheus-server/templates/alerts/_prometheus-thanos-rule.alerts.tpl
```diff
@@ -34,3 +34,37 @@ groups:
       description: Prometheus is scraping `{{`{{ $labels.pod }}`}}` pods in namespace `{{`{{ $labels.namespace }}`}}` multiple times. This is likely caused due to incorrectly placed scrape annotations. <https://{{ include "prometheus.externalURL" . }}/graph?g0.expr={{ urlquery `up * on(instance) group_left() (sum by(instance) (up{kubernetes_pod_name=~"PLACEHOLDER.*"}) > 1)` | replace "PLACEHOLDER" "{{ $labels.pod }}"}}|Affected targets>
       summary: Prometheus scrapes pods multiple times
   {{- end }}
+  {{- if and (eq $root.Values.vpaUpdateMode "Auto") $root.Values.alerts.thanos.enabled }}
```
viennaa, line 37:
```suggestion
  {{- if and (eq $root.Values.vpaUpdateMode "Auto") $root.Values.alerts.thanos.enabled }}
  {{/* This is covering all Prometheis but kubernetes. They don't have the vpa_butler metric and need to leverage thanos ruler instead. */}}
```

common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
```diff
@@ -249,6 +249,40 @@ groups:
       summary: Prometheus scrapes pods multiple times
   {{- end }}
 
+  {{- if and (eq $root.Values.vpaUpdateMode "Auto") (not $root.Values.alerts.thanos.enabled) }}
```
viennaa, line 252:
```suggestion
  {{- if and (eq $root.Values.vpaUpdateMode "Auto") (not $root.Values.alerts.thanos.enabled) }}
  {{/* Only affecting all prometheus-kubernetes and using Thanos alerts directive to distinguish */}}
```

common/prometheus-server/templates/alerts/_prometheus.alerts.tpl
viennaa: Thanks for elaborating. Suggesting two comments to remember future me of this nexus.

richardtief (approved):

PR #5870, opened 2024-01-08
[limes] Make rabbitmq user configurable

fwiesel:
We need to change user with credentials to enable a interruption free deployment.
First, roll-out new user with new password, then remove old user.

### DIFF
--- a/openstack/limes/templates/_utils.tpl
+++ b/openstack/limes/templates/_utils.tpl
@@ -20,7 +20,7 @@
       name: limes-secret
       key: rabbitmq_password
 - name: LIMES_AUDIT_RABBITMQ_USERNAME
-  value: "rabbitmq"
+  value: "{{ default "rabbitmq" .Values.limes.passwords.ccloud.rabbitmq_user }}"
 {{- end }}
 - name: LIMES_AUTHORITATIVE
   value: "true"


openstack/limes/templates/secret.yaml
```diff
@@ -6,5 +6,7 @@ data:
   os_password: {{ required ".Values.limes.passwords.ccloud.os_password" .Values.limes.passwords.ccloud.os_password | b64enc }}
   postgres_password: {{ required ".Values.postgresql.postgresPassword" .Values.postgresql.postgresPassword | b64enc }}
   {{- if .Values.limes.has_audit_trail }}
+  rabbitmq_user: {{ required ".Values.limes.passwords.ccloud.rabbitmq_user" .Values.limes.passwords.ccloud.rabbitmq_user | default "rabbitmq" | b64enc }}
```
SuperSandro2000, line 9: The default doesn't overwrite the required, does it?

openstack/limes/templates/_utils.tpl
```diff
@@ -20,7 +20,10 @@
       name: limes-secret
       key: rabbitmq_password
 - name: LIMES_AUDIT_RABBITMQ_USERNAME
-  value: "rabbitmq"
+  valueFrom:
+    secretKeyRef:
```
SuperSandro2000, line 24: I don't think we need to wrap the username in a secret.

fwiesel: Sure, it is not necessary, but more consistent. I've changed it though to a "direct" value.

PR #5877, opened 2024-01-10
allow individual mapping

viennaa:
specifying some dedicated values in our config repository lead to overwriting of the whole maxAllowed list. This uses a map instead of a list now and will prohibit this behaviour.


richardtief (approved):
Appears reasonable. Thanks

PR #5888, opened 2024-01-11
[maia] thanos compactor settings

notque:
Added the compactor settings, is this what you want for the config?


richardtief:
Generally yes, but these values are not included in the pod arguments. We need to fix this in the common chart. [#5890](https://github.com/sapcc/helm-charts/pull/5890) 

Note: If we activate this, the compactors will suddenly have a lot of work. We should introduce this in a supervised manner.

notque:
Closing, using the chart fix.

PR #5890, opened 2024-01-12
common/prometheus-server-pre7: fix thanos compactor values

richardtief:
The values were not included in the StatefulSet of the Thanos compactor. With this PR, reasonable default values are set.


richardtief:
Need to merge this since we wanna deploy this with Maia in QA today.

PR #5893, opened 2024-01-15
[Nova] Reduce cpu_allocation_ratio to 3

joker-at-work:
We've seen 4 times CPU overcommit being too high and leading to problems for VMs since the VMware driver also counts hyperthreading cores as full VCPUS.


dhoeller (approved):
LGTM

PR #5895, opened 2024-01-15
[cc3test] intially drop new phase label

thgrs:
as the introduction of the new phase label might result in partial false alerts until we have rolled out the change to all regions and updated the alerts accordingly, we will simply drop it initially and reenable it later when everything is prepared everywhere


ashifnihalb (approved):
lgtm

PR #5898, opened 2024-01-16
pgbackup: expose as service, postgres-ng: create backup before doing an upgrade

SuperSandro2000:


common/pgbackup/templates/service.yaml
```diff
@@ -0,0 +1,15 @@
...
+  annotations:
+    {{- if and $.Values.global.linkerd_enabled $.Values.global.linkerd_requested }}
+    linkerd.io/inject: enabled
+    {{- end }}
+spec:
+  selector:
+    name: "{{ .Release.Name }}-pgbackup"
+  ports:
+  - name: pgbackup
+    port: 5432
```
majewsky, line 14: This should probably be 8080 to match the container port. 5432 is for the postgres binary protocol.

PR #5899, opened 2024-01-17
nova/flavors: Separate instance quota for hana-flavors

grandchild:
Needed for Limes' upcoming 2024 quota revamp.

Only for nova-bigvm-enabled regions, since we only need it in qa-de-1 first.

### DIFF
--- a/openstack/sap-seeds/ci/test-values.yaml
+++ b/openstack/sap-seeds/ci/test-values.yaml
@@ -1,0 +1,1 @@
+hana_flavors_quota_separate: true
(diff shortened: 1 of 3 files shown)


fwiesel (approved):

grandchild:
we do this with the `nova-manage sap sync_instances_flavor` command executed from the conductor pod.

joker-at-work:
We only need this in qa and not in production, yet, right? How do we prevent it from going into production?

majewsky:
Yes, this should only be in QA for now, so maybe wrap this in `{{ if (.Values.global.region | contains "qa" ) }}`.

grandchild:
i was hoping the nova-pr would defuse that, but since it doesn't, i will have to make `if` wrappers.

grandchild:
```
sap-seeds/templates/_flavors_hana_no_exclusive.tpl:11:47: executing "sap-seeds/templates/_flavors_hana_no_exclusive.tpl" at <"qa-">: invalid value; expected string
```
erm... it _is_ a string. 🤔

grandchild:
I can't reproduce the linting failure locally. 🤷🏾

grandchild:
Tested and adds the specs in qa-de-1 and doesn't add them in ap-jp-1.

joker-at-work (approved):
I like the approach to make us able to configure it via secrets :+1:

PR #5900, opened 2024-01-17
keystone: refill credits every second

bbobrov:
It was decided not to block the user for 60 seconds, but rather for 1 second, because keystone can hold many requests by itself.


rajivmucheli (approved):
lgtm!

PR #5909, opened 2024-01-19
[ironic] switches to internal keystone endpoint

stefanhipfel:


stefanhipfel:
hm not sure if we need to change it at all:

**www_authenticate_uri**
`Complete “public” Identity API endpoint. This endpoint should not be an “admin” endpoint, as it should be accessible by all end users. Unauthenticated clients are redirected to this endpoint to authenticate. Although this endpoint should ideally be unversioned, client support in the wild varies. If you’re using a versioned v2 endpoint here, then this should not be the same endpoint the service user utilizes for validating tokens, because normal end users may not be able to reach that endpoint.`

sandzwerg:
hmmmkay let's ask Boris

sandzwerg:
So you are right, we are not supposed to use the internal endpoint of keystone. So let's close this.

PR #5923, opened 2024-01-23
[unbound][dnstap] -- let's run dnstap in a sidecar container

vssldmtrv:
As per https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/#sidecar-containers-and-pod-lifecycle

The dnstap container serves a clear supporting role.

Also, if enabled, it should be available before the main unbound container starts.

All in all, it ticks all the boxes as far as sidecar containers described in the above URL go.

The patch basically moves the dnstap container out of the "containers:" section
and into the "initContainers:" one and sets its restartPolicy to Always.

### DIFF
--- a/system/unbound/templates/deployment.yaml
+++ b/system/unbound/templates/deployment.yaml
@@ -36,6 +36,34 @@ spec:
               topologyKey: "kubernetes.io/hostname"
       nodeSelector:
         topology.kubernetes.io/zone: {{ .Values.global.region}}{{ .Values.unbound.failure_domain_zone}}
+      initContainers:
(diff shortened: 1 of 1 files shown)


mutax:
Hmm, 
do you want to run it as a sidecar or an initContainer?

Because you are moving it into the init container section:

> Init containers run and complete their tasks before the main application container starts. Unlike [sidecar containers](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers), init containers are not continuously running alongside the main containers.

And would mean that the unbound will only start after the dnstap container is terminating which is not what we want...

mutax (approved):
makes sense

PR #5932, opened 2024-01-24
[mariadb-galera] unittest for my cnf configmap

bashar-alkhateeb:


businessbean (approved):
lgtm. Let's test them in the pipeline

PR #5939, opened 2024-01-26
[hermes] activate linkerd only for logstash

Kuckkuck:
My proposal:

Disabling linkerd for the namespace and activating it only for logstash, because the namespace activation blocks the helm post-deploy and job pods. At the moment the post-deploy and job pods are using the toolbox image, which has no linkerd support. Switching to an own image is increasing the complexity.


notque (approved):
LGTM

PR #5946, opened 2024-01-30
update harvest: autoscale worker replicas 

chuan137:
changes due to netappsd updated to [v2](https://github.com/sapcc/netappsd/tree/v2), where netappsd master controls worker's replica, based on discovered filers.

### DIFF
--- a/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/etc/rest.custom_volume.yaml
+++ b/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/etc/rest.custom_volume.yaml
@@ -1,0 +1,109 @@
+# merge:
+# counters:
+#   - ^comment => comment
+#   - hidden_fields:
+#     - comment
+# export_options:
+#   instance_keys:
+#     - commen
+
+
+name: Volume
+query: api/storage/volumes
+object: volume
+counters:
+  - ^^uuid                                        => instance_uuid
+  - ^name                                         => volume
+  - ^svm.name                                     => svm
+  - ^aggregates.#.uuid                            => aggrUuid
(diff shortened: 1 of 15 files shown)


prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/etc/rest.custom_volume.yaml
```diff
@@ -0,0 +1,109 @@
+# merge:
+# counters:
+#   - ^comment => comment
+#   - hidden_fields:
+#     - comment
```
Carthaca, line 5: What is this section? Maybe some kind of `TODO`?

It matches some seemingly inconsistent config below: `comment` is in the counters as well as in hidden_fields, is that okay?

chuan137: For ZAPI configuration, extending is easier. One could just add the changed fileds in a custom file, and harvest will merge them. However, for REST, the entire file needs to be replaced. I requested a feature [here.](https://github.com/NetApp/harvest/issues/2607). So the section is about what would be needed when the merging feature is available for REST. But good catch, the section only applies to ZAPI conf.

chuan137: the hidden_fields is for requesting extra fields that is not returned otherwise.
https://github.com/NetApp/harvest/blob/26e5f2700f928cee206206d1049cbbaa4b5a72ad/docs/configure-rest.md?plain=1#L164

Carthaca (approved):

PR #5949, opened 2024-01-31
[maria db] Enable VPA butler

rhalina:
Add VPA butler annotation to unblock Erik Schuberts vpa-butler deployment. It is already enabled for the neutron chart, but mariadb still misses it.


viktor-kovacs (approved):
LGTM (if David approved it)

joker-at-work (commented):
This changes the default for all services. Why don't you enable it for Neutron's mariadb only?

businessbean (changes requested):
With the [nova](https://github.com/sapcc/helm-charts/pull/5949#pullrequestreview-1853492547) approach every service owner can decide if his DB should be autoscaled...

PR #5952, opened 2024-01-31
added tests for 5, 7 and 9 replicas

dusandordevicsap:


bashar-alkhateeb (dismissed):

businessbean (changes requested):
`mariadb.galera.pcrecovery` is not required for these tests. Please add the tests for 1 and 3 replicas too

PR #5955, opened 2024-01-31
common/redis: replace library/redis with base-images/redis

majewsky:
Also, simplify some templating to remove useless variability.

### DIFF
--- a/common/redis/templates/_helpers.tpl
+++ b/common/redis/templates/_helpers.tpl
@@ -12,31 +12,6 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | replace "_" "-" | trimSuffix "-" -}}
 {{- end -}}
 
-{{/*
-  The (contains "." $repo) checks if the chart user has overridden the
-  repository to refer to a custom registry. If not, the default Docker Hub
-  mirror gets used.
-*/}}
-{{- define "redis.image" -}}
-  {{- $repo := required ".Values.image.repository missing" .Values.image.repository -}}
-  {{- $tag := required ".Values.image.tag missing" .Values.image.tag -}}
-  {{- if contains "." $repo -}}
-    {{- $repo -}}:{{- $tag -}}
-  {{- else -}}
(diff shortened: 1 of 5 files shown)


common/redis/templates/deployment.yaml
```diff
@@ -31,8 +36,8 @@ spec:
     spec:
       containers:
       - name: {{ template "redis.fullname" . }}
-        image: {{ template "redis.image" . }}
-        imagePullPolicy: {{ default "" .Values.image.pullPolicy | quote }}
+        image: {{ $registry }}/base-images/redis:{{ .Values.image.tag }}
+        imagePullPolicy: {{ if contains "latest" .Values.image.tag }}Always{{ else }}IfNotPresent{{ end }}
```
SuperSandro2000, line 40: That's smart, I like it.

common/redis/templates/_helpers.tpl
```diff
@@ -12,31 +12,6 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | replace "_" "-" | trimSuffix "-" -}}
```
SuperSandro2000, line 12: We should combine those two functions tbh

common/redis/templates/_helpers.tpl
```diff
@@ -12,31 +12,6 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
 
-{{/*
-  The (contains "." $repo) checks if the chart user has overridden the
-  repository to refer to a custom registry. If not, the default Docker Hub
-  mirror gets used.
-*/}}
-{{- define "redis.image" -}}
-  {{- $repo := required ".Values.image.repository missing" .Values.image.repository -}}
-  {{- $tag := required ".Values.image.tag missing" .Values.image.tag -}}
-  {{- if contains "." $repo -}}
```
SuperSandro2000, line 23: What? If the repo contains a dot, we are just not using the mirror!?...

SuperSandro2000 (approved):

common/redis/templates/_helpers.tpl
majewsky: That's for image refs that already contain a hostname, e.g. Keppel forcing the alternate region mirrors.

PR #5956, opened 2024-01-31
castellum,limes,keppel,tenso: switch to postgres-ng

SuperSandro2000:


common/postgresql-ng/docker/entrypoint.sh
```diff
@@ -51,17 +51,21 @@ if [[ $(id -u) == 0 ]]; then
...
-  if [[ -e /data/PG_VERSION ]]; then
-    old_version=$(cat /data/PG_VERSION)
+  if [[ -e /data/data/PG_VERSION ]]; then
+    old_version=$(cat /data/data/PG_VERSION)
 
     # move to a temporary directory
     mkdir -p /data-old
-    mv /data/* /data-old/
+    mv /data/data/* /data-old/
+    rmdir /data/data
```
majewsky, line 60: Only seeing this now, but this will be very expensive and slow for the large DBs (Arc and Sentry) because you're moving off NFS into local storage and then back into NFS. Would it be possible to have the temp location within the NFS volume?

common/postgresql-ng/templates/deployment.yaml
```diff
@@ -52,6 +52,12 @@ spec:
         env:
         - name: DEBUG
           value: {{ .Values.debug | quote }}
+        - name: PERSISTENCE_ENABLED
+        {{- if .Values.persistence.enabled }}
+          value: true
+        {{- else }}
+          value: false
+        {{- end }}
```
majewsky, line 60:
```suggestion
          value: {{ .Values.persistence.enabled }}
```

Sometimes Go templates are less stupid than you think.

openstack/keppel/values.yaml
```diff
@@ -118,9 +118,13 @@ postgresql:
     enabled: true
     accessMode: ReadWriteOnce
     size: 100Gi
+  postgresDatabase: keppel
   sharedMemoryLimit: 256Mi
+  users:
+    keppel:
```
majewsky, line 124: Maybe just a matter of personal taste, but having an implicit null value makes it looks like you forgot to type something.
```suggestion
  users:
    keppel: {}
```

PR #5961, opened 2024-02-02
common/thanos: fix grpc and internal ingress hosts

richardtief:


common/thanos/templates/_thanos_helpers.tpl
```diff
@@ -42,53 +42,61 @@ thanos-{{- $name -}}
...
 {{- required ".Values.ingress.hostsFQDN must have at least one hostname set" $firstHost -}}
 {{- else -}}
-thanos-{{- $name -}}.{{- required "$root.Values.global.region missing" $root.Values.global.region -}}.{{- required "$root.Values.global.tld missing" $root.Values.global.tld -}}
+thanos-{{ $name }}.{{- required "$root.Values.global.region missing" $root.Values.global.region -}}.{{- required "$root.Values.global.tld missing" $root.Values.global.tld -}}
 {{- end -}}
 {{- end -}}
 
 {{/* External gRPC URL of this Thanos. */}}
-{{- define "thanos.externalGrpcURL" -}}
+{{- define "thanos.grpcURL" -}}
```
viennaa, line 50: Why the renaming, pointing out the "externality" as we do with the "internality" makes it a better read for others I'd suppose.

common/thanos/templates/_thanos_helpers.tpl
```diff
@@ -42,53 +42,61 @@ thanos-{{- $name -}}
...
 {{- $firstHost := first $root.Values.grpcIngress.hosts -}}
 {{- required ".Values.grpcIngress.hosts must have at least one hostname set" $firstHost -}}.{{- required ".Values.global.region missing" $root.Values.global.region -}}.{{- required ".Values.global.tld missing" $root.Values.global.tld -}}
+{{- else -}}
+thanos-{{ $name }}-grpc.{{- required ".Values.global.region missing" $root.Values.global.region -}}.{{- required ".Values.global.tld missing" $root.Values.global.tld -}}
+{{- end -}}
 {{- end -}}
 
-{{/* External gRPC URL of this Thanos. */}}
-{{- define "thanos.externalInternalURL" -}}
+{{/* External internal URL of this Thanos. */}}
```
viennaa, line 61: Isn't this description misleading now?

viennaa (commented):
Thanks for bringing in this change. It took me a bit to wrap my head around the wrappers but it looks good.

Added few comments.

common/thanos/templates/_thanos_helpers.tpl
richardtief: Yes - thought about the same. But I decided to just call it that, as `externalInternalURL` sound really weird to me. So just `internalURL` and `grpcURL`.

common/thanos/templates/_thanos_helpers.tpl
richardtief: copy paste error.
```suggestion
{{/* External URL to expose Thanos API only. */}}
```

common/thanos/templates/_thanos_helpers.tpl
viennaa: technically an overall `s/external//g`

the `externalInternalURL` was a little bit funny indeed :D

Ok, I guess we can go with it.

viennaa (approved):
Thanks again, details matter.

PR #5963, opened 2024-02-02
[global/gh-actions] Upgrade to autoscaling runner scale sets

notandy:
With the introduction of autoscaling runner scale sets, the existing autoscaling modes are now legacy.
Apart from active mainatanance by GitHub, the new autoscaling runner scale set controller has more features,
supports custom certificates way better and out of the box, and also integrates better with multiple organizations.

Also, it has now prometheus metrics.


auhlig (approved):

auhlig:
nice work!

PR #5964, opened 2024-02-05
Added recon_candidates_limit to overwrite default value of 5

Barun311:
Creating this PR to make changes in container-sharder to overwrite the default recon candidate value, which is 5 as of now. Setting value to -1 will help.


adic9000:
PR approved.

PR #5968, opened 2024-02-05
[hermes] fix logstash conf metis cleanup

notque:
The cleanup wasn't happening for the unavailable fields, leading to events going into an "unavailable" index which gives us two different dump indexes. This change removes what looks like an incorrect grouping of the domain_mapping if. Please take a look and validate.

### DIFF
--- a/openstack/hermes/templates/etc/_logstash.conf.tpl
+++ b/openstack/hermes/templates/etc/_logstash.conf.tpl
@@ -355,32 +355,32 @@ filter {
         }
       }
     }
-    # Cleanup
+  }
+  # Cleanup
+  mutate {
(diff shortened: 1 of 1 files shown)


openstack/hermes/templates/etc/_logstash.conf.tpl
```diff
@@ -355,32 +355,32 @@ filter {
         }
       }
     }
-    # Cleanup
-    mutate {
-      id => "f20d_removing_domain_mapping"
-      remove_field => ["[domain_mapping]"]
-    }
+  }
```
notque, line 358: Actually, it looks like this belongs in the above if. Adjusting.

PR #5971, opened 2024-02-06
compactor multi attach fixed

viennaa:
when rolling out new compactors they often will be stuck in ContainerCreating with multi-attach volume errors. Since they are not mission critical, we can easily take the short downtime and bring them down first, ensuring a working rollout.


richardtief (approved):
Looks good! Thanks

PR #5972, opened 2024-02-06
common/prometheus-server: align on prometheus label

richardtief:
If a prometheus label exists on a metric, then it should have the form `prometheus="<namespace/prometheus-name>"`, as each prometheus attaches an external label of the same name to each metric. Otherwise duplicate metrics will be created.


viennaa (approved):
Good catch, thx.

PR #5976, opened 2024-02-07
fetch latest crds

viennaa:


system/kube-system-global/Chart.lock
```diff
@@ -34,7 +34,7 @@ dependencies:
   version: 0.0.9
 - name: cert-manager
   repository: https://charts.jetstack.io
-  version: 1.13.3
+  version: v1.13.3
```
richardtief, line 37: is that intended?

viennaa: Thanks, it was not :)

richardtief (approved):

PR #5977, opened 2024-02-07
[tempest] Use external endpoint

misamoylov:


galkindmitrii (approved):
was it 'external' or 'public'?

openstack/tempest/tempest-base/templates/_tempest-pod.yaml.tpl
```diff
@@ -6,6 +6,10 @@ metadata:
   labels:
     system: openstack
     type: configuration
+  annotations:
```
Carthaca, line 9: Let's skip the linkerd annotation for now

I think we can treat tempest as an external client, not part of the internal mesh

Carthaca (changes requested):
Some services may have their own extra configuration, e.g. https://github.com/sapcc/helm-charts/blob/878af04a89badcb869f9b1bc5af837b84a5bc52a/openstack/tempest/manila-tempest/templates/etc/_tempest_extra_options.tpl#L37-L38 should be considered, too

Carthaca:
> was it 'external' or 'public'?

`public`, for instance see https://github.com/sapcc/tempest/blob/53f51d5048423d721d3058157290569b207b446d/tempest/config.py#L157

Carthaca (approved):

PR #5980, opened 2024-02-07
[mariadb-galera] db migration options added

businessbean:
- to allow migrations from another MariaDB service in the same K8s cluster
- chart version bumped
- gtidDomainId default set to 10817 to avoid server_id collisions for async replication setups
- gtidDomainId check removed, we want to support ids other than 1 or 2 to avoid async replication problems
- unit test rules updated
- DB logon removed from the liveness check to avoid pod restarts only because the credentials are wrong
- checkdbconnection added to still check if MariaDB provides a usable endpoint

### DIFF
--- a/common/mariadb-galera/helm/scripts/mariadb-galera/readiness.sh
+++ b/common/mariadb-galera/helm/scripts/mariadb-galera/readiness.sh
@@ -5,6 +5,22 @@ set -o pipefail
 
 source /opt/${SOFTWARE_NAME}/bin/common-functions.sh
 
+function checkdblogon {
+  mysql --protocol=socket --user=root --database=mysql --wait --connect-timeout={{ $.Values.readinessProbe.timeoutSeconds.database }} --reconnect --execute="STATUS;" | grep 'Server version:' | grep --silent "${SOFTWARE_VERSION}"
+  if [ $? -eq 0 ]; then
+    mysql --protocol=socket --user=root --batch --connect-timeout={{ $.Values.readinessProbe.timeoutSeconds.database }} --execute="SHOW DATABASES;" | grep --silent 'mysql'
+    if [ $? -eq 0 ]; then
+      echo 'MariaDB MySQL API usable'
+    else
+      echo 'MariaDB MySQL API not usable'
+      exit 1
+    fi
+  else
+    echo 'MariaDB MySQL API not usable'
+    exit 1
+  fi
+}
+
 function checkgaleraclusterstatus {
   mysql --protocol=socket --user=root --database=mysql --connect-timeout={{ $.Values.readinessProbe.timeoutSeconds.database }} --execute="SHOW GLOBAL STATUS LIKE 'wsrep_cluster_status';" --batch --skip-column-names | grep --silent 'Primary'
   if [ $? -eq 0 ]; then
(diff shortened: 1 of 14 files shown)


common/mariadb-galera/helm/scripts/mariadb-galera/readiness.sh
```diff
@@ -5,6 +5,22 @@ set -o pipefail
 
 source /opt/${SOFTWARE_NAME}/bin/common-functions.sh
 
+function checkdblogon {
```
bashar-alkhateeb, line 8: why not parameterising the common function and using it here?

common/mariadb-galera/helm/templates/_helpers.tpl
```diff
@@ -17,7 +17,11 @@ metadata:
   namespace: {{ .global.Release.Namespace }}
   {{- if eq .replica "notused" }}
     {{- if and (eq .type "frontend") (or (eq .component "database") (eq .component "proxysql") (eq .component "haproxy"))}}
+      {{- if .global.Values.mariadb.migration.enabled }}
```
bashar-alkhateeb, line 20: such changes increase the need for the unittests ;)

common/mariadb-galera/helm/scripts/mariadb-galera/readiness.sh
businessbean: We can do that refactoring in a dedicated story...

common/mariadb-galera/helm/templates/_helpers.tpl
businessbean: I have added 2 unit tests

common/mariadb-galera/helm/tests/02-services-mariadb-galera_test.yaml
```diff
@@ -974,3 +974,37 @@ tests:
...
+    asserts:
+      - isKind:
+          of: Service
+      - equal:
+          path: metadata.name
+          value: testrelease-mariadb-replica
+      - equal:
+          path: metadata.labels.app
+          value: testrelease
+  - it: database service name changed if migration option is disabled
```
bashar-alkhateeb, line 994: please add one empty line in between the tests ([it]s)

bashar-alkhateeb: if i got it correctly you are testing that the replica service is not created .. for that i think it is better to check that the name of the service, in the document index (where the replica service should be created if not working) , is not equal to (testrelease-mariadb-replica )

businessbean: I have added `notEqual` checks too

common/mariadb-galera/helm/tests/02-services-mariadb-galera_test.yaml
```diff
@@ -974,3 +974,45 @@ tests:
...
+      - notEqual:
+          path: metadata.name
+          value: testrelease-mariadb
+      - equal:
+          path: metadata.labels.app
+          value: testrelease
+
+  - it: database service name changed if migration option is disabled
+    template: service.yaml
+    documentSelector:
```
bashar-alkhateeb, line 1001: you are already checking for the document with name testrelease-mariadb i would use documentIndex instead 
like this 
https://github.com/sapcc/helm-charts/blob/b2428bb069158e5fc8e4b0378c04b755a8f52659/common/mariadb-galera/helm/tests/02-services-mariadb-galera_test.yaml#L222

businessbean: I have opted against `documentIndex` because it will break if the order of documents (because of new features) will change. My `documentSelector` is somewhat redundant, but it should be more stable...

bashar-alkhateeb (approved):

PR #5981, opened 2024-02-07
adding clusterLabel

viennaa:
clusterLabel can now be specified to allow cross cluster communication with versions >= v0.26.0 again


richardtief (approved):
Looks about right. Thanks!

PR #5988, opened 2024-02-08
rollback to last working version

viennaa:
everything above v0.70.0 is introducing subresources which are not working with VPA sticking with v0.70.0 for now
see https://github.com/prometheus-operator/prometheus-operator/issues/6291


richardtief (approved):
unfortunate, but necessary. Thanks

PR #5996, opened 2024-02-12
[plutono] get latest official k8s gitsync image via keppel

thgrs:


richardtief (approved):
Looks better. Thanks!

PR #6001, opened 2024-02-12
vpa fix and crd bump

viennaa:


richardtief (approved):
Thanks for fixing it.

PR #6002, opened 2024-02-12
[plutono] bump plutono and postgresql image versions

thgrs:


system/plutono/values.yaml
```diff
@@ -29,7 +29,7 @@ plutono:
   tls: true
   disco: true
   image: credativ/plutono
-  image_version: v7.5.26
+  image_version: v7.5.29
```
richardtief, line 32: As discussed, this can be `latest`.

system/plutono/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart for SAP's CCloud Plutono 7.5.x deployment
 name: plutono
-version: 0.2.6
+version: 0.2.7
 dependencies:
   - name: postgresql
     repository: file://../../common/postgresql
```
viennaa, line 7: unrelated, but why isn't this using `https://charts.eu-de-2.cloud.sap`

richardtief: https://github.com/sapcc/helm-charts/blob/641d765d4e7839a272f545625f824aefa0a2c357/common/postgresql/Chart.yaml#L2

system/plutono/values.yaml
thgrs: updated

richardtief (approved):
Thanks!

PR #6007, opened 2024-02-13
[arc-mosquitto] add livenessProbe to restart the service when the certificate changes

ArtieReus:


jknipper (approved):
Looks good to me!

openstack/arc/charts/mosquitto/templates/deployment.yaml
```diff
@@ -31,10 +31,22 @@ spec:
...
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         ports:
         - containerPort: {{ .Values.service.internalPort }}
+        livenessProbe:
+          exec:
+            command:
+            - cmp
+            - -s
+            - /mosquitto_tls/tls.crt
+            - /tmp/mosquitto_tls_crt.livenessprobe
```
databus23, line 49: I would add the some timeout/interval changes as from the example:
```
initialDelaySeconds: 60
periodSeconds: 60
timeoutSeconds: 3
```

We don't need to execute the check that often and the increased timeout (default: 1s) avoids false positive due to cpu/I/o contention.

ArtieReus: agreed and commited

PR #6029, opened 2024-02-19
enabling prometheus vpa

viennaa:
vpa activated for the leftover Prometheis. The maxAllowed values are giving a reasonable headroom to breathe as of the current workload.


richardtief (approved):
Looks promising! Lets go.

PR #6031, opened 2024-02-19
deployment for native vpa alerts

viennaa:
VPA alerts must not be installed in any prometheus but (kubernikus and kubernetes) or (the regional thanos ruler).
If thanos alerts are enabled, they are deployed to the thanos ruler.
Independently they still were deployed to prometheus-kubernetes if VPA is set to Auto. Explicitly prohibiting this will not deploy these unfulfillable alerts.


richardtief (approved):
Looks good. Thanks.

PR #6034, opened 2024-02-20
[vmware-monitoring] collect additional hostsystem metrics

ninePebbles:
Add CPU metrics:
* cpu_coreutilization_percentage
* cpu_corepeak_usage

Add network metrics:
* network_bytes_rx_kbps
* network_bytes_tx_kbps

Bump chart


ninePebbles:
@viennaa Unfortunately I don't have permissions to suggest reviewers, leaving a note here.

system/vmware-monitoring/templates/collector-configmap.yaml
```diff
@@ -176,6 +176,10 @@ data:
         key: "cpu|capacity_provisioned"
       - metric_suffix: "cpu_corecount_provisioned"
         key: "cpu|corecount_provisioned"
+      - metric_suffix: "cpu_corepeak_usage"
+        key: "cpu|peak_cpu_core_usage"      
+      - metric_suffix: "cpu_coreutilization_percentage"
+        key: "cpu|coreUtilization_average"
```
viennaa, line 182: Make sure you add the unit at the end, if possible. I suppose in these examples it is percentage, no?

```suggestion
      - metric_suffix: "cpu_peak_core_usage_percentage"
        key: "cpu|peak_cpu_core_usage"      
      - metric_suffix: "cpu_core_utilization_percentage"
        key: "cpu|coreUtilization_average"
```

viennaa (commented):
Thanks for providing this PR. I would try to stay as close as possible to the origin key and add the unit in the end. Suggestions attached, think about what fits best.

system/vmware-monitoring/templates/collector-configmap.yaml
ninePebbles: I wasn't sure about the unit in  'cpu|peak_cpu_core_usage', VROPS documentation is not clear about that. These look good to me, thanks.

viennaa: i haven't looked it up, so you better double check ;)

viennaa: Can spot it either, even when I am looking into an active vROps deployment. Question is, what do you expect from it, when you don't know, what it is about?

ninePebbles: Hopefully a percentage, if it's MHz instead, then we can compute the percentage ourselves...

viennaa:
@ninePebbles can you maintain your name in github please?

PR #6042, opened 2024-02-22
[Nova] Consume cell information from `nova-etc` Secret

joker-at-work:
These 2 patches are supposed to partly (rabbitmq, DB) prepare us for PCI 4 in the `nova-compute` pods.


joker-at-work:
Looks like using a list doesn't work in a `ConfigMap`. Need to find another solution to list the names. Could do a `cells: cell0,cell1` and then split on comma ...

fwiesel (approved):

PR #6049, opened 2024-02-26
[common][rabbitmq] optionally add detailed metrics

Carthaca:
This enables https://www.rabbitmq.com/docs/prometheus#detailed-endpoint
via podmonitor (not servicemonitor, because the metrics port is not
present at the service).

E.g. gives insights into the queue names via parameter
`family=queue_coarse_metrics` in new metrics called
`rabbitmq_detailed_XXX` which are added on top.


Carthaca:
The params are mandatory for the detailed endpoint, see https://www.rabbitmq.com/docs/prometheus#detailed-endpoint

> As mentioned earlier, using per-object metrics in environments with a lot of entitites is very computationally expensive. For example, /metrics/per-object returns all metrics for all entities in the system, even if many of them are not used by most clients (such as monitoring tools).

>This is why there is a separate endpoint for per-object metrics that allows the caller to query only the metrics they need:

>GET /metrics/detailed

>By default it does not return any metrics. All required metric groups and virtual host filters must be be provided as query parameters.

bashar-alkhateeb:
Oops! .. I think we forgot that in the servicemonitor resource
we will try to add this to the servicemonitor soon, as well as the related Alert.

PR #6051, opened 2024-02-26
[unbound] -- introducing the bind-rpz-proxy sidecar

vssldmtrv:
Sadly unbound does not support TSIG AXFR to pull RPZ zones.

As a workaround we will be starting a bind container that will be pulling the RPZ zones from the primaries per TSIG AXFR.

Unbound will be pulling those zones from bind "anonymously"

### DIFF
--- a/system/unbound/templates/deployment.yaml
+++ b/system/unbound/templates/deployment.yaml
@@ -127,6 +127,21 @@ spec:
           - name: dnstap-socket
             mountPath: {{ .Values.unbound.dnstap.socket_path | default "/run/dnstap/dnstap.sock" | dir }}
 {{ end }}
+{{ if .Values.unbound.bind_rpz_proxy.enabled }}
+      - name: bind-rpz-proxy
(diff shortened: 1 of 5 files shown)


system/unbound/templates/deployment.yaml
```diff
@@ -126,6 +126,21 @@ spec:
         volumeMounts:
           - name: dnstap-socket
             mountPath: {{ .Values.unbound.dnstap.socket_path | default "/run/dnstap/dnstap.sock" | dir }}
+{{ end }}
+{{ if .Values.unbound.bind_rpz_proxy.enabled }}
+      - name: bind-rpz-proxy
+        image: {{ required ".Values.global.registryAlternateRegion is missing" .Values.global.registryAlternateRegion }}/{{.Values.unbound.bind_rpz_proxy.image}}:{{ .Values.unbound.bind_rpz_proxy.image_tag}}
+        imagePullPolicy: {{ .Values.unbound.image_pullPolicy }}
+{{ if ge (int .Capabilities.KubeVersion.Minor) 29 }}
+        restartPolicy: Always
```
galkindmitrii, line 135: I don't think you need this, it is the default

galkindmitrii (approved):

PR #6055, opened 2024-02-27
prometheus-rules: alert when a pod cannot pull its images

SuperSandro2000:


majewsky (approved):
Please bump chart version.

PR #6067, opened 2024-02-29
[Neutron] Change HTTP Socket Timeout

sven-rosenzweig:
According to the uwsgi source code the default http-socket timemout is 60 seconds. After this time connection will be shutdown.


joker-at-work (approved):
I think this is a good idea.

sebageek (approved):

PR #6071, opened 2024-02-29
[hermes] route events through target project_id if no initiator

notque:


majewsky (approved):
I was surprised about the assertion that the audit event is sent only to the initiator domain, since I had for sure seen those audit events in the target project before. But you said in a comment in the internal issue tracker that the event is routed to both places.

Re the actual implementation, it looks fine to me, though admittedly I have not read the entire logic, so I cannot judge any potential wider impact.

For the specific usecase of matching just Keppel janitor events with synthetic initiators, you could also match on initiator type URI `service/docker-registry/janitor-task`. But there are also lots of other situations with cross-domain audit events, e.g. whenever Limes writes a quota somewhere or when Castellum does a resize somewhere (both have service accounts in the default domain). From what I can see, it makes sense to not litter the `service@Default` project with those audit events and only send them to the target projects instead.

notque:
I am going to spend time considering your comments. I can't really take an action on a friday regardless, so I will puzzle over it, and come back to it on Monday.

As you've rightfully pointed out, a lot of it is just deciding on the ideal situation. We already had a duplication procedure in some cases intentionally into the Default, but even that, should be questioned, which I did not do. 

I was mostly trying to improve it by continuing what existed without questioning if it really made sense. That's a useful step to do.

notque:
Ultimately this comes down to, we might as well put it in, and make it correct for this usecase, and punt the other longer term issues with it until we deal with those tasks.

PR #6074, opened 2024-03-01
ephemeral requests to avoid pod eviction

viennaa:
this is fixed with kubernetes>=1.28.5, but we are not there yet furthermore this is reasonable for scheduling


richardtief (approved):
Looks reasonable. Thanks.

PR #6084, opened 2024-03-04
[manila] add snapmirror label to openstack_manila_shares_size_gauge

chuan137:
Castellum needs to skip resizing a share  if it is snapmirror target.

### DIFF
--- a/openstack/manila/values.yaml
+++ b/openstack/manila/values.yaml
@@ -165,6 +165,7 @@ mysql_metrics:
         - "id"
         - "share_instance_id"
         - "status"
+        - "snapmirror"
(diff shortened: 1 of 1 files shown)


Carthaca (dismissed):
executed the query in a prod region that has shares with that metadata flag, looks good

Carthaca (changes requested):
I think this new label also needs to be added under the `labels` key (and maybe in the group by to keep the pattern)

Carthaca (approved):

chuan137:
Just a note, `share_metadata.value` should be the right field to add for snapmirror. But I think leaving it out is fine. We only need `share_id` in the Group by clause.

PR #6092, opened 2024-03-04
feat(prometheus-monitors): allow configuration of additional targets

IvoGoman:
Currently this helper chart gives some presets for adding very simple Pod|ServiceMonitors.
This change adds the option to add more than one endpoint to be scraped. In order to keep this compatible with existing usages the new value `additionalEndpoints` is introduced.

### DIFF
--- a/common/prometheus-monitors/Changelog.md
+++ b/common/prometheus-monitors/Changelog.md
@@ -1,4 +1,10 @@
-# 0.0.2
+# Changelog
+
+## 0.1.0
+
+- add support for additionalEndpoints. These additional endpoints are configured like the main endpoint, but under the `additionalEndpoints` key. This allows for the configuration of multiple endpoints for a single Pod|ServiceMonitor.
+
+## 0.0.2
 
 - drops default suffix from name of Pod|ServiceMonitor
 - adds explicit error for ServiceMonitor without selector
@@ -7,6 +13,6 @@
 - adds templating for basicAuth
 - fixes to ServiceMonitor template
 
-# 0.0.1
+## 0.0.1
 
 - initial release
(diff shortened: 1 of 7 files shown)


dusandordevicsap:
LGTM

dusandordevicsap (approved):
Looks good, I did not render chart though

common/prometheus-monitors/Changelog.md
```diff
@@ -1,4 +1,10 @@
-# 0.0.2
+# Changelog
+
+## 0.1.0
```
richardtief, line 3: The version upgrade is not reflected in `Chart.yaml`

common/prometheus-monitors/templates/podmonitor.yaml
```diff
@@ -32,33 +32,10 @@ spec:
   namespaceSelector:
     matchNames:
 {{  default (list .Release.Namespace ) .Values.namespaces | toYaml | indent 6 }}
-  
+
   podMetricsEndpoints:
-    - interval: {{ default "60s" .Values.scrapeInterval }}
-      scrapeTimeout: {{ default "55s" .Values.scrapeTimeout }}
-      port: {{ default "metrics" .Values.metricsPort }}
```
richardtief, line 39: I have templated this once. There seems to be something wrong with the yaml rendering. 
Screenshot 2024-03-06 at 15 05 58

richardtief (changes requested):
Thanks. Found some small things. Details below

common/prometheus-monitors/templates/podmonitor.yaml
IvoGoman: Thanks, there was still some diff between how the template was include in pod vs serviceMonitior

richardtief (approved):
Thanks!

PR #6098, opened 2024-03-05
nova/flavors: Separate instance quota for hana-flavors everywhere

grandchild:
Needed for Limes' upcoming 2024 quota revamp.

Now also for hana-exclusive regions.

### DIFF
--- a/openstack/sap-seeds/templates/_flavors_hana_exclusive.tpl
+++ b/openstack/sap-seeds/templates/_flavors_hana_exclusive.tpl
@@ -8,6 +8,10 @@
     "trait:CUSTOM_NUMASIZE_C48_M729": "required"
     "hw:cpu_cores": "24"  # used in nova-vmware as cores-per-socket (12pCPU = 24vCPU)
     "reservation:cpu": "24"
+    {{- if ( .Values.hana_flavors_quota_separate ) }}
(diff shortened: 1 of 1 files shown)


joker-at-work:
What about the delete flavors? How are they handled? I would think they need the same treatment.

grandchild:
Thanks, fixed.

grandchild:
Sorry for the back-and-forth, but Stefan just told me that billing is unaware of any of the deleted flavors, so they should be out of scope for now.

majewsky:
We need to move forward here. I need to have this deployed before my next call with Frank at 2PM and also I need to eat lunch at some point. Since Joker did not take issue with any of the changes in the hana_exclusive file, I'm going to merge that part now.

joker-at-work:
That means, Limes is fine with having `quota:separate` and non-`quota:separate` instances on the same `CUSTOM_HANA_EXCLUSIVE_HOST`-trait-having BBs. I didn't know what we would run that model in all regions - not just the ones without `CUSTOM_HANA_EXCLUSIVE_HOST`.

majewsky:
Well, "fine" is a strong word. It's just that the alternative is sending unexpected data to billing, which I cannot do right now without at least talking to them about it first.

PR #6101, opened 2024-03-05
add quota:separate etc. to selected old HANA flavors

majewsky:
Followup to #6098. We do in fact have prices for these specific 4 flavors, so we can put them as separate instance quotas without causing problems for Billing.

For the other flavors, Frank has tentatively agreed to do another push to get the customers off them.

I'm also defining a new extra spec for Limes here, which I will use to keep the capacity calculation from reporting more capacity on these flavors than absolutely necessary.

### DIFF
--- a/openstack/sap-seeds/templates/_flavors_deleted.tpl
+++ b/openstack/sap-seeds/templates/_flavors_deleted.tpl
@@ -410,6 +410,11 @@
     trait:CUSTOM_NUMASIZE_C48_M1459: required
     hw:cpu_cores: '24'  # cores-per-socket
     vmware:hw_version: vmx-18
+    {{- if .Values.hana_flavors_quota_separate }}
+    "quota:instance_only": "true"
+    "quota:separate": "true"
+    "limes:allow_growth": "false" # never report capacity above the current usage
+    {{- end }}
 
 - name: hana_c48_m1459
   id: "307"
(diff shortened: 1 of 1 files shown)


openstack/sap-seeds/templates/_flavors_deleted.tpl
```diff
@@ -410,6 +410,11 @@
     trait:CUSTOM_NUMASIZE_C48_M1459: required
     hw:cpu_cores: '24'  # cores-per-socket
     vmware:hw_version: vmx-18
+    {{- if .Values.hana_flavors_quota_separate }}
+    "quota:instance_only": "true"
+    "quota:separate": "true"
+    "limes:allow_growth": "false" # never report capacity above the current usage
```
grandchild, line 416: why not `quota:allow_growth`? i'd like to prevent section creep. "limes" is just a name, not a type/topic.

grandchild (changes requested):
would you add some of the info to the commit message? i often do blame to find out about motivations.

openstack/sap-seeds/templates/_flavors_deleted.tpl
majewsky: `quota:allow_growth` looks like something defined by Nova. I want to be very specific about this being a Limes-specific thing.

grandchild (approved):

PR #6116, opened 2024-03-07
[rabbitmq] add default container annotation

Carthaca:
useful to not have to specify the container name for
`kubectl logs` or `kubectl exec`

@Kuckkuck fyi


notque (approved):
Very much want this, looks good to me.

PR #6126, opened 2024-03-07
Pyroscope: Request linkerd

fwiesel:


joker-at-work (commented):
This doesn't seem to add any linkerd annotations I could see. Does it work like that? I thought we'd have to edit the `Deployment` and `Service` definitions to add the annotations.

fwiesel:
Sorry, I thought there was more magic happening. Now, I've set the values,  and checked it in a diff.

joker-at-work (approved):
The global option `linkerd_requested` has basically no functionality - shall we keep it as a hint that we have linkerd?

fwiesel:
I'd say yes. Mostly  so we can grep it.

PR #6130, opened 2024-03-08
[Hermes] switch to env vars from conf

notque:


openstack/hermes/templates/api-deployment.yaml
```diff
@@ -19,55 +19,60 @@ spec:
...
-          #   periodSeconds: 60
-          #   initialDelaySeconds: 60
-          #   readinessProbe:
-          #     httpGet:
-          #       path: /v1/
-          #       port: 80
-          #     timeoutSeconds: 5
-          #     periodSeconds: 5
+      - name: api
+        image: {{required ".Values.hermes.image is missing" .Values.hermes.image}}:{{.Values.hermes.image_tag | replace "string:" ""}}
```
IvoGoman, line 25: Is the `replace "string:" ""` still required? Checking our internal values I was not able to find a tag with this prefix.

openstack/hermes/templates/api-deployment.yaml
```diff
@@ -19,55 +19,60 @@ spec:
...
+            secretKeyRef:
+              name: hermes-es-api
+              key: HERMES_ES_USERNAME
+        - name: HERMES_ES_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: hermes-es-api
+              key: HERMES_ES_PASSWORD
+        - name: HERMES_OS_AUTH_URL
+          value: "{{.Values.hermes.auth_url}}"
```
IvoGoman, line 45: Duplicate of 53-54

openstack/hermes/templates/api-deployment.yaml
```diff
@@ -19,55 +19,60 @@ spec:
...
+          value: "/etc/hermes/policy.json"
+        volumeMounts:
+        - mountPath: /etc/hermes
+          name: config
+        resources:
+          requests:
+            memory: {{required ".Values.resources_requests_memory_api is missing" .Values.resources_requests_memory_api}}
+            cpu: {{required ".Values.resources_requests_memory_api is missing" .Values.resources_requests_cpu_api}}
+      volumes:
+      - name: config
```
IvoGoman, line 75: NIT: To make it obvious that the volume mount is now only used for the policy file
```suggestion
      - name: policy
```

openstack/hermes/templates/api-deployment.yaml
```diff
@@ -19,55 +19,60 @@ spec:
...
+          value: "service"
+        - name: HERMES_OS_PROJECT_DOMAIN_NAME
+          value: "Default"
+        - name: HERMES_API_LISTEN_ADDRESS
+          value: "0.0.0.0:80"
+        - name: HERMES_POLICY_FILE_PATH
+          value: "/etc/hermes/policy.json"
+        volumeMounts:
+        - mountPath: /etc/hermes
+          name: config
```
IvoGoman, line 69: NIT: To make it obvious that the volume mount is now only used for the policy file
```suggestion
          name: policy
```

openstack/hermes/templates/api-deployment.yaml
```diff
@@ -19,55 +19,60 @@ spec:
...
-          #     timeoutSeconds: 5
-          #     periodSeconds: 5
+      - name: api
+        image: {{required ".Values.hermes.image is missing" .Values.hermes.image}}:{{.Values.hermes.image_tag | replace "string:" ""}}
+        imagePullPolicy: {{required ".Values.hermes.image_pull_policy is missing" .Values.hermes.image_pull_policy}}
+        env:
+        - name: HERMES_DEBUG
+          value: '{{.Values.hermes.debug}}'
+        - name: HERMES_INSECURE
+          value: '{{.Values.hermes.insecure}}' # SSL certificates are hard :(
```
IvoGoman, line 31: Quick cross-check I did not find this set to true in any region. Maybe we can drop this altogether then?

openstack/hermes/templates/api-deployment.yaml
```diff
@@ -19,55 +19,60 @@ spec:
...
+          value: "{{.Values.hermes.auth_url}}"
+        - name: HERMES_OS_USERNAME
+          value: "{{.Values.hermes.username}}"
+        - name: HERMES_OS_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: hermes-api
+              key: HERMES_OS_PASSWORD
+        - name: HERMES_OS_AUTH_URL
+          value: "{{.Values.hermes.auth_url}}"
```
IvoGoman, line 54: NIT: These URLs could also be templated as such: https://github.com/sapcc/helm-charts/blob/40b53f86d5f7f555c946a52d536177de9728abd3/openstack/nova/templates/etc/_nova.conf.tpl#L99

openstack/hermes/templates/api-deployment.yaml
```diff
@@ -19,55 +19,60 @@ spec:
...
+        - name: HERMES_OS_CATALOG_URL
+          value: "{{.Values.hermes.catalog_url}}"
```
IvoGoman, line 56: AFAICS this setting is not used by the API

IvoGoman (changes requested):
The change in the API also adds `token_cache_time` and `memcached_servers` as env variables. Should these also be configured here?

The template `_hermes.conf.tpl` can be deleted as it is no longer used.

openstack/hermes/templates/api-deployment.yaml
```diff
@@ -19,55 +19,60 @@ spec:
...
+        - name: HERMES_OS_USER_DOMAIN_NAME
+          value: "Default"
+        - name: HERMES_OS_PROJECT_NAME
+          value: "service"
+        - name: HERMES_OS_PROJECT_DOMAIN_NAME
+          value: "Default"
```
IvoGoman, line 62: NIT: these are also defaulted to the same values in the config initialisation of the API. Since they are not configurable here, they could be removed to use the defaults by the API.

notque: Again, my habit of using Default values as documentation. Would you prefer I comment them out so they are clear as options? Or would you prefer they are removed?

openstack/hermes/templates/api-deployment.yaml
notque: It is for turning on when running local development. Which means it is not needed in this template.

notque:
> The change in the API also adds `token_cache_time` and `memcached_servers` as env variables. Should these also be configured here?
> 
> The template `_hermes.conf.tpl` can be deleted as it is no longer used.

We currently don't use memcached_servers, token_cache_time we use the default.

IvoGoman (approved):

PR #6133, opened 2024-03-08
[rabbitmq] fix permissions if persistence is enabled

Carthaca:
`fsGroupChangePolicy: "OnRootMismatch"`does not work for NFS mounts
(also see kubernetes/examples/issues/260)

### DIFF
--- a/common/rabbitmq/CHANGELOG.md
+++ b/common/rabbitmq/CHANGELOG.md
@@ -3,6 +3,26 @@ Rabbitmq CHANGELOG
 
 This file is used to list changes made in each version of the common chart rabbitmq.
 
+0.6.6
+-----
+maurice.escher@sap.com
+- fix permissions if persistence is enabled, does a recursive chown once
+- add owner info explicitly to pvc, clears helm diff from injected annotations
+
+0.6.5
+-----
+maurice.escher@sap.com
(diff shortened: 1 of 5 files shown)


fwiesel:
That can get very expensive... we need to do something like `OnRootMismatch` in the script.
Untested, but maybe like this:
```yaml
command:
- /bin/sh"
- -c
- |
- [ $(stat -c %u /var/lib/rabbitmq) == 999 ] || chown -R 999:999 /var/lib/rabbitmq
```

Carthaca:
> That can get very expensive... we need to do something like `OnRootMismatch` in the script. Untested, but maybe like this:
> 
> ```yaml
> command:
> - /bin/sh
> - -c
> - |
> - [ $(stat -c %u /var/lib/rabbitmq) -eq 999 ] || chown -R 999:999 /var/lib/rabbitmq
> ```

Good idea, I added this and tested in qa. Thanks!

fwiesel (approved):

bashar-alkhateeb (changes requested):
please add a change description to the CHANGELOG.md

Carthaca:
> please add a change description to the CHANGELOG.md

done

bashar-alkhateeb (approved):

PR #6134, opened 2024-03-08
[manila] take service status into account for rabbit alerts

Carthaca:
If a service is disabled we don't want to get false positives
on queuing messages.


chuan137 (approved):
lgtm

PR #6146, opened 2024-03-12
[openstack/utils] Support linkerd Job shutdown and add `job_finished_hook`

joker-at-work:
Please see the individual commits' messages for details.

### DIFF
--- a/openstack/utils/templates/_linkerd.tpl
+++ b/openstack/utils/templates/_linkerd.tpl
@@ -10,3 +10,17 @@ ingress.kubernetes.io/service-upstream: "true"
 nginx.ingress.kubernetes.io/service-upstream: "true"
 {{- end }}
 {{- end }}
+
+# Place this in job scripts when your script stops normally, but not abnormally
+# as this causes the side-car pod finish normally, but we need it for the re-runs
+{{- define "utils.linkerd.signal_stop_script" }}
+{{- if and $.Values.global.linkerd_enabled $.Values.global.linkerd_requested }}
+if command -v curl > /dev/null; then
+  curl -X POST http://localhost:4191/shutdown || true
+else
+  ( exec 3<>/dev/tcp/localhost/4191 &&
+    printf "POST /shutdown HTTP/1.0\r\n\r\n" 1>&3  &&
+    cat <&3 || true )
+fi
+{{- end }}
+{{- end }}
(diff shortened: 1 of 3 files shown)


openstack/utils/templates/_linkerd.tpl
```diff
@@ -10,3 +10,17 @@ ingress.kubernetes.io/service-upstream: "true"
 nginx.ingress.kubernetes.io/service-upstream: "true"
 {{- end }}
 {{- end }}
+
+# Place this in job scripts when your script stops normally, but not abnormally
+# as this causes the side-car pod finish normally, but we need it for the re-runs
+{{- define "utils.linkerd.signal_stop_script" }}
+{{- if and $.Values.global.linkerd_enabled $.Values.global.linkerd_requested }}
+if which curl > /dev/null; then
```
fwiesel, line 18: Maybe better `command -v curl > /dev/null`. Here is the [reasoning](https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script)

openstack/utils/templates/_linkerd.tpl
```diff
@@ -10,3 +10,17 @@ ingress.kubernetes.io/service-upstream: "true"
...
+  curl -X POST http://localhost:4191/shutdown || true
+else
+  exec 3<>/dev/tcp/localhost/4191 \
+  && printf "POST /shutdown HTTP/1.0\r\n\r\n" 1>&3 \
```
fwiesel, line 22: Why not straight?
```
printf "POST /shutdown HTTP/1.0\r\n\r\n" > /dev/tcp/localhost/4191
```
Or read from the file-descriptor the response and put it in a subshell to close the socket to have an orderly shutdown
```
( exec 3<>/dev/tcp/localhost/4191 &&
printf "POST /shutdown HTTP/1.0\r\n\r\n" 1>&3  &&
cat <&3 || true )
```

joker-at-work: Funnily enough, that doesn't work immediately:
> linkerd_app_core::serve: Connection closed error=connection closed before message completed client.addr=127.0.0.1:40368

I could add a sleep ala `(printf "POST …"; sleep 0.5) >/dev/tcp/…` (that works)

joker-at-work:
Tested the curl after a short sleep with a Pod in QA and it successfully shuts down linkerd and thus the Pod. Nova chart rendered successfully locally when using these changes.

PR #6154, opened 2024-03-13
[Nova] Use linkerd for Jobs

joker-at-work:
For using `linkerd` in Job-spawned Pods, we have to add a snippet at the end that will tell the `linkerd-proxy` sidecar to stop. This snippet is integrated - together with the proxysql snippet - into "utils.script.job_finished_hook" in the `utils` chart starting with version 0.14.6.

Since we need a version bump, we also get:
* add owner info explicitly to coordination pvc
* mount a trust-bundle at subpath


grandchild (approved):
LGTM

PR #6168, opened 2024-03-14
[manila] move configuration with credentials into own secrets

galkindmitrii:
- Configs with secrets will be projected into /etc/manila/manila.conf.d/
- And /etc/manila/backend-secret.conf
- Has to be tested

### DIFF
--- a/openstack/manila/templates/api-deployment.yaml
+++ b/openstack/manila/templates/api-deployment.yaml
@@ -79,7 +79,7 @@ spec:
           env:
           {{- if .Values.api_backdoor }}
             - name: COMMAND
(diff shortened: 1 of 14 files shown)


Carthaca (commented):
Many thanks for working on this!

I think the manila-scheduler command needs the additional config, too.

Regarding manila-api: I'm unsure how this is supposed to work, I don't see a way to specify the conf files/directory to use with wsgi

Carthaca:
I will test in our lab

Carthaca:
I added 2 commits - looks good in the lab

Carthaca (approved):

PR #6171, opened 2024-03-15
some postgresql ng additions

stanislav-zaprudskiy:
- **Let all user secrets be created**
- **Use Deployment name from Kubernetes manifest when running inside init script**
- **Allow overriding `postgres.fullname`**


SuperSandro2000:
I've cherry-picked the first two commits already into master

PR #6178, opened 2024-03-18
[mariadb] explicitly add owner labels to role and binding

Carthaca:
to not have the injected ones show up in helm diff each time


businessbean (approved):
lgtm

PR #6183, opened 2024-03-19
Add BTP_FP Domain to service seeds

rajivmucheli:


joker-at-work (approved):
Looked at Cinder, Neutron, Nova and didn't see anything problematic comparing to other deomains.

Carthaca (approved):
manila lgtm

majewsky (approved):
did not read, but this is usually fine

PR #6188, opened 2024-03-19
switches to chart v2 and prepares linkerd support

stefanhipfel:


fwiesel (approved):

openstack/kos-operator/templates/deployment.yaml
```diff
@@ -22,6 +22,8 @@ spec:
       name: kos-operator
   template:
     metadata:
+      annotations:
+      {{- include "utils.linkerd.pod_and_service_annotation" . | indent 8 }}
```
fwiesel, line 26: I far, that needs to be an `nindent 8` so we get the newline back that the `{{-` eats

fwiesel: I've added the commit. I hope you don't mind.

fwiesel: (I was mistaken, it seems. The template itself adds a newline)

sandzwerg (approved):

PR #6191, opened 2024-03-19
[netapp] fix metrics for offline volumes

chuan137:
In the aggregation rule `netapp_volume_exclusion_reason_offline`, we are currently reporting a share as `offline` if ANY of its volume is `offline`. In the DR or migration scenario, there can be multiple volumes for a share, and one of its volumes has `online` state and the others `offline`. We should drop such shares and allow them resizable by Castellum. Using the `unless` operator, we query metrics for `online` volumes on the right side and drop them if their labels (`share_id`, `project_id`) matched to the left.


Carthaca:
Do we need to keep `> 0` ?

chuan137:
> Do we need to keep `> 0` ?

I think not necessary. All the labels are transferred into the new metrics from the `count` vector

Carthaca (approved):

PR #6202, opened 2024-03-21
[rabbitmq] Adding support for new prometheus endpoints

dusandordevicsap:
This PR adds /metric/detailed and /metrics/per-object prometheus endpoints. They can be enabled/disabled from values.yaml, metrics section. It is done by creating new ServiceMonitor object


dusandordevicsap:
Initial proposal for this solution was this https://github.com/sapcc/helm-charts/pull/6033, but since I made mistake rebasing I created new PR. However other PR have valuable details in conversation that describe how we reached final solution

common/rabbitmq/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: monitoring.coreos.com/v1
```
bashar-alkhateeb, line 1: if the user keeps the default values you set, then no need to create this resources at all 
i think you need to put this resource under an if condition (if enableDetailedMetrics OR enablePerObjectMetrics )

bashar-alkhateeb (changes requested):

common/rabbitmq/templates/servicemonitor.yaml
dusandordevicsap: Done in latest commit

bashar-alkhateeb (approved):
Looks good for me

PR #6203, opened 2024-03-21
[hermes] mv logstash.conf to secrets

Kuckkuck:
Because of all the rabbitmq inputs, I have moved the logstash.conf to the secrets. Removed old env variables and this file is mounted next to the other logstash configs in the same directory.


notque (approved):
Works for me

PR #6209, opened 2024-03-21
[hermes] mv conf to secret

notque:


openstack/hermes/templates/api-secret.yaml
```diff
@@ -12,6 +12,7 @@ metadata:
 data:
   tls.crt: {{ b64enc $config.tls_crt }}
   tls.key: {{ b64enc $config.tls_key }}
+  hermes.conf: {{ include "hermes/templates/_hermes.conf.tpl" . | indent 4 }}
```
Kuckkuck, line 15: should be {{ include (print .Template.BasePath "/_hermes.conf.tpl" . |  b64enc }}

PR #6216, opened 2024-03-25
[Nova] Bump mysql_metrics chart for linkerd

joker-at-work:
We bump the chart to enable linkerd in it. With bumping the chart, we also move the config into a Secret and remove the prometheus annotations that would lead to double scraping with linkerd adding a second container.


Scsabiii (approved):
LGTM

PR #6218, opened 2024-03-25
cc3test: do not alert Limes when Keystone is down

majewsky:
As discussed in Slack: https://convergedcloud.slack.com/archives/CE5NDBQDC/p1711359842260729


viennaa (approved):
reasonable to me

PR #6221, opened 2024-03-25
Disable linkerd for metadata agent

sven-rosenzweig:
With metadata requests proxied through linkerd, metadata service does
 not work anymore. No packets are received on agent side. Temporarily we
 do not proxy those requests trough linkerd to make the service work
 again.


openstack/neutron/templates/statefulset-network-agent-apod.yaml
```diff
@@ -42,6 +42,7 @@ spec:
         configmap-etc-hash: {{ include (print $.Template.BasePath "/configmap-etc.yaml") $ | sha256sum }}
         configmap-etc-apod-hash: {{ include (print $.Template.BasePath "/configmap-etc-apod.yaml") $ | sha256sum }}
         {{- include "utils.linkerd.pod_and_service_annotation" $ | indent 8 }}
+        config.linkerd.io/skip-inbound-ports: 80
```
sebageek, line 45: This needs to be a string

sebageek (approved):

PR #6226, opened 2024-03-25
cc3test: do not alert Hermes or Maia when Keystone is down

notque:
Maia and Hermes, do not alert when Keystone is down

### DIFF
--- a/openstack/cc3test/alerts/ccloud-hermes.alerts
+++ b/openstack/cc3test/alerts/ccloud-hermes.alerts
@@ -9,11 +9,11 @@ groups:
       severity: critical
       support_group: observability
       tier: os
-      service: '{{ $labels.service }}'
+      service: hermes
       meta: 'CCloud Hermes API is down'
-      dashboard: 'cc3test-api-status?var-service={{ $labels.service }}'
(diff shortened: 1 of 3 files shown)


openstack/cc3test/alerts/ccloud-hermes.alerts
```diff
@@ -3,7 +3,7 @@ groups:
   rules:
   - alert: CCloudHermesApiDown
     expr: |
-        cc3test_status{service="hermes", type="api"} == 0
+        max(cc3test_status{service="hermes", type="api"} == 0) unless max(cc3test_status{service="keystone", type="api"} == 0)
```
viennaa, line 6: using the `max` statement here will filter out all the labels attached to `cc3test_status`. Meaning, everything with `{{ $labels.... }} will not work anymore.

I don't see any other mappings here but
```
type: api
service: hermes
```

which can be replaced in the alert definition.

viennaa (changes requested):
Thanks for adding the keystone dependency here too. Please replace all $label occurrences with their values, since `max` is soaking everything in.

viennaa:
And bump the chart please :)

viennaa (approved):
Thanks!

PR #6230, opened 2024-03-26
[nova] add openstack_compute_errored_live_migration_gauge metrics

leust:
This mysql metric exposes instances that are in error state because the last live-migration failed.

### DIFF
--- a/openstack/nova/values.yaml
+++ b/openstack/nova/values.yaml
@@ -679,6 +679,29 @@ mysql_metrics:
         WHERE deleted_at is NULL;
       values:
       - "version"
+    - help: Errored live-migrations
+      name: openstack_compute_errored_live_migration_gauge
+      labels:
+        - "instance_uuid"
+        - "migration_uuid"
+        - "vm_state"
+      query: |
+        WITH instance_migrations AS (
+            SELECT
+                m.instance_uuid,
+                m.status,
+                m.uuid,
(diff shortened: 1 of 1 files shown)


openstack/nova/values.yaml
```diff
@@ -679,6 +679,29 @@ mysql_metrics:
...
+                m.uuid,
+                ROW_NUMBER() OVER(PARTITION BY m.instance_uuid ORDER BY m.id DESC) AS rn
+            FROM migrations m
+            WHERE m.deleted = 0 AND m.migration_type='live-migration'
+        )
+        SELECT
+            i.uuid as instance_uuid,im.uuid as migration_uuid,i.vm_state,1 as gauge
+        FROM instances i
+        INNER JOIN instance_migrations im ON i.uuid = im.instance_uuid
+        WHERE i.deleted = 0 AND i.vm_state = 'error' AND im.rn = 1 AND im.status = 'error'
```
joker-at-work, line 702: Is `im.status` being `error` enough? Aren't there also `failed` Migrations? Would we want to see them here, too?

joker-at-work: hm ... for decom, I always have to set migrations in `failed` state to `error` state, because the service wouldn't delete otherwise. I also have to handle `running`. `failed` migrations often still have an allocation in `Placement`. So I would have thought, `error` is the stable state while `failed` is some non-rollbacked broken state.

joker-at-work (approved):

fwiesel (approved):

PR #6232, opened 2024-03-26
[cinder] move configuration with credentials into own secrets

galkindmitrii:
- Common secrets will be projected into /etc/cinder/cinder.conf.d/
- Vcenter operator might require more changes


joker-at-work:
> P.S. db-init-configmap.yaml already changed to db-init-secret.yaml, **but this requires the latest MariaDB chart dependency.**

This translates to "please don't merge because this PR can't possibly work" for me. Why didn't you bump the mariadb chart dependency if you need it bumped for this PR?

joker-at-work:
I've converted it to a draft so we don't merge it by accident until we a) verified it could work and b) have the mariadb bump in.

I do not know, yet, how the vcenter-operator is supposed to work, if we keep the masterpassword algorithm for now and if the vcenter-operator supports creating Secrets.

I do not like that all config for cinder-volume is moved into a Secret and thus no longer visible to us. Imho, these should be split.

galkindmitrii:
> > P.S. db-init-configmap.yaml already changed to db-init-secret.yaml, **but this requires the latest MariaDB chart dependency.**
> 
> This translates to "please don't merge because this PR can't possibly work" for me. Why didn't you bump the mariadb chart dependency if you need it bumped for this PR?

As discussed with @bashar-alkhateeb, there will be a separate PR bumping all shared dependencies per service at once.

hemna:
The updated requirements has merged and rolled to qa-de-1.

hemna:
So the rollout of the updated rabbitmq, mariadb and utils has broken the cinder deployment in qa-de-1.

utils moved the configmap to secrets and now cinder can't deploy.  I presume this PR adjusts for that change as well?

hemna:
This PR should also bump the cinder chart version from 0.1.2 to 0.2.0 since this is a major change.

hemna (commented):
I guess this looks ok, but my knowledge of helm and k8s is limited.   I would prefer to have Johannes give this a review before approving it.

openstack/cinder/templates/etc/_volume-secrets.conf.tpl
```diff
@@ -0,0 +1,23 @@
...
+    {{- range $key, $value := $values }}
+        {{- if kindIs "string" $value }}
+{{ $key }} = {{ $value | required (print "Please set the value for [" $name "]" $key) | replace "$" "$$" | quote }}
+        {{- else if kindIs "map" $value }}
+{{ $key }} = {{ $value | required (print "Please set the value for [" $name "]" $key) | toJson | squote }}
+        {{- else }}
+{{ $key }} = {{ $value | required (print "Please set the value for [" $name "]" $key) }}
+        {{- end }}
+    {{- end }}
+{{- end }}
```
joker-at-work, line 13: Why do we duplicate `cinder.iniValues`? I think that will be confusing.

joker-at-work (approved):
I think it's fine, even though most (currently all) cinder-volume pods still have all their config in Secret instead of only the Secret, because they are all created from the vcenter-template.

PR #6235, opened 2024-03-27
[Nova] Fix ironic pods not having cell config

joker-at-work:
When moving secrets into Secrets we missed to add cell-specific config to nova-compute-ironic pods. This lead to the driver not starting up because it couldn't reach rabbitmq.


Scsabiii (approved):
LGTM

PR #6238, opened 2024-03-27
[Nova] Add back nova_audit_map.yaml to nova-api-metadata

joker-at-work:
`nova-api-metadata` uses the same `api-paste.ini` as `nova-api` and thus requires the `nova_audit_map.yaml` file to be present.

In theory, we do not need a lot of the stuff in `api-paste.ini` for `nova-api-metadata` as the latter isn't really customer-facing. This could be changed in a follow-up. For now, we want things to work again.


Scsabiii (approved):
LGTM

PR #6239, opened 2024-03-27
[Nova] Cell2 conductor needs statsd-exporter.yaml

joker-at-work:
With moving to a projected configmap we accidentally got rid of `statsd-exporter.yaml`. We now change the cell2 conductor to read the file from `statsd-etc` volume like the cell1 conductor does.


Scsabiii (approved):
LGTM

PR #6240, opened 2024-03-27
kube-state-metrics: fix kube_pod_status_phased_normalized

richardtief:


Carthaca (approved):
Works for my concrete issue with having two pods of the same name in different namespaces

PR #6241, opened 2024-03-27
nova/flavors: Use fewest-possible sockets for 64vcpu+

grandchild:
Windows images have a socket limitation of 64, as hinted here:
https://learn.microsoft.com/en-us/archive/msdn-technet-forums/56adf991-952e-4337-8159-d632499db22b

Also, non-server windows images have even lower limits (2 or 4 sockets).

For all regular flavors with 64 vCPUS and more, set the cores-per-socket value to the vCPU/socket value of our (Cascade-Lake) hardware.

### DIFF
--- a/openstack/sap-seeds/templates/flavor-seed.yaml
+++ b/openstack/sap-seeds/templates/flavor-seed.yaml
@@ -204,6 +204,7 @@ spec:
     is_public: false
     extra_specs:
       {{- tuple . "vmware" | include "sap_seeds.helpers.extra_specs" | indent 6 }}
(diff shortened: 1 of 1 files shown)


joker-at-work:
Does that work? Looking at the hana_* flavors, we always set `hw:cpu_cores` to a value that can divide the flavor's `cpus` evenly. 64 Does not divide by 48, neither does 128. If you haven't tested this, please do so before we merge this as I fear VMware might not like non-evenly dividable values.

grandchild:
superseded by #6691

PR #6243, opened 2024-03-27
[nova] nova-compute-ironic to use the `nova` service user

leust:
nova-compute-ironic uses the ironic service user when talking to the Ironic API. Since this user is also used by Ironic itself, it's harder to coordinate password changes
(both Nova and Ironic need the new password at the same time). Additionally, it's not clear that Nova is talking to Ironic if the user is ironic.

### DIFF
--- a/openstack/nova/templates/etc-secret.yaml
+++ b/openstack/nova/templates/etc-secret.yaml
@@ -48,8 +48,8 @@ stringData:
     password = {{ required ".Values.global.nova_service_password is missing" .Values.global.nova_service_password | include "resolve_secret" }}
 
     [ironic]
-    username = ironic
-    password = {{ required ".Values.global.ironic_service_password is missing" .Values.global.ironic_service_password | include "resolve_secret" }}
+    username = nova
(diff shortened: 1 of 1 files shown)


openstack/nova/templates/seed.yaml
```diff
@@ -21,6 +21,7 @@ spec:
   - monsoon3/domain-neo-seed
   - monsoon3/domain-s4-seed
   - monsoon3/domain-wbs-seed
+  - monsoon3/glance-seed
```
joker-at-work, line 24: If this is just for the role `cloud_image_admin`, then I think we should just add it down below to the `roles` instead of creating this dependency. We opted to do that for all other roles, too.

joker-at-work (commented):
I think we need this as 2-step process or we need to seed the new roles manually upfront. The `nova` user will not have the roles until the seeder runs next after deployment, so any deployment in between may fail.

joker-at-work (approved):

grandchild (approved):

joker-at-work (approved):

PR #6247, opened 2024-03-28
enabling volumes

viennaa:
more and more compactors are fully utlizing the 25G eph storage.
This shall avoid ongoing maintenance efforts


richardtief (approved):
Thanks

PR #6249, opened 2024-03-28
[rabbitmq] fix secret to use encrypted data field

bashar-alkhateeb:
switch to use encrypted data field instead of stringData

Change-Id: Ic903c4d98b3bcea7984064d25599933ae1668ecd


dusandordevicsap (approved):
looks good to me

PR #6251, opened 2024-03-28
[vcenter-operator] Move secrets to Secret

joker-at-work:
We have to provide secrets via Secret instead of ConfigMap for PCI compliance. In doing so, we also clean up some no longer necessary configuration i.e. the OpenStack credentials and their seeeding and the `user_suffix`.

---

This depends on the `vcenter-operator` understanding the changes i.e. https://github.com/sapcc/vcenter-operator/pull/40 and https://github.com/sapcc/vcenter-operator/pull/41 to be rolled out in parallel.

Needs to be tested in a lab, first.


fwiesel (approved):

joker-at-work:
Oh, we still use the `variables` in the labs, That's too bad, now we have non-secrets in a Secret ...

PR #6265, opened 2024-04-02
[Nova] Provide vCenter username and password via Secret

joker-at-work:
Since we have to move out all passwords into Secret objects, the vcenter-operator-provided vCenter password needs to be moved out of the ConfigMap template and into a new Secret template.

### DIFF
--- a/openstack/nova/templates/hypervisors/hypervisors-vmware-secret-vct.yaml
+++ b/openstack/nova/templates/hypervisors/hypervisors-vmware-secret-vct.yaml
@@ -1,0 +1,35 @@
+{{- if or (.Capabilities.APIVersions.Has "vcenter-operator.stable.sap.cc/v1") (.Values.isImageTransportTemplating | default false) }}
+apiVersion: vcenter-operator.stable.sap.cc/v1
+kind: VCenterTemplate
+metadata:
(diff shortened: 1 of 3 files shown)


joker-at-work:
I tested just the `vcenter-operator` templating for the `else` case in a lab and it worked as in created the Secret object and the base64 looked fine.

fwiesel (approved):

openstack/nova/templates/hypervisors/hypervisors-vmware-secret-vct.yaml
```diff
@@ -0,0 +1,35 @@
...
+    labels:
+      system: openstack
+      type: configuration
+      component: nova
+      vcenter: {= host =}
+      datacenter: {= availability_zone =}
+      vccluster: {= cluster_name =}
+  data:
+    nova-compute-secrets.conf:{= " " =}
+      {%- filter b64enc %}
```
grandchild, line 25: I guess you preferred having the `filter` start on a separate line instead of

```suggestion
    nova-compute-secrets.conf: {% filter b64enc %}
```
?

If so, then 👍🏾

grandchild (approved):

PR #6291, opened 2024-04-04
[utils] add helm pre-upgrade weight to proxysql secret

bashar-alkhateeb:
add annotation to proxysql secret to set helm pre upgrade hook weit of -10 to proxysql secret, that to force the creation in case of a job needs the secret to be created before running.

option activated by set value `proxysql.forceSecretCreation` to true


BenjaminLudwigSAP (approved):
Thanks! LGTM.

galkindmitrii (approved):

PR #6294, opened 2024-04-04
[octavia] bump utils to 0.16.1, fix migration job

bashar-alkhateeb:
- bump up to use utils v0.16.1
- force creating proxysql secret before starting the migration job
- add an or condition to configure proxysql in the job if any values are set in vaules in proxysql map, that to fix the first run issue caused by the lookup function on K8S secret which is not created on K8S side while helm generating the templates

### DIFF
--- a/openstack/octavia/Chart.lock
+++ b/openstack/octavia/Chart.lock
@@ -13,12 +13,12 @@ dependencies:
   version: 0.6.11
 - name: utils
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.16.0
+  version: 0.16.1
 - name: owner-info
   repository: https://charts.eu-de-2.cloud.sap
   version: 0.2.3
 - name: linkerd-support
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
   version: 0.1.4
-digest: sha256:48e5ade2b1192d31655ae8c816c662ab36d6c1c59754097eef124535651fa3ee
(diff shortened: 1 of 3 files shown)


openstack/octavia/values.yaml
```diff
@@ -27,6 +27,7 @@ db_name: octavia
 
 proxysql:
   mode: ""
+  forceSecretCreation: true
```
fwiesel, line 30: This can be dropped as well, no?

bashar-alkhateeb: i think it is better to be kept for the future, if anything changed in the secret, the secret will get updated before the job gets triggered , or is there a reason  to trigger the job before updating the proxysql config ?

bashar-alkhateeb: the job has already a weight of -1 which force helm to change the normal upgrade order

bashar-alkhateeb: well in the case of renaming even  `"helm.sh/hook-delete-policy": before-hook-creation` cannot remove the resource 
so should i remove the hook and we live will the risk that the job will be triggered before the secret got updated ? or keep it and naming changes will not remove the old secret ?

fwiesel: I'd remove the hook in the secret. If a job is relying on the secret to be there, then it fails right away and we can fix it accordingly.

PR #6310, opened 2024-04-08
vpa support_group corrected

viennaa:


richardtief (changes requested):
Thanks. I know we always are, but can we make it consistent across the alerts files?

viennaa:
Yeah good catch, we should keep it configureable. Applied.

richardtief (approved):
Very much appreciate it. Thank you

PR #6318, opened 2024-04-09
helm refactoring

niranba:
split the config of cronus,nebula and rhea to secrets and config map

fix for:
https://github.wdf.sap.corp/cronus/issues/issues/486


openstack/cronus/templates/cronus/_config.yaml.tpl
```diff
@@ -108,26 +108,13 @@ cronus:
       errInterval: {{ .Values.cronus.tls.errInterval | default "60s" }}
 {{- end }}
   keystone:
-{{- if .Values.config.keystone }}
-{{- range $key, $value := .Values.config.keystone }}
```
corey-aloia, line 112: It looks like some of the values are config values. Would it make sense to list what it is? The loop was always a bit confusing to me, as it hides what the properties are.

For example
```
{{- if .Values.config.keystone }}
keystone:
    projectName: {{ .Values.config.projectName }}
    projectDomainName: {{ .Values.config.projectDomainName }}
    userDomainName:  {{ .Values.config.userDomainName }}
    username: {{ .Values.config.username }}
    ```

File: openstack/cronus/templates/cronus/config.yaml [hunk 2]
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.cronus.enabled -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cronus
+  namespace: {{ .Release.Namespace }}
+  annotations:
```
corey-aloia, line 7: The annotation can be removed

openstack/cronus/templates/cronus/config.yaml
```diff
@@ -0,0 +1,16 @@
...
+    cloud.sap/skip-template: "true"
+  labels:
```
corey-aloia, line 9: The only label worth having is the app label. Not sure why the secret has these labels. My guess would be due to some helm migration, and needing them explictily there. But I would remove labels except for `app`

corey-aloia (changes requested):
I left some comments. I should have done it as part of a github review.

Anyway, it looks like there are also still some secrets in the config. For example `rabbitmqUri`. This contains a password. Potentially a good way to see what all uses a password is to look through the specific environments values file, and look for all `*vault` references. Let me know if this doesn't make sense 👍

openstack/cronus/ci/test-values.yaml
```diff
@@ -10,7 +10,10 @@ rhea:
     tag: latest
 
 nebula:
+  sentryDsn:
```
corey-aloia, line 13: This whole file is still failing :) so sentryDsn is still nil, and I have a feeling there are many other fields that are nil.

I having a feeling this CI has always been failing and we just haven't paid attention. But you can also try locally via `helm template --values ci/test-values.yaml` and see where it gives you nil errors

openstack/cronus/templates/cronus/deployment.yaml
```diff
@@ -46,12 +46,27 @@ spec:
           args:
             - "-config=/cronus/config.yaml"
           env:
-          - name: GODEBUG
-            value: {{ .Values.cronus.godebug | quote }}
+            - name: SENTRY_DSN
```
corey-aloia, line 49: This whole deployment file is never used and should probably be deleted. But I think what we have here won't work as it is missing some env variables. In the sts we use envFrom. 

I think we should just delete this file

openstack/cronus/templates/cronus/secrets.yaml
```diff
@@ -2,16 +2,22 @@
...
     chart: cronus
     release: cronus
-    heritage: {{ .Release.Service }}
 type: Opaque
 data:
-  config.yaml: {{ include "cronus/templates/cronus/_config.yaml.tpl" . | b64enc }}
+  SENTRY_DSN: {{ .Values.cronus.sentryDsn | b64enc }}
+  RABBITMQ_URI: {{ printf "amqp://%s:%s@%s/" .Values.rabbitmq.users.default.user .Values.rabbitmq.users.default.password .Values.rabbitmq.host | b64enc }}
+  {{- $host := printf "%s.%s.%s:5672" "hermes-rabbitmq-notifications" $.Values.global.region $.Values.global.tld }}
+  RABBITMQ_URL: {{ printf "amqp://%s:%s@%s/" .Values.rabbitmq_notifications.users.default.user .Values.rabbitmq_notifications.users.default.password $host | b64enc }}
```
corey-aloia, line 16: I think this wont work on qa-de-2 as there used to be the logic:
```
{{ if .Values.config.cronusAuditSink.host }}{{ .Values.config.cronusAuditSink.host }}{{ else }}{{ $host }}{{ end }}
```

I think only qa-de-2 has `cronusAuditSink`. But looking at the value, and maybe it does work on qa-de-2 as the `cronusAuditSink` is similar. Lets try it like this, but just keep in mind that for qa-de-2 it could fail. In which case we can think of using a `hermesHost` in the values file or something.

openstack/cronus/templates/nebula/secrets.yaml
```diff
@@ -2,16 +2,31 @@
...
-    cloud.sap/skip-template: "true"
   labels:
     app: nebula
     chart: nebula
     release: nebula
     heritage: {{ .Release.Service }}
 type: Opaque
 data:
-  config.yaml: {{ include "cronus/templates/nebula/_config.yaml.tpl" . | b64enc }}
+  SENTRY_DSN: {{ .Values.nebula.sentryDsn | b64enc }}
```
corey-aloia, line 14: If I compare this with the config tpl above, most of these were wrapped in if statements:
`{{- if .Values.pki.enabled }}`
`{{- if .Values.notifier.enabled }}`
`{{- if .Values.nebula.sentryDsn }}`
`{{- if .Values.hermes }}`

But I think without these values cronus would never work. So having this in an if likely never made sense, as they are required properties. But @sapmli would know better

openstack/cronus/templates/rhea/secrets.yaml
```diff
@@ -11,5 +11,11 @@ metadata:
     heritage: {{ .Release.Service }}
 type: Opaque
 data:
-  config.yaml: {{ include "cronus/templates/rhea/_config.yaml.tpl" . | b64enc }}
-{{- end -}}
+  RABBITMQ_URI: {{ printf "amqp://%s:%s@%s/" .Values.rhea_rabbitmq.users.default.user .Values.rhea_rabbitmq.users.default.password .Values.rhea_rabbitmq.host | b64enc }}
+  {{- if .Values.config.keystone }}
```
corey-aloia, line 15: Same here... can rhea work without a keystone password? And can it work without a client_cert_ou?

corey-aloia (changes requested):

openstack/cronus/templates/nebula/secrets.yaml
corey-aloia: Looking at the nebula PR, and it indeed won't work without these there, so lets leave it how you have it. I will resolve my comment

corey-aloia (approved):

PR #6323, opened 2024-04-09
[Neutron] Move audit middleware to secrets

sven-rosenzweig:
Move audit middleware to secrets


sebageek (approved):
Would have deleted a second `\n` above, but works this way as well ;)

sebageek (approved):

PR #6326, opened 2024-04-09
[greenhouse] Extend alerts plugin adding values to the silenceTemplates

ArtieReus:


system/greenhouse-ccloud/templates/_silence.conf.tpl
```diff
@@ -0,0 +1,170 @@
+templates:
```
IvoGoman, line 1: The `templates:` makes this a `map` but the PluginDefinition requires a `list`

ArtieReus: You are right! I didn't see it. Changed

IvoGoman (approved):

PR #6332, opened 2024-04-10
[neutron] Allow to disable network agent locally

swagner-de:
The network agents are build, one deployment per apod from
$.Values.global.apods. This implements a temporary possibility to
disable an apod without having to touch the global variables, just for
the neutron deployment.


sebageek (approved):
Would probably make sense to sort this option under `agents` at some point, but let's go with this to get our ci up and running again.

PR #6337, opened 2024-04-10
[octavia] Remove driver-agent side-car because we do not use it anymore.

velp:
This PR removes `driver-agent` from Octavia because we disable it in https://github.com/sapcc/octavia-f5-provider-driver/pull/251


BenjaminLudwigSAP:
Rebased on master

PR #6341, opened 2024-04-11
[manila] increase max_shares_per_share_server to 50

Carthaca:
SVM migrate is supporting this, so we allow more volumes in
a vserver to avoid scattering volumes of the same customers
into too much different vservers on the same host.

This is especially interesting for customers that are creating many,
but small volumes.


kpawar-sap (approved):

chuan137 (approved):
lgtm

PR #6345, opened 2024-04-11
[barbican] upgrade mariadb to 0.10.1

businessbean:
- chart version bumped too


rajivmucheli:
Would ccroot need to be enabled by the service owner ?

businessbean:
> Would ccroot need to be enabled by the service owner ?

Hi Rajiv, yes it should be your decision which feature you want to enable.

rajivmucheli (approved):
lgtm

PR #6350, opened 2024-04-11
[keystone] upgrade mariadb to 0.10.1

businessbean:
- chart version bumped too

### DIFF
--- a/openstack/keystone/Chart.lock
+++ b/openstack/keystone/Chart.lock
@@ -1,7 +1,7 @@
 dependencies:
 - name: mariadb
   repository: https://charts.eu-de-2.cloud.sap
-  version: 0.9.2
+  version: 0.10.1
 - name: mariadb-galera
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
   version: 0.22.1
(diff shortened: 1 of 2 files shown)


rajivmucheli:
Would ccroot_user need to be enabled by the service owner ?

bbobrov:
Thanks. Some time ago we had an issue during upgrading because of mariadb. I suggest we schedule a time to deploy this, so that we all were ready to act if it happens again.

businessbean:
> Would ccroot_user need to be enabled by the service owner ?

Yes, you should decide which feature you want to enable.

businessbean:
> Thanks. Some time ago we had an issue during upgrading because of mariadb. I suggest we schedule a time to deploy this, so that we all were ready to act if it happens again.

Hi Boris, it is still MariaDB version [10.5.17](https://github.com/sapcc/helm-charts/blob/master/common/mariadb/values.yaml#L7), but yes let's monitor the rollout. Later we can be more confident because of the automated Galera integration tests.

bbobrov (approved):

bbobrov:
If mariadb version is not getting bumped, i am ok to deploy this normally, without special planning.

PR #6346, opened 2024-04-11
[cinder] upgrade mariadb to 0.10.1

businessbean:
- chart version bumped too

PR #6358, opened 2024-04-11
[Cinder] Enable ccroot user for mariadb

joker-at-work:
This user provides passwordless access to the DB from localhost, so we can administer the DB without having to read and rotate a secret.

With bumping the mariadb chart we also get some fixes regarding Secrets in.


businessbean:
https://github.com/sapcc/helm-charts/pull/6346 is related

Scsabiii (approved):

hemna (approved):

PR #6361, opened 2024-04-11
[utils] add helper to resolve secrets that need urlquery function

galkindmitrii:
- with a new secret-injector we cannot use urlquery over new secret vault references.
- with a new helper function we work around it by adding 'resolve' on the fly with optional urlquery at the end where it is needed.
- this patch also uses new helper for places where we had urlquery before. This would allow to simplify the values inside the secrets repo.

### DIFF
--- a/openstack/utils/Chart.yaml
+++ b/openstack/utils/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: utils
-version: 0.16.2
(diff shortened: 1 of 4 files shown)


s10 (approved):

Carthaca (approved):
I haven't used the merge and pick functions before, but thankfully the official documentation of the helm templating functions is pretty good so I could hopefully grasp what is done here with the dicts and the input validation.

PR #6362, opened 2024-04-11
[neutron] Enable ccroot user for db access

sebageek:


sebageek:
Closing in favor of  #6406

PR #6364, opened 2024-04-11
[archer] bump dependencies, switch to secrets-injector

notandy:
This PR switches archer to secrets injector.
Needs to be rolled out along with https://github.wdf.sap.corp/cc/secrets/pull/14154


m-kratochvil (approved):
lgtm

PR #6377, opened 2024-04-15
[octavia] Create secret as pre-upgrade hook (for migration job)

BenjaminLudwigSAP:
The migration job runs as a pre-upgrade hook and it needs the octavia-secrets secret. Therefore this secret needs to be created at pre-upgrade hook time as well, and with a lower weight than the migration job, so that it is guaranteed to exist before the migration job runs.


BenjaminLudwigSAP:
Force-pushed for better commit message

PR #6378, opened 2024-04-15
[manila] enable mariadb ccroot user

Carthaca:
This user provides passwordless access to the DB from localhost, so we
can administer the DB without having to read and rotate a secret.


grandchild (approved):
Compared with nova helm-chart. mariadb chart dependency version is up to date & config looks good.

chuan137 (approved):
lgtm

PR #6380, opened 2024-04-15
[openstack] helm charts with underscore renamed to dash

businessbean:
- to avoid naming violations in oci registries
- https://helm.sh/docs/chart_best_practices/conventions/#chart-names
- https://helm.sh/docs/chart_best_practices/conventions/#version-numbers
- helm dep update run for all related parent charts
- alias added to keep the old chart dependency variable names and to avoid names with dashes
- recent helm chart version pushed to Keppel with the new name


joker-at-work (commented):
Why do we need the `alias: mysql_metrics`? Sorry, it's in the description.
> alias added to keep the old chart dependency variable names and to avoid names with dashes

businessbean:
> Why do we need the `alias: mysql_metrics`? Sorry, it's in the description.
> 
> > alias added to keep the old chart dependency variable names and to avoid names with dashes
> 
> We already have names with dashes for the `owner-info` everywhere, I guess.

Yes, but sometimes it leads to ugly templating code if the variable contains a dash

joker-at-work:
> > Why do we need the `alias: mysql_metrics`? Sorry, it's in the description.
> > > alias added to keep the old chart dependency variable names and to avoid names with dashes
> > 
> > 
> > We already have names with dashes for the `owner-info` everywhere, I guess.
> 
> Yes, but sometimes it leads to ugly templating code if the variable contains a dash

Doesn't that mean, that the `mysql-metrics` chart requires any user of it to set settings with `mysql_metrics` instead? That's counter-intuitive and should be mentioned in the README in big fat red letters. (I assume the templating for `mysql_metrics` values happens mainly/only in the `mysql-metrics` chart.)

businessbean:
> > > Why do we need the `alias: mysql_metrics`? Sorry, it's in the description.
> > > > alias added to keep the old chart dependency variable names and to avoid names with dashes
> > > 
> > > 
> > > We already have names with dashes for the `owner-info` everywhere, I guess.
> > 
> > 
> > Yes, but sometimes it leads to ugly templating code if the variable contains a dash
> 
> Doesn't that mean, that the `mysql-metrics` chart requires any user of it to set settings with `mysql_metrics` instead? That's counter-intuitive and should be mentioned in the README in big fat red letters. (I assume the templating for `mysql_metrics` values happens mainly/only in the `mysql-metrics` chart.)

I have added the aliases to avoid having to change templating code like that
- https://github.com/sapcc/helm-charts/blob/2e340c853c162c1b4c2e14b41247f61c7f567b39/openstack/nova/values.yaml#L335
- https://github.com/sapcc/helm-charts/blob/2e340c853c162c1b4c2e14b41247f61c7f567b39/common/mysql_metrics/templates/etc/_config.yml.tpl#L6

It also removes the demand for hard to read `index` code in the templates
- https://github.com/helm/helm/issues/2192#issuecomment-293739257
- https://helm.sh/docs/chart_best_practices/values/
```text
chicken-noodle-soup: true # do not use hyphens in the name
```

joker-at-work:
Do as you like, it looks inconsistent to me and people stumble over inconsistencies like these, but I won't pressure that topic any more.

BerndKue:
Will the service owner of designate correct the linting error first?
`Error: failed linting charts: failed building dependencies for chart "designate => (version: \"0.2.2\", path: \"openstack/designate\")": failed waiting for process: exit status 1

failed linting charts: failed building dependencies for chart "designate => (version: \"0.2.2\", path: \"openstack/designate\")": failed waiting for process: exit status 1
`

businessbean:
> Will the service owner of designate correct the linting error first?
> 
> ```
> Error: failed linting charts: failed building dependencies for chart "designate => (version: \"0.2.2\", path: \"openstack/designate\")": failed waiting for process: exit status 1
> 
> failed linting charts: failed building dependencies for chart "designate => (version: \"0.2.2\", path: \"openstack/designate\")": failed waiting for process: exit status 1
> ```

Hi Bernd, yes we are working on that.

sandzwerg:
Can someone give a current state? I'm reluctant to ACK this as the chart fails to validate

businessbean:
closed because of review request bombing

PR #6382, opened 2024-04-16
[octavia] Fix creating secret as pre-upgrade hook

joker-at-work:
We tried to make the `octavia-secrets` Secret part of the pre-upgrade phase so the migration job can run as pre-upgrade hook in ce8cfe40fbe99c76de1bfe1d93f05ee670d88c86. This didn't work, because we put in the annotations as labels instead.


BenjaminLudwigSAP (approved):
Thank you!

PR #6383, opened 2024-04-16
[rabbitmq] Add standardised labels to all objects

dusandordevicsap:
Add standardised labels to all objects, according to https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

### DIFF
--- a/common/rabbitmq/CHANGELOG.md
+++ b/common/rabbitmq/CHANGELOG.md
@@ -6,7 +6,7 @@ This file is used to list changes made in each version of the common chart rabbi
 0.6.13
 ------
 dusan.dordevic@sap.com
--Adding standardized labels to all objects, according to https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels
+-Adding standardized labels to all objects, according to https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels and https://helm.sh/docs/chart_best_practices/labels/
 
 0.6.9
 -----
--- a/common/rabbitmq/CHANGELOG.md
+++ b/common/rabbitmq/CHANGELOG.md
@@ -3,6 +3,11 @@ Rabbitmq CHANGELOG
 
 This file is used to list changes made in each version of the common chart rabbitmq.
 
+0.6.13
+------
+dusan.dordevic@sap.com
+-Adding standardized labels to all objects, according to https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels
+
 0.6.9
 -----
 b.alkhateeb@sap.com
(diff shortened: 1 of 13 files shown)


common/rabbitmq/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rabbitmq
-version: 0.6.12
+version: 0.6.13
```
businessbean, line 3: Please bump the version to `0.7.0` because the PR introduces a new feature and not just a bugfix

common/rabbitmq/templates/_helpers.tpl
```diff
@@ -87,3 +87,12 @@ rabbit://{{ default "" $envAll.Values.global.user_suffix | print $rabbitmq.users
...
                 - reinstalling
 {{- end }}
+
+{{- define "sharedservices.labels" }}
+app.kubernetes.io/name: {{ .Chart.Name }}
+app.kubernetes.io/instance: {{ .Chart.Name }}-{{ .Release.Name }}
+app.kubernetes.io/version: {{ .Chart.Version }}
+app.kubernetes.io/component: RabbitMQ
+app.kubernetes.io/part-of: {{ .Release.Name }}
+app.kubernetes.io/managed-by: Helm
```
businessbean, line 97: Helm [sets this label](https://github.com/helm/helm/blob/14d0c13e9eefff5b4a1b511cf50643529692ec94/pkg/chartutil/create.go#L503) on its own. Please remove it to avoid patch/merge errors during upgrades. Helm does not configure that label for pods, but i think this is correct because they are not directly managed by Helm.

common/rabbitmq/CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ Rabbitmq CHANGELOG
 
 This file is used to list changes made in each version of the common chart rabbitmq.
 
+0.6.13
+------
+dusan.dordevic@sap.com
+-Adding standardized labels to all objects, according to https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels
```
businessbean, line 9: Please add the [Helm best practices for labels and annotations](https://helm.sh/docs/chart_best_practices/labels/) link too.

businessbean (changes requested):
lgtm. Just some minor changes are required.

common/rabbitmq/templates/_helpers.tpl
dusandordevicsap: Removed app.kubernetes.io/managed-by: {{ .Release.Service }}

common/rabbitmq/CHANGELOG.md
dusandordevicsap: Done

businessbean (approved):
lgtm

PR #6396, opened 2024-04-18
Add support of `reloader` annotations in postgresql and redis

stanislav-zaprudskiy:
The PR adds values to control [`reloader`](https://github.com/stakater/Reloader) annotations:
* of generated (user) password Secrets in `postgresql-ng` and `redis` charts
* and of the consuming Deployments in `pgbackup`, `pgmetrics`, `redis/metrics-deployment`

When both are configured, would cause the the consuming Deployments (`pgbackup`, `pgmetrics`, `redis/metrics-deployment`) to be restarted when the password secret changes (is regenerated) in `postgresql-ng` and `redis`.

### DIFF
--- a/common/pgbackup/Chart.yaml
+++ b/common/pgbackup/Chart.yaml
@@ -1,5 +1,8 @@
 name: pgbackup
-version: 1.1.12 # this version number is SemVer as it gets used to auto bump
+# 1.2.0
+# - Added `.Values.reloader.enabled` to control consumption of
+#   `stakater/Reloader` annotations on referenced Secrets and CMs.
+version: 1.2.0 # this version number is SemVer as it gets used to auto bump
 apiVersion: v2
 description: PostgreSQL backup/restore pod
 appVersion: "v0.10.0-111-gb31fd82" # of https://github.com/sapcc/backup-tools
--- a/common/pgbackup/templates/deployment.yaml
+++ b/common/pgbackup/templates/deployment.yaml
@@ -11,6 +11,10 @@ kind: Deployment
 
 metadata:
   name: "{{ .Release.Name }}-pgbackup"
+  {{- if .Values.reloader.enabled }}
+  annotations:
+    reloader.stakater.com/search: "true"
+  {{- end }}
 
 spec:
   replicas: 1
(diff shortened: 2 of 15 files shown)


SuperSandro2000 (commented):
We need to move this at least behind an value which is by default off as this could cause quite substantial unintended side effects and could easily cause downtime when eg the PVC is not automatically attaching.

majewsky:
> No description provided. 

Please provide a description. Which types of changes shall be triggered on, and why can this not be achieved with the conventional checksum mechanism?

stanislav-zaprudskiy:
I updated it to make the reloader turned off by default, and updated the PR description.

majewsky:
> For postgresql-ng it could be useful when Secret with password is generated externally (e.g. in umbrella chart) and was rotated. 

This is not supported.

stanislav-zaprudskiy:
> > For postgresql-ng it could be useful when Secret with password is generated externally (e.g. in umbrella chart) and was rotated.
> 
> This is not supported.

Is the rotation or an external Secret not supported? Why? Would be great to have it declared in the chart.

I have a case where the application working with `postgresql-ng` is managed by a CR and an operator. The CRD expects a database password to be provided to it as a Secret of a certain format, thus it needs to be known upon deployment. If not generating/setting the password in the umbrella chart of the app (to set it for both - the CR and `postgresql-ng`), it would then require to have `postgresql-ng` as a separate release happening before the app chart release, while the secret would be looked up and used during the release of the latter. I found the first option easier to deal with. But I'm open for suggestions.

majewsky:
The workflow implemented by the chart is that an init container on the Postgres deployment generates passwords for all users and writes them into Secrets if not done yet. The consuming charts then reference those secrets ([example](https://github.com/sapcc/limes/pull/431)). If you require a different naming convention somewhere there because the respective logic is not accessible to changes in your operator, we can make that configurable.

Our expectation is that, if password rotation is required, the respective secrets can be deleted, the Postgres container restarted, then all consumers restarted. This is a manual process, so no operator support would be needed.

stanislav-zaprudskiy:
@SuperSandro2000 @majewsky would you have a look at this? It's been re-purposed a bit - see the description and changes. 

Also, just for reference, I'm linking https://github.com/sapcc/helm-charts/pull/6465 to this.

majewsky (commented):
Yeah, we can do this. Please add `Chart.yaml` minor version bumps to all charts, then we can merge this.

majewsky (approved):

PR #6400, opened 2024-04-18
[ironic] use custom start/end string for jinja

sandzwerg:
Switch the variable string for jinja to something custom so there is no
clash with go-templates and helm tries to resolve things that are not
helm-related. We use the same string as cinder/nova.


joker-at-work (changes requested):
This will not work, because this is neither a `KosTemplate` nor a `VCenterTemplate`.

PR #6404, opened 2024-04-18
[ironic] change ipxe back to configmap

sandzwerg:
The ipxe config does not contain any secrets but handling it as a
k8s secret breaks helm as it will interpret jinja syntax that is
supposed to be interpreted by ironic later on.


joker-at-work (dismissed):

stanislav-zaprudskiy (commented):
After removing the Secret, aren't there any references to it left in other templates?

stanislav-zaprudskiy (approved):

BerndKue (approved):

PR #6405, opened 2024-04-18
[Ironic] Put the ipxe_config.template in a raw string

joker-at-work:
The `ipxe_config.template` is a jinja2 template interpreted by Ironic code. Since both Helm and jinja2 use `{{` and `}}` as deliminator for variables in the template, helm tries to interpret the jinja2 template. To prohibit that, we "escaped" the `{{` and `}}` by telling helm to put strings of `{{` and `}}` into the template.

With us not using much helm templating in this file, there's a better readable way to do it: raw strings. We put basically the whole file into a raw string, only ending the raw string where we need actual templating from helm, starting it again right afterwards.

See also https://go.dev/ref/spec#String_literals and https://pkg.go.dev/text/template#hdr-Examples

### DIFF
--- a/openstack/ironic/templates/etc/_ipxe_config.template.tpl
+++ b/openstack/ironic/templates/etc/_ipxe_config.template.tpl
@@ -1,7 +1,7 @@
 {{- define "ipxe_config_template" -}}
 {{- $conductor := index . 1 -}}
 {{- with index . 0 -}}
-#!ipxe
+{{`#!ipxe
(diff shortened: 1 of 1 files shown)


joker-at-work:
Rendering the chart before and after locally, the `diff` on the result showed that they're the same.

sandzwerg (approved):

fwiesel (approved):

PR #6407, opened 2024-04-19
fix references to already existing resources

databus23:


defo89 (approved):
lgtm, just chart bump was missing

PR #6412, opened 2024-04-19
[octavia] Fix pending LBs metric

BenjaminLudwigSAP:
The second selection (after UNION) is missing a field to compensate for the loadbalancer_project field introduced in #6233


notandy (approved):
lgtm

PR #6413, opened 2024-04-19
[maia] add api alert playbook

notque:


viennaa (approved):

viennaa (changes requested):
Thanks for adding the playbook, bump the chart version pls and it is good to go from my POV

PR #6414, opened 2024-04-22
[Keystone] Enabled Linkerd for jobs

rajivmucheli:


bbobrov:
Could you please describe here on in the commit message what specifically it does? I have no idea what this is about, although it directly affects keystone

bbobrov:
Thanks for the description. Do we also need a bump of the chart version since the chart is getting changed?

rajivmucheli:
For this commit, utils 0.14.6 and above is required. The current utils version for keystone is 0.15.0

bbobrov:
Right. But do we need to bump the version of the keystone chart, because the keystone chart is getting changed?

joker-at-work (changes requested):
Talked to Rajiv already that this most likely doesn't work.

rajivmucheli:
https://github.com/sapcc/helm-charts/pull/6460/files

PR #6423, opened 2024-04-23
common/alertmanager: config selector added

richardtief:


viennaa (approved):
Thanks for tidying up 👍

PR #6424, opened 2024-04-23
openstack/limes-auto: init at 0.1.0

majewsky:
The intention here is to deliver quota-overrides.json from the secrets repo into prod without human intervention (only gated on validation). Quota overrides are supposed to be maintained by the respective service owners (e.g. the network-api team for Neutron and Octavia quotas), so involving us in the deployment is an unnecessary complication.

This is a different deployment pattern than Limes itself, so it needs to be located in a different chart.

This first step does not do anything on its own, but after this chart is deployed, I will switch Limes to using the quota-overrides.json from here instead of from its own chart.

### DIFF
--- a/openstack/limes-auto/Chart.lock
+++ b/openstack/limes-auto/Chart.lock
@@ -1,0 +1,6 @@
+dependencies:
+- name: owner-info
+  repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
+  version: 0.2.3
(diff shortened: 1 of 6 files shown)


SuperSandro2000 (approved):

VoigtS (approved):

majewsky:
Extended with changes in the existing seed to add a new validation user for CI.

PR #6427, opened 2024-04-24
[memcached] Add standardised labels to objects

dusandordevicsap:
Add standardised labels to all Kubernetes objects according to 
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels https://helm.sh/docs/chart_best_practices/labels/


businessbean (approved):
lgtm

PR #6428, opened 2024-04-24
Add coredns rewrite config

jknipper:


Nuckal777:
iirc, this did not work.

PR #6433, opened 2024-04-24
[ironic] fix liveness probe with secrets

sandzwerg:
- the credentials that the liveness probe was using are no longer in the
  ironic.conf but now in the secrets.conf
- remove the vall to rabbitmq to not leak RO rabbitmq credentials to
  anyone having k8s read access to the pod. The openstack-agent-liveness
  should be enough for the liveness check

### DIFF
--- a/openstack/ironic/templates/_conductor-deployment.yaml.tpl
+++ b/openstack/ironic/templates/_conductor-deployment.yaml.tpl
@@ -94,10 +94,7 @@ spec:
         livenessProbe:
           exec:
             command:
-            - bash
-            - -c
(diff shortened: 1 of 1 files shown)


fwiesel (approved):
We should probably change eventually the "openstack-agent-liveness" to follow the openstack conventions of loading files.

joker-at-work:
> We should probably change eventually the "openstack-agent-liveness" to follow the openstack conventions of loading files.

I shortly looked into that and afaics this would mean getting some of the cmdline options to then tell oslo.config which service we belong to and by that in which directories to look for config files.

joker-at-work (approved):
There's a typo in the commit message: "vall" -> "call".

sandzwerg:
Fixed the typo, would love if the agent-liveness would follow the OpenStack convention of loading files.

fwiesel:
Here you go: https://github.com/sapcc/python-agentliveness/pull/5
Just missing one more review.

PR #6435, opened 2024-04-25
am-alertmanagerConfigMatcherStrategy

richardtief:
If set to `OnNamespace`, the operator injects a label matcher matching the  namespace of the AlertmanagerConfig object for all its routes and inhibition  rules. `None` will not add any additional matchers other than the ones specified in the AlertmanagerConfig. Default is `OnNamespace`.

### DIFF
--- a/common/prometheus-alertmanager-base/values.yaml
+++ b/common/prometheus-alertmanager-base/values.yaml
@@ -36,7 +36,7 @@ additionalPeers: []
 # If omitted it is generated automatically, which usually leads to different clusterLabels.
 clusterLabel: ccloud
 
-# OnNamespace or None to match all namespaces.
+# OnNamespace or None to match all namespaces. Default to None here to be independent from the deployed namespace since it would be locked to release namespace otherwise.
 alertmanagerConfigMatcherStrategy: None
 
 # Alertmanager mesh.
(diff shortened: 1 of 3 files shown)


common/prometheus-alertmanager-base/values.yaml
```diff
@@ -36,6 +36,9 @@ additionalPeers: []
 # If omitted it is generated automatically, which usually leads to different clusterLabels.
 clusterLabel: ccloud
 
+# OnNamespace or None to match all namespaces.
+alertmanagerConfigMatcherStrategy: None
```
viennaa, line 40: What is the reason for diverging from the default here?

viennaa (commented):
Thanks for this PR, I still need a little more context why this is needed.

common/prometheus-alertmanager-base/values.yaml
```diff
@@ -36,6 +36,9 @@ additionalPeers: []
 # If omitted it is generated automatically, which usually leads to different clusterLabels.
 clusterLabel: ccloud
 
+# OnNamespace or None to match all namespaces.
```
viennaa, line 39:
```suggestion
# OnNamespace or None to match all namespaces. Default to None here to be independent from the deployed namespace since it would be locked to prometheus-alertmanager otherwise.
```

common/prometheus-alertmanager-base/values.yaml
viennaa: Alright, adjusted comment a little bit.

viennaa (approved):
Great, thanks!

PR #6439, opened 2024-04-25
[keystone] disable disco

vssldmtrv:
Using disco results in doomed attemps to create DNS records in the wrong zone, e.g.
identity-3-qa.global.cloud.sap. in qa-de-1.cloud.sap.
or
identity-3.global.cloud.sap. in eu-de-1.cloud.sap.

These attempts will never succeed but are nevertheless retried over and over again.

Inspired by:
https://github.com/sapcc/helm-charts/commit/fc683cedc377d581f4154e41835b364130f60de6


vssldmtrv:
One more thing.

The keystone ingress in each region will try to have the regional DISCO operator register `identity-3.global.cloud.sap` as a CNAME to the default regional ingress hostname.

That's what we end up with:
EU-DE-1
```
1.7140546265485084e+09  INFO    disco   record does not exist in designate. creating it {"record": "monsoon3global/keystone-global", "zone": "eu-de-1.cloud.sap.", "name": "identity-3.global.cloud.sap.", "type": "CNAME", "records": "ingress.eu-de-1.cloud.sap.", "ttl": 1800}
```
EU-DE-2
```
1.7140547408795865e+09  INFO    disco   record does not exist in designate. creating it {"record": "monsoon3global/keystone-global", "zone": "eu-de-2.cloud.sap.", "name": "identity-3.global.cloud.sap.", "type": "CNAME", "records": "ingress.eu-de-2.cloud.sap.", "ttl": 1800}
```
EU-NL-1
```
1.7140547977877433e+09  INFO    disco   record does not exist in designate. creating it {"record": "monsoon3global/keystone-global", "zone": "eu-nl-1.cloud.sap.", "name": "identity-3.global.cloud.sap.", "type": "CNAME", "records": "ingress.eu-nl-1.cloud.sap.", "ttl": 1800}
```

Doesn't make a lot of sense.

bbobrov (approved):
I am fine with the change. But could it be something that keystone-global uses? @galkindmitrii could you please comment?

galkindmitrii:
We have dedicated ingress for global, but Disco does not know about it and thus tries to create records pointing to default internal ingress. We can disable it from my point of view.

galkindmitrii (approved):

vssldmtrv:
Alternatively, if one would really want to use disco then it could be done by assisting it, e.g.
```yaml
disco: "true"
disco/zone-name: "global.cloud.sap."
disco/record-type: "CNAME"
disco/record: "identity-3.ccloud-api-global.akadns.net."
```

That has some beauty to it as it ties the ingress definition and the corresponding DNS record more tightly together.

So, guys, it's up to you. You can merge and deploy this commit.
Or, if you want me to, I can prepare a new one implementing the suggestion above.

galkindmitrii:
> Alternatively, if one would really want to use disco then it could be done by assisting it, e.g.
> 
> ```yaml
> disco: "true"
> disco/zone-name: "global.cloud.sap."
> disco/record-type: "CNAME"
> disco/record: "identity-3.ccloud-api-global.akadns.net."
> ```
> 
> That has some beauty to it as it ties the ingress definition and the corresponding DNS record more tightly together.
> 
> So, guys, it's up to you. You can merge and deploy this commit. Or, if you want me to, I can prepare a new one implementing the suggestion above.

I think that won't work because we have 3 regions and aforementioned zone is only managed in one.

vssldmtrv:
Arghh, you're right of course.
Merge it is then.
I'll leave the deployment to you/whoever is responsible for the keystone deployments.

PR #6440, opened 2024-04-25
[manila-nanny] smoothen affinity alerts

Carthaca:
to avoid flapping because the metrics are not present on nanny restart
and take a while to be generated.

So we take rate([15m]) to look back over the values
of the last 15 minutes and remove duplicates by grouping by
share_id.


chuan137 (approved):
lgtm, thx!

kpawar-sap (approved):

PR #6442, opened 2024-04-25
[ironic] audit set default settings

sandzwerg:
We need to configure the central rabbitmq and set some sensible defaults
for other settings that we got from nova, apart from the queue size
which seems to be 100 for productive regions for most OpenStack
services.


notque (approved):
LGTM

BerndKue (approved):

PR #6443, opened 2024-04-26
plutono: secrets-injector adjustment

richardtief:
Moving ConfigMap content into a Secret to get the secrets injected properly.


viennaa (approved):
Thanks for moving this out of configmaps 👍

PR #6449, opened 2024-04-29
prometheus: lower retention across the releases

richardtief:


viennaa (approved):
go for standardisation!

PR #6451, opened 2024-04-29
[Limes] Align policies with elektra

VoigtS:
Please do not merge without approval from @majewsky !

This is a PR that aims to align the limes with the elektra policies.
Without the alignment we might have users with mismatching access rights that can access the API but not the UI or vise versa.

Done:
* Remove access rights not present in the elektra policies.
* Added a policy which exists in the elektra policies.

To note:
* The limes API rules for cluster/domain/project level that are not present in elektra are still available, since I'm not sure what exactly they are used for -> needs to be checked by the reviewer.


majewsky:
Outdated.

PR #6460, opened 2024-04-30
[Keystone] Enable linkerd jobs

rajivmucheli:
For using linkerd in Job-spawned Pods, we have to add a snippet at the
end that will tell the linkerd-proxy sidecar to stop. This snippet is
integrated - together with the proxysql snippet - into
"utils.script.job_finished_hook" in the utils chart starting with
version 0.14.6.

Since we need a version bump, we also get:

add owner info explicitly to coordination pvc
mount a trust-bundle at subpath
Ref : https://github.com/sapcc/helm-charts/commit/ed8e6ec64a39159a01105220d7512570419151c1

Closed PR : https://github.com/sapcc/helm-charts/pull/6414

### DIFF
--- a/openstack/keystone/templates/bin/_db-sync.tpl
+++ b/openstack/keystone/templates/bin/_db-sync.tpl
@@ -40,7 +40,7 @@ keystone-manage --config-file=/etc/keystone/keystone.conf --config-file=/etc/key
 echo "Keystone doctor:"
 keystone-manage --config-file=/etc/keystone/keystone.conf --config-file=/etc/keystone/keystone.conf.d/secrets.conf doctor
 
+{{- include "utils.script.job_finished_hook" . | trim }}
+
 # don't let the doctor break stuff (as usual not qualified enough and you allways need another opinion :P )
 exit 0
-
-{{- include "utils.script.job_finished_hook" . | trim }}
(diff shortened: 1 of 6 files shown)


openstack/keystone/templates/bin/_db-sync.tpl
```diff
@@ -41,5 +40,7 @@ keystone-manage --config-file=/etc/keystone/keystone.conf --config-file=/etc/key
 echo "Keystone doctor:"
 keystone-manage --config-file=/etc/keystone/keystone.conf --config-file=/etc/keystone/keystone.conf.d/secrets.conf doctor
 
+{{- include "utils.script.job_finished_hook" . | trim }}
```
joker-at-work, line 43: Same here. Need to switch `{{-` to `{{`.

joker-at-work (changes requested):
https://github.com/sapcc/helm-charts/blob/8740ab8d1c22b5db71516cce60e61713a1ffd3ec/openstack/keystone/templates/bin/_repair_assignments.tpl#L2 needs changing, too. contains the old proxysql shutdown stuff. I don't think it should contain it as it's run multiple times from the same cron-service-running pod.

joker-at-work (approved):

bbobrov (changes requested):
Please add a description to the pull request or to the commit messages describing why this change is necessary and what it does

bbobrov (approved):

PR #6461, opened 2024-05-01
Set deferred_deletion periodic callback value

kpawar-sap:


openstack/manila/templates/etc/_manila.conf.tpl
```diff
@@ -43,6 +43,8 @@ use_scheduler_creating_share_from_snapshot = {{ .Values.use_scheduler_creating_s
 
 is_deferred_deletion_enabled = {{ .Values.is_deferred_deletion_enabled | default false }}
 
+periodic_deferred_delete_interval == {{ .Values.periodic_deferred_delete_interval | default 300 }}
```
Carthaca, line 46:
```suggestion
periodic_deferred_delete_interval = {{ .Values.periodic_deferred_delete_interval | default 300 }}
```

should be an assignment

Carthaca (changes requested):

Carthaca (approved):

chuan137 (approved):

PR #6463, opened 2024-05-02
removing pre auto-update quota alerts

viennaa:
they are no longer needed since limes is auto increasing now. Leaving 99% as a safeguard


ashifnihalb (approved):
lgtm

PR #6464, opened 2024-05-02
nova/flavors: Add Sapphire-Rapids HANA flavors (v2)

grandchild:


joker-at-work:
So we use a new, non NUMA trait for the new saphire rapids after all? I was under the impression, we wouldn't do that. Does the buildup team know already?

grandchild:
We have to separate the hardware somehow. I looked into having nova talk to the cluster, check the hosts hypervisors' CPUID info and create a `COMPUTE_HW_` trait from that. But while possible it's not really implemented at all for VMware, and libvirt doesn't do hardware generations either (just AMD-SEV and Hyperthreading support is checked for the CPU). Doing it by NUMA size seems like mixing of concerns to me (what if a future hardware happens to have the same NUMA node size?).

No, the buildup team does not yet know. For the initial clusters, I will set the traits myself, and we can check if everything works as expected, before nailing down the process.

joker-at-work:
> Doing it by NUMA size seems like mixing of concerns to me (what if a future hardware happens to have the same NUMA node size?).

I'm annoyed by that as we had this topic in the last meeting and you didn't raise your concerns there. It's also in the meeting minutes if you want to check.

grandchild:
(sorry for the delay, laptop trouble)
i'm sorry to have annoyed you. :/ i must have missed that point in the discussion, it is in the minutes, as you say. i only thought of it this morning.

while i understand the frustration over lack of communication from my side, is there a technical argument against doing it this way? i would like to replace the custom trait with an automatic hardware trait in the long run, so that the buildup team doesn't have to be aware either, and we can simply match flavor generations with CPU generations.

joker-at-work:
No, nothing. It's more work and needs integration into tolling, but nothing technical.

grandchild:
After a quick talk with David I changed it to the NUMA trait version after all. Sorry again for the back-and-forth.
We can change it later to a HW-trait based solution if we want to and take the time.

grandchild:
We discussed that it's not _that_ important, which way we do it (NUMA trait, or manual HW trait), it's more important that we do it soon.
What tipped the scale was that:
1. we have experience with the NUMA-trait
2. we have to have NUMA traits anyway (2 different hw-NUMA configurations)

joker-at-work (commented):
asked a message in Slack

joker-at-work (approved):

dhoeller (approved):
Lgtm

PR #6465, opened 2024-05-02
postgres-ng: allow using custom secrets

SuperSandro2000:


common/postgresql-ng/templates/deployment.yaml
```diff
@@ -53,7 +53,7 @@ spec:
         - name: DEPLOYMENT_NAME
           value: {{ $deployment_name }}
         - name: USERS
-          value: {{ keys (.Values.users | required ".Values.users must be configured") | sortAlpha | join " " | quote }}
+          value: "{{ range $user, $settings := (.Values.users | required ".Values.users must be configured") }}{{ if and (typeIs "map[string]interface {}" $settings) (hasKey $settings "secretName") }}{{ $user }} {{ end }}{{ end }}"
```
majewsky, line 56: This will be unstable across different template runs and produce noise in the diffs. We should still iterate ove `keys (...) | sortAlpha`. I suggest moving that entire expression into a separate variable to improve readability.

Also, the template function `and` does not short-circuit like `&&` does, so this should use layered conditions to be safe. All in all:

```
{{- $allUserNames := keys (.Values.users | required ".Values.users must be configured") | sortAlpha }}
value: "{{ range $user := $allUserNames }}{{ $settings := index $.Values.users $user }}{{ if typeIs "map[string]interface {}" $settings }}{{ if hasKey $settings "secretName" }}{{ $user }} {{ end }}{{ end }}{{ end }}"
```

Yes, I realize that this is not very nice.

common/postgresql-ng/templates/deployment.yaml
```diff
@@ -87,8 +87,13 @@ spec:
         - name: USER_PASSWORD_{{ $user | replace "-" "_" }}
           valueFrom:
             secretKeyRef:
+        {{- if not (and (typeIs "map[string]interface {}" $settings) (and (hasKey $settings "secretName") (hasKey $settings "secretKey"))) }}
               name: {{ $.Release.Name }}-pguser-{{ $user }}
               key: postgres-password
+        {{- else }}
+              name: {{ $settings.secretName }}
+              name: {{ $settings.secretKey }}
```
majewsky, line 95:
```suggestion
              key:  {{ $settings.secretKey }}
```

majewsky (changes requested):
General approval, specific concerns:

common/postgresql-ng/templates/deployment.yaml
```diff
@@ -53,7 +53,8 @@ spec:
         - name: DEPLOYMENT_NAME
           value: {{ $deployment_name }}
         - name: USERS
-          value: {{ keys (.Values.users | required ".Values.users must be configured") | sortAlpha | join " " | quote }}
+          {{- $allUserNames := keys (.Values.users | required ".Values.users must be configured") | sortAlpha }}
+          value: "{{ range $user := $allUserNames }}{{ $settings := index $.Values.users $user }}{{ if typeIs "map[string]interface {}" $settings }}{{ if hasKey $settings "secretName" }}{{ $user }} {{ end }}{{ end }}{{ end }}"
```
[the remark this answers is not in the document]
SuperSandro2000: > The type match feels brittle. 

This complete construct in helms extremely limited templating engine is very brittle...

common/postgresql-ng/templates/deployment.yaml
```diff
@@ -90,8 +91,13 @@ spec:
         - name: USER_PASSWORD_{{ $user | replace "-" "_" }}
           valueFrom:
             secretKeyRef:
+        {{- if not (and (typeIs "map[string]interface {}" $settings) (and (hasKey $settings "secretName") (hasKey $settings "secretKey"))) }}
```
majewsky, line 94: another typeIs, same comment as above

majewsky (changes requested):

common/postgresql-ng/templates/deployment.yaml
SuperSandro2000: Isn't that already covered by line 58?

common/postgresql-ng/templates/deployment.yaml
```diff
@@ -90,8 +91,13 @@ spec:
         - name: USER_PASSWORD_{{ $user | replace "-" "_" }}
           valueFrom:
             secretKeyRef:
+        {{ $settings := $settings | default (dict) }}{{- if not (and (hasKey $settings "secretName") (hasKey $settings "secretKey")) }}
```
majewsky, line 94:
```suggestion
        {{- $settings := $settings | default (dict) }}{{ if not (and (hasKey $settings "secretName") (hasKey $settings "secretKey")) }}
```

Also, if there is a `not` at the top level of the `if`, it usually reads better to flip the branches.

majewsky (commented):
Please bump the chart version and as below. Then we're good to go.

majewsky (approved):

PR #6471, opened 2024-05-06
switch to read only and dedicated service accoutn

viennaa:


richardtief (approved):
Thanks

richardtief:
@viennaa this is still open.

PR #6473, opened 2024-05-06
[manila] fix share from snapshot config option

Carthaca:
see https://github.com/sapcc/manila/blob/3aff6a6621bf265780db5ef3370741d1aac915e4/manila/share/driver.py#L47

### DIFF
--- a/openstack/manila/templates/shares/_share-netapp.conf.tpl
+++ b/openstack/manila/templates/shares/_share-netapp.conf.tpl
@@ -93,10 +93,10 @@ reserved_share_extend_percentage = 100
 reserved_share_extend_percentage = {{ $share.reserved_share_extend_percentage | default 25 }}
 {{- end }}
 
-{{- if eq 100 (int $share.reserved_snapshot_percentage)}}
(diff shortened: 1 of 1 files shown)


kpawar-sap (approved):

chuan137 (commented):
thx and nice pick. Just remind me pls, what is the reason for the if-branch for value 100?

Carthaca:
> seems that we have used the fallback value of the `reserved_share_percentage`. we are going to set different values for this option now?

yes, that has always been the intention to have a lower reserve for shares from snapshot - we don't allow shares from snapshots to be scheduled somewhere else, so we allow those, even when we don't want regular new shares any more

Carthaca:
> thx and nice pick. Just remind me pls, what is the reason for the if-branch for value 100?

see 9c6dfb878c8a4fb79e0fa8838b9b831d941c6b15

we are not using this anymore, the reason this was needed was because of the `min` function, which would have discarded the `100`

chuan137 (approved):
lgtm

PR #6474, opened 2024-05-06
tidying up

viennaa:
cleaning up not longer existing or moved stuff


richardtief (approved):
Thanks. Flush it away!

PR #6475, opened 2024-05-06
Ceph: Add limits to osd pods

databus23:
Otherwise the osd will end up taking all memory until the node crashes


databus23:
@defo89

SchwarzM (approved):
lgtm

PR #6483, opened 2024-05-07
enable data volume for vmware prometheis

viennaa:
avoid crashing pods due to eph. storage overusage


richardtief (approved):
Looks reasonable. Thanks.

PR #6484, opened 2024-05-07
[netapp-exporter] remove manila inode alerts

Carthaca:
Storage team already got a monitoring for this, reducing duplicates.

We (compute-storage-api) keep this just as a backup,
therefore dropped to info and only after 2 days.


chuan137 (approved):
lgtm

PR #6499, opened 2024-05-13
[netapp] move credentials to secret

chuan137:


Carthaca (approved):
as far as I can tell this follows https://github.wdf.sap.corp/cc/secrets-injector#multiline-nested-templates and should work

PR #6503, opened 2024-05-13
Update cephobjectstore-placement-pools.yaml

ubi2go:


sumitarora2786 (commented):
I guess if condition needs to be defined as we have "replicasPerFailureDomain" & " subFailureDomain" only for one pool type: az-ab-4, @dmitri to verify

defo89 (approved):
all checks passed, lgmt

PR #6517, opened 2024-05-15
ceph: override the default auth order

kayrus:
This is undocumented option, at least it's not listed in https://docs.ceph.com/en/latest/radosgw/config-ref/
But this option is available in ceph source code:

```
ceph$ grep -r rgw_s3_auth_order
src/rgw/rgw_auth_s3.h:    boost::split(result, cct->_conf->rgw_s3_auth_order,
src/common/options/rgw.yaml.in:- name: rgw_s3_auth_order
```

### DIFF
--- a/system/cc-ceph/Chart.yaml
+++ b/system/cc-ceph/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: cc-ceph
 description: A Helm chart for the Rook / Ceph Objects inside the Storage Clusters
 type: application
-version: 1.0.28
+version: 1.0.29
 appVersion: "1.14.3"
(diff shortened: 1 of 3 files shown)


databus23:
What does this option do?

kayrus:
according to the yaml in ceph: https://github.com/ceph/ceph/blob/4a592ad7fe8914d62131da5bb4259c61bba82a1d/src/common/options/rgw.yaml.in#L883-L894

> Order of authentication strategies to try for s3 authentication, the
>   allowed options are a comma separated list of engines external, local. The default
>   order is to try all the externally configured engines before attempting local
>   rados based authentication

PR #6524, opened 2024-05-16
update .mgr pool to 2 zone setup

ubi2go:


defo89 (changes requested):
Please increase the chart version in same PR

PR #6527, opened 2024-05-16
nova/flavors: Add general-purpose sapphire-rapids flavors

grandchild:
Use a new CUSTOM_HW_SAPPHIRE_RAPIDS trait to distinguish the new generation. The flavor may change in the future, or use a true automatic COMPUTE_ flavor from nova, but use a manual trait to get started.


joker-at-work (approved):
Since we have mulitple NUMA traits for the sapphire rapids flavors, we cannot really use that trait as `required`, because Nova doesn't support `in:` syntax afaics and general purpose flavors are supposed to spawn on any NUMA sized sapphire rapid HV.

dhoeller (approved):
LGTM

PR #6530, opened 2024-05-16
[rabbitmq] sharedservices.labels component name fixed

businessbean:
- defined functions are shared across all subcharts
- hardcoded values will lead to wrong results
- chart version bumped


sandzwerg (approved):

dusandordevicsap (approved):
LGTM

PR #6535, opened 2024-05-21
[ironic] Switch all dell nodes to redfish + virtual-media

fwiesel:
This script takes care of migrating all dell nodes (Except the resource class zh2mlx1.large) to the redfish interface and a virtual-media boot

### DIFF
--- a/openstack/ironic/py/ironic-node-update.py
+++ b/openstack/ironic/py/ironic-node-update.py
@@ -1,0 +1,143 @@
+from contextlib import contextmanager
+from ironicclient import client as bmclient
+
+# import openstack # To test it locally
+# os = openstack.connect()
+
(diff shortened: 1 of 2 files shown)


fwiesel:
Before merging this, we need to make sure that the monitoring is in place.

fwiesel:
Certificates are rolled out to all dell nodes exposing the Redfish API.
Left is to clarify if we are also okay from operational perspective.

BerndKue:
There is a monitoring available in Prometheus: https://prometheus-infra-collector.qa-de-1.cloud.sap/graph?g0.expr=redfish_certificate_isvalid%3D%3D0&g0.tab=1&g0.stacked=0&g0.show_exemplars=0&g0.range_input=1h

BerndKue (approved):

PR #6541, opened 2024-05-21
[alertmanager] add capacity-ops support_group

geisslet:
@richardtief @majewsky


viennaa (approved):
capacity wants alerts? here we go

PR #6542, opened 2024-05-21
Add ceph service slack_by_cc_service receiver

databus23:
Looking at the receiver it seems this only works for non qa regions.
What do I need to do also get the alerts for qa in a dedicated slack channel?


kayrus (approved):

richardtief:
Only `support_groups` get their QA notifications posted in the Slack channel. The service owner usually does not want QA notifications in their channels as they can get quite noisy. 
Solutions:
1. use the support group correctly
2. use Greenhouse Supernova Dashboard filtered by QA and Service

PR #6547, opened 2024-05-21
manila: switch to better throttling metric

SuperSandro2000:


Carthaca (approved):
lgtm, found the same recommendation at https://stackoverflow.com/a/67316429/4537625

PR #6559, opened 2024-05-22
postgres-ng: fix rol drop when upgrading from 12 to 16

SuperSandro2000:
see https://stackoverflow.com/questions/51256454/cannot-drop-postgresql-role-error-cannot-be-dropped-because-some-objects-depe

### DIFF
--- a/common/postgresql-ng/Chart.yaml
+++ b/common/postgresql-ng/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: postgresql-ng
-version: 1.0.25 # this version number is SemVer as it gets used to auto bump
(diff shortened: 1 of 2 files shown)


majewsky (approved):

majewsky:
From the "I'm only reading this properly after the fact" department:

```sql
where not exists (SELECT rolname FROM pg_roles WHERE rolname='pg_read_all_data')
```

Will this not always be false in PostgreSQL 16 because pg_read_all_data is a system role that always exists? I suppose we should rather match for

```sql
where exists (SELECT rolname FROM pg_roles WHERE rolname='poor_mans_pg_read_all_data')
```

if that is allowed by the syntax.

SuperSandro2000:
#6565

PR #6561, opened 2024-05-22
[cc3test]: skipped few inbound and outbound ports for linkerd to make compute and other tests working

ashifnihalb:
ports that are skipped 
22,443,9125


thgrs (approved):
lgtm

PR #6576, opened 2024-05-23
[vmware-monitoring] Add host system properties metric

kevin-fischer:
- [x] Check if the data is available in the vROps REST API
- Include the no swap DS metric


viennaa (changes requested):
Thanks, please increase patch version.

viennaa (approved):

PR #6592, opened 2024-05-28
[neutron] Enable IPv6 RAs in DHCP agent

sebageek:
Add option to configure IPv6 router advertisements for our DHCP agents.


occamshatchet (approved):
lgtm

PR #6595, opened 2024-05-29
[Barbican] Introduce nanny to move secrets between projects

rajivmucheli:
Ref : https://github.com/sapcc/barbican/pull/8


openstack/barbican/values.yaml
```diff
@@ -307,3 +307,14 @@ hsm:
...
     enabled: true
+
+# barbican nanny
+barbican_nanny:
+  replicas: 1
+  enabled: false
+  # debug mode - no script run in the containers, just a sleep loop
+  debug: false
+  # run the nanny every n minutes
+  interval: 1
```
joker-at-work, line 318: imho better: Sleep time in minutes between runs

openstack/barbican/templates/scripts-configmap.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: barbican-scripts
+  labels:
+    system: openstack
+    type: configuration
+    component: barbican
+
+data:
+  barbican-db-check.sh: |
+{{ include (print .Template.BasePath "/scripts/_barbican-db-check.sh.tpl") . | indent 4 }}
```
joker-at-work, line 12: Why is it called `db-check`? Currently, it contains the secret move. Would a more generic `barbican-nanny` maybe be more appropriate?

openstack/barbican/templates/barbican-nanny-deployment.yaml
```diff
@@ -0,0 +1,112 @@
...
+            cpu: "100m"
+        volumeMounts:
+        - name: barbican-scripts
+          mountPath: /scripts/barbican-db-check.sh
+          subPath: barbican-db-check.sh
+          readOnly: true
+        - name: barbican-etc
+          mountPath: /etc/barbican/barbican-api.conf
+          subPath: barbican-api.conf
+          readOnly: true
```
joker-at-work, line 87: Running `barbican-manage`, I don't think it will find `barbican-api.conf`. Afaik, `oslo.config` will use the name of the command (i.e. `barbican-manage.conf`) and the name of the project (i.e. `barbican.conf`) automatically.

openstack/barbican/templates/barbican-nanny-deployment.yaml
```diff
@@ -0,0 +1,112 @@
...
+        prometheus.io/scrape: "true"
+        prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
+        {{- end }}
+        {{- include "utils.linkerd.pod_and_service_annotation" . | indent 8 }}
+    spec:
+      {{ include "utils.proxysql.pod_settings" . | indent 6 }}
+{{- if .Values.barbican_nanny.db_secret_move.enabled }}
+      containers:
+      - name: db-check
+        image: {{ required ".Values.global.registry is missing" .Values.global.registry }}/loci-barbican:{{.Values.imageVersionBarbicanApi | default .Values.imageVersionBarbicanApi | required "Please set barbican.imageVersion or similar"}}
```
joker-at-work, line 49: We're not using `barbican.imageVersion` but tell the user to set it. Instead, we use `imageVersionBarbicanApi` defaulting to the same if it's not set. That doesn't make sense either.

openstack/barbican/templates/barbican-nanny-deployment.yaml
```diff
@@ -0,0 +1,112 @@
...
+        checksum/etc-configmap.conf: {{ include (print $.Template.BasePath "/etc-configmap.yaml") . | sha256sum }}
+        secrets-hash: {{ include (print $.Template.BasePath "/secrets.yaml") . | sha256sum }}
+        {{- if .Values.proxysql.mode }}
+        prometheus.io/scrape: "true"
+        prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
+        {{- end }}
+        {{- include "utils.linkerd.pod_and_service_annotation" . | indent 8 }}
+    spec:
+      {{ include "utils.proxysql.pod_settings" . | indent 6 }}
+{{- if .Values.barbican_nanny.db_secret_move.enabled }}
```
joker-at-work, line 46: Does that `if` make sense here? It would remove all the `containers` currently. I propose to remove it and only add inidividual `if`s once you have more than one container in this nanny. Then, you'll probably also want to move the `interval` config into the individual nannies.

openstack/barbican/templates/barbican-nanny-deployment.yaml
```diff
@@ -0,0 +1,112 @@
...
+      labels:
+        app: barbican-nanny
+        name: barbican
+        alert-tier: os
+        alert-service: barbican
+        {{- include "utils.topology.pod_label" . | indent 8 }}
+      annotations:
+        chart-version: {{.Chart.Version}}
+        checksum/etc-configmap.conf: {{ include (print $.Template.BasePath "/etc-configmap.yaml") . | sha256sum }}
+        secrets-hash: {{ include (print $.Template.BasePath "/secrets.yaml") . | sha256sum }}
```
joker-at-work, line 38: It would make sense to also add the hash of the `barbican-scripts` configmap here, because any change to the script should re-create this pod.

openstack/barbican/values.yaml
rajivmucheli: its 1min, updating secrets could be done any time know. Should i increase it ?

openstack/barbican/templates/scripts-configmap.yaml
rajivmucheli: i preferred `move-secrets` for container name as the pod name is barbican-nanny.

openstack/barbican/templates/barbican-nanny-deployment.yaml
rajivmucheli: could you please elaborate ? it works well in qa-de-3.

openstack/barbican/templates/barbican-nanny-deployment.yaml
rajivmucheli: we have plan to add vcenter-barbican nanny hence this was introduced.

openstack/barbican/templates/barbican-nanny-deployment.yaml
```diff
@@ -0,0 +1,113 @@
...
+      labels:
+        app: barbican-nanny
+        name: barbican
+        alert-tier: os
+        alert-service: barbican
+        {{- include "utils.topology.pod_label" . | indent 8 }}
+      annotations:
+        chart-version: {{.Chart.Version}}
+        checksum/etc-configmap.conf: {{ include (print $.Template.BasePath "/etc-configmap.yaml") . | sha256sum }}
+        checksum/etc-configmap.conf: {{ include (print $.Template.BasePath "/scripts-configmap.yaml") . | sha256sum }}
```
joker-at-work, line 38: This needs another key, `checksum/scripts-configmap.conf` sounds right to me.

openstack/barbican/templates/barbican-nanny-deployment.yaml
joker-at-work: Ok, but then this `if` needs to go below `containers:` as either vcenter or move-secrets or both can be enabled and enabling both would otherwise duplicate the `containers` key or enabling one would be missing the `containers` key.

openstack/barbican/templates/barbican-nanny-deployment.yaml
joker-at-work: If it runs well, maybe it doesn't need any setting from the `barbican-api.conf`. Did you check that any of these settings take effect?
To my knowledge, `barbican-api.conf` will not be consumed by an executable called `barbican-manage` unless explicitly given via `--config-file` option.

openstack/barbican/templates/barbican-nanny-deployment.yaml
```diff
@@ -0,0 +1,113 @@
...
+          requests:
+            memory: "250Mi"
+            cpu: "25m"
+          limits:
+            memory: "250Mi"
+            cpu: "100m"
+        volumeMounts:
+        - name: barbican-scripts
+          mountPath: /scripts/barbican-db-check.sh
+          subPath: barbican-db-check.sh
```
joker-at-work, line 83: I propose to rename the script to match what it does and the container name: `move-secrets.sh`.

openstack/barbican/values.yaml
joker-at-work: I was talking about the comment. It's not true that the nanny runs every minute if it takes 5min to run. It will then only run every 6min. Therefore, I think replacing the comment with "Sleep time in minutes between runs" would be more accurate.

openstack/barbican/templates/barbican-nanny-deployment.yaml
```diff
@@ -34,7 +34,6 @@ spec:
         {{- include "utils.topology.pod_label" . | indent 8 }}
       annotations:
         chart-version: {{.Chart.Version}}
-        checksum/etc-configmap.conf: {{ include (print $.Template.BasePath "/etc-configmap.yaml") . | sha256sum }}
         checksum/etc-configmap.conf: {{ include (print $.Template.BasePath "/scripts-configmap.yaml") . | sha256sum }}
```
joker-at-work, line 37: Removing the first `checksum/etc-configmap.conf` is an error. The second `checksum/etc-configmap.conf` should have been renamed to `checksum/scripts-configmap.conf`.

PR #6610, opened 2024-05-30
common/prometheus-server: add new alert label `incident_group_key`

richardtief:
Introducing a new alert label that makes it possible to group alerts in ServiceNow by alertname, region and cluster. This ensures that multiple incidents are no longer generated for one alert, which also led to multiple tickets and notifications etc.

Requested by @geisslet


viennaa (approved):
If this is needed on Service Now side, ok.

I suppose we should add it to the pre7 chart too, for consistency, no?

richardtief:
It's only Maia still on the pre7 chart. So no need to do this

PR #6612, opened 2024-05-30
mute all cc3test api down alerts in a region when keystone api is down

artherd42:


richardtief (changes requested):
Looks fine. Please bump the chart and then it's good to go.

richardtief:
Please reset the patch version if you increase the minor. `4.3.0`

richardtief (approved):
Thanks.

PR #6615, opened 2024-05-31
Fix false positive ceph alerts

databus23:
* disable MTU alert until we come up with a better query for it
* Split PGimbalance alert into two alerts for apod not storage ndoes
* exclude /boot from fs full alert


kayrus (approved):

sumitarora2786 (approved):
LGTM..

PR #6617, opened 2024-06-03
[DS] Switch External Domains to new Seeder

rajivmucheli:


rajivmucheli:
https://github.com/sapcc/helm-charts/pull/6669

PR #6619, opened 2024-06-03
Update cephcluster.yaml

sumitarora2786:
podantiaffinity is required rule when host network is used or when AllowMultiplePerNode is false.  good to have 2 mgr pods for HA.


defo89 (changes requested):
pls bump the chart version within the PR

PR #6630, opened 2024-06-04
upgrade prometheus across releases

richardtief:


viennaa (approved):
Thanks. Please specify the change brought in with this version bump roughly in the commit message when rebasing/squashing to make it visible in the commit history without looking up the version

PR #6631, opened 2024-06-04
[CODEOWNERS] mysql metrics

Carthaca:
adding current name with underscore until #6380 merged
and adding Birk and Bashar as owners


businessbean (approved):
lgtm

PR #6632, opened 2024-06-04
[ceph]: add metrics-server chart

kayrus:
Add metrics-server chart for the `kubectl top pod`


databus23 (changes requested):
I don't think the metrics-server should be sub chart of the cc-ceph chart.
We need to add this to kube-monitmoring greenhouse chart or a seperate greenhouse plugin

Nuckal777:
We also have the great greenhouse-extras-storage chart. metrics-server feels unrelated to ceph itself.

PR #6649, opened 2024-06-10
[ironic] Increase timeout for ramdisks

fwiesel:
The default of 15minutes is quite sporty for our use-case. Increasing the the timeout to 1.5hours to be on the safe-side.


fwiesel:
Sorry, need to fix it. That's not working

PR #6652, opened 2024-06-11
[keystone] blocklist dns_hostmaster, dns_zonemaster, dns_mailmaster

vssldmtrv:
A normal project admin should not be able to give themselves and others these roles.

The CoreDNS team is granting them per ticket:
https://documentation.global.cloud.sap/docs/customer/networking/dns/dns-start-roles/


rajivmucheli (approved):

bbobrov:
Would you please share some details about these roles: when were they created? Which of two keystones do they affect? Who designed the roles, and why they have no `cloud_` prefix in them? 

Please also do not forget that this is a public github, and links to company docs should be avoided

galkindmitrii (approved):

PR #6671, opened 2024-06-14
[mariadb]: tcp_keepalive options added

businessbean:
- to support network setups that close idle connections after some time
- default values kept at `0`
- documentation snippets added
- tested locally with `helm template`
- chart version bumped


bashar-alkhateeb (approved):
looks good !

fwiesel (approved):

PR #6674, opened 2024-06-14
[netapp] switch to harvest for maia

chuan137:
prepare aggregation rules for maia using harvest metrics 
fallback to old maia metrics from capacity exporter

rolled out manually to qa

### DIFF
--- a/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
+++ b/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
@@ -1,33 +1,76 @@
-
+# document: https://documentation.global.cloud.sap/docs/customer/monitoring/operational-metrics--maia/metrics-metrics/#nfs-manila---capacity
 groups:
   - name: manila-maia
     rules:
-
-    - record: netapp_volume_size_total:manila
-      expr: netapp_volume_size_total{app="netapp-harvest-exporter-manila"}
-
-    - record: netapp_volume_size_used:manila
-      expr: netapp_volume_size_used{app="netapp-harvest-exporter-manila"}
-
-    - record: netapp_volume_size_available:manila
-      expr: netapp_volume_size_available{app="netapp-harvest-exporter-manila"}
-
-    - record: netapp_volume_space_logical_used:manila
-      expr: netapp_volume_space_logical_used{app="netapp-harvest-exporter-manila"}
+    #
+    # capacity metrics
+    #
 
     # sum available_bytes and logical_used_bytes when only space reporting is enabled
     # details see https://operations.global.cloud.sap/docs/operation/storage/manila/300-logical-space/
-    - record: netapp_volume_total_bytes:maia2
-      expr: (
-        (netapp_volume_size_available:manila + netapp_volume_space_logical_used) * on (app, aggr, svm, volume) group_left()
-        netapp_volume_labels{is_space_reporting_logical="true", is_space_enforcement_logical="false"}
-        ) or netapp_volume_size_total:manila
+    - record: netapp_volume_total_bytes:pre
+      expr: label_replace((netapp_volume_size_available{app="netapp-harvest-exporter-manila"} + netapp_volume_space_logical_used)
+        * on (app, host, svm, volume) group_left() netapp_volume_labels{is_space_reporting_logical="true", is_space_enforcement_logical="false"}
+        or netapp_volume_size_total{app="netapp-harvest-exporter-manila"}, "vserver", "$1", "svm", "(.*)")
 
     # use logical used bytes when logical space reporting is enabled
+    - record: netapp_volume_used_bytes:pre
+      expr: label_replace(netapp_volume_space_logical_used{app="netapp-harvest-exporter-manila"}
+        * on (app, host, svm, volume) group_left() netapp_volume_labels{is_space_reporting_logical="true"}
+        or netapp_volume_size_used{app="netapp-harvest-exporter-manila"}, "vserver", "$1", "svm", "(.*)")
+
+    - record: netapp_volume_snapshot_reserved_bytes:pre
+      expr: label_replace(netapp_volume_snapshot_reserve_size{app="netapp-harvest-exporter-manila"} * on (app, host, svm, volume) group_left() netapp_volume_labels, "vserver", "$1", "svm", "(.*)")
+
+    - record: netapp_volume_snapshot_available_bytes:pre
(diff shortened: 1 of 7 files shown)


prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
```diff
@@ -36,19 +86,79 @@ groups:
...
+
+    - record: netapp_volume_inode_files_used:maia
+      expr: netapp_volume_inode_files_used:maia2 or netapp_volume_inode_files_used:maia1{project_id!=""}
+
+    - record: netapp_volume_inode_files_used_percentage:maia
+      expr: netapp_volume_inode_files_used_percentage:maia2 or netapp_volume_inode_files_used_percentage:maia1{project_id!=""}
+
+    - record: netapp_volume_total_ops:maia
+      expr: netapp_volume_total_ops:maia2 or netapp_volume_total_ops:maia1{project_id!=""}
+
```
Carthaca, line 164: `netapp_volume_avg_latency_miliseconds:maia` is missing

prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
```diff
@@ -1,33 +1,83 @@
...
+    - record: netapp_volume_write_latency_milliseconds:pre
+      expr: netapp_volume_write_latency{app="netapp-harvest-exporter-manila"} * on (filer, svm, volume) group_left(project_id, share_id, share_name, share_type) netapp_volume_labels{app="netapp-harvest-exporter-manila"}
+
+
+    #
+    # only keep necessary labels visible to users
+    #
+
+    # We need to take average of the metrics scraped from different exporters, to avodid exporting multiplied values
```
Carthaca, line 75: 'different exporters' here means having multiple pods of harvest-exporter that would export from the same filer, right?

Carthaca (changes requested):
In addition to the inline comments:

- the label replacement from svm to vserver feels clunky. Could this also be done by config in the harvester itself?
- I think the label `availability_zone` should be kept in maia
- I'm not sure about the label `aggregate`, I think we can do a scream test = i.e. not adding it and see if someone wants it back

prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
chuan137: maybe `multiple harvest worker` is easier to read. I meant that it is possible same filer is scraped by multiple harvest workers due to some errors in netappsd.

chuan137:
> In addition to the inline comments:
> 
> * the label replacement from svm to vserver feels clunky. Could this also be done by config in the harvester itself?

yes, it is possible to change the labels in harvest metric config files. But harvester uses `svm` for all metrics, I will need to override every config files  in this list  https://github.com/sapcc/helm-charts/blob/29c6eb1f25ff0ad2b3c8dc7761b44484ff82263c/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/etc/rest.limited.yaml#L5-L12
If you prefer, it's also doable

> * I think the labels `availability_zone` and `volume_type` should be kept in maia

ok

> * I'm not sure about the label `aggregate`, I think we can do a scream test = i.e. not adding it and see if someone wants it back

agree :)

prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
chuan137: I wanted to replace it with 'netapp_volume_read_latency_milliseconds:maia` and `netapp_volume_write_latency_milliseconds:maia`. Avg latency alone is probably not very interesting.

prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
Carthaca: Yes, that would make it clearer 👍

Carthaca:
> yes, it is possible to change the labels in harvest metric config files. But harvester uses svm for all metrics, I will need to override every config files in this list

I imagined it would be easier over there. Let's keep it like it is, then

prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
Carthaca: Makes sense. That means we need a documentation update

chuan137:
for example, https://prometheus-storage.qa-de-1.cloud.sap/graph?g0.expr=%7B__name__%3D~%22.*%3Amaia%22%2C%20share_name%3D%22CCQA97BWHDB01_DATA%22%7D&g0.tab=1&g0.display_mode=lines&g0.show_exemplars=0&g0.range_input=15m

prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
```diff
@@ -36,19 +79,78 @@ groups:
...
+      expr: netapp_volume_total_bytes:maia2 or sum (netapp_volume_total_bytes:maia1{project_id!=""}) without (app, instance, job, filer, aggregate, node, snapshot_policy, volume_state)
+
+    - record: netapp_volume_used_bytes:maia
+      expr: netapp_volume_used_bytes:maia2 or sum (netapp_volume_used_bytes:maia1{project_id!=""}) without (app, instance, job, filer, aggregate, node, snapshot_policy, volume_state)
+
+    - record: netapp_volume_available_bytes:maia
+      expr: netapp_volume_available_bytes:maia2 or sum (netapp_volume_available_bytes:maia1{project_id!=""}) without (app, instance, job, filer, aggregate, node, snapshot_policy, volume_state)
+
+    - record: netapp_volume_used_percentage:maia
+      expr: netapp_volume_used_percentage:maia2 or sum (etapp_volume_used_percentage:maia1{project_id!=""}) without (app, instance, job, filer, aggregate, node, snapshot_policy, volume_state)
```
Carthaca, line 135: typo `(etapp_volume` ..., similarly below

chuan137: done

Carthaca (approved):
🌽 👍

PR #6675, opened 2024-06-14
create default rgw pools

### LINKED ISSUE #6670: manual assignment of crushrule for default radosgw pools
the radosgw creats default pools if they are not exists. 
The pool creation should be controlled by a helm chart:

create a helm template for:
```.rgw.root
<zone>.rgw.control
<zone>.rgw.meta
<zone>.rgw.log
<zone>.rgw.buckets.index
<zone>.rgw.buckets.data
<zone>.rgw.buckets.non-ec
```
pool settings:

- all pools have to use the same crush_rule wie defined for region-4: rgw.buckets.index 
-  .rgw.root,rgw.control, rgw.meta pool should have 8 pgs 
- rgw.log pool should have 32 pgs
- .rgw.buckets.index and .rgw.buckets.data and .rgw.buckets.non-ec should have 8 pgs

https://github.com/sapcc/helm-charts/tree/master/system/cc-ceph/templates

sources:
https://docs.ceph.com/en/latest/radosgw/layout/#appendix-compendium

arttor:
Closes #6670


whutsap (approved):

kayrus (approved):

sumitarora2786 (changes requested):
failureDomian = host
and
subfailureDomain also = host

is it correct?

if we want to keep the crush rule for default pools equal to replicated size = 3, and wants replicas to spread across the availibility zones, crushRoot will be = Default/region, in that case failuredomain should be zone and subfailuredomain should be host, no?

arttor:
@sumitarora2786 i've changed default failureDomain to zone and set size to 4 which should be more correct. But the thing is that this pools are disabled by default and @ubi2go  planned to override all values properly for each environment in kube-secrets repo.

sumitarora2786 (changes requested):
replicasPerFailureDomain: 1 should be 2

ubi2go:
the naming of the pool is wrong, the prefix was hidden in the issue by md.
zonename is https://github.wdf.sap.corp/cc/kube-secrets/blob/master/values/helm/storage/na-us-2/st1-na-us-2/cc-ceph.yaml#L12

arttor:
@ubi2go done, added `objectstore.name` as prefix

ubi2go:
@arttor looks good

sumitarora2786 (commented):
lgtm

PR #6691, opened 2024-06-18
nova/flavors: Minimal socket count for flavors >= 64vCPUs

grandchild:
We now want larger general purpose flavors to keep their cores together instead of spreading them across sockets.

The underlying hypervisor topology is unrelated to the values that are settable: Valid cores-per-socket numbers are the ones that result in integer socket counts with the total number of CPUs, with a maximum of 64 cores.

E.g. with 96 vCPUs possible cores-per-socket values are:
48, 32, 24, 16, 12, 8, 6, 4, 3, 2, 1

The resulting socket-size can be larger than the socket-size of the underlying hardware. The number of _NUMA nodes_ is influenced by the underlying hardware, and the number of NUMA nodes can curiously enough be greater than the number of sockets.

### DIFF
--- a/openstack/sap-seeds/templates/flavor-seed.yaml
+++ b/openstack/sap-seeds/templates/flavor-seed.yaml
@@ -208,6 +208,7 @@ spec:
     is_public: false
     extra_specs:
       {{- tuple . "vmware" | include "sap_seeds.helpers.extra_specs" | indent 6 }}
+      "hw:cpu_cores": "64"  # used in nova-vmware as cores-per-socket (32pCPU = 64vCPU)
   - name: "m_c4_m64"
     id: "100"
     vcpus: 4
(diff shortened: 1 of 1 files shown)


openstack/sap-seeds/templates/flavor-seed.yaml
```diff
@@ -430,6 +434,7 @@ spec:
     is_public: true
     extra_specs:
       {{- tuple . "vmware" | include "sap_seeds.helpers.extra_specs" | indent 6 }}
+      "hw:cpu_cores": "48"  # used in nova-vmware as cores-per-socket (24pCPU = 48vCPU)
```
joker-at-work, line 437: We also have older hardware with only 44 cores per socket still in use.

grandchild: I did some more testing and the underlying socket-size is mostly irrelevant. Details in the commit message.

I also fixed some invalid cpus-per-socket values to reflect the new findings.

grandchild:
I tested with both Linux and Windows (11) with a 128 vCPU flavor and the number of sockets is correctly reflected inside the VM.

joker-at-work (approved):
Te

PR #6698, opened 2024-06-19
[mariadb-galera] passwordless ccroot_user option added 

businessbean:
- to support local root access without password
- unit tests added
- removeuser function added to also use prepared statements to drop a user from the DB
- startup probe check order fixed: tcp connect checks before testing the API
- mariadb version check uses SOFTWARE_VERSION_CLEAN to avoid problems with version string changes in newer MariaDB releases
- Galera Dockerfile simplified
  - ADD instead of curl during RUN
  - permissions and ownership set with the ADD step
  - org.opencontainers.image labels
  - one layer for all ENV vars
  - one layer for all labels
- helm chart option names fixed for container images
- software versions bumped


dhoeller (approved):
LGTM

bashar-alkhateeb (approved):
LGTM

PR #6699, opened 2024-06-19
nova/flavors: Re-enable CPU reservation for hana-exclusive

grandchild:


grandchild:
Testing is still underways, please review the contents of the config anyway: I will only merge when both review approvals _and_ testing is positive.

joker-at-work (approved):
Why are we only doing this for hanas on hana-exclusive hosts?

grandchild:
> Why are we only doing this for hanas on hana-exclusive hosts?

Because CPU-reservation on mixed-workload clusters is still an open issue because of CPU overcommit. We haven't implemented a solution to "take away" the additional `reserved_cores * (overcommit - 1)` from the capacity.

grandchild:
Testing completed successfully: Created a 48-vcpu VM & the reserved MHz in VMware * the MHz-per-vCPU value for the host == 45vCPUs, i.e. the 95%-reservation value set for the flavor.

PR #6707, opened 2024-06-19
[keystone]pump up mariadb-galera dependencies

bashar-alkhateeb:
pump up dependencies to mariadb-galera 0.27.1
 - add feature to fix persistent volume fs permissions

Change-Id: Ic78187936d14c1c4f7b44f5e590917ea95ea908f

### DIFF
--- a/openstack/keystone/Chart.lock
+++ b/openstack/keystone/Chart.lock
@@ -4,7 +4,7 @@ dependencies:
   version: 0.10.1
 - name: mariadb-galera
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-  version: 0.27.0
+  version: 0.27.1
 - name: memcached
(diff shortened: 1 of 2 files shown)


bbobrov (commented):
Should we increase the keystone chart version too?

businessbean (approved):
lgtm

businessbean (changes requested):
the keystone chart version has to be bumped to indicate the change

businessbean (approved):
the keystone chart version bump does not seem to be required...

bashar-alkhateeb:
AFAIK we dont increase the wrapper helm chart when we pump up the dependencies .. @bbobrov if you think it is better for keystone we can increase the version , i am totally fine with it ?

bbobrov (approved):

PR #6719, opened 2024-06-21
[openstack] helm charts with underscore renamed to dash

businessbean:
- to avoid naming violations in oci registries
  - https://helm.sh/docs/chart_best_practices/conventions/#chart-names
  - https://helm.sh/docs/chart_best_practices/conventions/#version-numbers
- `helm dep update run` for all related parent charts
- alias added to keep the old chart dependency variable names and to avoid names with dashes


sandzwerg (commented):
What about the conflicts?

PR #6724, opened 2024-06-24
[keystone]option to switch to mariadb-galera

bashar-alkhateeb:
Add a new value `databaseKind` which when set to galera it will build the databae connection string to use mariadb-galera service instead of percona or legacy mariadb mono instance.
When set to empty string the connection string will be set as usual even if mariadb-galera is enabled, that to support the phase of migrating from
  old database to mariadb-galerai.

Change-Id: I2bbf96d823d41aaa8665a33a6bc3606fbcd31719


businessbean (approved):
lgtm. The local `helm template` tests have provided the expected results.

rajivmucheli (approved):

PR #6725, opened 2024-06-24
[nannies] wip - cleanup of unused stuff

thgrs:


openstack/nannies/values.yaml
```diff
@@ -2,9 +2,7 @@ global:
   linkerd_requested: false
 
 nannies:
-  vcenter_username: DEFINED-IN-REGION-CHART
-  netapp_username: DEFINED-IN-REGION-CHART
-  netapp_password: DEFINED-IN-REGION-CHART
+  master_password: DEFINED-IN-REGION-CHART
```
joker-at-work, line 5: DEFINED-IN-REGION-CHART is not a valid default value and by that an anti-pattern. Better comment it out.

thgrs: ok - i commented it out as suggested now

PR #6733, opened 2024-06-26
[octavia] Create additional LBStuckinPending alert

rhalina:
OctaviaLoadBalancerStuckinPending alerts are considered critical alerts and get paged. Yet, due to some specific workloads regarding kubernikus e2e tests and shoot networks the alarm fires a lot without indicating an actual data plane problem.
We create a specific alert targeting e2e and shoot LBs and consider assign them severity "warning". Thus they won't get paged, but will still be visible in the #alert-network-api-warning and #cc-os-octavia slack channels

We also remove the redundant waiting 1s condition as the metric already implies seconds.


BenjaminLudwigSAP (approved):
lgtm

m-kratochvil (approved):
lgtm

PR #6737, opened 2024-06-26
Add postfixMgmt config to _config.yaml.tpl

raphting:


openstack/cronus/templates/nebula/_config.yaml.tpl
```diff
@@ -145,6 +145,8 @@ nebula:
     zoneID: {{ .Values.postfixDNS.zoneID }}
     auth:
       auth_url: {{ .Values.postfixDNS.auth.auth_url }}
+  postfixMgmt:
+    url: {{ .Values.postfixMgmt.url }}
```
corey-aloia, line 149: Would it make sense to have it under `.Values.postfix`? I see the other properties are not, so it is fine, but IMO `.Values.postfix.management.url` make sense. Up to you 👍

corey-aloia (approved):

openstack/cronus/templates/nebula/_config.yaml.tpl
raphting: Yes, valid point!
In the future, we will remove postfixDNS from Nebula altogether. This will be a good moment to reconsider this structure.

PR #6738, opened 2024-06-26
[mariadb-galera] deploy storageclasses only if requested

businessbean:
- to avoid unexpected and not required storageclasses in the cluster
- unit tests added
- chart version bumped

### DIFF
--- a/common/mariadb-galera/helm/templates/_helpers.tpl
+++ b/common/mariadb-galera/helm/templates/_helpers.tpl
@@ -315,3 +315,37 @@ app.kubernetes.io/version: {{ $.Chart.Version }}
 app.kubernetes.io/component: MariaDB-Galera
 app.kubernetes.io/part-of: {{ $.Release.Name }}
 {{- end }}
+
+{{- define "storageclassCinder" }}
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: {{ include "commonPrefix" $ }}-cinder
+  labels:
+    {{- include "sharedservices.labels" $ | indent 4 }}
+parameters:
+  type: vmware
+provisioner: cinder.csi.openstack.org
+reclaimPolicy: Delete
+volumeBindingMode: WaitForFirstConsumer
+allowVolumeExpansion: true
+{{- end }}
+
+{{- define "storageclassNfs" }}
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: {{ include "commonPrefix" $ }}-nfs
+  labels:
+    {{- include "sharedservices.labels" $ | indent 4 }}
+parameters:
+  pathPattern: "${.PVC.namespace}-${.PVC.name}"
+  onDelete: "delete"
+  archiveOnDelete: "true"
+provisioner: cluster.local/nfs-subdir-external-provisioner
+reclaimPolicy: Delete
+volumeBindingMode: WaitForFirstConsumer
+allowVolumeExpansion: true
+{{- end }}
--- a/common/mariadb-galera/helm/templates/storageclass.yaml
+++ b/common/mariadb-galera/helm/templates/storageclass.yaml
@@ -1,28 +1,58 @@
----
-apiVersion: storage.k8s.io/v1
-kind: StorageClass
-metadata:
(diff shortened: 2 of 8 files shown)


common/mariadb-galera/helm/templates/storageclass.yaml
```diff
@@ -1,28 +1,58 @@
...
-  pathPattern: "${.PVC.namespace}-${.PVC.name}"
-  onDelete: "delete"
-  archiveOnDelete: "true"
-provisioner: cluster.local/nfs-subdir-external-provisioner
-reclaimPolicy: Delete
-volumeBindingMode: WaitForFirstConsumer
-allowVolumeExpansion: true
+{{- $cinderEnabled := false }}
+{{- $nfsEnabled := false }}
+{{- range $volumeClaimTemplatesKey, $volumeClaimTemplatesValue := $.Values.volumeClaimTemplates }}
```
bashar-alkhateeb, line 3: why not looping over $.Values.volumeMounts instead repeating the conditions 3 times?

bashar-alkhateeb:
in my point of view that massive amount of checks is over killing, we need to think about re-factoring the values file, simplifies this helm chart and calculate some values instead of giving all these possibilities to the user,  we have a very defined use cases.
in each loop we are trying to solve the problems of improper values because the user needs to add information dome times in two different places e.g 
` if eq $volumeClaimTemplatesKey $volumeMountsValue.claimName`

businessbean:
> in my point of view that massive amount of checks is over killing, we need to think about re-factoring the values file, simplifies this helm chart and calculate some values instead of giving all these possibilities to the user, we have a very defined use cases. in each loop we are trying to solve the problems of improper values because the user needs to add information dome times in two different places e.g ` if eq $volumeClaimTemplatesKey $volumeMountsValue.claimName`

This PR does not extend the `values.yaml`. It just adds some logic to configure things only if required. This make life easier for the user of this chart, but this also means more (template) code on our side.

common/mariadb-galera/helm/templates/storageclass.yaml
businessbean: The [depth is not the same](https://github.com/sapcc/helm-charts/blob/0e874cc3d35dbe4d19901e8c7aea960929d7b113/common/mariadb-galera/helm/values.yaml#L1044).
```yaml
volumeMounts:
  database:
    data:
    log:
  proxy:
    proxysql:
      data:
  backup:
    kopia:
      data:
```
We should refactor [volumeMounts](https://github.com/sapcc/helm-charts/blob/0e874cc3d35dbe4d19901e8c7aea960929d7b113/common/mariadb-galera/helm/values.yaml#L1044) and [volumeClaimTemplates](https://github.com/sapcc/helm-charts/blob/0e874cc3d35dbe4d19901e8c7aea960929d7b113/common/mariadb-galera/helm/values.yaml#L1095C1-L1095C21) by merging them into one option set. This means less flexibility because you than cannot reuse a volumeClaimTemplate anymore, but this is anyhow not used currently.

bashar-alkhateeb (approved):
Approved .. to not block the deployment in QA but it needs indeed refactoring

common/mariadb-galera/helm/templates/_helpers.tpl
```diff
@@ -315,3 +315,37 @@ app.kubernetes.io/version: {{ $.Chart.Version }}
...
+  labels:
+    {{- include "sharedservices.labels" $ | indent 4 }}
+parameters:
+  pathPattern: "${.PVC.namespace}-${.PVC.name}"
+  onDelete: "delete"
+  archiveOnDelete: "true"
+provisioner: cluster.local/nfs-subdir-external-provisioner
+reclaimPolicy: Delete
+volumeBindingMode: WaitForFirstConsumer
+allowVolumeExpansion: true
```
galkindmitrii, line 350: In other charts we normally don't insert long optional yaml blocks into `_helpers.tpl`, rather keep those in `.yaml` and enable or disable based on values. Like for example with PVC for MariaDB: https://github.com/sapcc/helm-charts/blob/master/common/mariadb/templates/pvc.yaml#L1-L19

galkindmitrii (approved):

PR #6749, opened 2024-07-02
[rabbitmq] Add an option to enable all stable feature flags in the RabbitMQ instance

s10:
Add an option to execute rabbitmqctl enable_feature_flag all on rabbitmq-server startup.

It only needed and might be used on instances with persistence enabled. This also makes downgrades impossible.

PR #6752, opened 2024-07-02
[rabbitmq] Add support for RABBITMQ_FEATURE_FLAGS env variable in RabbitMQ

s10:
Populate it by default with all stable feature flags from 3.13.0

### DIFF
--- a/common/rabbitmq/CHANGELOG.md
+++ b/common/rabbitmq/CHANGELOG.md
@@ -3,6 +3,10 @@ Rabbitmq CHANGELOG
 
 This file is used to list changes made in each version of the common chart rabbitmq.
 
+0.7.3
+-----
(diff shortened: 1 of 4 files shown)


galkindmitrii (approved):
Please note that the chart version and changelog now collide with PR: https://github.com/sapcc/helm-charts/pull/6749

bashar-alkhateeb:
the only thing i dont like in ENV variables that any change needs a pod restart,
and i was thinking if one of the features changed to be required in the new version that means 2 downtimes 
1- activate the feature flag by redeploy the old version on rabbitmq with new set of values (not sure if that can even work because the activation of the feature will be done during spining up rabbitmq and the persistent volumes will speak against that because the feature was not active in the previous rabbitmq instance which has been destroyed )
2- deploy the new version of rabbitmq (newer helm-chart version)

bashar-alkhateeb (approved):
Please rebase !

PR #6755, opened 2024-07-02
[barbican] enable get audit events

notque:
@rajivmucheli This is the configuration to test the read audit events for Barbican. I need this deployed in QA first before we deploy this completely to test and validate. We need to see what the load is, if it works correctly, and if we can handle the load. 

It may also be that we only require a subset of events with read, and then will need to add to the logstash conf to drop other events. 

So next steps are, can you please deploy this change just to a QA region for evaluation? If it makes sense, and we agree, we can deploy to all regions. If it has issues, or we need to make additional config changes, we can do so.

### DIFF
--- a/openstack/barbican/templates/etc/_barbican-api-paste.ini.tpl
+++ b/openstack/barbican/templates/etc/_barbican-api-paste.ini.tpl
@@ -86,7 +86,7 @@ config_file = /etc/barbican/watcher.yaml
 [filter:audit]
 paste.filter_factory = auditmiddleware:filter_factory
 audit_map_file = /etc/barbican/barbican_audit_map.yaml
-ignore_req_list = GET
+ignore_req_list = HEAD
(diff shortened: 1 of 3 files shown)


notque:
@rajivmucheli I added the config in the wrong place, it's a config for the paste. I've updated the PR, can you deploy this to qa-de-2 for us to test?

notque:
We need a full list of events to drop due to Barbican creating multiple events per action. I've added an example that will need the list of each criteria, and we'll drop them at the beginning of the pipeline

notque:
I'd like to get this deployed out to all regions. Looking at the created events in qa-de-1, my belief is the correct setup is to drop read/list and read/get and leave only read events without an extension. 

I will update this PR with that change, and I'm hoping we can get this deployed out into all regions @rajivmucheli

rajivmucheli (approved):

PR #6758, opened 2024-07-03
[keystone] optionally enable auth and service ips for memcached

galkindmitrii:
- service ips are needed for global setup
- memcached auth can optionally be enabled from 0.5.0 chart version


fwiesel (approved):

openstack/keystone/Chart.yaml
bbobrov, line 1: This is good, but please also increase the keystone chart version (line 12), because this is a significant change. I suggest 0.7.0

bbobrov (changes requested):

bbobrov (approved):

PR #6760, opened 2024-07-03
[rabbitmq] Remove support of the insecure rabbitmq-exporter sidecar

s10:
Remove support of the insecure rabbitmq-exporter sidecar container:
1. Sidecar uses plaintext credentials in environment variables
2. Sidecar utilises Management API metrics
3. Management API metrics will not be supported in future RabbitMQ releases

Always use rabbitmq prometheus plugin instead

### DIFF
--- a/common/rabbitmq/CHANGELOG.md
+++ b/common/rabbitmq/CHANGELOG.md
@@ -3,6 +3,13 @@ Rabbitmq CHANGELOG
 
 This file is used to list changes made in each version of the common chart rabbitmq.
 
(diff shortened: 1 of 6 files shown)


bashar-alkhateeb (changes requested):
please increase the version and add a section in CHANGELOG.md

s10:
Need to migrate Hermes metrics/alerts first. It currently uses rabbitmq-exporter metrics.

businessbean (approved):
lgtm

PR #6761, opened 2024-07-03
[neutron] Remove ucs driver from chart

sebageek:
The driver is no longer in use in our cloud, we don't need it in our helm-charts anymore.


mutax (approved):
lgtrm

PR #6763, opened 2024-07-03
[rabbitmq] Fix RabbitMQRPCUnackTotal alert

s10:
RabbitMQRPCUnackTotal alert should support both sidecar exporter metric name and prometheus plugin


galkindmitrii (approved):
nice!

bashar-alkhateeb (approved):
looks good

PR #6785, opened 2024-07-10
[rabbitmq][Hermes] Don't use metrics sidecar container in Hermes

s10:
Bump rabbitmq dependency chart to 0.7.2

Update RabbitMQRPCUnackTotal alert to use prometheus plugin metric names


notque (approved):
Lgtm

PR #6787, opened 2024-07-10
[rabbitmq][Cronus] Use prometheus plugin for rabbitmq metrics

s10:
bump rabbitmq dependency chart to 0.7.3

Fix for RabbitMQRPCUnackTotal alert, add standard labels

### DIFF
--- a/openstack/cronus/Chart.yaml
+++ b/openstack/cronus/Chart.yaml
@@ -1,14 +1,14 @@
 apiVersion: v2
 name: cronus
 description: proxying hyperscaler statless services
-version: 0.1.13
+version: 0.1.14
 dependencies:
   - name: rabbitmq
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-    version: 0.6.12
+    version: 0.7.3
   - name: rabbitmq
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-    version: 0.6.12
+    version: 0.7.3
     alias: rhea_rabbitmq
   - name: owner-info
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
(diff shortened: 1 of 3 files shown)


openstack/cronus/Chart.yaml
```diff
@@ -1,14 +1,14 @@
...
-version: 0.1.13
+version: 0.1.14
 dependencies:
   - name: rabbitmq
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
     version: 0.6.12
   - name: rabbitmq
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-    version: 0.6.12
+    version: 0.7.2
```
corey-aloia, line 11: It looks like only 1 rabbitmq was upgraded. Is this expected?

corey-aloia (changes requested):
Previously when upgrading, we came across an issue where we needed to manually exec into the pods of each environment and enable the rabbitmq features. Is this fixed? I.E. would there be any risk in this upgrade?

FWIW we use persistent volumes for rabbitmq which is why we have a different behavior in upgrades than other teams

openstack/cronus/Chart.yaml
s10: Unexpected. Added version bump for the second one.

corey-aloia (approved):
Looks fine to me as long as the upgrade is low risk. I would rather @dhalimi have a look first though :)

s10:
> Previously when upgrading, we came across an issue where we needed to manually exec into the pods of each environment and enable the rabbitmq features. Is this fixed? I.E. would there be any risk in this upgrade?
> 
> FWIW we use persistent volumes for rabbitmq which is why we have a different behavior in upgrades than other teams

This issue only affects clusters which are running version <3.12.0 and upgrading to 3.12.0+. So this enable_feature_flags was needed to run once on clusters with version 3.11.* (and, as I see, it was done).

For any future updates, we decided to enable feature flags on upgrade automatically, this is being done in helm chart rabbitmq 0.7.3+

bashar-alkhateeb (approved):

PR #6791, opened 2024-07-10
system/tenso: convert to new image value structure

SuperSandro2000:


system/tenso/images.yaml
```diff
@@ -0,0 +1,5 @@
+containerImages:
+  tenso:
+    account: ccloud
+    repository: tenso
+    tag: "20240719041832"
```
SuperSandro2000, line 5: Instead of the trailers we want to the following to easily see the old commit and allow to further process that:

```suggestion
    tag: "20240719041832"
    sources:
      - repo: https://github.com/sapcc/tenso.git
        commit: 01a1ac291b84e19ce373a63a09d973a0c8b6f6a9
```

PR #6792, opened 2024-07-11
[mariadb] Seed root-password

fwiesel:
The liveness-probe uses the root-password for
authentication. If it is not seeded, the probe
will fail.


fwiesel:
When trying to deploy a service from scratch, the mariadb did never came up, presumably due to this missing piece.
Just a draft, because it would be good to verify that in a more isolated fashion.

PR #6810, opened 2024-07-15
nova/flavors: Add g_c24_m96 flavor on request from S4

grandchild:


joker-at-work (approved):
Commit message is misleading as it sounds like the flavor will only be available for S4.

grandchild:
Fixed commit message, and as we learned yesterday, that keeps the approval \o/

PR #6828, opened 2024-07-18
[openstack-exporter] Update CinderShardFreeSpace alerts

kevin-fischer:
- Triggers an alert if either the capacity ratio falls below 30%
or if the number of available storage units per shard meets the specified threshold


himanip94 (approved):
lgtm

PR #6830, opened 2024-07-19
[Nova] Add weight for AggregateMultitenancyIsolation weigher

joker-at-work:
We want BBs dedicated to certain customers to be preferred by these customers over generally available BBs. This makes sure that as much as possible generally available resources are available. We still want BBs marked for decommissioning to be freed, though.


dhoeller (approved):
LGTM

PR #6836, opened 2024-07-22
[netapp] add lif metrics and alert for manila filers

chuan137:


Carthaca (approved):
looks good, numbers match with `network interface show` cmd on ontap cli

PR #6837, opened 2024-07-22
common/alertmanager: receiver name on metrics

richardtief:
This PR adds a second label to all notification metrics so that we not only have the type of notification, but also the `receiver_name` to which the notification belongs. Due to the proliferation of Slack channels, this is necessary to properly diagnose errors based on metrics - currently we get one metric for all Slack channels, regardless of the fact that they may represent completely different notification paths.


viennaa (approved):
Cool, I want it!

PR #6838, opened 2024-07-22
GkVulnerableImages: add "status" field to object identity

majewsky:
This makes the vulnerability status accessible as a label on the DOOP metrics.

I am using a rather generic name here, rather than the more specific "vulnerability_status" or "vuln_status". This label will now have to appear on all DOOP metrics, so I want to use a name that might be useful for other types of violations as well.

### DIFF
--- a/system/gatekeeper/tests/gatekeeper-policies/suite.yaml
+++ b/system/gatekeeper/tests/gatekeeper-policies/suite.yaml
@@ -676,16 +676,16 @@ tests:
     object: fixtures/vulnerable-images/deployment-vulnerable.yaml
     assertions:
     - violations: 1
-      message: '^\{"support_group":"foo-group","service":"dummy"\} >> image keppel.example.com/vulnerability-unpinned:low for container "dummylow" has "X-Keppel-Vulnerability-Status: Low"$'
+      message: '^\{"status":"Low","support_group":"foo-group","service":"dummy"\} >> image keppel.example.com/vulnerability-unpinned:low for container "dummylow" has "X-Keppel-Vulnerability-Status: Low"$'
     - violations: 1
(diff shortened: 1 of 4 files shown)


system/gatekeeper/tests/gatekeeper-policies/suite.yaml
```diff
@@ -676,16 +676,16 @@ tests:
...
+      message: '^\{"status":"Medium","support_group":"foo-group","service":"dummy"\} >> image keppel.example.com/vulnerability-unpinned:medium for container "dummymedium" has "X-Keppel-Vulnerability-Status: Medium"$'
     - violations: 1
-      message: '^\{"support_group":"foo-group","service":"dummy"\} >> image keppel.example.com/vulnerability-unpinned:high for container "starter" has "X-Keppel-Vulnerability-Status: High"$'
+      message: '^\{"status":"High","support_group":"foo-group","service":"dummy"\} >> image keppel.example.com/vulnerability-unpinned:high for container "starter" has "X-Keppel-Vulnerability-Status: High"$'
   - name: pod-unclear
     object: fixtures/vulnerable-images/deployment-unclear.yaml
     assertions:
     - violations: 1
-      message: '^\{"support_group":"foo-group","service":"dummy"\} >> image keppel.example.com/vulnerability-unpinned:unclear for container "dummy" has "X-Keppel-Vulnerability-Status: Unclear"$'
+      message: '^\{"status":"Unclear","support_group":"foo-group","service":"dummy"\} >> image keppel.example.com/vulnerability-unpinned:unclear for container "dummy" has "X-Keppel-Vulnerability-Status: Unclear"$'
```
SuperSandro2000, line 688: Isn't this Unknown in trivy?

majewsky: "Unclear" is a special value that the policy uses to report "we did not get an X-Keppel-Vulnerability-Status header at all".

SuperSandro2000 (approved):

PR #6847, opened 2024-07-23
[designate-nanny] adjusting memory due to OOMs observed

mutax:
There have been frequent OOM kills observed. This should give enough headroom.


galkindmitrii (approved):

rhalina (approved):
@dhoeller is on vacation, hence I'm substituting

LGTM

PR #6852, opened 2024-07-24
[bind] set service externalTrafficPolicy to Local

defo89:


galkindmitrii (approved):
that make service ips to be advertised as /32

mutax (approved):

PR #6855, opened 2024-07-24
[rabbitmq] provide option to set service externalTrafficPolicy and type

defo89:


bashar-alkhateeb (commented):
would you kindly add a section for this change to the CHANGELOG.md other wise looks good

common/rabbitmq/templates/service.yaml
```diff
@@ -21,6 +21,10 @@ metadata:
     linkerd.io/inject: enabled
 {{- end }}
 spec:
+  type: {{ default "ClusterIP" .Values.service.type }}
```
s10, line 24: I get this error when generating template with test values
```
Error: template: rabbitmq/templates/service.yaml:24:38: executing "rabbitmq/templates/service.yaml" at <.Values.service.type>: nil pointer evaluating interface {}.type
helm.go:84: [debug] template: rabbitmq/templates/service.yaml:24:38: executing "rabbitmq/templates/service.yaml" at <.Values.service.type>: nil pointer evaluating interface {}.type
```
The same error in ci/concourse/test-helm-chart

fwiesel: I'd move the default straight to the values. I.e.
Here simply: `{{ .Values.service.type }}`
And in the values
```
service:
  type: "ClusterIP"
```
That fixes the problem and personally, I'd prefer my defaults to be in the values instead being spread over the templates.

defo89: hope the last fix addresses all issues

s10 (approved):
lgtm

common/rabbitmq/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rabbitmq
-version: 0.9.0
+version: 0.9.1
```
businessbean, line 3: Because of the added feature it has to be a `0.10.0`.

businessbean (changes requested):
lgtm. Just change the version...

common/rabbitmq/Chart.yaml
defo89: sure, changed the version

businessbean (approved):
lgtm

PR #6858, opened 2024-07-24
[nova-vspc] set service externalTrafficPolicy to Local

defo89:
to advertise the /32 IP to upstream


joker-at-work:
Could you explain a little more what this means? What is "upstream" in this context?

defo89:
> Could you explain a little more what this means? What is "upstream" in this context? Why did it work before and needs a change now?

I have updated the commit with more details and tagged you in the relevant slack thread for more context

joker-at-work (approved):

joker-at-work:
Reading up on it, this seems to have the disadvantage of not distributing the load as good when multiple pods are running on the same node. Since `nova-vspc` is currently not in-use except for some debugging scenarios, we can live with that.

Scsabiii (approved):

PR #6869, opened 2024-07-25
[hermes] set rabbitmq service to type LB and override traffic policy

defo89:


notque:
Will deploy from the branch to test. If everything is reasonable, will merge and deploy on Monday.

notque:
Merging, will deploy to the rest of the regions on Monday

PR #6872, opened 2024-07-26
[ironic] upgrade dependencies

sandzwerg:
- mariadb
- memcached
- mysql_metrics
- rabbitmq
- utils

### DIFF
--- a/openstack/ironic/Chart.lock
+++ b/openstack/ironic/Chart.lock
@@ -1,27 +1,27 @@
 dependencies:
 - name: mariadb
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-  version: 0.11.1
+  version: 0.14.0
 - name: memcached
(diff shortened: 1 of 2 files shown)


joker-at-work (commented):
These look like quite significant version bumps. It would be nice if you can give a rough summary of what one should expect to change e.g. in the pipeline diff and also to know if restarts are necessary.

sandzwerg:
Some examples:
maridadb:

- Additionally add single quotation to support usage of all special symbols
- update mysqld-exporter to version 0.15.1

rabbitmq:

- Update RabbitMQ to version 3.13.6-management
- add ability to change service type and set externalTrafficPolicy
- Remove support of the insecure rabbitmq-exporter sidecar container

Do you want these as part of the commit as well?

joker-at-work (approved):

stefanhipfel (approved):

PR #6875, opened 2024-07-29
[rabbitmq] Update RabbitMQ to 3.13.6, add helm.sh/chart label

s10:
- Update RabbitMQ to version 3.13.6-management
- Add `helm.sh/chart` label
- `app.kubernetes.io/version` is now an application version


businessbean (approved):
lgtm

PR #6896, opened 2024-07-31
Create am-config-inhibit-rules-kubernetes.yaml

sumitarora2786:
alert inhibition rule for ceph


richardtief:
Could you please remove these lines in the other yaml: 
https://github.com/sapcc/helm-charts/blob/441de060025b6609d57a337ea88c0626c42c84e1/global/prometheus-alertmanager-operated/templates/am-config-inhibit-rules-base.yaml#L63-L83

Otherwise it will create duplicate rules.

global/prometheus-alertmanager-operated/templates/am-config-inhibit-rules-kubernetes.yaml
```diff
@@ -0,0 +1,39 @@
...
+       - matchType: "="
+         name: alertname
+         value: NodeInMaintenance
+      targetMatch:
+        - matchType: "="
+          name: inhibited_by
+          value: node-maintenance
+      equal:
+        - 'node'
+    - sourceMatch:
```
richardtief, line 30: Just one nitpicking, then we're good.
```suggestion
     # If the alert NodeInMaintenance is firing other alerts with the label inhibited-by: cluster-maintenance are being      inhibited on the same cluster.
    - sourceMatch:
```

richardtief (changes requested):

PR #6908, opened 2024-08-02
Base certificate expiry alert upon percentage of certificate lifetime

Nuckal777:


viennaa:
Why was this needed? Do we have certs which are always under 30d expiry?

Nuckal777:
We have some certs with 30d lifetime. For them the alert would alway be active.

PR #6914, opened 2024-08-02
feat(cloudprober): add service to owner-info

IvoGoman:


viennaa (changes requested):
The atlas change slipped in here too, this is in #6915 already.

viennaa (approved):
Good to go from my side.

PR #6915, opened 2024-08-02
feat(atlas): add service to owner-info

IvoGoman:


system/atlas/values.yaml
```diff
@@ -2,6 +2,7 @@ global:
   region:
 owner-info:
   support-group: observability
+  service: prometheus
```
viennaa, line 5: Why prometheus? This service is with @stefanhipfel :)

IvoGoman: It's already in our support group. Since it provides a Prometheus SD it felt fitting

viennaa (approved):
And go.

PR #6917, opened 2024-08-05
[mariadb] mysqld-exporter scrapes localhost pod-local mysql target

s10:
No need to target cluster service for pod-only mysqld-exporter scrapes


businessbean (approved):
lgtm

PR #6933, opened 2024-08-07
[mysql_metrics] Use resolve for mysql-metrics connection string

s10:
We use single quoted literal strings to avoid any unnecessary escaping

go-mysql DSN itself doesn't require any escaping for username and password

### DIFF
--- a/common/mysql_metrics/Chart.yaml
+++ b/common/mysql_metrics/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart for MySQL Metrics
 name: mysql_metrics
-version: 0.3.4
+version: 0.3.5
 dependencies:
   - name: linkerd-support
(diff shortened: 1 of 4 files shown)


joker-at-work:
What if we have an @ in the password though? Other mysql URLs use `| urlquery` for this.

s10:
> What if we have an @ in the password though? Other mysql URLs use `| urlquery` for this.

go-mysql-driver doesn't support escaping for username/password, see:
https://github.com/go-sql-driver/mysql/blame/master/README.md#L134
https://github.com/go-sql-driver/mysql/blob/master/dsn.go#L372

So the first colon in the username would be considered the start of the password and the last @ would be the start of the address.

joker-at-work (approved):

galkindmitrii (approved):

PR #6936, opened 2024-08-08
fix misplaced owner-info

majewsky:
The owner-info dependency belongs in the top-level chart, and only there. At best, putting it in a subchart will be ignored. At worst, it will override the owner-info of the top-level chart.


defo89 (approved):

richardtief (approved):
Thanks.

PR #6937, opened 2024-08-08
[manila] metrics for castellum's exclusion resons

chuan137:
Metrics to list shares that are in "error" status and shares that are
snapmirror targets. Currently, Castellum calls manila API's to filter
those shares. We provides these metrics for better performance.


Carthaca (approved):
lgtm

PR #6939, opened 2024-08-08
[memcached] Use resolve for  sasl-db secret string

s10:
Use urlquery both for username and password

* These values are consumed literally in urlqueried form by memcached
* These values should be urlqueried in `[cache]` section of the services configuration and consumed literally there, because they can't contain `$` sign.

### DIFF
--- a/common/memcached/Chart.yaml
+++ b/common/memcached/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: memcached
-version: 0.5.2
+version: 0.5.3
 description: Free & open source, high-performance, distributed memory object caching system.
 home: http://memcached.org/
 sources:
--- a/common/memcached/ci/test-values.yaml
+++ b/common/memcached/ci/test-values.yaml
@@ -4,3 +4,6 @@ global:
 auth:
   username: test
   password: secure
+
+alerts:
+  support_group: test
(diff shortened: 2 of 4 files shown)


joker-at-work (changes requested):
Missing the version bump in Chart.yaml

joker-at-work (commented):
Do you know if we need to escape any characters here? afaics, we have a `:` with special meaning.

Looks good otherwise.

s10:
> Do you know if we need to escape any characters here? afaics, we have a `:` with special meaning.

memcached doesn't perform any unescaping and considers the colon as a delimiter
https://github.com/memcached/memcached/blob/1.6.28/sasl_defs.c#L74

So, if the password contains a colon (e.g., 'username:pass:word'), only the part before the colon would be considered a password ('pass').

Currently, for the password in memcached sasl configuration, we perform urlquery transformation, so this password is equal both in memcached and service configuration, but not necessarily equal to the string in a secret store.

galkindmitrii:
Memcached with SASL only enabled for single service so far. Maybe we just add a warning note about the `:`

galkindmitrii (approved):

s10:
> Why do we `urlquery` now? I thought we just add a warning about not using a `:` and thus only need `resolve` without `urlquery`?

I misunderstood that part. Currently, we (1) don't enforce the prohibition of these special characters anywhere and (2) use resolve with urlquery in values for the memcached password.
I changed the resolve in this chart back to plain resolve without urlquery and added a comment about our choice.

joker-at-work (approved):

PR #6944, opened 2024-08-12
limes: add liquid-neutron, liquid-octavia

majewsky:


openstack/limes/templates/ingress-liquids.yaml
```diff
@@ -1,6 +1,6 @@
 {{- range $name, $config := .Values.limes.local_liquids }}
 
-{{- $domain := printf "liquid-%s.%s.%s" $name $.Values.global.region $.Values.global.tld -}}
+{{- $domain := printf "liquid-%s.%s.%s" $name $.Values.global.region $.Values.global.tld }}
```
majewsky, line 3: FYI, this was necessary to fix a whitespace error that only occurs when rendering multiple Ingress objects in this loop.

VoigtS (approved):

PR #6947, opened 2024-08-12
[manila][castellum] move castellum rules to manila chart

chuan137:
We have castellum rules in two different prometheus services. Keep them
in the same chart.


Carthaca (approved):
Clarified offline: there will be a transition phase with some duplicated metrics

PR #6954, opened 2024-08-12
[manila] use harvest metrics in castellum rules

chuan137:
Follow up of migrating netapp exporters to netapp harvest

The rules are duplicated in previous PR. The identical rules does not produce multiple metrics. With this pr, we migrate the rules to use harvest metrics, producing different set of labels. And there will be two set of castellum metrics.

### DIFF
--- a/openstack/manila/aggregates/storage/castellum.rules
+++ b/openstack/manila/aggregates/storage/castellum.rules
@@ -5,56 +5,64 @@ groups:
     rules:
       # Case one: default provision style, where snapshot reserve is allocated AS PART OF the target share size and snapshot reserve percentage is 5.
       # Logical space reporting and enforcement are always disabled for them. Once resized, the share will be converted to case two or three.
-      # Share size = netapp_volume_total_bytes + netapp_volume_snapshot_reserved_bytes
-      # Share usage = netapp_volume_used_bytes
-      # Share minimal size = netapp_volume_used_bytes + netapp_volume_snapshot_reserved_bytes
+      # Share size = NetApp Volume size = netapp_volume_size_total +  netapp_volume_snapshot_reserve_size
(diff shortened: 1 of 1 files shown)


openstack/manila/aggregates/storage/castellum.rules
```diff
@@ -5,56 +5,64 @@ groups:
...
       # New provision style means snapshot reserve is allocated side by side to the share space.
       # The volume is provisioned with double of share size and sanpshot reserve percentage is set to 50.
       # To avoid snapshot spill, share's minimal size must be larger than the snapshot used size or share used size, whichever is larger.
-      # Share size = netapp_volume_total_bytes
+      # Share size = netapp volume total size (snapshot reserve excluded)
       # Share usage = netapp_volume_used_bytes
       # Share minimal size = max(netapp_volume_used_bytes, netapp_volume_snapshot_used_bytes)
       - record: netapp_volume_provision_case_two
-        expr: (netapp_volume_percentage_snapshot_reserve{share_id!="", volume=~"share_[0-9a-f]{8}_[0-9a-f]{4}_[0-9a-f]{4}_[0-9a-f]{4}_[0-9a-f]{12}", volume_type="rw"} / 50 == 1) * ((1 + netapp_volume_is_space_enforcement_logical) == 1)
+        expr: (netapp_volume_snapshot_reserve_percent{share_id!="", type="rw",
```
Carthaca, line 33: `type` should not be specified two times

Carthaca (changes requested):

openstack/manila/aggregates/storage/castellum.rules
chuan137: fixed

Carthaca (approved):

PR #6956, opened 2024-08-13
[Nova] Support secrets-injector

joker-at-work:
With the recent version bump we got a new helper in the `utils` chart that makes it possible to support `secrets-injector` secrets will still maintaining compatibility with not-yet-migrated secret values. In this commit, we use this helper to make the `Nova` chart compatible with `secrets-injector`.

We also bump the version of the `utils` chart again, because there is a change to support resolving more usernames. It doesn't matter for us, since we don't read them from Vault, yet, but we might in the future.

Since we implement custom helpers in `Nova` to support our cell2, we also had to update them. For this, we use the available helpers from other charts as far as possible so common functionality in generating URLs is not duplicated.


openstack/nova/templates/seed.yaml
```diff
@@ -71,9 +71,9 @@ spec:
         role: cloud_compute_admin
       - project: admin
         role: cloud_compute_admin
-    - name: {{ .Values.global.nova_service_user | default "nova" }}{{ .Values.global.user_suffix | default "" }}
+    - name: '{{ .Values.global.nova_service_user | default "nova" | include "resolve_secret" }}'
       description: Nova Service
-      password: {{ required ".Values.global.nova_service_password is missing" .Values.global.nova_service_password }}
+      password: '{{ required ".Values.global.nova_service_password is missing" .Values.global.nova_service_password | include "resolve_secret" }}'
```
s10, line 76: According to the secrets-injector documentation, it works only for "kind: Secret". So, I suppose, we have a problem with all OpenStack Seeds that create users.

joker-at-work: No. https://github.wdf.sap.corp/cc/secrets-injector#other-resource-kinds
>  In Converged Cloud controlplanes this is configured for kind OpenstackSeed.

I also validated the same with Erik.

s10 (approved):

Scsabiii (approved):

grandchild (approved):

PR #6958, opened 2024-08-13
[cronus][rabbitmq] Add support for secrets-inejctor resolve for RabbitMQ URIs

s10:
* Add helper helm functions with secrets-injector resolve support
* Pass username, password and host as arguments for the helper function
* Add CI test-values
* Update rabbitmq helm dependency to 0.11.0: full secrets-injector support

Note: rabbitmq chart 0.7.3 -> 0.11.0 upgrade could require service rabbitmq statefulset removal because this chart contained the bug in version 0.6.13-0.7.5, which made upgrade impossible.

### DIFF
--- a/openstack/cronus/Chart.lock
+++ b/openstack/cronus/Chart.lock
@@ -1,15 +1,15 @@
 dependencies:
 - name: rabbitmq
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-  version: 0.7.3
+  version: 0.11.0
 - name: rabbitmq
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-  version: 0.7.3
+  version: 0.11.0
 - name: owner-info
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
   version: 0.2.0
 - name: maillog
   repository: file://sub-charts/maillog
   version: 0.1.0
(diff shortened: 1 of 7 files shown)


openstack/cronus/templates/nebula/_helpers.tpl
```diff
@@ -0,0 +1,15 @@
+{{- define "nebula.resolve_secret" -}}
+    {{- $str := . -}}
+    {{- if (hasPrefix "vault+kvv2" $str ) -}}
+        {{"{{"}} resolve "{{ $str }}" {{"}}"}}
```
bashar-alkhateeb, line 4: don't we need to add `urlquery` escaping here ?

openstack/cronus/templates/rhea/_helpers.tpl
```diff
@@ -0,0 +1,15 @@
+{{- define "rhea.resolve_secret" -}}
+    {{- $str := . -}}
+    {{- if (hasPrefix "vault+kvv2" $str ) -}}
+        {{"{{"}} resolve "{{ $str }}" {{"}}"}}
```
bashar-alkhateeb, line 4: and here ?

bashar-alkhateeb (changes requested):
don't forget the bump up the chart version

bashar-alkhateeb (approved):
Looks good to me 
@corey-aloia would you check form your end ?

dhalimi1975 (approved):
Looks okay by me

PR #6964, opened 2024-08-14
keppel-auto: init

majewsky:
Similar to the limes-auto chart, this is for automatic deployments of certain configuration files, specifically the seed configuration for managed accounts.

On its own, this deployment will do nothing until we change the main Keppel chart to consume this configmap.

### DIFF
--- a/openstack/keppel-auto/values.yaml
+++ b/openstack/keppel-auto/values.yaml
@@ -1,0 +1,12 @@
+owner-info:
+  support-group: containers
+  service: keppel
+  maintainers:
+    - Stefan Majewsky
+    - Sandro Jäckel
+    - Stefan Voigt
+  helm-chart-url: https://github.com/sapcc/helm-charts/tree/master/openstack/keppel-auto
+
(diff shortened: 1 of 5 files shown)


openstack/keppel-auto/values.yaml
```diff
@@ -0,0 +1,12 @@
+owner-info:
+  support-group: containers
+  service: keppel
+  maintainers:
+    - Stefan Majewsky
+    - Sandro Jäckel
+    - Stefan Voigt
+  helm-chart-url: https://github.com/sapcc/helm-charts/tree/master/openstack/keppel-auto
+
+# See docstring on keppel-managed-accounts.yaml in the secrets repo for details.
```
SuperSandro2000, line 10: Can't we link the keppel doc instead?

majewsky: No, the format of the `accounts` section deviates in some ways from the config file that Keppel sees, because it is written in a region-agnostic way.

PR #6974, opened 2024-08-19
chore(cc3test): warn and info for cert alerts

viennaa:
I want the alert to be warning only if there is a more immanent action required. Default goes with info and this is visible in our dashboard already, as info alerts are taken care of too.


ashifnihalb (approved):
lgtm

PR #6985, opened 2024-08-20
[manila] maia metrics for snapshot size induced by snapmirror relationships

chuan137:
This metrics provide a overview to customers of the extra snapshot
usages, in additional to manila snapshots, i.e., backend snapshots
created for snapmirror relationships.

### DIFF
--- a/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
+++ b/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
@@ -160,6 +160,17 @@ groups:
     - record: netapp_snapmirror_destination_snapshot_count:maia
       expr: '{__name__="netapp_snapmirror_destination_snapshot_count:federated"}'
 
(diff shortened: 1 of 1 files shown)


majewsky:
How (and if so, why) is this different from `limes_project_usage{resource="snapmirror_capacity"}`, as computed with [these queries](https://github.com/sapcc/limes/blob/master/internal/plugins/manila.go#L540-L541)?

chuan137:
This is for explaining the extra physical snapshot sizes, especially when customer does not have Manila Snapshots.
The limes metrics you quotes is the netapp volumes set up (not via manila) for EC2 backups.

majewsky:
Do we have a chart somewhere to explain all the different types of usage and mirroring and snapshots and what not? I'm way beyond confused at this point.

chuan137:
@notque had a meeting last week with Stefan and Maurice, I will pick this up

chuan137:
docu https://github.wdf.sap.corp/cc/documentation-operation/pull/2996 to explain the snapmirror related metrics.

sumitarora2786 (approved):

PR #6990, opened 2024-08-21
[designate] support secret injector

galkindmitrii:
With the recent version bump we got a new helper in the `utils` chart that makes it possible to support `secrets-injector` secrets will still maintaining compatibility with not-yet-migrated secret values. In this commit, we use this helper to make the `Designate` chart compatible with `secrets-injector`.

- deps bumped to get the required functionality also across sub-charts


openstack/designate/templates/etc/_secrets.conf.tpl
```diff
@@ -1,9 +1,9 @@
 [DEFAULT]
-transport_url = rabbit://{{ .Values.rabbitmq.users.default.user | default "rabbitmq" }}:{{ .Values.rabbitmq.users.default.password }}@{{ include "rabbitmq_host" . }}:{{ .Values.rabbitmq.port | default 5672 }}/
+transport_url = rabbit://{{ .Values.rabbitmq.users.default.user | default "rabbitmq" | include "resolve_secret_urlquery" }}:{{ .Values.rabbitmq.users.default.password | include "resolve_secret_urlquery" }}@{{ include "rabbitmq_host" . }}:{{ .Values.rabbitmq.port | default 5672 }}/
 
 [keystone_authtoken]
 username = {{ .Values.global.designate_service_user | default "designate" }}
-password = {{ .Values.global.designate_service_password }}
+password = {{ .Values.global.designate_service_password | include "resolve_secret" }}
 
 [storage:sqlalchemy]
```
s10, line 8: Unfortunately, resolve_secret hasn't been added for secrets in db_url_pxc in utils helm chart...

galkindmitrii (commented):
In such case we have to make the file also a template:
https://github.com/sapcc/helm-charts/blob/master/openstack/designate/templates/etc/_test.html

s10:
> In such case we have to make the file also a template: https://github.com/sapcc/helm-charts/blob/master/openstack/designate/templates/etc/_test.html

Renamed this file in commit https://github.com/sapcc/helm-charts/pull/6990/commits/a61b24fc67d4f5157bb1552c6426c85ac031b920

vssldmtrv (approved):

PR #6991, opened 2024-08-21
[neutron] Honor override_vlan_pool in fabric config

sebageek:
When a hostgroup overrides its vlan pool we need to pick that name for the list of network_vlan_ranges instead of the switchgroup name. This might lead to duplicate names in the list for now, but as in some cases we might also override the vlan ranges we can't just deduplicate this. As this is only relevant for qa for now we'll leave it as is and rethink this once it becomes a problem.


occamshatchet (approved):
LGTM

swagner-de (approved):

PR #6994, opened 2024-08-22
[pxc-operator] Add basic pxc-operator chart

s10:


system/pxc-operator/values.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+logLevel: "INFO"
+
+watchNamespace: ""
+# watchNamespace: "monsoon3,monsoon3global"
+
+ccFixPemissions: "true"
+
+resources:
+  limits:
```
galkindmitrii, line 19: I'd set those higher directly from the start.

PR #6996, opened 2024-08-22
[utils] Add support for secrets-injector for db_url_pxc function

s10:
Use resolve secret with urlquery, because it's an oslo.db connection


galkindmitrii (approved):
Thanks!

joker-at-work (approved):

PR #6997, opened 2024-08-23
[utils] Change kubernetes-entrypoint image repo

joker-at-work:
The old version of kubernetes-entrypoint was based on a rotten image and not updated in years. We now switch to a new upstream and rebuild the base image for security updates regularly through our "shared-app-images" pipeline. The tag of the image contains the version and a build number.

Since PCI forces us to do staged deployments for feature changes, we track the version in the `utils` chart. This means we need a `utils` chart bump to get a new version in, which will be transported through the stages by the individual services' pipelines.

The build version, that's only incremented for base OS updates, can be updated in all regions at the same time. Therefore, we track the latest build version - and with that the one we want to deploy in a region - through a region-local values file.

The idea would be to follow up with a pipeline that updates the build version in the values file automatically every time a new image is built.


openstack/utils/values.yaml
```diff
@@ -6,6 +6,9 @@
 global:
   user_suffix: ""
   master_password: ""
+  # defined in values files so it can be updated automatically
+  # kubernetes_entrypoint_build_version:
+  #   4fbcf7ce324dc66e78480f73035e31434cfea1e8: latest
```
fwiesel, line 11: If the version is practically non-human readable (a hash versus an actual version), I think we are better off using the content hash right away.

joker-at-work: I don't think I follow. Do you mean the content-hash of the snippet?
The "version" is defined in the shared-app-images pipeline as that's the "version" (whatever version means here) that's pulled from the upstream repository. We also have to manually bump it to comply with PCI staged deployments.

fwiesel: Sorry, the correct term would be "image digest".

To expand: The proposal is to specify the image in the form:

`image: <repo>:<image-tag>` -> `image: repo:somefairlylonghash-dateandtimetothesecond`.

To my eyes, a tag in this format isn't really human-readable anymore, so we might as well go with:

`image: <repo>@<digest>` -> `image: repo@sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2`

I know, we do lose version pinning, but that container is mostly in maintenance mode anyway, so we might roll the whole thing forward as just one blob.

joker-at-work: Then we lose the automatic update on security-rebuild if we need to keep the hash here. The point of this proposal was to have explicit version updates by utils chart bumping while getting security updates (i.e. the last part of the tag updated) automatically.

If we don't want the automatic security updates of the base OS for the same upstream version rolled out automatically, I don't care how we track the image - if by tag or digest.

fwiesel: I'll close this, as the topic seems to be more generic than how to store the tag / version.

fwiesel:
> Since PCI forces us to do staged deployments for feature changes, we track the version in the utils chart.

AFAIK, PCI requires us to have an engineering policy and that one requires a staged deploy. 
Reading the policy, I do not see any distinction between features and security updates (or configuration for that matter). Personally, I would advocate against making such a distinction in general (Though I see a point in this particular case, see next paragraph).

> The build version, that's only incremented for base OS updates, can be updated in all regions at the same time. Therefore, we track the latest build version - and with that the one we want to deploy in a region - through a region-local values file.

While I wrote  that I am not a fan of that, I have to agree that the OS updates for that particular image are very likely not causing any problems, because it is mostly the result of them choosing a large base image with binaries not being used at all.

But for that exists another solution: Build our own images from the repository with a minmal (i.e. distroless / ko build) process.

fwiesel:
It seems I was mistaken: https://github.com/airshipit/kubernetes-entrypoint/blob/master/images/Dockerfile.ubuntu_focal#L32-L35

Some OS tools are needed after all.

joker-at-work:
Some OS tools are integrated. I wouldn't immediately assume they are actually used. Building our own image is definitely an option. Then we would "just" switch the version to the latest specific tag from the shared-app-images to get away from the rotten old image and follow-up with creating our own image and then updating the `utils` chart again for that new repo/tag.

fwiesel:
The package was added in the following commit: https://github.com/airshipit/kubernetes-entrypoint/commit/f3ccc07e866eba463a5a6e6f3c9441fba68bdc78

Looking at the [code](https://github.com/airshipit/kubernetes-entrypoint/blob/master/kubernetes-entrypoint.go#L31-L41) it seems it happens when you explicitly pass `COMMAND=echo`, which is presumably their particular use-case (ours is to pass `COMMAND=true`).
We could simply leave `COMMAND` empty and live with the warning (my preferred option), or add an empty file as `/usr/bin/true`.

fwiesel:
I've created some [PRs](https://github.com/airshipit/kubernetes-entrypoint/pulls/fwiesel) on the upstream repo, in the hope they will take them. If that is the case, it could reduce our maintenance work.

joker-at-work:
Do they not use gerrit like the rest of opendev?

fwiesel:
There is also a gerrit repo... but honestly, I don't know what workflow they follow. The CONTRIBUTION.md looked more like a github workflow, but it is a bit ambiguous.

fwiesel:
They do have a [gerrit project](https://review.opendev.org/q/project:airship/kubernetes-entrypoint), so I moved the PRs there.

PR #7008, opened 2024-08-28
nova/flavors: Add m_c4_m32 for BTP

grandchild:
BTP requested this flavor with 8-to-1 RAM-to-CPU ratio.

While it is outside of the hardware RAM-CPU ratio, it's a small flavor, and it's high-memory which is not as TCO-sensitive as high-CPU flavors.

Create m_c4_m32_v2 for Sapphire-Rapids clusters as well.


dhoeller (approved):
LGTM

PR #7009, opened 2024-08-28
[manila] alert on broken db purge

Carthaca:
have a look at the individual commit messages


kpawar-sap (approved):

Carthaca:
adjusted the query so that the ID is the ID of the oldest share and not just the first entry in the db, thanks to Chuan

PR #7014, opened 2024-08-29
[pxc-db] helm chart created

s10:


businessbean:
Let's merge the work into https://github.com/sapcc/helm-charts/pull/7076

businessbean:
Let's merge the work into https://github.com/sapcc/helm-charts/pull/7076

common/pxc-db/values.yaml
```diff
@@ -0,0 +1,389 @@
...
+    annotations:
+      config.linkerd.io/opaque-ports: "3306,3307,3009,4444,4567,4568,33060,33062"
+  autoRecovery: true
+  configuration: |
+    [mysqld]
+    wsrep_provider_options          = "gcache.size=1G; gcache.recover=yes"
+    wsrep_slave_threads             = 4
+    wsrep_retry_autocommit          = 3
+    binlog_format                   = ROW
+    max_connections                 = 1024
```
galkindmitrii, line 127: jfyi: we have higher limits for nova/neutron at least in large regions.

common/pxc-db/values.yaml
```diff
@@ -0,0 +1,389 @@
...
+    innodb_flush_log_at_trx_commit  = 0
+    innodb_flush_method             = O_DIRECT
+    innodb_file_per_table           = 1
+    innodb_autoinc_lock_mode        = 2
+    innodb_buffer_pool_size         = 1024M
```
galkindmitrii, line 132: the same, will need to be adjusted per service/region basis.

common/pxc-db/values.yaml
```diff
@@ -0,0 +1,389 @@
+# Default values for percona-xtradb-cluster.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+linkerd:
+  enabled: true
+
+owner_info:
+  enabled: false
```
galkindmitrii, line 8: I don't think we need it, because owner_info comes as dependency from openstack charts.

common/pxc-db/Chart.yaml
```diff
@@ -0,0 +1,30 @@
...
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.15.0"
+
+dependencies:
```
galkindmitrii, line 26: It is not enabled for https://github.com/sapcc/helm-charts/blob/master/common/mariadb/Chart.yaml and comes as openstack chart dependency.

common/pxc-db/values.yaml
s10: Removed this dependency, because it's no longer needed in labs

common/pxc-db/Chart.yaml
s10: Removed this dependency, because it's no longer needed in labs for independent chart install

common/pxc-db/values.yaml
s10: Moved all these options to hash `pxc.configuration.options`. Now these options can be overwritten individually.

```
pxc:
  configuration:
    options:
      max_connections: 2048
      innodb_buffer_pool_size: 2048M
```

galkindmitrii (approved):

businessbean (approved):
lgtm

PR #7021, opened 2024-09-02
[unbound] toggle the rpz-signal-nxdomain-ra flag for the RPZ zones

vssldmtrv:
That will unset the RA bit in the NXDOMAIN response for a blocked domain.

That will help clients deduce whether an NXDOMAIN is a result of an RPZ match (RA unset) or an actual, authoritative NXDOMAIN (RA set).


galkindmitrii (approved):

mutax (approved):
lgtm

PR #7027, opened 2024-09-02
Add owner-label-injector-staging config to knownBrokenReleases

onuryilmaz:
These namespace and charts are used to test owner-label-injector in qa-de-1


majewsky (approved):
I constricted the

PR #7029, opened 2024-09-03
elektra: use postgresql-ng version from library chart

majewsky:
We recommend not pinning the image version and using the default that is hardcoded into the library chart. Updates can be received by bumping the chart dependency with `helm dep up` (which can be automated with the respective shared CI task).


ArtieReus (approved):
lgtm Thanks!!!

PR #7031, opened 2024-09-03
redis: make Deployment roll out on configuration changes

stanislav-zaprudskiy:
The mechanics of how the annotation causes the deployment to roll out is explained here - https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments.

Without this PR any changes of Redis configuaration or password require additional intervention (e.g. manual `kubectl rollout restart deployment`) to make them picked up, which doesn't scale well on frequent changes or with numerous deployments.

### DIFF
--- a/common/redis/templates/deployment.yaml
+++ b/common/redis/templates/deployment.yaml
@@ -29,6 +29,12 @@ spec:
         {{- if and $.Values.global.linkerd_enabled $.Values.global.linkerd_requested }}
         linkerd.io/inject: enabled
         {{- end }}
+        {{- if .Values.config }}
+        checksum/configmap: {{ include (print .Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- end }}
+        {{- if .Values.redisPassword }}
+        checksum/secrets: {{ include (print .Template.BasePath "/secrets.yaml") . | sha256sum }}
+        {{- end }}
     spec:
       containers:
       - name: {{ template "redis.fullname" . }}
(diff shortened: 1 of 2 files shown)


common/redis/templates/deployment.yaml
```diff
@@ -29,6 +29,12 @@ spec:
         {{- if and $.Values.global.linkerd_enabled $.Values.global.linkerd_requested }}
         linkerd.io/inject: enabled
         {{- end }}
+        {{- if .Values.config }}
+        checksum/configmap: {{ include (print .Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- end }}
+        {{- if .Values.redisPassword }}
+        checksum/secrets: {{ include (print .Template.BasePath "/secrets.yaml") . | sha256sum }}
+        {{- end }}
```
SuperSandro2000, line 37: I've looked through our other things and sometimes we use Template.BasePath and sometimes not. I leave it up to @majewsky what we use.

I don't thinkwe need to be conditional here, worst case we hash the empty files.

```suggestion
        checksum/configmap: {{ include "redis/templates/configmap.yaml" . | sha256sum }}
        checksum/secrets: {{ include "redis/templates/secrets.yaml" . | sha256sum }}
```

majewsky: I believe the form with `.Template.BasePath` is necessary for library charts that are included as dependencies elsewhere, so the full path would be something like `my-app/charts/redis/templates/configmap.yaml`.

majewsky (changes requested):
We can take the current form with `.Template.BasePath`, but please bump the version in Chart.yaml before merging.

majewsky (approved):

PR #7034, opened 2024-09-03
[rabbitmq] remove unused unacknowledged alert from rename

notque:


s10 (approved):

bashar-alkhateeb (approved):
looks good as we removed the support of metrics side container in version [0.8.0 ](https://github.com/sapcc/helm-charts/blob/master/common/rabbitmq/CHANGELOG.md#080)

PR #7037, opened 2024-09-05
Add scheduler seed

PhilippMatthes:
cc @auhlig 

WIP


openstack/scheduler/templates/_helpers.tpl
```diff
@@ -0,0 +1,16 @@
...
+{{- end -}}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+*/}}
+{{- define "fullname" -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
```
majewsky, line 16: If you don't need multiple installations of this chart in the same namespace (which approximately no one ever does), I recommend to follow YAGNI and get rid of all of this.

openstack/scheduler/templates/domain-ccadmin-seed.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: "openstack.stable.sap.cc/v1"
+kind: "OpenstackSeed"
+metadata:
+  name: scheduler-seed
+  labels:
+    app: {{ template "fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
```
majewsky, line 9: Same with this stuff. Helm automatically adds annotations to track provenance of the objects that it provisions, so those can be used to trace back the seed to its containing Helm release.

auhlig (approved):

majewsky (approved):

PR #7050, opened 2024-09-09
[manila] cleanup migration job

Carthaca:
remove condition for secret config.
All secrets have been moved, so it is no longer needed for
'pre-upgrade'.
And the condition is wrong for 'post-install',
because the lookup happens before install and the migration script
would fail to run without the secrets on a fresh installation now.


kpawar-sap (approved):

chuan137 (approved):
lgtm

PR #7052, opened 2024-09-09
chore(kube-monitoring): use version constraint for oomkill

IvoGoman:
version constraints was introduced for scaleout, this add them for the other contexts as well.
This makes sure the routing fix for `PodOOMKill` will be made available for the other clusters as well. 
If you feel this is to pushy, I am also fine with replacing the `appVersion` with `version` and setting it to latest.

### DIFF
--- a/system/kube-monitoring-admin-k3s/Chart.yaml
+++ b/system/kube-monitoring-admin-k3s/Chart.yaml
@@ -26,7 +26,7 @@ dependencies:
     version: "^2.5"
   - name: oomkill-exporter
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-    version: 0.5.11
+    version: "^1.x"
   - name: ping-exporter
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
     version: 0.1.1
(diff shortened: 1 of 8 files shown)


Kuckkuck (approved):

system/kube-monitoring-admin-k3s/Chart.yaml
```diff
@@ -26,7 +26,7 @@ dependencies:
     version: "^2.5"
   - name: oomkill-exporter
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-    version: 0.5.11
```
viennaa, line 29: Wouldn't this break if a new version is pulled in, since no `helm dep up` is run?

viennaa (commented):
Thanks for updating these deps. Comment inline.

system/kube-monitoring-admin-k3s/Chart.yaml
```diff
@@ -26,7 +26,7 @@ dependencies:
...
+    version: "^1.x"
```
viennaa, line 29: Wouldn't this break if a new version is pulled in, since no helm dep up is run?

IvoGoman: My understanding is that `helm dep up` will update the lock file with the newest version.
`helm diff` & `helm upgrade` should use the `Chart.lock` to lookup the dependency version, no?

PR #7061, opened 2024-09-11
[designate] Add managed_resource_tenant_id variable into designate.conf

misamoylov:


openstack/designate/templates/etc/_designate.conf.tpl
```diff
@@ -104,7 +104,7 @@ default_pool_id = '794ccc2c-d751-44fe-b57f-8894c9f5c842'
 
 # Tenant ID to own all managed resources - like auto-created records etc.
 #managed_resource_tenant_id = 123456
```
galkindmitrii, line 106: Please remove duplicated setting here.

galkindmitrii (approved):

galkindmitrii (approved):

openstack/designate/templates/etc/_designate.conf.tpl
```diff
@@ -103,8 +103,7 @@ default_pool_id = '794ccc2c-d751-44fe-b57f-8894c9f5c842'
 #managed_resource_email = hostmaster@example.com.
 
 # Tenant ID to own all managed resources - like auto-created records etc.
-#managed_resource_tenant_id = 123456
-
+managed_resource_tenant_id = {{ .Values.managed_resource_tenant_id | default "00000000-0000-0000-0000-000000000000" }}
```
mutax, line 106: I would prefer to have an if statement surrounding the config value that checks if its set instead of introducing a magic number 000...

galkindmitrii: the default is hardcoded in Designate code
https://github.com/sapcc/designate/blob/129e5fabeaeb3909018497af4cfd9d8b23adea0c/designate/conf/central.py#L40
https://docs.openstack.org/designate/latest/admin/samples/config.html

PR #7064, opened 2024-09-12
[Keystone] upgrade to openstack bobcat

rajivmucheli:


openstack/keystone/values.yaml
```diff
@@ -194,7 +194,7 @@ tempest:
   enabled: false
   image: "rally-openstack"
   #imageTag: "ccloud-1.5.0"
-  imageTag: "ccloud-zed"
+  imageTag: "ccloud-bobcat"
```
bbobrov, line 197: does this tag exist? I suggest we set the value to "", and let decide the tag in per-region values.

openstack/keystone/values.yaml
```diff
@@ -31,7 +31,7 @@ run_db_migration: true
 skipRepairRoleAssignments: false
 
 # current keystone release
-release: zed
+release: bobcat
```
bbobrov, line 34: lets remove this too? I think it is not used anywhere at all.

BerndKue (approved):

PR #7070, opened 2024-09-12
alertmanager: fix accidental cross-service silences

majewsky:


global/prometheus-alertmanager-operated/templates/am-config-inhibit-rules-base.yaml
```diff
@@ -52,10 +52,16 @@ spec:
         - matchType: "=~"
           name: severity
           value: 'critical|warning'
+        - matchType: "=~"
+          name: context
+          value: '.+'
```
IvoGoman, line 57:
```suggestion
        - matchType: "!="
          name: context
          value: ''
```

global/prometheus-alertmanager-operated/templates/am-config-inhibit-rules-base.yaml
```diff
@@ -52,10 +52,16 @@ spec:
...
       targetMatch:
         - matchType: "="
           name: severity
           value: 'info'
+        - matchType: "=~"
+          name: context
+          value: '.+'
```
IvoGoman, line 64:
```suggestion
        - matchType: "!="
          name: context
          value: ''
```

global/prometheus-alertmanager-operated/templates/am-config-inhibit-rules-base.yaml
```diff
@@ -31,17 +31,17 @@ spec:
...
       targetMatch:
         - matchType: "="
           name: severity
           value: 'warning'
-        - matchType: "="
+        - matchType: "=~"
           name: context
           value: '.+'
```
IvoGoman, line 46:
```suggestion
        - matchType: "!="
          name: context
          value: ''
```

global/prometheus-alertmanager-operated/templates/am-config-inhibit-rules-base.yaml
```diff
@@ -31,17 +31,17 @@ spec:
...
-         value: 'critical'
-       - matchType: "="
-         name: context
-         value: '.+'
+        - matchType: "="
+          name: severity
+          value: 'critical'
+        - matchType: "=~"
+          name: context
+          value: '.+'
```
IvoGoman, line 39:
```suggestion
        - matchType: "!="
          name: context
          value: ''
```

IvoGoman (changes requested):
I have manually validated the matcher on our alertmanager.
For an Alert **without** the context label it did still match with `context =~ '.*'`. With `context != ''` these alerts are getting filtered out.

global/prometheus-alertmanager-operated/templates/am-config-inhibit-rules-base.yaml
majewsky: I'm assuming that means you are confident that `!=` is allowed here. I was going to use that first, but did not find good documentation on which match types are allowed, so I kept to what was already used in the file.

IvoGoman: Yes, you can double check within the link I shared in Company Slack

viennaa: The Matcher is allowed:

https://github.com/prometheus-operator/prometheus-operator/blob/9092aa95a3acfd4c5caedacd0afffd315cf5a154/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L1039-L1044

viennaa:
Can you elaborate, when exactly did these accidental cross service silences happen? I can get the justification out of the name.

viennaa:
One more thing: bump patch version of the chart pls.

IvoGoman:
> Can you elaborate, when exactly did these accidental cross service silences happen? I can get the justification out of the name.

I have cc'ed you on the slack thread

majewsky:
> One more thing: bump patch version of the chart pls.

Done.

viennaa (approved):
Since context is missing on many alerts, we don't want the unexpected inhibits. Taking suggestion.

IvoGoman (approved):

PR #7079, opened 2024-09-16
keystone: remove references to cloud_dns_resource_admin role

majewsky:
This role was deleted in 48e630076c812e920ae3495830c6ba47f13e4064.


openstack/keystone/templates/etc/_policy.yaml.tpl
```diff
@@ -31,7 +31,6 @@
 "blocklist_roles": "'resource_service':%(target.role.name)s or
   'cloud_registry_admin':%(target.role.name)s or
   'cloud_registry_viewer':%(target.role.name)s or
-  'cloud_dns_resource_admin':%(target.role.name)s or
```
bbobrov, line 34: Are you completely sure that the role is gone and does not give any special access? What is the worst case, if the role is somehow recreated, and will get used by a user?
If you not sure, we can leave the role here for now, it costs nothing for it to be here; maybe with a TODO to remove it later.

majewsky: > Are you completely sure that the role is gone

I deleted it in all regions, which you can confirm by finding no results for `openstack_assignments_per_role{role_name="cloud_dns_resource_admin"}` in the global Thanos.

> and does not give any special access

It was only accepted in the Limes policy, where I have removed it in all regions. GitHub code search (both in this org and the internal org) does not yield any results beyond UI strings and blocklist entries like this one.

> maybe with a TODO to remove it later.

The thing is that I always forget to come back to these. :)

bbobrov (approved):

galkindmitrii (approved):

PR #7080, opened 2024-09-16
feat(kube-monitoring): don't manage preset externally

viennaa:
* preset should be maintained with the cluster selector and the cluster

* this is obfuscating what actually matches

* switch to generic label


richardtief (approved):
Thanks.

PR #7086, opened 2024-09-17
castellum: use thanos-metal for Manila queries

majewsky:
As discussed with @Carthaca @chuan137 @viennaa in Slack.


viennaa (approved):
Go fly 🚀

PR #7087, opened 2024-09-17
feat(thanos): adjusting ingress for thanos

viennaa:
* introducing metrics.st1 urls


richardtief (approved):
Ok. I'm not happy about the extra stuff - but it's necessary because of the operations. We need to fill in the gaps afterwards. The Thanos plugin should provide an Ingress resource that can be configured with additional annotations, for example.

PR #7094, opened 2024-09-17
Add new PriorityClasses

Nuckal777:


galkindmitrii:
Are we keeping the old `openstack-service-critical` priority class or replacing it after?
https://github.com/search?q=repo%3Asapcc%2Fhelm-charts%20openstack-service-critical&type=code

Nuckal777:
I'd like to replace that in a follow-up PR.

PR #7116, opened 2024-09-23
[hermes] log malformed target project_id

notque:
Log and correct malformed project_ids, trying to get more detail on where the data is coming from that is causing a problem with arrays in target project_id


notque:
this is now functioning, and not causing errors with the structure.

PR #7117, opened 2024-09-24
[Keystone] remove tempest setup in keystone helm-chart

rajivmucheli:
The new setup is located at https://github.com/sapcc/helm-charts/tree/master/openstack/tempest/keystone-tempest


rajivmucheli:
This change was introduced before tempest tests were green. Let's take a step back, merge the previous tempest PR, validate it for few days and look into this PR.

stanislav-zaprudskiy (changes requested):
Removing references to the deleted template files is also required in [`templates/configmap-etc.yaml`](https://github.com/sapcc/helm-charts/blob/remove_tempest_keystone/openstack/keystone/templates/configmap-etc.yaml) (e.g. `rally_deployment_config.json`)

stanislav-zaprudskiy:
> @bbobrov shall we remove this spec https://github.com/sapcc/helm-charts/blob/remove_tempest_keystone/openstack/keystone/values.yaml#L190-L198 now that all PR's are merged and working ?

I believe it would be failing during deployments where `.Values.tempest` hash is not provided, e.g. while performing this check here https://github.com/sapcc/helm-charts/blob/remove_tempest_keystone/openstack/keystone/templates/etc/_policy.yaml.tpl#L24.

rajivmucheli:
> I believe it would be failing during deployments where `.Values.tempest` hash is not provided, e.g. while performing this check here https://github.com/sapcc/helm-charts/blob/remove_tempest_keystone/openstack/keystone/templates/etc/_policy.yaml.tpl#L24.

When i ran it locally it worked and the idea is to control this in secrets (tempest is only enabled in qa).

stanislav-zaprudskiy (approved):

openstack/keystone/values.yaml
```diff
@@ -189,14 +189,6 @@ cache:
...
 tempest:
   enabled: false
-  image: "rally-openstack"
-  #imageTag: "ccloud-1.5.0"
-  imageTag: ""
-  adminPassword: ""
-  userPassword: ""
-  # tempest domain-id  -- do not set in production environments!
-  domainId: not_set_this_in_production
-  adminProjectId: not_set_in_production
```
bbobrov, line 199: This value is still needed

openstack/keystone/values.yaml
```diff
@@ -189,14 +189,6 @@ cache:
 
 tempest:
   enabled: false
-  image: "rally-openstack"
-  #imageTag: "ccloud-1.5.0"
-  imageTag: ""
-  adminPassword: ""
-  userPassword: ""
-  # tempest domain-id  -- do not set in production environments!
-  domainId: not_set_this_in_production
```
bbobrov, line 198: This value is still needed

bbobrov (changes requested):
This must also have an approval from @misamoylov

openstack/keystone/values.yaml
rajivmucheli: this value is enabled in secrets.

openstack/keystone/values.yaml
rajivmucheli: this value is enabled in secrets.

bbobrov (commented):
This is good, but it still must be approved by @misamoylov

PR #7123, opened 2024-09-24
[Cinder]ignore down pools in cinder alert expr

bashar-alkhateeb:
Fix cinder CinderShardMaxVolumeSize alert to not count pools in state=down while checking for free pools
that to trigger the alert even if down pools have capacity.

Change-Id: I5a2c7e9a49e26454cadeda8bba925cf7a6653e44


hemna (approved):
lgtm thank you!

joker-at-work (approved):

viennaa (changes requested):
Thanks for improving this alert.

Please increase minor version when altering Chart content.

bashar-alkhateeb:
done!

viennaa (approved):
Thanks!

PR #7125, opened 2024-09-24
[tempest] Disable domain scope token for keystone tests

misamoylov:


rajivmucheli (approved):

openstack/tempest/keystone-tempest/templates/bin/_tempest-start-and-cleanup.sh.tpl
```diff
@@ -17,6 +17,7 @@ function cleanup_tempest_leftovers() {
   export OS_PROJECT_DOMAIN_NAME=tempest
   export OS_DOMAIN_NAME=tempest
```
bbobrov, line 18: This must be removed, because out of `OS_PROJECT_NAME` and `OS_DOMAIN_NAME` only 1 must be provided.

PR #7129, opened 2024-09-25
[ironic] add flavor bm_s4_c32_m1024_v0 for HD&A servers

BerndKue:


BerndKue:
Need the information about the right disk size as soon as the first server arrives.

sandzwerg:
I think this is superseeded and can be closed @BerndKue ?

PR #7138, opened 2024-09-25
add alert: OpenstackLimesIneffectiveQuotaOverride

majewsky:
Followup to sapcc/limes#569.


openstack/limes/alerts/openstack/api.alerts
```diff
@@ -293,3 +293,19 @@ groups:
...
+      severity: warning
+      support_group: containers
+      tier: os
+      service: limes
+    annotations:
+      summary: Quota override is ineffective
+      description: |
+        In the project {{ $labels.project_id }} ({{ $labels.domain }}/{{ $labels.project }}), the quota override for {{ $labels.service }}/{{ $labels.resource }} is ineffective.
+        The actual quota is {{ $value }}, which differs from the configured quota override.
+        Please check the Resource Management UI of the project. If the quota override cannot be enforced because commitments or usage are higher than it, the override needs to be removed (from quota-overrides.yaml in cc/secrets).
```
SuperSandro2000, line 310: Can we add a deep link?

VoigtS (approved):
A deeplink sandro mentioned would be fine, but I think the description already tells us everything we need to know for this case. I'm ok with either approach.

PR #7141, opened 2024-09-26
Fix incorrect indention for prevented host

swagner-de:
If multiple host were prevented (due to miscabling) the first one would
print with a correct indent, due to the contnrol instruction being
indented, yet the second host would have a different indent. That lead
to a YAML parse error. To fix that, we are using `nindent` which inserts
a line break and then start with the values indented by `nindent`s
value.


occamshatchet (approved):
LGTM

ikatza-at-work (approved):

nikatza (approved):

rhalina (approved):

PR #7147, opened 2024-09-27
fix(prometheus): remove no longer needed thanos rule

viennaa:
* metrics are all available to Prometheus itself, hence the switch makes no sense anymore.

* fixing the unsoftened condition for prometheus-kubernetes (was still on trigger > 0.1) which actually triggered permanent alerts

* removing thanos condition, has no impact anymore

* adjusting time from 4d to hours, since 4d could never be met anymor (12h retention)


common/prometheus-server/templates/alerts/_prometheus-thanos-rule.alerts.tpl
```diff
@@ -34,38 +34,3 @@ groups:
       description: Prometheus is scraping `{{`{{ $labels.pod }}`}}` pods in namespace `{{`{{ $labels.namespace }}`}}` multiple times. This is likely caused due to incorrectly placed scrape annotations. <https://{{ include "prometheus.externalURL" . }}/graph?g0.expr={{ urlquery `up * on(instance) group_left() (sum by(instance) (up{kubernetes_pod_name=~"PLACEHOLDER.*"}) > 1)` | replace "PLACEHOLDER" "{{ $labels.pod }}"}}|Affected targets>
       summary: Prometheus scrapes pods multiple times
   {{- end }}
-  {{- if and (eq $root.Values.vpaUpdateMode "Auto") $root.Values.alerts.thanos.enabled }}
-  {{/* This is covering all Prometheis but kubernetes. They don not have the vpa_butler metric and need to leverage thanos ruler instead. */}}
```
richardtief, line 38: This is still true, isn't it?

richardtief (commented):
Thanks. One question below.

viennaa:
Scratch this PR. Follow up in #7157

PR #7154, opened 2024-10-01
[pxc-db] Remove hook annotations from pxc-db secrets resources

s10:


galkindmitrii (approved):

bashar-alkhateeb:
do we have any Jobs that read this secrets on helm-upgrade ?

s10:
> Do we have any Jobs that read these secrets on helm-upgrade ?

We have a job "initdb-job.yaml", but it no longer uses these secrets, changed in this PR.

bashar-alkhateeb (approved):

PR #7157, opened 2024-10-01
chore(prometheus|thanos): default retention up and alert condition re…

viennaa:
…fined

### DIFF
--- a/common/prometheus-server/values.yaml
+++ b/common/prometheus-server/values.yaml
@@ -34,7 +34,7 @@ name:
 # names:
 
 # Defines how long data is stored. Format: `[0-9]+(ms|s|m|h|d|w|y)`
-retentionTime: 12h
+retentionTime: 25h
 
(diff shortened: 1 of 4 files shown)


richardtief (commented):
All good. One question regarding the retention.

common/prometheus-server/values.yaml
```diff
@@ -34,7 +34,7 @@ name:
 # names:
 
 # Defines how long data is stored. Format: `[0-9]+(ms|s|m|h|d|w|y)`
-retentionTime: 12h
+retentionTime: 25h
```
richardtief, line 37: This is volume issue safe?

viennaa: I don't expect big changes in volume size here. For Thanos this has no impact where volume size usually has the bigger impact.

richardtief (approved):

PR #7173, opened 2024-10-04
Adding flavor for Lenovo AMD SR655 v3

degricar:


openstack/ironic/templates/seed.yaml
```diff
@@ -517,6 +517,18 @@ spec:
       "resources:CUSTOM_HV_S8_C32_M2048_V0": "1"
       {{- tuple . "baremetal" | include "ironic.helpers.extra_specs" | indent 6 }}
 
+  # new Lenovo AMD servers
+  - name: "hv_s1_c24_m196_v0"
+    id: "2038"
```
BerndKue, line 522: Has to be one higher than the last flavor. Would be 2039 then.

BerndKue (changes requested):

openstack/ironic/templates/seed.yaml
degricar: done

BerndKue (approved):

sandzwerg (approved):

PR #7174, opened 2024-10-04
[rabbitmq_cluster] Add support for secrets-injector

s10:
Create custom secret in required format and use it for rabbitmq-cluster


s10:
rabbitmq-operator should be updated to v2.1.0+ for this to work

galkindmitrii (approved):
Even if that is not used right now, let's get this merged. Thanks!

businessbean (approved):
lgtm

PR #7178, opened 2024-10-07
[kube-system-metal] set nging-external service externalTrafficPolicy

defo89:


defo89:
FYI @galkindmitrii following up with service I missed to change last time

PR #7183, opened 2024-10-07
[utils][pxc-db] Add support for multiple ProxySQL secrets creation

s10:
If proxysql.multiSecret is set and multiple database types are enabled, then proxysql_secret helper will additionally create two ProxySQL secrets:
- for all pxc-db dependencies
- for all mariadb dependencies


galkindmitrii (approved):

fwiesel:
But where are they used? And why not just add the secrets as additional fields to the existing secret?

PR #7188, opened 2024-10-08
ironic watcher

sandzwerg:
- **ironic: enable watcher in testing**

### DIFF
--- a/openstack/ironic/templates/etc/_ironic.conf.tpl
+++ b/openstack/ironic/templates/etc/_ironic.conf.tpl
@@ -120,21 +120,20 @@ port_setup_delay = {{ .Values.neutron_port_setup_delay }}
 [oslo_middleware]
 enable_proxy_headers_parsing = True
 
-{{- if .Values.watcher.enabled }}
+{{ if .Values.watcher.enabled -}}
 [watcher]
 enabled = true
 service_type = baremetal
 config_file = /etc/ironic/watcher.yaml
-{{ end }}
+{{- end }}
 
-{{ if .Values.audit.enabled }}
+{{ if .Values.audit.enabled -}}
 [audit]
 enabled = True
 audit_map_file = /etc/ironic/api_audit_map.yaml
 ignore_req_list = GET, HEAD
 record_payloads = {{ if .Values.audit.record_payloads -}}True{{- else -}}False{{- end }}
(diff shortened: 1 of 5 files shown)


openstack/ironic/templates/etc/_ironic.conf.tpl
```diff
@@ -120,21 +120,20 @@ port_setup_delay = {{ .Values.neutron_port_setup_delay }}
 [oslo_middleware]
 enable_proxy_headers_parsing = True
 
-{{- if .Values.watcher.enabled }}
+{{ if .Values.watcher.enabled -}}
```
fwiesel, line 123: Generally, I have made good experiences with `{{-` practically everywhere.

sandzwerg: But here it would remove the empty line, so the result looks like this:

```
[oslo_middleware]
enable_proxy_headers_parsing = True
[watcher]
enabled = true
service_type = baremetal
config_file = /etc/ironic/watcher.yaml
```

instead of 

```
[oslo_middleware]
enable_proxy_headers_parsing = True

[watcher]
enabled = true
service_type = baremetal
config_file = /etc/ironic/watcher.yaml
``` 

The first block is from a live pod, so that's the current state, and from my understanding it should more look like the second example. From my reading of https://helm.sh/docs/chart_template_guide/control_structures/#controlling-whitespace that should do that and local templating confirms it.

sandzwerg:
OK, @stefanhipfel reasons that the conductor has no things that the watcher needs to watch, so we should just remove it altogether. The issue is that the same ironic.conf is used for API and conductors.

openstack/ironic/templates/etc/_ironic.conf.tpl
fwiesel: You can add the empty line to the block you include...
```
{{- if something someting }}

[block]
{{-  end }}
```

fwiesel:
You could add an `ironic-api.conf`.  By default that one should be loaded automatically and merged  the "global" `ironic.conf` by the ironic-api process, but not ironic-conductor

sandzwerg:
The latest commits should have done something similar. I moved the watcher config in its own file as suggested by @joker-at-work Please check if everything fits. I did my best but I can't rule out I missed some reference.

fwiesel (approved):

joker-at-work (approved):

stefanhipfel (approved):

PR #7193, opened 2024-10-08
ironic: enable audit as default

sandzwerg:
We somehow apparently never enabled audit. Now we enable it by default.


notque:
Connected PR to update the audit map https://github.com/sapcc/helm-charts/pull/7194

BerndKue (approved):

PR #7195, opened 2024-10-09
Update seed.yaml

degricar:
Fix memory value for AMD and set Superdome BM flavor to public


BerndKue (approved):

BerndKue:
The old seeder was not able to update properties of existing flavors when changing a property in the seeds.
I suppose you need to change the flavors in the regions where the Ironic pipeline already rolled them out.

sandzwerg (approved):

PR #7200, opened 2024-10-09
adding alert for exhausted PVC

viennaa:


majewsky (approved):
We discussed this in the office. The existing `KubernetesHighPVCUsagePredicted` is too unpredictable (and thus likely to be ignored or silenced), and also too unreliable for very slow growth. I was hesitant to add a general nearly-full alert because some PVCs may legitimately be rather full without it causing a problem, but 99% is a sufficiently high bar to ensure that false positives should be minimal.

SuperSandro2000:
Why not delete the https://github.com/sapcc/helm-charts/blob/dc5c91ce65d563b21ffc225f4b4ae660d77566b7/prometheus-rules/prometheus-controlplane-rules/alerts/kubernetes-node.alerts#L122 then if it is really that bad?

viennaa:
If we delete it we should introduce another stage with like > .9 usage and throw this as an info alert.

viennaa:
@majewsky @Nuckal777 @SuperSandro2000 if no objections, I would merge.

Nuckal777 (approved):

PR #7221, opened 2024-10-11
[system/metis] Update mariadb chart to 0.14.2

fwiesel:
This is needed for the switch to secrets-injector. It will cause a reboot of the db due to a different sidecar image version.


fwiesel:
Already done. Sorry about the noise.

PR #7222, opened 2024-10-11
[kmip] Update mariadb chart to 0.14.2

fwiesel:
This is needed for the switch to secrets-injector. It will cause a reboot of the db due to a different sidecar image version.


rajivmucheli:
linter errors are fixed here : https://github.com/sapcc/helm-charts/pull/7212

PR #7224, opened 2024-10-11
PX IPv6 support

swagner-de:
- **Use StatefulSet instead of deployments**
- **Simplify PX charts: Region names not put all over all places**
- **Use different naming for statefulset**
- **[PX] Set bird as default container**
- **Remove metricRelabelings as the same labels are already created**
- **Support legacy config file path and new config file path**
- **Use different naming for statefulset**
- **Use 1 statefulset with replicas:2 instead of 2 statefulsets**
- **Let router-id be determined from init-network**
- **Add bird IPv6 statefulsets once an IPv6 network was supplied.**


px/lg/templates/configmap.yaml
```diff
@@ -6,6 +6,15 @@ metadata:
     name: {{ $lg }}
 data:
     "lg_config.cfg": |
-{{ include "lg_conf" $lg_config | indent 8 }}
+        DEBUG = False
+        LOG_FILE="/var/log/lg.log"
+        LOG_LEVEL="INFO"
+
+        BIND_IP = "0.0.0.0"
+        BIND_PORT = 80
```
nikatza, line 14: Do we have any rules regarding unencrypted HTTP traffic?

nikatza (approved):

px/bird/templates/configmap-init-network.yaml
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        if [[ $(ip_assigned) == "unassigned" ]]; then
+            assign_ip_address
+        fi
+        
+        echo "Setting promiscious mode on $PX_INTERFACE"
+        ip link set $PX_INTERFACE promisc on
+
+    calculate_ip_by_statefulset_ordinal.py: |
```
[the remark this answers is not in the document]
swagner-de: The bird container has python anyway as we use a python executable to monitor the config reload.

px/bird/values.yaml
```diff
@@ -66,47 +66,37 @@ looking_glass:
 probe_interval_ms: 5000
 probe_timeout_ms: 3000
 
+bird_replicas: 2
```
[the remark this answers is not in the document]
swagner-de: 2 replicas means 2 route-servers, in our old lingo `instances`. If we do more replicas customers have to span more BGP sessions. I didn't check, but for v4 we might even run into IP collisions.

px/lg/templates/configmap.yaml
swagner-de: This is just the service that is serving unencrypted. The ingress does https.

px/bird/values.yaml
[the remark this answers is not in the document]
nikatza: I see. The naming is rather confusing for the new eyes. Thanks for clarifying.

px/lg/templates/configmap.yaml
nikatza: Yeap, i am aware. Just wondering if it goes against any policy that enforces any traffic (even inside) to be encrypted (it really doesn't matter much since this is the LG anyway). Just asking :)

swagner-de: For pod to pod encryption we would usually use linkerd if credentials are transmitted/requried. I am unaware of any policy from ingress to service head end.

nikatza: Ack, just making sure. Thanks!

rhalina (approved):

PR #7227, opened 2024-10-11
greenhouse added; alerting refined

viennaa:
* added more precise alert with warning severity for non reachable store

* adding generic alert for non queryable Greenhouse Thanos


richardtief (approved):
Thanks! Ship it.

PR #7231, opened 2024-10-11
[mariadb] remove user and password from readyness and lifeness probes

mutax:
With the credentials for the root user present in /root/.my.cnf we do not need to give them as parameters for the probes, avoiding problems with shell escaping and showing passwords in the processlist of the k8s nodes.

### DIFF
--- a/common/mariadb/templates/deployment.yaml
+++ b/common/mariadb/templates/deployment.yaml
@@ -60,7 +60,7 @@ spec:
         lifecycle:
           postStart:
             exec:
-              command: ["sh", "-c", "while ! mysqladmin ping -uroot -p$MYSQL_ROOT_PASSWORD --silent; do sleep 1; done; mysql_upgrade"]
+              command: ["sh", "-c", "while ! mysqladmin ping --silent; do sleep 1; done; mysql_upgrade"]
         env:
         - name: MYSQL_ROOT_PASSWORD
           valueFrom:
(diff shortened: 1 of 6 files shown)


businessbean:
What about just using the `ccroot` user for these kind of internal connections? https://github.com/sapcc/helm-charts/pull/6297

mutax:
> What about just using the `ccroot` user for these kind of internal connections? #6297

The probes will use whatever credentials are set in the root users config file, so if we add the ccroot user there instead of root that would work as well. In any case the password should not be present in any commandline/process list. So when changing to a different user we should come up with another solution anyway.

s10 (approved):

galkindmitrii (approved):

common/mariadb/templates/deployment.yaml
```diff
@@ -60,7 +60,7 @@ spec:
         lifecycle:
           postStart:
             exec:
-              command: ["sh", "-c", "while ! mysqladmin ping -uroot -p$MYSQL_ROOT_PASSWORD --silent; do sleep 1; done; mysql_upgrade"]
+              command: ["sh", "-c", "while ! mysqladmin ping --silent; do sleep 1; done; mysql_upgrade"]
```
s10, line 63: mysql_upgrade still fails for passwords with `\` and `'`

PR #7232, opened 2024-10-11
Neutron resolve secrets

sven-rosenzweig:


openstack/neutron/templates/etc/_ssh_host_id_ec.tpl
```diff
@@ -1 +1 @@
-{{ .Values.sftp.server_key | required "Please set .Values.sftp.server_key" }}
+{{ .Values.sftp.server_key | required "Please set .Values.sftp.server_key" | include "resolve_secret" }}
```
fwiesel, line 1: That shouldn't be needed, as the template is simply used in `openstack/neutron/templates/sftp-secrets.yaml` directly,
and the secret-injector should resolve those values automatically.

mutax:
I think there are still some missing -- e.g.  .Values.interconnection.password  in seed.yaml

openstack/neutron/templates/secret-neutron-swift-injector.yaml
```diff
@@ -8,4 +8,4 @@ metadata:
 type: Opaque
 data:
   OS_USERNAME: {{ $.Values.global.neutron_service_user | default "neutron" | replace "$" "$$" | b64enc }}
-  OS_PASSWORD: {{ $.Values.global.neutron_service_password | default "" | replace "$" "$$" | b64enc }}
+  OS_PASSWORD: {{ $.Values.global.neutron_service_password | default "" | replace "$" "$$" | include "resolve_secret" | b64enc }}
```
fwiesel, line 11: I think, that won't work (as intended).
` replace "$" "$$" ` should escape the `$` in the password, now it escapes a `$` in the secrets injector url.
AFAIK, there is no way to escape the password anymore, so you might as well replace it with
`{{ $.Values.global.neutron_service_password }}`.
Our passwords should not contain any `$` anymore.

openstack/neutron/templates/secret-nsxv3-logstash.yaml
```diff
@@ -7,5 +7,5 @@ metadata:
     application: {{ .Release.Name }}
 type: Opaque
 data:
-  NEUTRON_DB_PASSWORD: {{ coalesce .Values.dbPassword .Values.global.dbPassword .Values.mariadb.root_password | required ".Values.mariadb.root_password is required!" | b64enc }}
+  NEUTRON_DB_PASSWORD: {{ coalesce .Values.dbPassword .Values.global.dbPassword .Values.mariadb.root_password | required ".Values.mariadb.root_password is required!" | include "resolve_secret" | b64enc }}
```
fwiesel, line 10: No need to change, it should automatically be resolved as it is a plain secret value.

openstack/neutron/templates/sftp-secrets.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
     release: "{{.Release.Name}}"
 type: Opaque
 data:
-  password: {{ required ".Values.sftp.password missing" .Values.sftp.password | b64enc | quote }}
+  password: {{ required ".Values.sftp.password missing" .Values.sftp.password  | include "resolve_secret" | b64enc | quote }}
```
fwiesel, line 9: Dito, no need to change.

fwiesel (approved):
I "should work", but I would drop the `resolve` from the plain secret values,
as well the `replace "$" "$$"` as they would not work anymore.

openstack/neutron/templates/etc/_ssh_host_id_ec.tpl
mutax: Hmm, true, but it also doesn't hurt?
I would prefer to rather have a secret get piped through the injector twice but have all templates look consistent, which helps in reviewing and working with them, than miss one.
Also when later someone decides to include that template somewhere else in a different way, it will not get resolved.

openstack/neutron/templates/secret-neutron-swift-injector.yaml
fwiesel: "After" would mean a go-template string for the secret-injector to evaluate like the [urlquery here](https://github.com/sapcc/helm-charts/blob/9c42bd414f0b3524514cd5b0317d02daf5527038/openstack/utils/templates/_hosts.tpl#L13)

mutax (approved):

openstack/neutron/templates/sftp-secrets.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
     release: "{{.Release.Name}}"
 type: Opaque
 data:
-  password: {{ required ".Values.sftp.password missing" .Values.sftp.password | b64enc | quote }}
+  password: {{ required ".Values.sftp.password missing" .Values.sftp.password  | b64enc | quote }}
```
fwiesel, line 9: That change could go.

fwiesel (approved):

mutax (approved):

openstack/neutron/templates/etc/_ssh_host_id_ec.tpl
fwiesel: No, it doesn't hurt. Matter of taste.
I would prefer to make clear that actually there is no go template processing of the output happening there, as the string is replaced on server-side.

openstack/neutron/templates/secret-neutron-swift-injector.yaml
fwiesel: Not quite:
`include "resolve_secret | replace "$" "$$"` is the same as `replace "$" "$$" | include "resolve_secret"`
Both get evaluated on the client-side before there is any secret involved, which may contain a `$`.

That was the motivation behind the macro, which could be changed like this

```
{{- define "_resolve_secret" -}}
    {{- $str := index . 0 -}}
    {{- $add_urlquery := index . 1 -}}
    {{- $escape_dollar := index . 2 -}}
    {{- if (hasPrefix "vault+kvv2" $str) -}}
        {{"{{"}} resolve "{{ $str }}" {{ if $add_urlquery }}| urlquery {{ end }}{{"}}"}}  {{ if $add_urlquery }}| urlquery {{ end }}{{ if $escape_dollar }} | replace "$" "$$" {{ end }} {{"}}"}}
    {{- else -}}
        {{ $str }}
{{- end -}}
{{- end -}}
```

But now that I read the macro, the else case doesn't do any escaping, which is maybe a bug.

rhalina (approved):

PR #7233, opened 2024-10-14
adding safeguard alert

viennaa:
* there was no critical alerting when vROps is gone entirely
* playbook still needs to be added here before merging

### The last part is crucial, please add the playbook before merging!

### DIFF
--- a/prometheus-rules/prometheus-vmware-rules/Chart.yaml
+++ b/prometheus-rules/prometheus-vmware-rules/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A collection of Prometheus alert rules.
 name: prometheus-vmware-rules
-version: 1.0.7
+version: 1.0.8
 dependencies:
   - name: owner-info
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
--- a/prometheus-rules/prometheus-vmware-rules/alerts/vrops.alerts
+++ b/prometheus-rules/prometheus-vmware-rules/alerts/vrops.alerts
@@ -22,6 +22,26 @@ groups:
         which indicates that vrops is reporting internal server errors or is unreachable. 
         Check if vrops is running and healthy.
 
+  - alert: VropsAPIDownEntirely
+    expr: |
+      abesent(vrops_api_response)
+    for: 10m
+    labels:
+      severity: critical
+      tier: vmware
+      service: compute
+      support_group: compute
+      context: vrops
+      playbook: docs/devops/alert/vcenter/vrops#vrops_api_down_entirely
+      dashboard: vrops-exporter-status
+      meta: "Vrops API is down. All collectors are not reporting anymore."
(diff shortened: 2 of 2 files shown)


richardtief (commented):
I would vote for adding the absent condition to the alert above. There is a VropsAPIDown alert already.

viennaa:
Initially I thought the same, however I would not like to add it to the existing alert anymore, since it has `$labels.target` used in meta and summary, which won't be filled in case of the absent condition hitting.

viennaa:
Another thought: Sunny wanted this as a warning. So we can give it the same name with warning, different messages and would be good to go.

himanip94:
I am not sure why are we keeping the alert as a warning if all collectors are down. Alternatively, we could change the previous alert to warning and set the new one to critical, using the same alert name?

viennaa:
To me both of them are worth critical. The recent case was without noticing since October, 9th.
I would generally tend to increase the for condition to 2h here, so it is a rather slow firing alert.

Sunny's argument was, that this can easily fire during maintenance activities. If these activities would take more than 2h, it would be wise, to silence the alert beforehand. I can give a quick guidance how to do this preventatively.

himanip94:
I agree to silencing alerts before maintenance activities, that way can keep both as critical.

sunnyrarora:
Silencing is one way, but here i believe we are increasing every next alert to critical in that sense, it will delete purpose of monitoring. InfraOps onduty primary task is to handle MIMs, i dont think it would be wise to involve him/her on troubelshooting monitoring tasks. here secondary can assist better. thats my two cents, you are free to keep it primary, but again one needs to sure if they are ready to handle it.

viennaa:
Just to keep in mind: there will be no critical alerts fired from affected vc if a node is down or anything else happens. Technically nothing in production is affected at this time, but if something would be affected from then on, no alerting takes place.

Maybe you can take this into your meeting and come back with a decision here.

viennaa:
any decision?

himanip94:
Will discuss this in Tuesday monitoring call

himanip94:
@sunnyrarora, lets keep both API down alerts in warning with 10 minutes threshold, raise both to critical after 90 minutes. This gives Secondary InfraOps enough time to resolve this alert and if missed, Primary InfraOps will be paged.

viennaa:
Feel free to push the changes to the branch.

himanip94 (changes requested):
As discussed, please keep both API down alerts in warning channel with 10 minutes threshold and raise to critical after 90 minutes.

sunnyrarora:
Hi @viennaa I am trying to push the changes, hitting permission issue. if you can please assign me permissions.

sunnyrarora:
I have push the changes, Please review. thanks !

viennaa:
Looks good to me, but can not approve my own PR :D

richardtief (approved):
Thanks!

PR #7238, opened 2024-10-14
[utils][proxysql] Set query_retries_on_failure to 1 explicitly

s10:
Default value is 1, but sometimes we might want more retries instead of 500

This could be done for SELECT queries by configuration:
```
proxysql:
  query_rules:
    retries_on_select: 3
```

### DIFF
--- a/openstack/utils/templates/snippets/_proxysql.cfg.tpl
+++ b/openstack/utils/templates/snippets/_proxysql.cfg.tpl
@@ -42,6 +42,7 @@ mysql_variables =
     connect_retries_on_failure = {{ default 1000 .global.Values.proxysql.connect_retries_on_failure }}
     connect_retries_delay = {{ default 100 .global.Values.proxysql.connect_retries_delay }} {{- /* The default is 1ms, and that means we will run through the retries on failure in no time */}}
     connect_timeout_server_max = {{ default 100000 .global.Values.proxysql.connect_timeout_server_max }}
+    query_retries_on_failure = {{ default 1 .global.Values.proxysql.query_retries_on_failure }} {{- /* The ProxySQL default is 1, and that means queries are being retried once */}}
 }
 
 mysql_servers =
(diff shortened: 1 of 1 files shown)


fwiesel:
I am not sure, if we can enable that feature without risk.
The scenario I am thinking of is that the (modifying) query actually successfully executed, it just failed to send the information back to the client. (In terms of the two-phase commit protocol, we had an error on the "ack").
My understanding is, that the query will be re-executed, and if the constraints are properly set, we will "simply" have different error (a constraint violation vs connection reset error), but in the worst case, we have now a record twice, which we shouldn't have.

fwiesel (approved):

openstack/utils/templates/snippets/_proxysql.cfg.tpl
```diff
@@ -81,5 +82,14 @@ mysql_query_rules =
         cache_ttl = 3600000,
         match_pattern = "^SELECT 1$",
     },
+{{- if and .global.Values.proxysql.query_rules .global.Values.proxysql.query_rules.retries_on_select }}
+    {
+        rule_id = 1,
+        active = 1,
+        apply = 1,
+        retries = {{ default 10 .global.Values.proxysql.query_rules.retries_on_select }},
+        match_digest = "^SELECT",
```
businessbean, line 91: With [SELECT FOR UPDATE](https://dev.mysql.com/doc/refman/8.4/en/select.html) this can be a DML command and than the retries can be risky if something has been changed in between, or?

fwiesel: My assumption was, that retries are only done, if it isn't within a transaction. `SELECT FOR UPDATE` only creates a lock within a transaction.
I've checked the code, and it looks to me, my assumption is [correct](https://github.com/sysown/proxysql/blob/5cc418bb6c59aefca2bc21232e21de6f240a49cf/lib/MySQL_Session.cpp#L4667C56-L4667C75).

A retry of a single statement within an aborted transaction wouldn't make sense.

s10: Yes, looks like it might retry such SELECT FOR UPDATE requests, if they weren't running in the transaction, see  https://github.com/sysown/proxysql/blob/v2.5.5-branch/lib/MySQL_Session.cpp#L4234

PR #7247, opened 2024-10-15
[memcached] version bumped to 1.6.31-alpine3.20

businessbean:
- Memcached [1.6.31 Release Notes](https://github.com/memcached/memcached/wiki/ReleaseNotes1631)
- added alpine version tag to enforce exactly this version
- version info added to labels
  - to allow gatekeeper rules based on them
- old (non-standard) labels removed
- chart version bumped

## label example
```yaml
labels:
  app.kubernetes.io/name: memcached
  app.kubernetes.io/instance: keystone-memcached
  app.kubernetes.io/component: memcached-deployment-kvstore
  app.kubernetes.io/part-of: keystone
  app.kubernetes.io/version: 1.6.31
  app.kubernetes.io/managed-by: "helm"
  helm.sh/chart: memcached-0.6.0
```

## removed labels
```yaml
labels:
   app: keystone-memcached
   component: memcached
   chart: "memcached-0.5.3"
   release: "keystone"
   heritage: "Helm"
   type: configuration
   application: keystone
```

### DIFF
--- a/common/memcached/CHANGELOG.md
+++ b/common/memcached/CHANGELOG.md
@@ -1,0 +1,36 @@
+# Changelog
+
+## v0.6.0 - 2024/10/15
+* version info added to labels
+  * to allow gatekeeper rules based on them
+* old (non-standard) labels removed
+* memcached version bumped to `1.6.31-alpine3.20`
+  * added alpine version to the tag to enforce exactly this version
+* chart version bumped
+
+### label example
+```yaml
+labels:
+  app.kubernetes.io/name: memcached
+  app.kubernetes.io/instance: keystone-memcached
+  app.kubernetes.io/component: memcached-deployment-kvstore
+  app.kubernetes.io/part-of: keystone
+  app.kubernetes.io/version: 1.6.31
+  app.kubernetes.io/managed-by: "helm"
+  helm.sh/chart: memcached-0.6.0
+```
+### removed labels
+```yaml
+labels:
+   app: keystone-memcached
+   component: memcached
+   chart: "memcached-0.5.3"
+   release: "keystone"
+   heritage: "Helm"
+   type: configuration
+   application: keystone
+```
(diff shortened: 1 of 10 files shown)


common/memcached/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+## v0.6.0 - 2024/10/15
+* version info added to labels
+  * to allow gatekeeper rules based on them
+* old (non-standard) labels removed
```
s10, line 6: These non-standard labels are used extensively in all helm charts (just search "app: "), so I'm not sure that their removal in shared helm charts would be beneficial and wouldn't introduce too much confusion.

Another problem is that these labels are being used in alerts not only in these dependency charts but also in parent charts, see, e.g., OpenstackManilaMariadbRowLocks or OpenstackKeystoneDatabaseScrapeMissing.

businessbean: Yes, this change needs some clean ups. We have introduced these standard labels a while ago and with this release the phase of this standardization will be done.

s10 (approved):
lgtm

Such custom labels as `system` and `type` are present everywhere, but aren't being used anywhere.
`application` and `component` label is being used in `helm-toolkit.snippets.kubernetes_metadata_labels` helper, but has no meaning here.

common/memcached/templates/service.yaml
```diff
@@ -3,12 +3,7 @@ kind: Service
 metadata:
   name: {{ template "fullname" . }}
   labels:
-    app: {{ template "fullname" . }}
```
s10, line 6: Below the necessary change to service selector is missing, see
```
spec:
  selector:
```

s10 (changes requested):

s10 (approved):

dhoeller (approved):
LGTM

PR #7248, opened 2024-10-15
 [rabbitmq] version bumped to 3.13.7-management

businessbean:
- RabbitMQ [3.13.7 Release Notes](https://github.com/rabbitmq/rabbitmq-server/releases/tag/v3.13.7)
- version info added to labels
  - to allow gatekeeper rules based on them
- old (non-standard) labels removed
- chart version bumped

## label example
```yaml
labels:
  app.kubernetes.io/name: rabbitmq
  app.kubernetes.io/instance: keystone-rabbitmq
  app.kubernetes.io/component: rabbitmq-prometheusrule-alert
  app.kubernetes.io/part-of: keystone
  app.kubernetes.io/version: 3.13.7
  helm.sh/chart: rabbitmq-0.12.0
```


s10 (approved):
lgtm

Such custom labels as `system` and `type` are present everywhere, but aren't being used anywhere.

common/rabbitmq/templates/service.yaml
```diff
@@ -3,14 +3,7 @@ kind: Service
 metadata:
   name: {{ template "fullname" . }}
   labels:
-    app: {{ template "fullname" . }}
```
s10, line 6: Below the necessary change to service selector is missing, see
```
spec:
  selector:
```

s10 (changes requested):

common/rabbitmq/templates/service.yaml
businessbean: thanks for the hint

notque:
The volumeClaimTemplates in the statefulset.yaml is still using the old labels.

businessbean:
> The volumeClaimTemplates in the statefulset.yaml is still using the old labels.

Thanks for the hint. https://github.com/sapcc/helm-charts/pull/7248/commits/c307f4abb7eb89cd13a1c818d2eaf93fdb33674f fixes that

s10 (approved):

PR #7250, opened 2024-10-16
[Utils] Append urlencode in resolve_secret_urlencode

sven-rosenzweig:
Using an old vault reference '*vault(path:..)' as input for resolve_secret_urlencode does not append an urlencode as the function name might indicate.

During the migration from old vault-injector to the new secret-injector, we observed this problem with some Neutron passwords. Removing the urlencode from our Chart (replacing it with a 'resolve_secret_urlencode') lead to an unencoded password. Simply appending an 'urlencode' back in our Chart does not work either if we migrate to the new vault reference would
urlencode the string '{{ resolve vault+kvv2:///some/path }}'. Assuming the secret-injector does not pick this up.

### DIFF
--- a/openstack/utils/templates/_hosts.tpl
+++ b/openstack/utils/templates/_hosts.tpl
@@ -11,9 +11,11 @@
     {{- $add_urlquery := index . 1 -}}
     {{- if (hasPrefix "vault+kvv2" $str) -}}
         {{"{{"}} resolve "{{ $str }}" {{ if $add_urlquery }}| urlquery {{ end }}{{"}}"}}
-    {{- else -}}
-        {{ $str }}
-{{- end -}}
+    {{- else if $add_urlquery }}
+        {{- $str | urlquery }}
+    {{- else }}
+        {{- $str  }}
+    {{- end }}
 {{- end -}}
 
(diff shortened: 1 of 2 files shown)


openstack/utils/templates/_hosts.tpl
```diff
@@ -12,7 +12,7 @@
     {{- if (hasPrefix "vault+kvv2" $str) -}}
         {{"{{"}} resolve "{{ $str }}" {{ if $add_urlquery }}| urlquery {{ end }}{{"}}"}}
     {{- else -}}
-        {{ $str }}
+        {{ {{ $str }} {{ if $add_urlquery }}| urlquery {{ end }} }}
```
fwiesel, line 15: I think you can do
``{{ {{ pipeline }} }}``

I'd suggest the following:
```
{{- else if $add_urlquery }}
        {{ $str | urlquery }} 
{{ - else }}
        {{ $str  }} 
{{ - end }}
```

joker-at-work (approved):

fwiesel (approved):

galkindmitrii (approved):

PR #7258, opened 2024-10-16
feat(greenhouse): Add pod/exec teamrole

uwe-mayer:
This adds a CCloud specific `pod/exec` `TeamRole` to the ccloud org in Greenhouse as this role is not part of the open source default roles.


auhlig (approved):
Since this is a subset of permissions for pods, it might be good to have
```
- apiGroups: [""]
   resources: [pods]
   verbs: ["get", "list", "watch"]
``` 

I'm aware, this might be part of the developer role to which this one is aggregated to.
However, a self-contained set of permissions would make the pod exec role more versatile in its usage.

I'm aware this might be part of the role i

IvoGoman (approved):

PR #7262, opened 2024-10-16
[utils] Remove whitespace in _resolve_secrets

sven-rosenzweig:


joker-at-work:
Whitespace inside the `{{ }}` show up in the output - just not the last one before the `}}`?

fwiesel:
That shouldn't be the case.

PR #7274, opened 2024-10-17
nova/flavors: Add KVM v1 (Cascade-Lake) flavors

grandchild:
As agreed in ADR core/nova/0008 "KVM flavor naming scheme" the KVM flavors (and eventually, but not in this change, the VMware flavors too) gain a new hypervisor field after the g/c/m purpose field. It's "k" for KVM/Qemu (and will be "v" for VMware).

Also the flavors will be suffixed with a "v1" hardware generation number for Cascade Lake right away.

So, the KVM counterpart for "g_c4_m16" is named "g_k_c4_m16_v1".

A per-region config flag `kvm_enabled` is introduced, defaulting to false, to allow per-region enablement of KVM flavors.

---

Note that there are a few flavors still commented out, those are legacy flavors with the AWS naming schema that are still in use, and we should decide whether we still want some of these or close variants thereof.

### DIFF
--- a/openstack/sap-seeds/templates/_flavors_hana_exclusive.tpl
+++ b/openstack/sap-seeds/templates/_flavors_hana_exclusive.tpl
@@ -225,3 +225,111 @@
     "quota:instance_only": "true"
     "quota:separate": "true"
     {{- end }}
+
+# KVM HANA Flavors
+{{- if .Values.kvm_enabled }}
+- name: "hana_c30_m240_k1"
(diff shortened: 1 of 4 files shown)


grandchild:
Instead of the `kvm_enabled` config flag, we could also check for the presence of certain traits, like we check for any required CUSTOM_NUMA_ traits to decide whether to seed a HANA flavor. But the simple per-region flag seems less overhead for now.

PR #7290, opened 2024-10-22
[pxc-db] Change default timeouts

s10:
* Increase default net_read_timeout and net_write_timeout

net_write_timeout limits the max size of binlog sent to mysqlbinlog

180s in my tests were enough for 1GB binlog file

* Set connect_timeout to 30s

* Set wait_timeout and interactive_timeout to the same values as in MariaDB


galkindmitrii (approved):
Should default `wait_timeout` be adjusted too?

PR #7304, opened 2024-10-23
[manila] set reserve down to 25%

Carthaca:
storage team is rebalancing on 70% usage, so we have a 5% buffer
for new share creations, e.g. to fulfil anti-affinity constraints.

Without the buffer we would get too much 'no valid host' situations.


kpawar-sap (approved):

chuan137 (approved):
lgtm

PR #7305, opened 2024-10-23
chore(violation.alert): encode support_group label with ~S

ArtieReus:
### Description

This commit modifies the Helm chart annotations to include the `label_replace()` function for encoding the `support_group` label with the `~S`. The change impacts the annotations for the Greenhouse dashboard view link, ensuring that the `support_group` value follows the required encoding format.

#### Key Changes:
- Used `label_replace()` to append `~S` to the `support_group` label in the description annotation.
- Improved the clarity of variable usage in the Greenhouse dashboard link by encoding support group labels consistently across deployments.

### DIFF
--- a/system/doop-central/alerts/infra-frontend/violation.alerts
+++ b/system/doop-central/alerts/infra-frontend/violation.alerts
@@ -23,7 +23,7 @@ groups:
         annotations:
           summary: "Significant vulnerable images used in prod deployments"
           description: |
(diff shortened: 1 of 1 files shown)


majewsky (changes requested):
label_replace() is a Prometheus function. It needs to be used in the Prometheus expression a few lines above in `expr: ...`.

The interpolation here, as performed by the alertmanager, does not support any templating functions beyond the bare minimum syntax, esp. not any string replacement functions.

PR #7315, opened 2024-10-25
compactor pvc added

viennaa:
eu-de-2 compactor started to crashloop due to 25GiB ephemeral not being enough storage.

Tackling this globally.


richardtief (approved):
Looks good.

PR #7318, opened 2024-10-28
[barbican] whitelist ceph user from rate-limiting

rajivmucheli:


openstack/barbican/templates/etc/_ratelimit.yaml.tpl
```diff
@@ -5,6 +5,7 @@ whitelist:
 
 whitelist_users:
   - TS4_S4_SMTP_01
+  - ceph
```
kayrus, line 8: it should be ceph-barbican: https://github.com/sapcc/helm-charts/blob/master/openstack/ceph-seed/templates/seed.yaml#L32
```suggestion
  - ceph-barbican
```

rajivmucheli: ah ok, presumed this first but wasnt sure.

kayrus (approved):

stefanhipfel (approved):

PR #7323, opened 2024-10-28
common/redis: switch to valkey

SuperSandro2000:


common/redis/Chart.yaml
```diff
@@ -13,4 +13,10 @@ name: redis
 # from 1.4.x
 # - the deployment will be rolled out (restarted) on password and configuration
 #   Secrets changes.
-version: 1.5.3 # this version number is SemVer as it gets used to auto bump
+#
+# from 1.5.x
+# - the reployment now rolls out on configuration changes and secrets changes
+#
+# from 1.6.x
+# - switched from redis to valkey
```
majewsky, line 21: You parsed the list semantics incorrectly. We are upgrading to 1.6.x, not from 1.6.x. Also, I suggest clarifying what the switch entails.
```suggestion
# from 1.5.x
# - Switched from Redis to its fork Valkey because of the Redis license change.
#   The image now comes from ccloud/shared-app-images/valkey.
#   If you have image.tag pins (which we recommend against), please update them accordingly.
```

majewsky (changes requested):

majewsky (approved):

PR #7332, opened 2024-10-30
cronjob-monthly-ticket-analysis.yaml

sunitha-reddy:
Add CronJob template for monthly ticket analysis repor


system/jupyter-cron-reports/templates/cornjob-monthly-ticket-analysis-report.yaml
```diff
@@ -0,0 +1,71 @@
...
+              valueFrom:
+                secretKeyRef:
+                  name: jupyter-cron-reports-secrets
+                  key: swift_url
+            - name: swift_password
+              valueFrom:
+                secretKeyRef:
+                  name: jupyter-cron-reports-secrets
+                  key: swift_password
+            - name: vault_token
```
timojohlo, line 40: Here you are referencing a token that should be stored in the regional secrets file. `vault_token` is not a valid reference. Also if you want to run the notebook with a cronjob, I would recommend using the approle which is already part of the regional secrets file.

system/jupyter-cron-reports/templates/cornjob-monthly-ticket-analysis-report.yaml
```diff
@@ -0,0 +1,71 @@
...
+            {{- if not (and $.Values.global.linkerd_enabled $.Values.global.linkerd_requested) }}
+            - name: LINKERD_AWAIT_DISABLED
+              value: "Linkerd was not enabled or not requested"
+            {{- end }}
+            - name: swift_url
+              valueFrom:
+                secretKeyRef:
+                  name: jupyter-cron-reports-secrets
+                  key: swift_url
+            - name: swift_password
```
timojohlo, line 35: Here you are referencing `swift_password` that should be stored in the regional secrets file if you want to use it that way. You could also retrieve it from vault within the notebook like [this](https://github.wdf.sap.corp/cc/jupyterhub-dev/blob/72724a71985f2db02c53a5eb7dc61188b7d30196/cronjobs/api-availability.ipynb).

timojohlo (changes requested):
Hi,
Overall this looks good. I have added some comments below about the secrets and I have send you a PM regarding the .ipynb file. Feel free to contact me if you have any questions.

system/jupyter-cron-reports/templates/cornjob-monthly-ticket-analysis-report.yaml
sunitha-reddy: Thank you. will update this

PR #7340, opened 2024-11-04
[rabbitmq] Add config.linkerd.io/opaque-ports annotation

s10:
Add config.linkerd.io/opaque-ports annotation to avoid protocol detection delay, 10s on every connection to RabbitMQ

https://linkerd.io/2.16/features/protocol-detection/


fwiesel (approved):

businessbean (approved):
lgtm. As described in the [docs](https://linkerd.io/2.16/features/protocol-detection/#marking-ports-as-opaque) we have to set it on the service and the workloads used as endpoint for the service...

PR #7342, opened 2024-11-04
[keystone] Disable apache2 server-status endpoint

s10:
Disables the server status page of Apache.
This page contains internal information about working process.

### DIFF
--- a/openstack/keystone/Chart.yaml
+++ b/openstack/keystone/Chart.yaml
@@ -9,7 +9,7 @@ maintainers:
 name: keystone
 sources:
   - https://github.com/sapcc/keystone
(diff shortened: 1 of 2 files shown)


fwiesel (approved):

rhalina (approved):

bbobrov (changes requested):
Please describe why the change is needed, preferably in the commit message

bbobrov:
> This page contains internal information about working process.

Could you please describe, how could i access this page from my work computer? With curl or with a browser, for example. I am trying to understand the impact of the issue and why we need to act at all.

bbobrov (approved):
thanks

PR #7361, opened 2024-11-06
[designate] increase number of rpc central workers and other fixes

s10:
* Increase default number of rpc workers per pod from 2 to 4
* Remove unused rpc_workers option
* Enable oslo_rpc_server_ping rpc endpoint
* Increase rabbitmq heartbeat interval from 15 to 30 seconds
* Remove deprecated min_pool_size variable, it's not used since Rocky
* Move max_pool_size and max_overflow option to effective section
* Lower db_max_retry_interval from 10 to 1 second
* Set heartbeat_emitter interval, increase it from 10 to 30 seconds


galkindmitrii (approved):
lgtm, discussed changes in chat

PR #7372, opened 2024-11-07
fix(lyra-ingress): reference a custom secret within the ingress

ArtieReus:


openstack/lyra/templates/secret-tls.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Release.Name }}-tls
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+type: Opaque
```
IvoGoman, line 7:
```suggestion
type: kubernetes.io/tls
```

PR #7381, opened 2024-11-08
castellum, keppel, limes: use policy.json instead of policy.yaml

majewsky:
This does not entirely remove the YAML parser library from these applications yet, but it's a prerequisite to getting there and reducing our dependency tree a little bit.

This change is immediately deployable because JSON is a subset of YAML to a sufficient extent.


openstack/limes/files/policy.yaml
SuperSandro2000, line 1: Why not save the file as json here, too?

majewsky: Looks like a copy-paste error. Fixed.

SuperSandro2000 (approved):

PR #7382, opened 2024-11-08
replace with global logger.

notque:
testing on branch first in qa. slightly difficult because we're catching malformed events, and it triggers when we hit one, but hopefully it being correct as it's global, and checking everything, will solve the issue regardless.


notque:
works, looks good. I have other regions down at the moment, so I'm going to move fast through this, and will expect to be available to deal with errors as oncall for this issue since it's critical, and i'm pushing on friday.

PR #7386, opened 2024-11-11
[px] feat: spread pods in multiple AZs

nikatza:
* Currently we have a set of nodeAffinity to pin PX pods to apods that are cabled for PX. We have podAntiAffinity to make sure that a PX service does not run with a pod of the same service on the same phyiscal node. Another podAntiAffinity makes sure that an PX instance (PX has 2 instances per domain and service) of the same PX service never runs in the same AZ. This leads to the each PX Domain being spread over at least 2 AZs. 
  See original issue: https://github.wdf.sap.corp/cc/px/issues/181
* Add topology spread contraints such that pods are spread over more than 2 AZs, considering each time only the pods  that belong to the same service and AFI. 
  See: https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints

### DIFF
--- a/px/bird/templates/_deployment-bird.tpl
+++ b/px/bird/templates/_deployment-bird.tpl
@@ -24,6 +24,7 @@ spec:
       annotations:
         k8s.v1.cni.cncf.io/networks: '[{ "name": "vlan{{ .domain_config.multus_vlan }}", "interface": "vlan{{ .domain_config.multus_vlan }}"}]'
     spec:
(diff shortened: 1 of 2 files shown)


swagner-de:
I am merging this into my branch that will undergo 4 eye review anyway

swagner-de:
Implements https://github.wdf.sap.corp/cc/px/issues/181

PR #7390, opened 2024-11-12
Allow max quota editing for project level

VoigtS:
This PR aims to publish max quota setting to project level.


VoigtS:
As discussed with Stefan M. the project max quota should be its own managed field in limes that does not influence the domain quota set.

PR #7401, opened 2024-11-14
[bind] introducing secret_env

vssldmtrv:
Some of our bind images are started using env vars.
Some of these are secret.
We now have a dedicated secret to pull these from.

### DIFF
--- a/system/bind/templates/deployment.yaml
+++ b/system/bind/templates/deployment.yaml
@@ -86,6 +86,11 @@ spec:
         {{ if .Values.extra_env }}
 {{ toYaml .Values.extra_env | indent 8 }}
         {{ end }}
+
+        envFrom:
+          - secretRef:
+            name: {{ .Release.Name }}-secret-env
+
         volumeMounts:
           - name: persistent-storage
             mountPath: {{ .Values.storage_mountpoint | default "/var/lib/bind" }}
--- a/system/bind/templates/etc/_secret_env.tpl
+++ b/system/bind/templates/etc/_secret_env.tpl
@@ -1,0 +1,7 @@
+{{- if .Values.secret_env }}
+{{- range .Values.secret_env }}
+{{ .name }}: {{ .value | include "resolve_secret" }}
+{{- end }}
+{{- else }}
+{}
+{{- end }}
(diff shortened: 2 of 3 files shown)


system/bind/templates/deployment.yaml
```diff
@@ -86,6 +86,11 @@ spec:
         {{ if .Values.extra_env }}
 {{ toYaml .Values.extra_env | indent 8 }}
```
galkindmitrii, line 87: I think this part has to go.

system/bind/templates/deployment.yaml
```diff
@@ -86,6 +86,11 @@ spec:
...
         {{ end }}
```
galkindmitrii, line 88: and this `end` move down under `envFrom` section.

system/bind/templates/etc/_secret_env.tpl
```diff
@@ -0,0 +1,7 @@
+{{- if .Values.secret_env }}
+{{- range .Values.secret_env }}
+{{ .name }}: {{ .value | include "resolve_secret" }}
+{{- end }}
+{{- else }}
+{}
```
galkindmitrii, line 6: what's the point of this `else` condition?

system/bind/templates/etc/_secret_env.tpl
```diff
@@ -0,0 +1,7 @@
+{{- if .Values.secret_env }}
+{{- range .Values.secret_env }}
```
galkindmitrii, line 2: I think there is an extra `env` key in the dictionary.

system/bind/templates/deployment.yaml
vssldmtrv: ```envFrom``` is unconditional. If there are no secret env vars then it will just stay empty.

system/bind/templates/etc/_secret_env.tpl
vssldmtrv: Not under the ```secret-env:``` key.

system/bind/templates/etc/_secret_env.tpl
vssldmtrv: Aesthetics, mostly. In case there are no secret env vars.
IOW avoid an empty, dangling ```data:``` vs the more explicit ```data: {}```.
Both are valid YAML.

galkindmitrii (approved):

PR #7404, opened 2024-11-14
[utils][proxysql] Set default transaction and query timeouts

s10:
Currently, default values are the following:
- transaction: 4h
- query: 24h

In very rare cases under high load caused by locust we've faced an issue when some sqlalchemy transactions acted as follows:
1 Began
2 Some blocking query was executed (they held a row lock)
3 No commit/rollback follows
4 Transaction continues to hang indefinitely in an idle state
5 All parallel requests modifying these rows were blocked because of (2)

This deadlock could be resolved by:
- killing this idle transaction on DB side manually
- killing application process
- killing transactions on the ProxySQL side

The proposed solution is to introduce query and transaction timeouts in the ProxySQL sidecar with the following values:
- transaction: 60s
- query: 90s

From the application perspective, after reaching these timeouts the error would be:
- transaction: `DBConnectionError: (pymysql.err.OperationalError) (2013, 'Lost connection to MySQL server during query')`
- query: `sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError) (1907, 'Query execution was interrupted, query_timeout exceeded')`

### DIFF
--- a/openstack/utils/Chart.yaml
+++ b/openstack/utils/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: utils
-version: 0.19.7
+version: 0.20.0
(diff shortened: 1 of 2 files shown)


fwiesel (approved):
I wouldn't mind even lower limits.

galkindmitrii:
> I wouldn't mind even lower limits.

Those are set to 60 and 90 seconds for the transaction and query after this change.

I think we need an explicit OK from Neutron folks.

galkindmitrii (approved):

Carthaca:
We sometimes have long running slow queries in manila. I'm indecisive wether I should keep this new default now and watch how the clients deal with it or set different thresholds.

PR #7411, opened 2024-11-18
[Glance] remove kubernetes_entrypoint from nanny

rajivmucheli:


openstack/glance/templates/glance-nanny-deployment.yaml
```diff
@@ -50,11 +50,13 @@ spec:
         image: {{ required ".Values.global.registry is missing" .Values.global.registry }}/loci-glance:{{.Values.imageVersionGlanceApi | default .Values.imageVersion | default .Values.image_version | required "Please set glance.imageVersion or similar"}}
         imagePullPolicy: IfNotPresent
         command:
-        - dumb-init
-        - kubernetes-entrypoint
-        env:
-        - name: COMMAND
-          value: "{{ if not .Values.glance_nanny.debug }}/bin/bash /scripts/glance-db-consistency-and-purge.sh{{ else }}sleep inf{{ end }}"
+{{- if not .Values.glance_nanny.debug }}
+        - /bin/bash
```
fwiesel, line 54: Don't we need `dumb-init` anymore? In the past, we needed it, because k8s would only terminate the `PID 1` and rely the "init" process to clean up, otherwise we end up with zombie processes.

An honest question, I do not know, if that is still the case.

fwiesel (approved):

stefanhipfel (approved):

PR #7412, opened 2024-11-18
[Glance] set ttlSecondsAfterFinished: 3600 for migration job

rajivmucheli:


fwiesel (changes requested):
We need the opposite: This is exactly what would break a restart of a pod.
The job would get cleaned up after 60 minutes, and if you restart then a pod, the `initContainer` [here](https://github.com/sapcc/helm-charts/blob/master/openstack/glance/templates/deployment.yaml#L47) would wait for that job, but it has been removed.

Leaving it unset would be better, unless we move the job to a hook, and remove the dependencies from the deployment.

fwiesel:
I am not sure what the cause of the failure was, because it was waiting for a job of a different release. The job of the expected release was still there.

rajivmucheli:
okay, i will raise a Glance Jira ticket for further investigation , closing this PR for now.

PR #7419, opened 2024-11-18
[pxc-db] priorityClassName updated to new naming convention

businessbean:
- default value added


s10 (approved):
lgtm.

The move from camel case to snake case for variable name is questionable, because in pxc-db chart everywhere the camel case is being used, except the compat options.

galkindmitrii (approved):

PR #7426, opened 2024-11-20
[pxc-db] Lower db net timeouts back to defaults

s10:


common/pxc-db/values.yaml
```diff
@@ -131,20 +131,22 @@ pxc:
       wsrep_retry_autocommit: 3
       binlog_format: ROW
       binlog_expire_logs_seconds: 345600  # default 30 days -> 4 days
-      net_read_timeout: 120
-      net_write_timeout: 300
+      sync_binlog: 0
```
businessbean, line 134: `sync_binlog=1` and `innodb_flush_log_at_trx_commit=1` provide the greatest level of data safety. Do we really want to lower that level for the Openstack services that heavily rely on the consistency of the configuration data even between the services? 
- [innodb_flush_log_at_trx_commit](https://mariadb.com/kb/en/innodb-system-variables/#innodb_flush_log_at_trx_commit)
- [sync_binlog](https://mariadb.com/kb/en/replication-and-binary-log-system-variables/#sync_binlog)

s10: This PR is about returning some options to default values or to values similar to the current MariaDB configuration, so we have a comparable performance-wise configuration.

sync_binlog is 0 in all current MariaDB instances https://mariadb.com/docs/server/ref/mdb/system-variables/sync_binlog/ , and they are not even clustered, so I am not really convinced about the data safety argument for this option at the expense of crippling the write performance.

The same applies to innodb_flush_log_at_trx_commit; its default value for non-galera setups is 1, but currently, we have an unsafe 0, and the recommended safe value in Galera setup is 2, not 0.

s10: @businessbean  , I changed both options to the safe one (1).

businessbean (approved):
lgtm, i have rebased it from the current master...

PR #7440, opened 2024-11-22
[liquid-ceph]: Add support for —allow-domain cli flag

databus23:


system/liquid-ceph/templates/deployment.yaml
```diff
@@ -44,6 +44,9 @@ spec:
           - --ceph-conf=/etc/ceph/ceph.conf
           - --target-zone={{ required ".Values.ceph.targetZone missing" .Values.ceph.targetZone }}
           - --valid-pools={{ required ".Values.ceph.validPools missing" .Values.ceph.validPools }}
+          {{- range .Values.ceph.allowedDomains }}
+          - --domain-allow={{ join "," . }}
+          {{- end }}
```
majewsky, line 49:
```suggestion
          {{- if .Values.ceph.allowedDomains }}
          - --domain-allow={{ join "," .Values.ceph.allowedDomains }}
          {{- end }}
```

databus23: Ah god catch. I did in wrong. I wanted to have the same flag multiple times and forgot to remove the `join` command

databus23: @majewsky I updated the PR, can you have another look

PR #7480, opened 2024-11-27
fix(gh-actions): fix chart-dirs in ct config to detect changes

kengou:
Remove the `.`  in [ci/config.yaml](https://github.com/sapcc/helm-charts/blob/master/ci/config.yaml) since it's not detecting correctly the chart changes. 
Directly referring  to the top level directories does the trick and detects the chart changes correctly. Tested locally gives the correct results

### DIFF
--- a/ci/config.yaml
+++ b/ci/config.yaml
@@ -10,7 +10,13 @@ check-version-increment: true
 
 # Consider helm charts in all directories.
 chart-dirs:
-  - .
+  - common
+  - global
+  - openstack
+  - prometheus-exporters
+  - prometheus-rules
+  - px
+  - system
 
 # Don't look here.
 excluded-charts:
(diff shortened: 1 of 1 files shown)


ci/config.yaml
```diff
@@ -8,9 +8,15 @@ validate-yaml: true
 validate-maintainers: false
 check-version-increment: true
 
-# Consider helm charts in all directories.
+# Consider helm charts in directories. '.' as chart-dir does not work.
 chart-dirs:
```
auhlig, line 12: Or maybe remove the explicit whitelist to have it run by default?

Exclusion (below) is always possible.

kengou: Tested locally, we need the explicit list sadly, because of our folder structure.
See issues:
https://github.com/helm/chart-testing/issues/633
https://github.com/helm/chart-testing/issues/226

auhlig (approved):

PR #7496, opened 2024-12-02
Elektra: more soft probes

hgw77:
This change should help prevent an Elektra pod from being restarted if we experience more than 10 concurrent requests that take longer than 5 seconds to process. By relaxing the readiness probe parameters, we allow for temporary spikes in response time without unnecessarily marking the pod as unready.

We can play also with the `timeoutSeconds` and `initialDelaySeconds`

related issue https://github.com/sapcc/elektra/issues/1466


ArtieReus (commented):
I'm not convinced about this change. Increasing the "killing period" from 10 seconds to 30 seconds doesn't seem to offer significant benefits. While it might address this specific case, it won't resolve the firing alarm event issue since the same number of requests will still pile up. Additionally, in other scenarios, it could prolong recovery time.

andypf (approved):

PR #7505, opened 2024-12-04
[utils][proxysql] Configure Galera monitor in ProxySQL sidecars

s10:
* Enables monitor if DB is pxc
* Adds all cluster members to `mysql_servers` configuration
* Configures `mysql_galera_hostgroups`
* Remove `multiDestination` option: the same could be done with `force_enable`

### DIFF
--- a/openstack/utils/templates/snippets/_proxysql.cfg.tpl
+++ b/openstack/utils/templates/snippets/_proxysql.cfg.tpl
@@ -15,11 +15,21 @@ mysql_variables =
     {{- $monitorEnabled := false }}
     {{- $monitorUsername := "" }}
     {{- $monitorPassword := "" }}
+    {{- $monitorGalera := false }}
     {{- /*
     Check that proxysql_monitor user exists in all DB instances
     Check that proxysql_monitor credentials are the same in all DB instances
     */}}
     {{- range $dbKey, $db := $dbs }}
+        {{- if hasKey $db "pxc" }}
+            {{- $monitorEnabled = true }}
+            {{- $monitorUsername = "monitor" }}
+            {{- $monitorPassword = $db.system_users.monitor.password }}
+            {{- $monitorGalera = true }}
+            {{- if (ne $monitorPassword $db.system_users.monitor.password) }}
+                {{- fail (printf "system_users.monitor.password needs to be the same in %s as in other databases" $dbKey) }}
+            {{- end }}
+        {{- end }}
         {{- if and $db.users $db.users.proxysql_monitor }}
             {{- if not $monitorEnabled }}
                 {{- $monitorEnabled = true }}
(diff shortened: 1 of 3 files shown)


galkindmitrii (approved):
cannot judge on the settings, such as `max_transactions_behind`, etc. but otherwise lgtm.

s10:
@fwiesel , could you please take a look?

openstack/utils/templates/snippets/_proxysql.cfg.tpl
```diff
@@ -15,11 +15,21 @@ mysql_variables =
...
     {{- $monitorPassword := "" }}
+    {{- $monitorGalera := false }}
     {{- /*
     Check that proxysql_monitor user exists in all DB instances
     Check that proxysql_monitor credentials are the same in all DB instances
     */}}
     {{- range $dbKey, $db := $dbs }}
+        {{- if hasKey $db "pxc" }}
+            {{- $monitorEnabled = true }}
+            {{- $monitorUsername = "monitor" }}
```
fwiesel, line 26: Shouldn't the username also come from the config?
I.e. 
```
{{- $monitorUsername = $db.system_users.monitor.user | default "monitor"  }}
```

openstack/utils/templates/snippets/_proxysql.cfg.tpl
```diff
@@ -15,11 +15,21 @@ mysql_variables =
...
     {{- $monitorUsername := "" }}
     {{- $monitorPassword := "" }}
+    {{- $monitorGalera := false }}
     {{- /*
     Check that proxysql_monitor user exists in all DB instances
     Check that proxysql_monitor credentials are the same in all DB instances
     */}}
     {{- range $dbKey, $db := $dbs }}
+        {{- if hasKey $db "pxc" }}
+            {{- $monitorEnabled = true }}
```
fwiesel, line 25: Is it required to have monitoring enabled?
If not, I'd enable it conditionally, depending on whether `$db.system_users.monitor.password` exists or not.

openstack/utils/templates/snippets/_proxysql.cfg.tpl
```diff
@@ -49,12 +59,75 @@ mysql_variables =
...
+        max_connections = 20
+        weight = 101
+    }
+*/}}
+    {{- $db := get $dbs $dbKey }}
+    {{- if hasKey $db "pxc" }}
+    {{- range $i := until (int $db.pxc.size) }}
+    {
+        address = "{{ $db.name }}-db-pxc-{{ $i }}.{{ include "svc_fqdn" $.global }}"
+        hostgroup = {{ add 1 (mul 10 (add1 $index)) }}
```
fwiesel, line 76: The implication here is that `$db.pxc.size < 10`, correct?
I'd at least test that and raise an error, if not.

fwiesel (changes requested):
The only thing that is really important for me would be 
- addressing the assumption behind `mul 10`. 
- the monitor username from the config.
-

openstack/utils/templates/snippets/_proxysql.cfg.tpl
s10: In general, a cluster shouldn't be bigger than 7 members.
I will add a check to pxc-db chart and here.

openstack/utils/templates/snippets/_proxysql.cfg.tpl
s10: These system users' usernames ('root', 'monitor', 'operator', 'xtrabackup', 'replication' and 'proxyadmin') in pxc-db are fixed and can't be changed, so it always will be a `monitor`, so there would be misleading to use this variable like we can use other user as a monitor user.

PR #7506, opened 2024-12-04
[netapp-harvest] set alert label 'thanos-ruler'

chuan137:


Carthaca (approved):

prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/templates/alerts.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   labels:
     {{- include "netapp-harvest.labels" . | nindent 4 }}
     type: alerting-rules
-    prometheus: metal
+    thanos-ruler: metal
```
viennaa, line 10: Usually I'd expect these alerts in `prometheus: storage`.

Why do we need to query the whole metal clusters here?

chuan137: https://github.com/sapcc/helm-charts/blob/cf8121f7c34e2430f7609b8bf2a98031301f4fcb/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/alerts/netappsd.yaml#L64
this includes metrics from openstack prometheus and storage prometheus.
we are monitoring if all backend hosts of Manila service are scraped by netapp exporter

chuan137: updated, please have a look new commit

PR #7507, opened 2024-12-05
liquid-ceph: add support for per-az usage and quota

jecluis:
Signed-off-by: Joao Eduardo Luis \<joao@clyso.com>


system/liquid-ceph/templates/configmap.yaml
```diff
@@ -23,3 +23,10 @@ data:
 
     [client.admin]
     keyring = /etc/ceph-keyring/keyring
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: liquid-ceph-zone-mapping
+data:
+  zone-mapping.yaml: ""
```
Nuckal777, line 32: I think the value here needs to be templated.

system/liquid-ceph/templates/deployment.yaml
```diff
@@ -70,21 +71,13 @@ spec:
...
-              secretKeyRef:
-                name: {{ .Values.ceph.rgwSecretName }}
-                key: SecretKey
           securityContext:
             runAsNonRoot: true
           volumeMounts:
           - mountPath: /etc/liquid
             name: policy
+          - mountPath: /etc/liquid
+            name: zone-mapping
```
Nuckal777, line 80: I think the mount points collide. I would propose to move `zone-mappings.yaml` and `policies.yaml` in a single ConfigMap (e.g. called just `liquid-ceph`). Then mounting that single ConfigMap would provide both files.

Nuckal777 (changes requested):

system/liquid-ceph/templates/configmap.yaml
majewsky, line 32: Why do you try to make the configmap name the thing that is configured? The standard approach is to take the zoneMapping config itself via `values.yaml` and render it out here (Ctrl-F `toYaml` on this repo for many many examples of how to do this). Then hardcode the configMap name at the usage site.

majewsky (changes requested):

system/liquid-ceph/templates/configmap.yaml
databus23, line 32: To directly specify the yaml in the values file and render it out here I think this should do it.
```suggestion
  zone-mapping.yaml: | {{- toYaml .Values.ceph.zoneMapping | nindent 4 }}
```

system/liquid-ceph/values.yaml
```diff
@@ -3,13 +3,10 @@ image:
...
 ceph:
-  endpoint: "" # "http://ceph-rgw.ceph.svc.cluster.local:80"
-  targetZone: ""
-  validPools: "" # regex
-  resourceName: "objectstore-ceph-region-3-size"
   secretName: "" # ceph.conf and keyring file
-  rgwSecretName: "" # AccessKey and SecretKey fields
-  allowedDomains: [] 
+  allowedDomains: []
+  zoneMapping: "" # zone mapping file
```
databus23, line 8: file is misleading there needs to be yaml there
```suggestion
  zoneMapping: {} # zone mapping yaml
```

system/liquid-ceph/ci/test-values.yaml
```diff
@@ -6,7 +6,8 @@ global:
   linkerd_requested: false
 ceph:
   secretName: ceph-secret
-  endpoint: "http://ceph-rgw.ceph.svc.cluster.local:80"
-  targetZone: the-zone
-  validPools: the-pools
-  rgwSecretName: rgw-secret
+  zoneMapping: zone-mapping
```
databus23, line 9:
```suggestion
  zoneMapping:
    az-a-3:
      resource: local-standard
      az: a
```

system/liquid-ceph/templates/deployment.yaml
```diff
@@ -39,17 +39,15 @@ spec:
           image: "{{required ".Values.global.registry is missing" .Values.global.registry }}/{{ .Values.image.repository }}:{{required ".Values.image.tag is missing" .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
+          {{ .Values.ceph.enableDebug }}
```
Nuckal777, line 42: Missing `if`.

system/liquid-ceph/templates/deployment.yaml
```diff
@@ -39,17 +39,15 @@ spec:
           image: "{{required ".Values.global.registry is missing" .Values.global.registry }}/{{ .Values.image.repository }}:{{required ".Values.image.tag is missing" .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
+          {{ if .Values.ceph.enableDebug }}
```
databus23, line 42:
```suggestion
          {{- if .Values.ceph.enableDebug }}
```

system/liquid-ceph/templates/deployment.yaml
```diff
@@ -39,17 +39,15 @@ spec:
...
+          - -d
+          {{ end }}
```
databus23, line 44:
```suggestion
          {{- end }}
```

databus23 (approved):

system/liquid-ceph/ci/test-values.yaml
```diff
@@ -6,7 +6,11 @@ global:
   linkerd_requested: false
 ceph:
   secretName: ceph-secret
-  endpoint: "http://ceph-rgw.ceph.svc.cluster.local:80"
-  targetZone: the-zone
-  validPools: the-pools
-  rgwSecretName: rgw-secret
+  zoneMapping:
+    az-a-3:
```
kayrus, line 10: can we reuse the `rgwTargetPlacements` and `premiumPlacements` mappings from the system/cc-ceph chart? duplicating zone mappings adds flexibility, but also a complexity and a room for a human error.

kayrus: cc @sumitarora2786

databus23 (approved):

PR #7512, opened 2024-12-05
upgrade cam link for elektra

hgw77:
this will update the cam link in elektra

### DIFF
--- a/openstack/elektra/values.yaml
+++ b/openstack/elektra/values.yaml
@@ -19,7 +19,7 @@ monsoon_dashboard_mail_user: defined_in_secrets
 monsoon_dashboard_mail_password: defined_in_secrets
 
 monsoon_dashboard_avatar_url: https://avatars.wdf.sap.corp/avatar/#{current_user.name}?size=24x24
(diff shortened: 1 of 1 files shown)


openstack/elektra/values.yaml
```diff
@@ -19,7 +19,7 @@ monsoon_dashboard_mail_user: defined_in_secrets
 monsoon_dashboard_mail_password: defined_in_secrets
 
 monsoon_dashboard_avatar_url: https://avatars.wdf.sap.corp/avatar/#{current_user.name}?size=24x24
-monsoon_dashboard_cam_url: https://spc.ondemand.com/sap/bc/webdynpro/a1sspc/cam_wd_central
+monsoon_dashboard_cam_url: https://cam.int.sap/cam/ui/admin?#
```
andypf, line 22: Hey, is `?#`really needed at the end?

andypf (commented):
Please address my comments

openstack/elektra/values.yaml
hgw77: this the link that was provided and it worked but you are right I guess we do not need that

PR #7514, opened 2024-12-05
[vmware-monitoring] Modify host system properties metric

kevin-fischer:
- [x] Check if the data is available in the vROps REST API
- Change the no swap DS metric to NVMe


himanip94 (approved):
lgtm

PR #7517, opened 2024-12-09
[unbound] accept both externalIP (singular) and externalIPs (plural)

vssldmtrv:
Keeping externalIP for backwards compatibility's sake. We might think about retiring it completely later on.

The externalIP value is expected to be a string, i.e. a single IP. The externalIPs value is expected to be a list.

If both are defined then the values will be merged. If none are defined then we error out.

Q: Why we need this?
A: There might be a situation where we need to expose unbound over multiple IPs.

### DIFF
--- a/system/unbound/templates/service.yaml
+++ b/system/unbound/templates/service.yaml
@@ -23,4 +23,13 @@ spec:
       protocol: UDP
       port: 53
       targetPort: dns-udp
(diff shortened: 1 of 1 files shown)


galkindmitrii (commented):
I'd rather keep it simple by using list in the (regional) values directly and keep here the values reference without the list brackets `[ ... ]`.

PR #7520, opened 2024-12-09
[PX] Use correct python interpreter for lgproxy

swagner-de:
With the new looking glass image, we use a distroless image. Instead of
running and installing everything into system python as before, we need
to specify the python in the venv that we copied into the runtime
container at the build stage.

This wasn't uncovered in testing as we only ever tested that image in
the lg pod where we would rely on the ENTRYPOINT directive from the
dockerfile. As a sidecar, we start a different pyhton file though but
still called that from system python.


occamshatchet (approved):
Looks good.

nikatza (approved):

PR #7530, opened 2024-12-11
[octavia] Fix OctaviaApiFailing alert: Only fires at falling edge

BenjaminLudwigSAP:
We want an alert that triggers immediately, but `changes` only triggered at the falling edge.
Use `max` instead and alert on any metric presence. If this doesn't work like I expect, I will change it again.


m-kratochvil (approved):
lgtm

PR #7546, opened 2024-12-16
maia: ignore metrics from limes-global

majewsky:


viennaa (commented):
Two things:

How fast does this need to go live?

Increase patch version of the Chart please.

majewsky:
> How fast does this need to go live?

The delivery target is Q4, so I need to push this out before my vacation starts on Thursday. I'm going for a deployment to eu-de-2 tomorrow. Sorry that I didn't think about Maia before; this topic has been stuck in limbo for half a year and I did not anticipate the sudden panic that set in around this topic a week ago.

> Increase patch version of the Chart please.

Done. (Even though chart versions are irrelevant for top-level charts.)

viennaa (approved):
Looks good to me. Approval from @artherd42 or @notque pls

Chart versioning matters for rollbacks :)

notque (approved):
LGTM - although I'm now very concerned I didn't get the memo about the priority of this increasing for my stuff.

PR #7556, opened 2024-12-16
newer traefik requires h2c

viennaa:
without this option, requests which are coming in from outside the cluster will have a malformed header in their reply.


richardtief (approved):
Thanks. Looks good

PR #7562, opened 2024-12-17
[unbound] introducing multiple external ports

vssldmtrv:
We can now expose the unbound ports (53/tcp, 53/udp) over multiple external ports.

That might came handy if the clients connecting to the unbounds are behind a load balancers and get NAT-ed. A lot of clients could cause the LB NAT tables to get exhausted.

Using multiple ports and/or multiple service IPs we can increase the service endpoints as required in order to accomodate those clients.


mutax (approved):
lgtm

mbrowny (approved):

PR #7564, opened 2024-12-17
[netapp-harvest] exclude particular filer for alert

chuan137:
The name for this filer (stnpca3-cp001) follows old name convention. We
use alternative name in manila-vendor.yaml than how it is configured in
netbox. Replace host label for it to avoid false alert.

### DIFF
--- a/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/alerts/thanos-metal/netapp.yaml
+++ b/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/alerts/thanos-metal/netapp.yaml
@@ -1,9 +1,17 @@
 groups:
   - name: netapp-harvest
     rules:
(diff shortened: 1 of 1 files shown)


Carthaca (commented):
What do you think about replacing the value instead of ignoring this host?

e.g.
```
label_replace(manila_total_capacity_gb,"share_backend_fqdn","stnpca3-cp001.cc.eu-de-1.cloud.sap","share_backend_fqdn","stnpca3.cc.eu-de-1.cloud.sap")
```

chuan137:
> What do you think about replacing the value instead of ignoring this host?
> 
> e.g.
> 
> ```
> label_replace(manila_total_capacity_gb,"share_backend_fqdn","stnpca3-cp001.cc.eu-de-1.cloud.sap","share_backend_fqdn","stnpca3.cc.eu-de-1.cloud.sap")
> ```

thanks, updated based on your idea

Carthaca (approved):

PR #7569, opened 2024-12-17
[unbound] make unbound listen on multiple ports

vssldmtrv:
The goal is to use 1-to-1 service port mappings as having multiple external ports mapping to the same backend port is not a supported feature. See:
https://github.com/kubernetes/kubernetes/issues/47222

Also, we use the interface unbound config clause as it can be specified multiple times as opposed to the port clause.


mutax (approved):
lgtm

mbrowny (approved):

PR #7578, opened 2024-12-18
[ceph] support multi-instance rgw

arttor:
add support for [Rook rgw multi-instance  setup](https://rook.io/docs/rook/latest-release/Storage-Configuration/Object-Storage-RGW/object-storage/#object-multi-instance).


arttor:
@kayrus i've copied all resources like cert, record, service for extra RGW. Did i miss anything to support tls and domains for new RGW instances?

PR #7579, opened 2024-12-19
[ironic] Set agent relevant conductor parameters in api too

fwiesel:
The ironic-python-agent gets some config values over the lookup api as 'config' in the json.
And two values are unintuitively taken from the
conductor config section.
So, we have to set them in the api node as well,
in order to have the settings take hold in the agent.


fwiesel:
The PR itself doesn't have the message: This disables the image verification in the the conductor.
We need a new ironic-python-agent which does that instead.

sandzwerg (approved):

PR #7583, opened 2024-12-20
[keystone][pxc-db] Add support for PXC galera cluster

s10:
* Add support for PXC galera cluster URL
* Update mysql-metrics and utils charts to support galera cluster
* Remove mariadb-galera support
* Remove unused helm helper functions db_host and keystone_db_host
* Update linkerd-support and owner-info to release versions

### DIFF
--- a/openstack/keystone/Chart.lock
+++ b/openstack/keystone/Chart.lock
@@ -2,26 +2,26 @@ dependencies:
 - name: mariadb
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
   version: 0.16.9
-- name: mariadb-galera
+- name: pxc-db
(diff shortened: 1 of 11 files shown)


rajivmucheli:
is this tested in qa regions ? qa-de-3 maybe ?

s10:
> is this tested in qa regions ? qa-de-3 maybe ?

Don't we have a chicken-egg problem here? This couldn't be tested before this PR is merged.

However, this PR itself shouldn't produce any diff outside of mysql_metrics version bump without introducing to secrets:
```
pxc:
  enabled: true
```
After that new pxc cluster will be created alongside mariadb database, and no switch-over will be performed if dbType variable isn't set to `pxc-db`

galkindmitrii (approved):

stanislav-zaprudskiy (approved):

PR #7602, opened 2025-01-02
fix not existing metric

viennaa:
the older fix did not quite cut it.
it needs to use count and bucket and one metric was still pointing to the old, non-existing one.


richardtief (approved):
Thank you for keeping an eye on these changes.

PR #7603, opened 2025-01-03
[unbound] moved the prometheus annotations to the deployment

vssldmtrv:
This is making sure the scrape targets are not discovered twice.

We've also removed the "prometheus.io/port" annotation and relying on the metrics port definition.

This way the prometheus service discovery will be ending up with the metric container.

The monitors will have to be adjusted.


mbrowny (approved):

galkindmitrii (commented):
why discarding the port?
`prometheus.io/port: "{{$.Values.unbound.port_unbound_exporter}}"`

PR #7604, opened 2025-01-03
[pxc-operator] Set correct target for the OperatorNotReady alerts

s10:
* Set correct target for the OperatorNotReady alerts
* Update maintainers list

Pods metrics from kube-state-metrics present only in kubernetes prometheus


galkindmitrii (approved):
jfyi: since alerts are a template the `support_group` could be taken from values `owner-info: support-group`

PR #7605, opened 2025-01-03
[octavia][pxc-db] Add support for PXC galera cluster: part 1

s10:
* Add support for PXC galera cluster URL
* Update mysql-metrics and utils charts to support galera cluster
  * mysql-metrics 0.3.5 -> 0.4.2, minimum 0.3.6 is required because of the https://github.com/sapcc/helm-charts/pull/7130
  * utils  0.19.7 -> 0.23.0, minimum 0.21.0 is required because of the https://github.com/sapcc/helm-charts/pull/7505

### DIFF
--- a/openstack/octavia/ci/test-values.yaml
+++ b/openstack/octavia/ci/test-values.yaml
@@ -1,9 +1,11 @@
+---
 global:
   registry: myRegistry
   dbPassword: secret!
   master_password: topSecret!
   octavia_service_password: topSecret!!
-  dockerHubMirrorAlternateRegion: some_region
+  registryAlternateRegion: other.docker.registry
+  dockerHubMirrorAlternateRegion: other.dockerhub.mirror
   availability_zones:
     - foo
     - bar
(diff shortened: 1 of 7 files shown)


openstack/octavia/templates/etc/_secrets.conf.tpl
```diff
@@ -11,7 +11,7 @@ username = {{ .Release.Name }}
 password = {{ .Values.global.octavia_service_password | replace "$" "" }}
 
 [database]
-connection = {{ include "db_url_mysql" . }}
+connection = {{ include "utils.db_url" . }}
```
BenjaminLudwigSAP, line 14: The [definition](https://github.com/sapcc/helm-charts/blob/master/openstack/utils/templates/_hosts.tpl#L59-L67) reads to me like this config would default to `utils._db_url_mariadb`, because `dbType` isn't set, right? Not sure whether that's what you want.

openstack/octavia/ci/test-values.yaml
```diff
@@ -1,9 +1,11 @@
+---
 global:
   registry: myRegistry
   dbPassword: secret!
   master_password: topSecret!
   octavia_service_password: topSecret!!
-  dockerHubMirrorAlternateRegion: some_region
+  registryAlternateRegion: other.docker.registry
+  dockerHubMirrorAlternateRegion: other.dockerhub.mirror
```
BenjaminLudwigSAP, line 8: What are these needed for? Why is it okay to use dockerhub directly instead of keppel?

BenjaminLudwigSAP (commented):
I don't know how pxc-db works, so I have to trust you on this, but I guess it looks okay to me. I just have a few questions.

openstack/octavia/templates/etc/_secrets.conf.tpl
s10: This is the expected behavior. If `dbType` isn't set, then the this value would be rendered the same as if `dbType=mariadb`, eg current behavior.

openstack/octavia/ci/test-values.yaml
s10: We currently use dockerub mirror for upstream mariadb, memcached and rabbitmq images. The same is with database images like `percona/percona-xtradb-cluster:8.0.39-30.1` in `pxc-db` chart.

galkindmitrii (approved):

s10:
@BenjaminLudwigSAP , could you please take a look at this PR again?

PR #7607, opened 2025-01-03
fix broken playbook link

artherd42:


timojohlo (approved):
Thanks for fixing this!

PR #7608, opened 2025-01-06
Add sr 4s2t hv flavor

degricar:
we currently have only the bm flavor for our Thinksystems SR850 v3. Adding the hv flavor

### DIFF
--- a/openstack/ironic/templates/seed.yaml
+++ b/openstack/ironic/templates/seed.yaml
@@ -541,6 +541,18 @@ spec:
       "resources:CUSTOM_BM_S16_C16_M8192_V0": "1"
       {{- tuple . "baremetal" | include "ironic.helpers.extra_specs" | indent 6 }}
 
+  # New Sapphire Rapids 4S2T as Hypervisor in qa
+  - name: "hv_s4_c32_m2048_v0"
+    id: "2041"
+    vcpus: 128
+    ram: 2097152
+    disk: 447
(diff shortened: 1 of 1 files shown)


sandzwerg (approved):

openstack/ironic/templates/seed.yaml
```diff
@@ -541,6 +541,18 @@ spec:
       "resources:CUSTOM_BM_S16_C16_M8192_V0": "1"
       {{- tuple . "baremetal" | include "ironic.helpers.extra_specs" | indent 6 }}
 
+  # New Sapphire Rapids 4S2T as Hypervisor in qa
+  - name: "hv_s4_c32_m2048_v0"
+    id: "2041"
```
BerndKue, line 546: Since we already have the corresponding BM flavor 22035 https://github.com/sapcc/helm-charts/blob/1f1837f528d60ab9b89b4a5c14fd716172b3ebd4/openstack/ironic/templates/seed.yaml#L473
the right ID would be 2035

BerndKue (changes requested):

openstack/ironic/templates/seed.yaml
degricar: wasn't aware of this naming convention.
I've adjusted it accordingly

BerndKue (approved):

sandzwerg (approved):

PR #7613, opened 2025-01-06
chore: bump image version

artherd42:
The base image has reached End of Support Life and will not receive further security updates.


prometheus-exporters/cloudprober/Chart.yaml
```diff
@@ -1,10 +1,13 @@
 apiVersion: v2
 name: cloudprober
-version: 0.12.2
+version: 0.13.9
```
viennaa, line 3:
```suggestion
version: 0.13.0
```

viennaa (changes requested):
Thanks for the image bump!

I've only found one breaking change in v0.13.2 which affects `targets.endpoints` which I could not find to be used here. So this change is good to go. Some little version thing below.

viennaa (approved):

PR #7615, opened 2025-01-06
adding bm flavor for 8s8t

degricar:
lenovo sr950 v3

### DIFF
--- a/openstack/ironic/templates/seed.yaml
+++ b/openstack/ironic/templates/seed.yaml
@@ -557,6 +557,18 @@ spec:
       "resources:CUSTOM_BG_S2_C48_M2048_V0": "1"
       {{- tuple . "baremetal" | include "ironic.helpers.extra_specs" | indent 6 }}
 
+  # Sapphire Rapids 8S8T BM servers
+  - name: "bm_s8_c32_m8192_v0"
+    id: "2042"
+    vcpus: 256
+    ram: 8388608
+    disk: 480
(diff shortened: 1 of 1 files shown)


FThimm (approved):

sandzwerg (approved):

openstack/ironic/templates/seed.yaml
```diff
@@ -557,6 +557,18 @@ spec:
       "resources:CUSTOM_BG_S2_C48_M2048_V0": "1"
       {{- tuple . "baremetal" | include "ironic.helpers.extra_specs" | indent 6 }}
 
+  # Sapphire Rapids 8S8T BM servers
+  - name: "bm_s8_c32_m8192_v0"
+    id: "2042"
```
BerndKue, line 562: BM Flavor should have an additional 2 in the ID: https://github.com/sapcc/helm-charts/blob/76c0f1b88a61041adcbf6347ce95b96e10e9439a/openstack/ironic/templates/seed.yaml#L70

22042 would be the right ID.

BerndKue (changes requested):

openstack/ironic/templates/seed.yaml
degricar: or as the hv flavor was already present as 2037 it should rather be 22037 for the bm one
adjusted it accordingly

BerndKue (approved):

PR #7632, opened 2025-01-09
[cinder] Change default name of storage_profile

hemna:
This patch updates the default name of the FCD storage profile from vmware-fcd to cinder-fcd.


joker-at-work (approved):

dhoeller (approved):
LGTM

PR #7633, opened 2025-01-09
fix(ci): disable helm lint & helm push

kengou:
Disable helm linting for now


SuperSandro2000 (changes requested):
Please add a comment why we commented that or we forget what we did here.

sandzwerg (approved):
I'd suggest squashing this on merge

PR #7642, opened 2025-01-10
[cinder][pxc-db] Add support for PXC galera cluster: part 1

s10:
Add default values for pxc-db chart dependency


galkindmitrii (approved):

s10:
@hemna , could you please take a look at this PR?

hemna:
Please bump the Chart.yml version for cinder since this is a significant change

hemna (approved):

PR #7644, opened 2025-01-13
adjust support_group

viennaa:


system/metis/values.yaml
```diff
@@ -7,7 +7,7 @@ global:
 
 alerts:
   prometheus: "infra-collector"
-  supportGroup: "observability"
+  supportGroup: "services"
```
richardtief, line 10: This should be: 
```suggestion
  supportGroup: "containers"
```

richardtief (changes requested):
Please check with @IvoGoman. Thanks!

IvoGoman (approved):
Thanks :)

system/metis/values.yaml
```diff
@@ -7,7 +7,7 @@ global:
 
 alerts:
   prometheus: "infra-collector"
-  supportGroup: "observability"
+  supportGroup: "containers"
 
 owner-info:
   support-group: observability
```
IvoGoman, line 13:
```suggestion
  support-group: containers
```

IvoGoman (changes requested):

system/metis/values.yaml
richardtief: This is another occurrence in the `alerts` directive.

IvoGoman (approved):

PR #7653, opened 2025-01-13
[octobus-query-exporter] fix jumpserver alert, mv alert to regional thanos ruler

Kuckkuck:
…thanos-ruler

### DIFF
--- a/prometheus-exporters/octobus-query-exporter/values.yaml
+++ b/prometheus-exporters/octobus-query-exporter/values.yaml
@@ -16,6 +16,8 @@ octobus_query_exporter:
         type: prometheus
       - name: scaleout
         type: thanos-ruler
+      - name: region
+        type: thanos-ruler
   auditSources:
     enabled: false
   listen_port: 9206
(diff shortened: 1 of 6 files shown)


prometheus-exporters/octobus-query-exporter/values.yaml
```diff
@@ -16,6 +16,8 @@ octobus_query_exporter:
         type: prometheus
       - name: scaleout
         type: thanos-ruler
+      - name: region
```
viennaa, line 19: Please also adjust the folder from `region` to `regional`

```suggestion
      - name: regional
```

prometheus-exporters/octobus-query-exporter/vendor/octobus-query-exporter/alerts/region/jump.alerts
```diff
@@ -0,0 +1,16 @@
+groups:
+  - name: jumplogs
+    rules:
+      - alert: OctobusJumpserverLogShipping
```
viennaa, line 4: Can you remove this from the other alerts defintion?

viennaa (changes requested):
Thanks. Little typo inside and we need to remove the other alert condition.

viennaa (approved):
Thanks, very much appreciated.

PR #7666, opened 2025-01-14
increase ironic liquid to 128MiB

VoigtS:
It was confirmed that there were baremetal deployments for a new flavor today. This strained the ironic liquid. 
The increase to 128Mi should be sufficient for now. The peaks are about 100MiB.


Nuckal777 (approved):

majewsky:
We will go with https://github.com/sapcc/limes/pull/644 instead and reduce the page limit in affected regions.

PR #7668, opened 2025-01-15
[unbound] create proper PrometheusRule objects

vssldmtrv:
No need to be that dynamic and work with *.alert files.

Also, recovered the UnboundEndpointNotAvailable alert (previously known as OpenstackDesignateDnsUnboundEndpointNotAvailable).


mbrowny (approved):

system/unbound/templates/prometheus-alerts.yaml
```diff
@@ -0,0 +1,93 @@
...
+---
+apiVersion: "monitoring.coreos.com/v1"
+kind: "PrometheusRule"
+metadata:
+  name: unbound-kubernetes-alerts
+  labels:
+    app: unbound
+    tier: os
+    type: alerting-rules
+    prometheus: openstack
```
galkindmitrii, line 73: this should be `kubernetes`

galkindmitrii (approved):

mbrowny (approved):

PR #7670, opened 2025-01-15
seeding a new user for backup service

chuan137:


openstack/manila/templates/seeds.yaml
```diff
@@ -64,6 +64,8 @@ spec:
   - name: cloud_sharedfilesystem_viewer
   - name: sharedfilesystem_admin
   - name: sharedfilesystem_viewer
+{{- Role `cloud_objectstore_admin` needed for backup to cloud service }}
+  - name: cloud_objectstore_admin
```
Carthaca, line 68: This is seeded in `swift-seed`, so maybe we should add a dependency to that above under `spec.requires`.
But this line here should be removed

Carthaca (changes requested):

openstack/manila/templates/seeds.yaml
```diff
@@ -64,6 +64,8 @@ spec:
   - name: cloud_sharedfilesystem_viewer
   - name: sharedfilesystem_admin
   - name: sharedfilesystem_viewer
+{{- Role `cloud_objectstore_admin` needed for backup to cloud service }}
```
Carthaca, line 67: This is not a valid format for a comment, maybe add it to the role assignment L118 below at the end after putting `# `

openstack/manila/templates/seeds.yaml
```diff
@@ -99,13 +100,23 @@ spec:
         role: service
       - project: service
         role: cloud_network_admin
+    - name: manilabackup
+      description: Manila Backup Service
+      password: '{{.Values.global.manila_bakcup_password | include "resolve_secret"}}'
```
Carthaca, line 105: typo
```suggestion
      password: '{{.Values.global.manila_backup_password | include "resolve_secret"}}'
```

Carthaca (changes requested):

Carthaca (approved):

PR #7678, opened 2025-01-16
Update vmware_esxi_nic_mellanox_issue.cfg

atcyuth:
added another search string : command failed: IO was aborted

### DIFF
--- a/prometheus-exporters/octobus-query-exporter/vendor/octobus-query-exporter/files/queries/vmware/vmware_esxi_nic_mellanox_issue.cfg
+++ b/prometheus-exporters/octobus-query-exporter/vendor/octobus-query-exporter/files/queries/vmware/vmware_esxi_nic_mellanox_issue.cfg
@@ -10,7 +10,8 @@ QueryJson = {
           "bool": {
               "must": [
                   { "match_phrase": { "syslog_message": "Health: Miss counters detected" }},
-                  { "match": { "syslog_message": "NMLX_ERR" }}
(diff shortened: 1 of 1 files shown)


prometheus-exporters/octobus-query-exporter/vendor/octobus-query-exporter/files/queries/vmware/vmware_esxi_nic_mellanox_issue.cfg
```diff
@@ -10,7 +10,8 @@ QueryJson = {
           "bool": {
               "must": [
                   { "match_phrase": { "syslog_message": "Health: Miss counters detected" }},
-                  { "match": { "syslog_message": "NMLX_ERR" }}
+                  { "match": { "syslog_message": "NMLX_ERR" }},
+                  { "match": { "syslog_message": "command failed: IO was aborted" }}
```
kevin-fischer, line 14: The syntax here won't quite work like this!
@max-len is still working on the issue and will rewrite it.

```suggestion
```

kevin-fischer (changes requested):

kevin-fischer (commented):
@max-len, it seems that you also reverted the filter range
from `now-15m` to `now-1h` again.
See here: --> https://github.com/sapcc/helm-charts/commit/6c51fc192972d3bab061e9fef95f435ed262860e.

Was this intentional?
If not, could we please change it back before merging?

kevin-fischer (changes requested):

PR #7684, opened 2025-01-16
services/andromeda: liquid server sidecar definition

ronchi-oss:


ronchi-oss:
After discussing with Andy, this will be recreated with the Liquid Server as its own pod instead of sidecar container.

PR #7686, opened 2025-01-16
[ironic] add states power to audit middleware mapping

notque:
This corrects the unmapped path for states/power in ironic audit events.


notque:
I've also added a commit for attempting to handle the specifics that in ironic, a name or a uuid can be used. If this doesn't work, we will have to edit the openstack-audit-middleware's current setup which only matches UUID, to a regex that will also match a naming convention for ironic nodes.

BerndKue (commented):
Lgtm

PR #7687, opened 2025-01-17
[hermes] upgrade rabbitmq chart 0.13.0

notque:


notque:
I'm not sure I need to actually update the chart version since really, I'm just upgrading subdependencies. 

If this is an issue, just let me know

artherd42 (approved):
lgtm

viennaa (approved):
Thanks for updating old stuff :).
 
Updating the overall chart version is very reasonable and highly encouraged by me. In the end we always want to be able to easily roll back in case something goes down and it would be hard to distinguish, which of the `0.1.2` versions has the updated subcharts and which one hasn't. Especially after one rollback, another deployment and another rollback. Then we wind up with 5 times the same version in `helm history`.

Always bump, always good 🚀

PR #7690, opened 2025-01-17
[manila] add metrics for limes capacity plugin

Carthaca:
we move the function definition away from limes and take
ownership in manila


chuan137 (approved):
lgtm

kpawar-sap (approved):

PR #7700, opened 2025-01-20
cornjob-monthly-ticket-analysis-report.yaml

sunitha-reddy:
This is a cornjob template for ticket analysis script that runs monthly.


timojohlo (approved):
Its a cool use case, LGTM!

PR #7701, opened 2025-01-20
[KMIP] Support secret-injector

rajivmucheli:


openstack/kmip/templates/deployment.yaml
```diff
@@ -93,27 +93,27 @@ spec:
             - name: OS_AUTH_URL
               value: "https://{{ include "keystone_api_endpoint_host_public" . }}:443/v3"
             - name: OS_USERNAME
-              value: {{ .Values.kmip.openstack_env.username | quote }}
+              value: {{ .Values.kmip.openstack_env.username | include "resolve_secret" }}
```
s10, line 96: This wouldn't work, because `secret-injector` supports only `Secret` resource kind. So all these secrets should be moved to `Secret` resource first.

s10 (approved):

PR #7702, opened 2025-01-20
[unbound] definde the prometheus alerts per unbound deployment

vssldmtrv:
There are two unbound deployments - unbound1 and unbound2.

We could have a separate deployment just for the alerts or integrate the alerts into the actual unbound deployment. The latter is what we choose to do.

Also, replaced the DnsUnboundEndpointNotAvailable alert with DnsUnbound{1,2}UnexpectedNumberOfEndpoints.

The new alert will fire if the number of endpoints is not what we expect. We expect a specific number based on the number of ports exposed, the protocols and replicas.

For example, a deployment with two ports, two replicas using both TCP and UDP, should always result in 8 endpoints.

### DIFF
--- a/system/unbound/templates/prometheus-alerts.yaml
+++ b/system/unbound/templates/prometheus-alerts.yaml
@@ -2,7 +2,7 @@
 apiVersion: "monitoring.coreos.com/v1"
 kind: "PrometheusRule"
 metadata:
-  name: unbound-openstack-alerts
+  name: {{ $.Values.unbound.name }}-openstack-alerts
   labels:
     app: unbound
     tier: os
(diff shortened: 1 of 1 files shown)


system/unbound/templates/prometheus-alerts.yaml
```diff
@@ -13,59 +13,43 @@ spec:
...
       annotations:
-        description: {{` 'Recursor {{ $labels.app }} returns lots of SERVFAIL responses in {{ $labels.region }} region.' `}}
-        summary: {{` '{{ $labels.app }} returned a lot of SERVFAIL responses in the last hour. Check the logs.' `}}
+        description: 'Recursor {{ $.Values.unbound.name }} returns lots of SERVFAIL responses in {{ $.Values.global.region }} region.'
+        summary: '{{ $.Values.unbound.name }} returned a lot of SERVFAIL responses in the last hour. Check the logs.'
 
-    - alert: DnsUnbound1Down
-      expr: absent(unbound_up{app="unbound1"}) == 1 or unbound_up{app="unbound1"} != 1
+    - alert: Dns{{ $.Values.unbound.name | title }}Down
+      expr: absent(unbound_up{app="{{ $.Values.unbound.name }}"}) == 1 or unbound_up{app="{{ $.Values.unbound.name }}"} != 1
```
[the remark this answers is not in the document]
vssldmtrv: You're probably right. And my tests seem to confirm that as well.
The one with `==1 ` is working fine, too, and since the goal of these PRs is to migrate the alerts from designate to unbound, I am not touching things that function. We can improve them going forward.

system/unbound/templates/prometheus-alerts.yaml
```diff
@@ -76,18 +60,22 @@ spec:
   groups:
   - name: unbound.alerts
     rules:
-    - alert: DnsUnboundEndpointNotAvailable
-      expr: max(kube_endpoint_address{namespace="dns-recursor"}) BY (region,endpoint) < 1
+    - alert: Dns{{ $.Values.unbound.name | title }}UnexpectedNumberOfEndpoints
+{{- $num_replicas := 2 }}
```
galkindmitrii, line 64: replicas might be a bit misleading here, because we have 2 Unbound deployments with 2 replicas each == 4 pods.
maybe `num_deployments` or `num_services` could be used instead.

galkindmitrii (approved):

system/unbound/templates/prometheus-alerts.yaml
vssldmtrv: I use the word replicas as intended though. Since each alert is dealing with its own deployment, we're not really dealing with the number of deployments.

galkindmitrii: Right, we have independent alert now.

mbrowny (approved):

PR #7704, opened 2025-01-20
limes: add liquid-nova deployment

majewsky:
@Varsius Please double-check the liquid-nova configuration. I'm especially unsure whether we need `ignored_traits` from the start.

### DIFF
--- a/openstack/limes/values.yaml
+++ b/openstack/limes/values.yaml
@@ -30,6 +30,7 @@ limes:
     designate: { skip: false, cpu_request:  20m, memory_limit:  64Mi }
     ironic:    { skip: false, cpu_request:  20m, memory_limit:  64Mi }
     manila:    { skip: false, cpu_request:  20m, memory_limit:  64Mi }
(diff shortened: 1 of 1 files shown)


VoigtS (approved):

Varsius:
We implemented `ignored_traits` for the case that some pooled flavors require different values of the trait `trait:CUSTOM_HW_SAPPHIRE_RAPIDS`. There was no such error when I tried reporting capacity on the qa-de-1 deployment.

The remaining part of the configuration looks good to me 👍🏻

Varsius:
> We implemented `ignored_traits` for the case that some pooled flavors require different values of the trait `trait:CUSTOM_HW_SAPPHIRE_RAPIDS`. There was no such error when I tried reporting capacity on the qa-de-1 deployment.

It turns out that all pooled flavors in qa-de-1 require the trait CUSTOM_HW_SAPPHIRE_RAPIDS, thats why there was no error regarding the agreement of common extra specs between pooled flavors. However, this lead to ignoring some hypervisors without this trait in the capacity calculation. Follow up PR: #7806

PR #7707, opened 2025-01-20
[nova] Fix shellinabox console

fwiesel:
Some functions were moved common/token_lookup.lua without the references for shellinabox.
The common part now also needs to trigger an update of the deployment of shellinabox deployments.

### DIFF
--- a/openstack/nova/console/common/token_lookup.lua
+++ b/openstack/nova/console/common/token_lookup.lua
@@ -44,7 +44,7 @@ end
 -- the url, or from the referrer
 -- Caches the value, as it won't change over the request
 -- and apparently ngx.var lookups are not the cheapest
-local get_token = store_in_ctx("token", function()
+_M.get_token = store_in_ctx("token", function()
     local var_token = ngx.var.token
     if var_token then return var_token end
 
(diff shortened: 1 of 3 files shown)


openstack/nova/console/common/token_lookup.lua
```diff
@@ -164,24 +164,24 @@ end
...
+        return lookup_db(token_hash)
+    else
+        local entry, err = lookup_memc(token_hash)
+
+        if err then
+            ngx.log(ngx.ERR, "could not retrieve user: ", err)
+            return ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
+        end
+        return entry
+    end
```
joker-at-work, line 184: optional: This could be more readable if we removed the `else`:
```lua
if not use_memc then
    return lookup_db(token_hash)
end

local entry, err = lookup_memc(token_hash)
if err then
    ngx.log(ngx.ERR, "could not retrieve user: ", err)
    return ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
end
return entry
```

joker-at-work (approved):

sandzwerg (approved):

PR #7714, opened 2025-01-20
[nova] Fixup 0a699e7d Fix shellinabox console

fwiesel:


sandzwerg (approved):

openstack/nova/console/shellinabox/revproxy.lua
```diff
@@ -3,7 +3,7 @@ local _M = {}
 local token_lookup = require "token_lookup"
 local get_token = token_lookup.get_token
 local lookup = token_lookup.lookup
-local host_from_entry = token_lookup.token_lookup.lookup
+local host_from_entry = token_lookup.fix_nova_shellinabox_console
```
joker-at-work, line 6: Shouldn't this be `token_lookup.host_from_entry`?

joker-at-work (changes requested):

openstack/nova/console/shellinabox/revproxy.lua
fwiesel: Sorry you are right

joker-at-work: This is still not correct. It now points to `token_lookup.lookup` instead of `token_lookup.host_from_entry`.

fwiesel: You're right... maybe I better lie down. Sorry for the extra work.

joker-at-work (approved):

PR #7724, opened 2025-01-22
Tempest kvm

misamoylov:


galkindmitrii:
Please add respective a entry to codeowners: https://github.com/sapcc/helm-charts/blob/master/.github/CODEOWNERS#L109-L119

PR #7732, opened 2025-01-23
WIP: KVM Seeds

fwiesel:


fwiesel:
The alternative to that I had in mind  would be to move that logic out of the helm charts and add a generator in our configuration repo to generate all the variants, and have this chart just render the values for the flavors ([like this)](https://github.com/sapcc/helm-charts/pull/7732/files#diff-22848d73caa0caa8dc03ce8ed5b388114cdd9b3e1c63fc9e575e73a35df4fc9eR31-R34).

I mean, I would intend to move the values out of the chart into the configuration repo anyway, but the question here is more how much logic should be in this chart.

PR #7741, opened 2025-01-24
[andromeda] switch to secrets-injector.

ronchi-oss:
To be rolled out with https://github.wdf.sap.corp/cc/secrets/pull/15992

* [X] Apply basic `sed` transformations to `values.yaml` files
* [X] Inject`DATABASE_CONNECTION` in all pods that need it
* [x] Migrate `edgerc`

### DIFF
--- a/global/andromeda/templates/deployment.yaml
+++ b/global/andromeda/templates/deployment.yaml
@@ -67,6 +67,13 @@ spec:
                 secretKeyRef:
                   name: andromeda-secret
                   key: service_user_password
+            {{- if .Values.mariadb.enabled }}
+            - name: DATABASE_CONNECTION
+              valueFrom:
+                secretKeyRef:
+                  name: andromeda-secret
+                  key: database_connection
+            {{- end }}
             {{- if .Values.audit.enabled }}
             - name: AUDIT_TRANSPORT_URL
               valueFrom:
--- a/global/andromeda/templates/deployment.yaml
+++ b/global/andromeda/templates/deployment.yaml
@@ -44,6 +44,10 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           args: ["--config-file", "/etc/andromeda/andromeda.yaml", "--config-file", "/etc/andromeda/akamai.yaml", "--port", "8000"]
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          lifecycle:
+            poststart:
+              exec:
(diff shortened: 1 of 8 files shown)


global/andromeda/templates/deployment.yaml
```diff
@@ -62,6 +62,18 @@ spec:
                 secretKeyRef:
                   name: sentry
                   key: {{ include "andromeda.fullname" . }}.DSN
+            - name: OS_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: andromeda-secret
+                  key: service_user_password
+            {{- if .Values.audit.enabled }}
+            - name: AUDIT_TRANSPORT_URL
```
notandy, line 71: `AUDIT_TRANSPORT_URL` env is not yet supported by andromeda, I've created a PR for that: https://github.com/sapcc/andromeda/pull/722

global/andromeda/templates/etc/_andromeda.yaml.tpl
notandy, line 1: `database.connection` includes database credentials from vault, which we need to put into kubernetes secrets.
https://github.com/sapcc/andromeda/pull/722 includes the env variable `DATABASE_CONNECTION` which can be used in a similar way as `AUDIT_TRANSPORT_URL` to render the whole connection string in a secret and then inject it as an environment variable to the andromeda-api, housekeeper and migration pod. 
The akamai agent doesn't need it since it doesn't connect via SQL but uses only NATS.

notandy (changes requested):
needs https://github.com/sapcc/andromeda/pull/722 and the database.connection adapation, rest looks good!

notandy:
Also just noticed, we need to move the edgerc to secrets: https://github.com/sapcc/helm-charts/blob/master/global/andromeda/templates/etc/_edgerc.tpl

could be just mounted like this: https://www.jeffgeerling.com/blog/2019/mounting-kubernetes-secret-single-file-inside-pod

global/andromeda/templates/etc/_andromeda.yaml.tpl
ronchi-oss: I still haven't checked *what* is using the `root` password though. Whatever uses it should also be broken at the moment.

ronchi-oss: Is it possible that the MariaDB `root_password` is not in use?

```sh
find -s ./global/andromeda -type f | xargs grep root_password
```

ronchi-oss: Nevermind, the `root_password` definition should be fine.

PR #7742, opened 2025-01-24
add rook admin user for prysm

arttor:
Use rook `CephObjectStoreUser` for prysm.

@kayrus @fa-at-pulsit please take a look.


kayrus (commented):
you need to bump the Chart.yaml version.
another concern is that we still don't have a proper working configuration for the multi instance environment, rook operator tries to connect with the external endpoint instead of the internal k8s service endpoint.

arttor:
@kayrus for QA, we will also have to set `.Values.objectstore.prysm.user.store` to admin rgw name.

system/cc-ceph/templates/rgw-exporter-user.yaml
```diff
@@ -0,0 +1,23 @@
...
+    metadata: "*"
+    zone: "*"
+    roles: "*"
+    info: "*"
+    amz-cache: "*"
+    bilog: "*"
+    mdlog: "*"
+    datalog: "*"
+    user-policy: "*"
+    ratelimit: "*"
```
kayrus, line 22:
```suggestion
    usage: "read"
    metadata: "read"
    zone: "read"
```

system/cc-ceph/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: cc-ceph
 description: A Helm chart for the Rook / Ceph Objects inside the Storage Clusters
 type: application
-version: 1.1.16
+version: 1.1.17
```
kayrus, line 5:
```suggestion
version: 1.1.19
```

sumitarora2786 (approved):

PR #7745, opened 2025-01-24
[wip][neutron][sentry] adding config map for using new sentry logger

mutax:
This commit adds a new logging config and a switch to the neutron server deployment to use that logging config. This enabled us to switch to the new sentry logger one deployment at a time.


sebageek (approved):
Let's try it out in qa!

PR #7758, opened 2025-01-28
[sap-seeds] Move flavours out of sap-seeds

fwiesel:
Instead of using templating feed them in throught values. This gives us more flexibility in defining the flavors. For starters, we can simply combine flavors by using different sets of value files.
It also open the door to generate these values.


fwiesel:
We can only merge that after having set up the values correctly.

joker-at-work (approved):

dhoeller (approved):
LGTM

PR #7769, opened 2025-01-29
[placement][pxc-db] Add support for PXC galera cluster: part 1

s10:
Add default values for pxc-db chart dependency
Update utils charts to support galera cluster

Add an explicit dbType option to be able to force the usage of service DB.

Without setting pxc_db.enabled=true this PR causes no change in deployment.

### DIFF
--- a/openstack/placement/ci/test-values.yaml
+++ b/openstack/placement/ci/test-values.yaml
@@ -13,11 +13,38 @@ imageVersion: myTag
 
 mariadb:
   enabled: false
+  root_password: topSecret
+  users:
+    placement:
+      password: topSecret
   backup:
     enabled: false
   backup_v2:
     enabled: false
 
+pxc_db:
+  enabled: false
+  users:
+    placement:
+      password: topSecret!
+  system_users:
+    root:
+      password: topSecret!
+    xtrabackup:
+      password: topSecret!
+    monitor:
+      password: topSecret!
+    proxyadmin:
+      password: topSecret!
+    operator:
+      password: topSecret!
+    replication:
+      password: topSecret!
+  backup:
+    s3:
+      secrets:
+        aws_access_key_id: topSecret!
+        aws_secret_access_key: topSecret!
 
 api_db:
   user: a_name
--- a/openstack/placement/templates/_helpers.tpl
+++ b/openstack/placement/templates/_helpers.tpl
@@ -1,4 +1,4 @@
-{{- define "job_name" }}
+{{- define "placement.job_name" }}
   {{- $name := index . 1 }}
   {{- with index . 0 }}
     {{- $all := list (include "utils.proxysql.job_pod_settings" . ) (include "utils.proxysql.volume_mount" . ) (include "utils.proxysql.container" . ) (include "utils.proxysql.volumes" .) (tuple . (dict) | include "utils.snippets.kubernetes_entrypoint_init_container") (include "utils.trust_bundle.volume_mount" . ) (include "utils.trust_bundle.volumes" .) | join "\n" }}
@@ -8,9 +8,10 @@
 {{- end }}
 
 
-{{- define "db_name" }}
-  {{- if .Values.mariadb.enabled -}}
-"{{ .Values.mariadb.name }}-mariadb"
+{{- define "placement.db_name" }}
+  {{- if or (eq .Values.dbType "mariadb") (eq .Values.dbType "pxc-db") }}
+{{- include "utils.db_host" . -}}
+{{/* Use nova-api-mariadb database, if no database is selected explicitly */}}
   {{- else -}}
 "nova-api-mariadb"
   {{- end }}
--- a/openstack/placement/values.yaml
+++ b/openstack/placement/values.yaml
@@ -1,8 +1,11 @@
+---
 # Default values for placement
 # This is a YAML-formatted file.
 # Declare name/value pairs to be passed into your templates.
 # name: value
 global:
+  registryAlternateRegion: other.docker.registry
+  dockerHubMirrorAlternateRegion: other.dockerhub.mirror
   placementApiPortAdmin: '8778'
   placementApiPortInternal: '8778'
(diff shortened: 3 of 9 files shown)


galkindmitrii (approved):
lgtm, `utils` can probably be bumped again in the meantime.

openstack/placement/values.yaml
```diff
@@ -1,8 +1,11 @@
+---
 # Default values for placement
 # This is a YAML-formatted file.
 # Declare name/value pairs to be passed into your templates.
 # name: value
 global:
+  registryAlternateRegion: other.docker.registry
+  dockerHubMirrorAlternateRegion: other.dockerhub.mirror
```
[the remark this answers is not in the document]
s10: These values should've been added to ci/test-values.yaml, they ended up here by mistake...

s10:
@joker-at-work , could you please take a look at this PR again?

openstack/placement/ci/test-values.yaml
```diff
@@ -13,11 +14,38 @@ imageVersion: myTag
 
 mariadb:
   enabled: false
+  root_password: topSecret
+  users:
+    placement:
+      password: topSecret
```
joker-at-work, line 20: How come this is necessary even though `mariadb.enabled` is `false`?

openstack/placement/ci/test-values.yaml
```diff
@@ -13,11 +14,38 @@ imageVersion: myTag
...
+      password: topSecret!
+    operator:
+      password: topSecret!
+    replication:
+      password: topSecret!
+  backup:
+    s3:
+      secrets:
+        aws_access_key_id: topSecret!
+        aws_secret_access_key: topSecret!
```
joker-at-work, line 48: same question here: it's disabled, so why do we need all these values?

openstack/placement/templates/_helpers.tpl
```diff
@@ -8,9 +8,10 @@
 {{- end }}
 
 
-{{- define "db_name" }}
-  {{- if .Values.mariadb.enabled -}}
-"{{ .Values.mariadb.name }}-mariadb"
+{{- define "placement.db_name" }}
+  {{- if or (eq .Values.dbType "mariadb") (eq .Values.dbType "pxc-db") }}
```
joker-at-work, line 12: Why don't we check on `mariadb.enabled` or `pxc_db.enabled` here? Is that for migrating from one to the other where we would need both enabled?

This can have a problem if neither `enabled` is set, but `dbType` is set. Hopefully, we know not to do that. Alternatively, we could set `dbType = "mariadb"` right now, but keep it disabled and check for `enabled` here.

openstack/placement/values.yaml
```diff
@@ -47,11 +48,14 @@ uwsgi:
...
 max_overflow: 5
 
-api_db: # Only used when mariadb.enabled=False to connect to the nova-api-mariadb
+api_db:  # Only used when mariadb.enabled=False to connect to the nova-api-mariadb
   name: placement
   user: placement
   password: null
 
+dbType: ""
+dbName: "placement"
```
joker-at-work, line 57: Do we need this if we set the `databases` and the `name` in the individual DB-specific values dicts?

openstack/placement/values.yaml
```diff
@@ -67,6 +71,32 @@ mariadb:
...
+  databases:
+    - placement
+  users:
+    placement:
+      name: placement
+      grants:
+        - "ALL PRIVILEGES on placement.*"
+  pxc:
+    persistence:
+      size: 10Gi
```
joker-at-work, line 90: Is this service-specific? How would I compute that? Any hint we can leave here for the placement chart maintainers?

s10: This is not service-specific, this specific value is the same as a default value in `mariadb` or new `pxc-db` chart, it's just added here explicitly to make it more transparent.

openstack/placement/values.yaml
s10: `dbName` isn't needed, `dbType` is, because `utils.db_url` and `proxysql` configuration logic depends on it

openstack/placement/templates/_helpers.tpl
s10: The data migration is supposed to be done with a transitional period when both databases are enabled.

The migration workflow ideally is close to the following:
1. `pxc_db.enabled` is set to `true`, after that in the following pipeline run a new cluster is created. It runs in parallel with the existing MariaDB.
2. We run `go-maria-sync` deployment to sync MariaDB data with PXC with the following switch-over dbType=`pxc-db` or a job with mysqldump with dbType=`pxc-db`
3. If everything is ok, then mariadb could be disabled.

Additionally to this, for `placement` it's a bit more complicated, because this chart needs to support the current deployment scheme, where we have `placement` service using `placement` database in `nova-api-mariadb`.

So, basically, we have `dbType=""` for a `placement` database using `nova-api-mariadb`, `dbType=mariadb` for a placement database in `placement-mariadb `, and `dbType=pxc-db` for placement database in `placement-pxc`.

openstack/placement/values.yaml
joker-at-work: `dbType` defaults to `mariadb` anyways, right?

When looking that the values as service-owner, it's confusing to have the db name in 3 places. Hence me trying to reduce "clutter" :) If you say it's necessary, we can keep it.

openstack/placement/templates/_helpers.tpl
joker-at-work: If we go forward with `placement-pxc`, we will make no effort to migrate to `placement-mariadb` first.

joker-at-work: Anyways, it works as is, it just doesn't make much sense to me and could lead to errors if someone just goes and sets it to `mariadb` while we have `mariadb.enabled` being `False`. But I guess we'll catch that in qa soon enough.

joker-at-work (approved):

PR #7770, opened 2025-01-29
introduce workload management support group

auhlig:
This PR introduces the `workload-management` support group to the alertmanager and gatekeeper.


system/gatekeeper-config/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: gatekeeper-config
 description: Config objects for Gatekeeper rules
 type: application
-version: 0.1.0
+version: 0.1.1
```
SuperSandro2000, line 5:
```suggestion
version: 0.1.0
```
we don't do that

SuperSandro2000 (approved):

PR #7779, opened 2025-01-30
[andromeda] Liquid Server pod

ronchi-oss:
Tested on qa-de-3, looks good.

### DIFF
--- a/global/andromeda/templates/etc/_liquid-policy.json.tpl
+++ b/global/andromeda/templates/etc/_liquid-policy.json.tpl
@@ -1,6 +1,8 @@
 {
-  "liquid:get_info": "",
-  "liquid:get_capacity": "",
-  "liquid:get_usage": "",
-  "liquid:set_quota": ""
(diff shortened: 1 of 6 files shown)


global/andromeda/templates/etc/_liquid-policy.json.tpl
```diff
@@ -0,0 +1,6 @@
+{
+  "liquid:get_info": "",
+  "liquid:get_capacity": "",
+  "liquid:get_usage": "",
+  "liquid:set_quota": ""
+}
```
notandy, line 6: I think we should set this to sensible defaults, should be able to c&p from here: https://github.com/sapcc/helm-charts/blob/master/system/liquid-ceph/templates/configmap.yaml#L7

notandy (changes requested):

global/andromeda/templates/etc/_liquid-policy.json.tpl
ronchi-oss: Change applied as per recommendation.

PR #7791, opened 2025-01-31
[rabbitmq-operator] Update rabbitmq uperators

s10:
* Update rabbitmq uperators
    * Update CRDs
    * Add owner-info


galkindmitrii (approved):
note: not currently in use.

PR #7794, opened 2025-01-31
[netapp-harvest-exporter] add snapmirror destinations

Carthaca:
Regular snapmirror metrics do not include the destinations.
I think that is the case, because traditionally snapmirrors are
observed from the target cluster.
But in certain scenarios, e.g. cross-region setups, we want to monitor
from the source cluster, too.

Manually deployed to qa, you may have a look there at the new metric netapp_snapmirror_destination_labels

### DIFF
--- a/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
+++ b/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/maia.yaml
@@ -158,6 +158,9 @@ groups:
     #
     # snampirror relationship
     #
+    # on(filer, destination_volume):
+    #     leads to only looking at volumes on destination side,
+    #     because on source side volume filer and snapmirror filer will not match
 
     - record: netapp_snapmirror_labels:maia
       expr: |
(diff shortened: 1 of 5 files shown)


chuan137:
thanks for the PR, I discussed with Harsha that customers are confused that the `netapp_snapmirror_labels` does not include the snapmirror relationships in the source projects. What do you think we add these metrics to the same metric name instead of using different name. `netapp_snapmirror_labels` will contain snapmirror relationships in both direction, the ones with current project as destination will have more labels, and the ones as source will have fewer. Therefore it's quite easy to tell them apart, without introducing another metrics. I think it can be confusing when users need to remember to check two set of metrics.

Carthaca:
I thought the other way, I think it is confusing to have a different set of labels - but if the customers would like to have it combined in one metric, fine with me :)

chuan137:
I will ask for feedback, there is a meeting with customer this Thursday

Carthaca:
Another perspective: for maia we have anyhow extra recording rules, so joining these two different metrics together could also be done at that place

chuan137 (approved):
lgtm, expecting another pr for maia metrics

kpawar-sap (approved):

PR #7737, opened 2025-01-23
[gatekeeper] add rules for otel

timojohlo:

PR #7805, opened 2025-02-03
GkPodSecurity: restrict otel exception to the relevant clusters

majewsky:
Followup for #7737. @timojohlo Does this look good to you?


timojohlo (approved):
Should be fine for our use-case. Thanks!

PR #7816, opened 2025-02-04
[KMIP] Add disco/record, disco/zone-name in ingress

rajivmucheli:


openstack/kmip/templates/ingress.yaml
```diff
@@ -16,6 +16,8 @@ metadata:
   {{- if .Values.tlsacme }}
     kubernetes.io/tls-acme: "true"
     disco: "true"
+    disco/record: kmip.cc.{{ .Values.global.region }}.{{ .Values.global.tld }}.
+    disco/zone-name: "cc.{{ .Values.global.region }}.{{ .Values.global.tld }}"
```
stefanhipfel, line 20: last may miss a dot at the end.

PR #7817, opened 2025-02-04
prometheus-kubernetes-rules: add PodNotSchedulable alert

majewsky:
This is intended to replace the last remaining alerts based on event-exporter metrics:
- PodSchedulingInsufficientCPU
- PodSchedulingInsufficientMemory

As defined in `prometheus-exporters/event-exporter/alerts/scheduling.alerts`

### DIFF
--- a/prometheus-rules/prometheus-kubernetes-rules/alerts/health.alerts.tpl
+++ b/prometheus-rules/prometheus-kubernetes-rules/alerts/health.alerts.tpl
@@ -149,6 +149,20 @@ groups:
       description: "The pod {{`{{ $labels.namespace }}`}}/{{`{{ $labels.pod }}`}} is not ready for more then 2h."
       summary: "Pod not ready for a long time"
 
+  - alert: PodNotSchedulable
+    expr: max by (namespace, pod) (kube_pod_status_unschedulable > 0) * on (namespace, pod) group_left(label_alert_tier, label_alert_service, label_ccloud_support_group, label_ccloud_service) (max without (uid) (kube_pod_labels))
+    for: 1h
+    labels:
+      tier: {{ include "alertTierLabelOrDefault" .Values.tier }}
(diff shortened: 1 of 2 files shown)


prometheus-rules/prometheus-kubernetes-rules/alerts/health.alerts.tpl
```diff
@@ -149,6 +149,20 @@ groups:
       description: "The pod {{`{{ $labels.namespace }}`}}/{{`{{ $labels.pod }}`}} is not ready for more then 2h."
       summary: "Pod not ready for a long time"
 
+  - alert: PodNotSchedulable
+    expr: max by (namespace, pod) (kube_pod_status_unschedulable > 0) * on (namespace, pod) group_left(label_alert_tier, label_alert_service, label_ccloud_support_group, label_ccloud_service) (max without (uid) (kube_pod_labels))
```
richardtief, line 153: I assume that might not only trigger on image pull issues, but also e.g. insufficient resources? Are we not able to check the Kubernetes events to detect pull issues?

richardtief (commented):
Looks good. One question below

prometheus-rules/prometheus-kubernetes-rules/alerts/health.alerts.tpl
majewsky: > not only trigger on image pull issues

This does not trigger on image pull issues at all. Image pull fails on the kubelet, but we don't get to that point if scheduling is already failing.

> check the Kubernetes events

That's the thing. We are only using the event-exporter for this exact one thing, and we want to get rid of the component in order to reduce maintenance burden.

For what it's worth, this alert will indeed match on more stuff than the old alerts do (e.g. in qa-de-1, I saw a pod that's unschedulable because it's referencing a nonexistent PersistentVolumeClaim). But I think that's overall a good thing.

viennaa (approved):
Bye

PR #7818, opened 2025-02-04
[mariadb] Change default binary log format from MIXED to ROW

s10:
mariadb: Change default binary log format from MIXED to ROW
This will increase the size of binary logs but will make them more consistent and usable for mariadb->galera migration when AUTO_INCREMENT is being used.
Note: the overall size of binary logs and binary log backups will increase after this change

### DIFF
--- a/common/mariadb/CHANGELOG.md
+++ b/common/mariadb/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.36.0 - 2026/05/11
+* default binary log format changed from MIXED to ROW
+* chart version bumped
+
 ## v0.35.1 - 2026/05/06
 * fixed pvc storage_class templating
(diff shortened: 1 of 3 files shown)


galkindmitrii (dismissed):
lgtm, but I think @IvoGoman should have a look

IvoGoman:
This should be fine from `go-maria-sync`-side, as long as this is `RowsEventV1`. I was not yet able to start working on the `RowsEventV2` support.

I faintly remember some issues with enabling [binlog_annotate_row_events](https://mariadb.com/kb/en/replication-and-binary-log-system-variables/#binlog_annotate_row_events) but since this seems to be already enabled by default for quite some time there should be no issue. 

In case it's not the v2 just let me know once you start rolling this out in QA then I could double-check if the replication still works :)

PR #7821, opened 2025-02-04
[nova] Disable compute node identification

CordulaGuder:


joker-at-work:
I don't think we need this anymore.

PR #7822, opened 2025-02-04
keystone: add cloud_masterdata_{admin,viewer} to blocklist_roles

majewsky:
We are planning to add support for these roles in the CBR masterdata API.

I'm not sure if `policy.json` or `policy.yaml` is the relevant place, so I added the roles to both places just to be safe.


rajivmucheli (approved):
policy.yaml is the file. lgtm

stanislav-zaprudskiy (approved):

PR #7824, opened 2025-02-04
[CODEOWNERS] update 

Carthaca:
- manila and netapp exporter: add some DevOps colleagues from Storage Operations
- remove Thomas Graichen

There are more errors, but I can't/don't want to judge those


dhoeller (approved):
LGTM

PR #7825, opened 2025-02-04
[maia] enable honor_timestamps: true in scrape config

notque:
This is considered a potential fix to our out of order timestamps. But the default setting is true, so really, this shouldn't change anything except make it explicit this is in use. 

There are some other options as well, but I feel like we should do each one individually and validate.


viennaa:
Thanks for bringing up this idea. 

I've checked the config of the maia Prometheus and the `honor_timestamps` are set for every job already. This won't fix nor add something.

I'm closing this for now, let's continue this discussion in our ticket.

PR #7827, opened 2025-02-05
[opensearch-logs] global nodepool variable for all kubernetes jobs

Kuckkuck:


system/opensearch-logs/templates/install-ds-templates-job.yaml
```diff
@@ -24,6 +24,15 @@ spec:
         app.kubernetes.io/instance: {{ .Release.Name | quote }}
         helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     spec:
+      affinity:
```
joluc, line 27: We could also reuse it from the [secrets](https://github.wdf.sap.corp/cc/secrets/blob/master/scaleout/s-qa-de-1/values/opensearch-logs.yaml#L343-L350) repo :) But also fine like that

joluc (approved):

PR #7832, opened 2025-02-05
VMFS/NFS Datastore alerts are added.

macitnasif:


viennaa (changes requested):
Thanks for adding this PR.

Bump the patch version please:
https://github.com/sapcc/helm-charts/blob/master/prometheus-rules/prometheus-vmware-rules/Chart.yaml#L4

kevin-fischer (approved):
+1

PR #7849, opened 2025-02-07
[memcached] version bumped to 1.6.36-alpine3.21

businessbean:
- because of a critical bugfix
- bumped memcached version
- chart version bumped


s10 (approved):
The change looks good

PR #7852, opened 2025-02-07
[neutron] add rbac related parts for ovn

toanju:
This adds the following kinds used by ovn:

* ClusterRole
* ClusterRoleBinding
* Role
* RoleBinding
* ServiceAccount

### DIFF
--- a/openstack/neutron/templates/rbac-ovn-db-raft.yaml
+++ b/openstack/neutron/templates/rbac-ovn-db-raft.yaml
@@ -1,0 +1,67 @@
+{{- if ((.Values).ovn).enabled | default false }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+    name: ovnkube-db
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+    name: ovnkube-db
+roleRef:
+    name: ovnkube-db
+    kind: ClusterRole
+    apiGroup: rbac.authorization.k8s.io
+subjects:
+    - kind: ServiceAccount
+      name: ovnkube-db
+      namespace: {{ .Release.Namespace }}
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+    name: ovnkube-db-ep
+roleRef:
+    name: ovnkube-db-ep
+    kind: Role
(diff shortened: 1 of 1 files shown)


openstack/neutron/templates/rbac-ovn-kube-master.yaml
```diff
@@ -0,0 +1,124 @@
...
+      resources:
+          - events
+      verbs: ["create", "patch", "update"]
+    - apiGroups: [""]
+      resources:
+          - nodes/status
+          - pods/status
+          - services/status
+      verbs: [ "patch", "update" ]
+
```
fwiesel, line 113: I suspect, we do not need that, as we do not run the ovnkube-master, as it is only for running OVN as a CNI.

openstack/neutron/templates/rbac-ovn-db-raft.yaml
```diff
@@ -0,0 +1,67 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+    name: ovnkube-db
+rules:
+    - apiGroups: [""]
+      resources:
+          - nodes
+          - namespaces
+      verbs: [ "get", "list", "watch" ]
```
fwiesel, line 44: I suspect we can patch that out: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/d3b10e87f7fffa38fdf4ad52f98bc8ba998df6c2/dist/images/ovnkube.sh#L736-L743,
and then do not need access to the node api either.

openstack/neutron/templates/rbac-ovn-kube-master.yaml
toanju: indeed, had this in earlier, but forgot to remove this

openstack/neutron/templates/rbac-ovn-db-raft.yaml
toanju: for now I'd like to keep it and do as less patching as needed since we should rewrite that with a proper operator anyway

openstack/neutron/templates/rbac-ovn-kube-master.yaml
toanju: removed

fwiesel (approved):

rhalina (approved):

PR #7854, opened 2025-02-07
[rabbitmq] Add credential-updater sidecar

s10:
* Add [user-credential-updater](https://github.com/sapcc/rabbitmq-user-credential-updater) sidecar container
* Use sidecar container for runtime password updates


notque:
Where are we at with this? I need to have this in, and test it, understand how to use it. Thanks.

galkindmitrii (approved):

businessbean (approved):
lgtm

PR #7865, opened 2025-02-10
[limes] campfire - add owner references

VoigtS:
While building the pipeline I realized why we need that.


openstack/limes-campfire/values.yaml
```diff
@@ -5,3 +5,12 @@ campfire:
   image_tag: null # defined in values file
   cpu_request: 20m
   memory_limit: 64Mi
+
+owner-info:
+  support-group: containers
+  service: limes-campfire
```
majewsky, line 11:
```suggestion
  service: limes
```
Otherwise pod alerts will not show up in the limes alert channel.

PR #7872, opened 2025-02-11
[utils] Add missing default value for mem_queue_size audit option

s10:
Add missing default value for mem_queue_size audit option.
In general, values in utils templates should be either `required` or they should have a `default` value set to avoid misconfiguration errors.

### DIFF
--- a/openstack/utils/templates/_ini_sections.tpl
+++ b/openstack/utils/templates/_ini_sections.tpl
@@ -70,7 +70,7 @@ driver = messagingv2
                     {{- include "ini_sections._transport_url" (tuple . $data) }}
                 {{- end }}
             {{- end }}
-mem_queue_size = {{ .Values.audit.mem_queue_size }}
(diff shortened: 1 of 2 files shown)


openstack/utils/templates/_ini_sections.tpl
```diff
@@ -70,7 +70,7 @@ driver = messagingv2
                     {{- include "ini_sections._transport_url" (tuple . $data) }}
                 {{- end }}
             {{- end }}
-mem_queue_size = {{ .Values.audit.mem_queue_size }}
+mem_queue_size = {{ .Values.audit.mem_queue_size | default 1000 }}
```
businessbean, line 73: please add `| int }}` to ensure that Helm is really using that as an integer and not as something else...

businessbean (changes requested):

businessbean (approved):
lgtm

galkindmitrii (approved):

PR #7877, opened 2025-02-11
[manila] add netapp_flexgroup_pools option

Carthaca:
allow to control aggregate selection for flexgroups.
This also means the aggr-list-multiplier will default to 4
(the default of that setting, currently not changeable)
instead of something else that would be automatically set by ONTAP.
Leading to 4 times number of aggregates constituents
(i.e. often 8 in our case).


kpawar-sap (approved):

chuan137 (approved):
lgtm

PR #7878, opened 2025-02-11
[absent-metrics]: add support for passing arguments

databus23:


system/absent-metrics-operator/values.yaml
```diff
@@ -2,3 +2,5 @@ global:
   registry: DEFINED_IN_VALUES_FILE
 
 imageTag: "20240930123552"
+# args:
+# - --keep-labels=app
```
SuperSandro2000, line 6:
```suggestion
# - --keep-labels=app

```

system/absent-metrics-operator/templates/deployment.yaml
```diff
@@ -31,6 +31,10 @@ spec:
         - name: operator
           image: {{ .Values.global.registry }}/absent-metrics-operator:{{ .Values.imageTag }}
           imagePullPolicy: IfNotPresent
+          {{- with .Values.args }}
+          args:
+           {{- toYaml . | nindent 10 }}
+          {{- end }}
```
majewsky, line 37: Does this work if `.Values.args` is not given, and `toYaml` renders just nothing? I recommend rendering with `toJson` instead as a more defensive alternative:
```suggestion
          args: {{ toJson .Values.args }}
```

Then `values.yaml` should also have `args: []` as a default to ensure that the value has the correct type.

databus23: the `{{with}}` stanza does not generate anything if the argument is nil or false.
Thats kind of the best practise according the stuff generated by `helm create [CHART]`:


Excerpt from the default chart [CHART]/templates/deployment.yaml:
```
[...]
      {{- with .Values.volumes }}
      volumes:
        {{- toYaml . | nindent 8 }}
      {{- end }}
      {{- with .Values.nodeSelector }}
      nodeSelector:
        {{- toYaml . | nindent 8 }}
      {{- end }}
      {{- with .Values.affinity }}
      affinity:
        {{- toYaml . | nindent 8 }}
      {{- end }}
      {{- with .Values.tolerations }}
      tolerations:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

databus23: > Does this work if .Values.args is not given, and toYaml renders just nothing?

Quoting:https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with:

> the block after with only executes if the value of PIPELINE is not empty.

PR #7886, opened 2025-02-12
Added VMWare alert: VCClusterMemoryDistrubutionOverVMs

christophrichtersap:
- Added alert for Memory consumption of VMs within Cluster > 90 (48h existence)
- The alerts is fired to the `alert-vmware-info` channel

### DIFF
--- a/prometheus-rules/prometheus-vmware-rules/alerts/vccluster.alerts
+++ b/prometheus-rules/prometheus-vmware-rules/alerts/vccluster.alerts
@@ -117,3 +117,20 @@ groups:
     annotations:
       description: "The HANA-exclusive {{ $labels.vccluster }} cluster DRS configuration is not set to partially automated in the *{{ $labels.datacenter }}* availability zone\nLink to the vCenter: --> <https://{{ $labels.vcenter }}|{{ $labels.vcenter }}>"
       summary: "The HANA-exclusive {{ $labels.vccluster }} cluster DRS configuration is not set to partially automated in the *{{ $labels.datacenter }}* availability zone\nLink to the vCenter: --> <https://{{ $labels.vcenter }}|{{ $labels.vcenter }}>"
+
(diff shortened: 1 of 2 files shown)


prometheus-rules/prometheus-vmware-rules/alerts/vccluster.alerts
```diff
@@ -117,3 +117,24 @@ groups:
...
+  - alert: VCClusterMemoryDistrubutionOverVMs
+    expr: |
+      (
+        sum by (vccluster)(vrops_virtualmachine_memory_consumed_kilobytes{vccluster=~"productionbb.*"}) 
+      unless on (hostsystem) vrops_hostsystem_runtime_maintenancestate{state="inMaintenance"}
+      ) / 
+      ( 
+        sum by (vccluster)(vrops_hostsystem_hardware_memory_size_kilobytes{vccluster=~"productionbb.*"}) unless on (hostsystem) vrops_hostsystem_runtime_maintenancestate{state="inMaintenance"}
+      ) * 100 > 90
+    for: 48h
```
viennaa, line 130: This should be decreased. if there is a gap in the metric within the 48 hours it won't fire. I'd choose a reasonable shorter amount.

viennaa (commented):
Thanks for bringing in the first bedrock alert. Little nit inside.

prometheus-rules/prometheus-vmware-rules/alerts/vccluster.alerts
christophrichtersap: Is this the case when a single metric value is lost or just in case on metric value is below the defined threshold?

now-48h always >= value 

Rahul mentioned the alert should be triggered when the state is persistent for more than 48h, as there is actual no action in place to mitigate the alert...

christophrichtersap:
> g in the first bedrock alert. Little nit inside.

it isn't bedrock alert :D

PR #7888, opened 2025-02-12
gate-keeper: add constraint to restrict the deployment of new releases to monsoon3 namespace

Varsius:


system/gatekeeper-config/templates/constraint-restrict-monsoon3-deployments.yaml
```diff
@@ -0,0 +1,19 @@
...
+  labels:
+    severity: 'error'
+spec:
+  enforcementAction: deny
+  parameters:
+    patterns:
+      {{- if contains "qa-" .Values.global.region }}
+      - "bogus-pattern-for-testing"
+      {{- else }}
+      - "todo"
```
Varsius, line 17: What should be the pattern of the "allowed list"?

system/gatekeeper/templates/constrainttemplate-restrict-monsoon3-deployments.yaml
```diff
@@ -0,0 +1,36 @@
...
+        package restrictnewdeployments
+
+        import future.keywords.in
+
+        iro = input.review.object
+        n := iro.metadata.namespace
+        releaseName := iro.metadata.name
+
+        violation[{"msg": msg}] {
+          n == "monsoon3"
```
majewsky, line 32: I just realized that this needs to apply to the namespace `monsoon3global`. (That's where monsoon3 components for the "global" region are deployed to.)

system/gatekeeper/templates/constrainttemplate-restrict-monsoon3-deployments.yaml
```diff
@@ -0,0 +1,36 @@
...
+          some pattern in input.parameters.patterns
+          not regex.match(pattern, releaseName)
```
majewsky, line 34: This will always fail if you have multiple patterns that are mutually exclusive. For example, consider `input.parameters.patterns = [ "foo", "bar" ]`. This looks like the input `bar` ought to be allowed, but the policy as written will reject is because there is some pattern (`foo`) that does not match the input.

Generally, if you want to express an allowlist in Rego, you will need to go through an additional level of indirection (see e.g. GkPodSecurity as an example of how to do this), but in this case, it's way easier to just leverage the ability of regexes to encode multiple alternatives at once: `input.parameters.patterns = "foo|bar"` (or rather, `input.parameters.patterns = "^(?:foo|bar)$"` to avoid accidental partial matches).

This note and the previous one tie into the question you asked:
> What should be the pattern of the "allowed list"?

Since one regex per namespace is enough, but multiple namespaces must be supported, I suggest structuring the input values as an object mapping namespace names to regexes, e.g.:
```yaml
parameters:
  patterns:
    monsoon3: "^(?:foo|bar)$"
    monsoon3global: "^(?:foo-global|bar-global)$"
```

system/gatekeeper/tests/gatekeeper-policies/fixtures/restrict-monsoon3-deployments/configmap-allowed.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: bogus-pattern-for-testing
+  namespace: monsoon3
+data:
+  test-key: "test-value"
+  another-key: "another-value"
```
majewsky, line 8: I don't want for this policy to match on everything. That will require an excessive level of detail in the object name regexes. We should only match on Helm releases. Please review the existing policy GkOwnerInfoOnHelmReleases for how to match Helm releases, specifically:

- the match expression on the constraint: https://github.com/sapcc/helm-charts/blob/master/system/gatekeeper-config/templates/constraint-owner-info-on-helm-releases.yaml#L12
- the tests for that policy that construct Helm release objects out of a specification file

Helm release objects are always `iro.kind == "Secret"` and are named according to the pattern `sh.helm.release.v1.$NAME.v$VERSION`, so you can match the name to extract the release name without having to parse the full release manifest.

majewsky (changes requested):

system/gatekeeper/templates/constrainttemplate-restrict-monsoon3-namespace.yaml
```diff
@@ -0,0 +1,55 @@
...
+      libs:
+        - |
+          {{ .Files.Get "lib/helm-release.rego" | nindent 10 }}
+      rego: |
+        package restrictnewnamespace
+        import data.lib.helm_release
+        
+        iro := input.review.object
+        n := iro.metadata.namespace
+        release := helm_release.parse_k8s_object(iro, input.parameters.helmManifestParserURL)
```
majewsky, line 34: Parsing is very expensive and will not be needed here. We can match on the secret name only.

majewsky (approved):
I have enumerated existing Helm releases and put regexes in the config accordingly. Also, I downgraded `severity: error` to `severity: debug` in order to not make violations show up until we have confirmed that all is good.

If this looks good to you, please deploy this to prod and check if any unexpected violations come up, including in QA. If all is well, we can set `severity: error` and `enforcementAction: deny` in a subsequent PR.

PR #7898, opened 2025-02-12
[keystone] don't use template for credentials secret and use b64enc

s10:


rajivmucheli (approved):

rajivmucheli:
maybe same for fernet keys ? i just saw this in lint. Could you please raise another PR

PR #7902, opened 2025-02-12
Revert "GkPodSecurity: restrict otel exception to the relevant clusters"

olandr:
This reverts commit c256c1a64f8a0820b5c04785b34b56c8e511e8d9.

Instead of #7891


majewsky (approved):
Context: There was a misunderstanding when I applied this restriction. This deployment does in fact need to go into all clusters.

PR #7905, opened 2025-02-12
[mariadb] maintenance job added

businessbean:
- maintenance job added
  - currently the [analyze table](https://mariadb.com/kb/en/analyze-table/) task is supported
  - the [README.md](README.md#analyzetable) is updated with the new maintenance job details
- chart version bumped

### DIFF
--- a/common/mariadb/CHANGELOG.md
+++ b/common/mariadb/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.16.2 - 2025/02/14
+- maintenance job added
+  - currently the [analyze table](https://mariadb.com/kb/en/analyze-table/) task is supported
+  - the[README.md](README.md#analyzetable) is updated with the new maintenance job details
+- chart version bumped
+
 ## v0.16.1 - 2025/02/18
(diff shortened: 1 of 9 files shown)


joker-at-work:
What's the reason for not enabling analyze table for all tables by default if no explicit table is given? Does it have a disadvantage?

businessbean:
> What's the reason for not enabling analyze table for all tables by default if no explicit table is given? Does it have a disadvantage?

In the past(not the case with recent versions) this operation has put a read lock on the table and depending on the table size it could take minutes/hours to release the lock. Because of that history i want to be careful with enabling that feature. I can build in a `.allTables` option later, but first let us gather some experience in QA and production.

- https://www.percona.com/blog/analyze-table-is-no-longer-a-blocking-operation/
- https://www.percona.com/blog/is-analyze-table-safe-on-a-busy-mysql-database-server/
- https://mariadb.com/kb/en/innodb-limitations/#table-analysis
- https://jira.mariadb.org/browse/MDEV-7901

dhoeller (commented):
LGTM

PR #7918, opened 2025-02-13
add rgw custom config to ceph chart

arttor:
supprt helm templates for rgw instace config overrides.
See: https://rook.io/docs/rook/v1.16/CRDs/Object-Storage/ceph-object-store-crd/?h=rgwcon#advanced-configuration

```yaml
spec:
  gateway:
    # ...
    rgwConfig:
      debug_rgw: "10" # int
      # debug-rgw: "20" # equivalent config keys can have dashes or underscores
      rgw_s3_auth_use_ldap: "true" # bool
    rgwCommandFlags:
      rgw_dmclock_auth_res: "100.0" # float
      rgw_d4n_l1_datacache_persistent_path: /var/log/rook/rgwd4ncache # string
      rgw_d4n_address: "127.0.0.1:6379" # IP string
```

### DIFF
--- a/system/cc-ceph/Chart.yaml
+++ b/system/cc-ceph/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: cc-ceph
 description: A Helm chart for the Rook / Ceph Objects inside the Storage Clusters
 type: application
-version: 1.1.36
+version: 1.1.37
 appVersion: "1.16.2"
 dependencies:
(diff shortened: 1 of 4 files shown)


sumitarora2786 (approved):

kayrus (approved):
I'd like to see also the transition from a `templates/ceph-config-override.yaml` to a new CRD

PR #7921, opened 2025-02-14
[designate] Add rate limiting

misamoylov:


openstack/designate/values.yaml
```diff
@@ -439,3 +439,60 @@ vpa:
...
+rate_limit:
+  enabled: false
+  rate_limit_by: target_project_id
+  max_sleep_time_seconds: 0
+  log_sleep_time_seconds: 10
+  backend_timeout_seconds: 1
+  whitelist:
+    - Default/service
+    - Default/admin
+    - ccadmin/cloud_admin
```
galkindmitrii, line 453: I think we need to also add ccadmin/master and projects used by locust/tempest

openstack/designate/Chart.yaml
```diff
@@ -42,3 +42,8 @@ dependencies:
   - name: linkerd-support
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
     version: 1.1.0
+  - name: redis
+    alias: api-ratelimit-redis
+    repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
+    version: 1.5.3
```
galkindmitrii, line 48: There is a new redis chart version 2.0.0 which supports automatic password generation.
Please update accordingly. See example: https://github.com/sapcc/helm-charts/pull/7971/files

galkindmitrii (changes requested):
The rate limiting is not enabled for the API, it needs `{{- include "rate_limit_pipe" . }}`
Please check https://github.com/sapcc/helm-charts/commit/9254ec4707d5c2f74573a696bb8ee587058f7e90#diff-60ecb59ceb7281a5a75f5d46b395df488e87861c0890e11caf371f09240043e2R26

galkindmitrii:
Still missing {{- include "rate_limit_pipe" . }} in API v2: https://github.com/sapcc/helm-charts/blob/master/openstack/designate/templates/etc/_api-paste.ini.tpl#L16-L19

galkindmitrii (approved):

mutax (approved):

PR #7922, opened 2025-02-14
add ingress annotations to fix policy check

artherd42:


richardtief (changes requested):
Please bump the chart version then it's good to go.

richardtief (approved):
Thanks.

PR #7923, opened 2025-02-14
[andromeda] liquid service user

ronchi-oss:


ronchi-oss:
will be recreated without the extra user.

PR #7926, opened 2025-02-14
feat(opensearch-logs-query-exporter): add secondary auth fallback mechanism

joluc:
Add two more `env` for failover credentials used in https://github.com/sapcc/prometheus-es-exporter/pull/2


timojohlo (approved):
lgtm

PR #7934, opened 2025-02-14
[HSM] route all alerts to hsm slack channel

rajivmucheli:


richardtief (changes requested):
Looks good. We usually bump the version in `Chart.yaml`. Then it's good to go.

richardtief (approved):

PR #7942, opened 2025-02-17
alertmanager: fix bedrock slack route

richardtief:
`msg="skipping alertmanagerconfig" error="route[35]: json: cannot unmarshal bool into Go struct field Matcher.matchers.value of type string"`


richardtief:
fyi @christophrichtersap 

Please adjust your template if necessary.

PR #7946, opened 2025-02-17
Add choice to upload backup to AWS

benjamin-sap:
Currently the backups of a MariaDB are pushed to AWS and Swift.

This PR adds the possibility to choose, if a backup should be pushed to AWS or not.
It defaults to `enabled: true`, to not introduce a breaking change for other users of this chart.


stefanhipfel:
lgtm

PR #7950, opened 2025-02-17
chore(elektra): increase resource limits

andypf:
This PR increases the resource limits to address the ElektraPredictHighNumberOfDatabaseConnections - High alert and ensure stable application performance under increased workload.

## Changes

**From:**
```yaml
resources:
  limits:
    memory: 128Mi
    cpu: 100m
  requests:
    memory: 64Mi
    cpu: 50m
```

**To:**
```yaml
  resources:
    limits:
      memory: 256Mi
      cpu: 500m
    requests:
      memory: 128Mi
      cpu: 250m
```

### DIFF
--- a/openstack/elektra/values.yaml
+++ b/openstack/elektra/values.yaml
@@ -87,7 +87,6 @@ pgbackup:
     service: elektra
 
 pgmetrics:
-  isPostgresNG: true
   db_name: monsoon-dashboard_production
   alerts:
     support_group: containers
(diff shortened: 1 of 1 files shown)


majewsky (changes requested):
Why do you believe this to be necessary? These pods currently consume between 10-20 MiB RAM and between 2-6 milli CPU in all prod regions. I find it very likely that you're misinterpreting the cause of the alert.

andypf:
not needed!

PR #7959, opened 2025-02-18
[neutron] Use IP instead of DNS for OVN DB connection

fwiesel:
OVS only can resolve DNS names, when the optional unbound
package is installed, as it wants to do its own DNS resolution
for some reason.


toanju (approved):

joker-at-work (changes requested):
The required doesn't trigger, because there's a default in the chart. Additionally, an empty list (the current default) is also accepted.

joker-at-work (approved):

rhalina (approved):

PR #7960, opened 2025-02-18
[unbound] introducing the DNSUnbound{1,2}LowTraffic prometheus alert

vssldmtrv:
It will fire if the traffic in the past minute is less than 10% of the 24h average.

### DIFF
--- a/system/unbound/templates/prometheus-alerts.yaml
+++ b/system/unbound/templates/prometheus-alerts.yaml
@@ -45,6 +45,22 @@ spec:
         description: 'DNS {{ $.Values.unbound.name }} recursor is down.'
         summary: DNS {{ $.Values.unbound.name }} recursor is down. DNS resolution might be handled by another region.
 
+    - alert: Dns{{ $.Values.unbound.name | title }}LowTraffic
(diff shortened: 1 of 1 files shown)


system/unbound/templates/prometheus-alerts.yaml
```diff
@@ -45,6 +45,22 @@ spec:
         description: 'DNS {{ $.Values.unbound.name }} recursor is down.'
         summary: DNS {{ $.Values.unbound.name }} recursor is down. DNS resolution might be handled by another region.
 
+    - alert: Dns{{ $.Values.unbound.name | title }}LowTraffic
+      expr: avg(rate(unbound_queries_total{app="unbound2", namespace="dns-recursor"}[1m]))/avg(rate(unbound_queries_total{app="unbound2", namespace="dns-recursor"}[24h])) < 0.10
```
galkindmitrii, line 49: Looks like the app name is hardcoded to `unbound2`

vssldmtrv: right.. and fixed

mbrowny (approved):

galkindmitrii (approved):

PR #7964, opened 2025-02-18
[glance][mariadb] use ROW binlog format

s10:
* use ROW binlog format instead of MIXED

This should help to avoid getting RETURNING statements in binary logs


rajivmucheli (approved):

businessbean (approved):
lgtm

PR #7966, opened 2025-02-18
common/redis: cleanup clutter, add password autogeneration on first startup

majewsky:
Most of the changes are as described in the `Chart.yaml` diff.

Removing the label clutter from the metrics deployment podspec cannot be done as an in-place edit (in Deployment objects, `.spec.selector.matchLabels` is immutable for good reason), so this renames the deployment from "redis-metrics" to "redis-exporter" to allow the cleanup to proceed.

The password autogeneration capability is mostly copied from the respective capability in the `common/postgresql-ng` chart. The name pattern is slightly different in case users want to deploy multiple separate Redis instances (i.e., multiple instances of this chart) inside a single application chart.

### DIFF
--- a/common/redis/bin/init-generate-secrets.sh
+++ b/common/redis/bin/init-generate-secrets.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env ash
+# shellcheck shell=ash
+# shellcheck disable=SC3010
+
+# This is the entrypoint script for the "generate-secrets" init container of the redis pod.
+
+set -eou pipefail
+[[ ${DEBUG:-} != false ]] && set -x
+
+for USER in ${USERS:-}; do
+  SECRET="${DEPLOYMENT_NAME}-user-$USER"
+
+  # if we already have a secret, we can stop here
+  if [[ "$(kubectl get secrets "$SECRET" --ignore-not-found)" != "" ]]; then
+    continue
+  fi
+
+  # create new secret with randomly generated password
+  # NOTE: make sure that the generated password contains no newline
+  echo -n "
+    apiVersion: v1
+    kind: Secret
+    metadata:
+      name: $SECRET
+      ownerReferences:
+        - apiVersion: apps/v1
+          blockOwnerDeletion: true
+          kind: Deployment
+          name: $DEPLOYMENT_NAME
+          uid: $(kubectl get deployment "$DEPLOYMENT_NAME" -o jsonpath='{.metadata.uid}')
+    data:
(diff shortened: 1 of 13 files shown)


common/redis/bin/init-generate-secrets.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env ash
+# shellcheck shell=ash
+# shellcheck disable=SC3010
+
+# This is the entrypoint script for the "generate-secrets" init container of the postgres pod.
```
SuperSandro2000, line 5: We are not in postgres 😉

common/redis/templates/deployment.yaml
```diff
@@ -32,6 +32,20 @@ spec:
         checksum/secrets: {{ include (print .Template.BasePath "/secrets.yaml") . | sha256sum }}
         {{- end }}
     spec:
+      {{- if not .Values.redisPassword }}
+      initContainers:
+      - name: generate-secrets
+        image: {{ $registry }}/shared-app-images/alpine-kubectl:3.21-latest
```
SuperSandro2000, line 38: I would use latest-latest here

common/redis/templates/rbac.yaml
```diff
@@ -0,0 +1,49 @@
...
+  # This service account is used by the redis pod itself.
+  name: {{ $fullname }}
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ $fullname }}-secrets
+rules:
+  # The init container of the postgres pods generates the
```
SuperSandro2000, line 16: more copy paste errors

common/redis/templates/rbac.yaml
```diff
@@ -0,0 +1,49 @@
...
+    resourceNames:
+      - {{ $fullname }}-user-default
+
+  # "create" permission cannot be restricted to specific resourceNames
+  # Ref: <https://stackoverflow.com/a/65203104>
+  - apiGroups: [ "" ]
+    resources: [ secrets ]
+    verbs:     [ create ]
+
+  # need the UID of the postgres deployment to render an owner reference
```
SuperSandro2000, line 30: more

SuperSandro2000 (changes requested):

common/redis/templates/deployment.yaml
majewsky: This is copied from postgresql-ng. Should we switch to latest-latest over there as well?

SuperSandro2000 (approved):

PR #7975, opened 2025-02-18
[ironic] move dhparams to secrets

sandzwerg:
The dhparams were missed when we moved the secrets from configmaps to
secrets. This should make them work with the new secrets-injector. I
also renamed the nginx config from default.conf to the default name
nginx.conf. The emptydir is needed to be able to mount the secret and
possible other configs as well, even when we currently have no other
configs.


BerndKue (approved):

fwiesel (approved):

sandzwerg:
OK, even if that does work it is not enough.
```
k describe cm ironic-console | grep -n kvv
17:vault+kvv2:///secrets/qa-de-2/ironic/console/default/ssl_dhparam
68:        secure_link_md5 "$secure_link_expires$1 vault+kvv2:///secrets/qa-de-2/ironic/console/default/password";
k describe cm ironic-conductor-testing-etc | grep -n kvv
131:url_auth_digest_secret = vault+kvv2:///secrets/qa-de-2/ironic/console/default/password
```

So more secrets were missed.

BerndKue (approved):

BerndKue (approved):

fwiesel (approved):

PR #7982, opened 2025-02-19
[manila] rework share type seeding

chuan137:
This PR moves customisation of share type's extra_specs to value files.
We merge `share_type_extra_specs` into `share_types.*.extra_specs`, with latter takes precedence.
The share type name is used as `share_backend_name` if otherwise not specified.
We can also set bool value to any item in the share type list,
but then default `share_type_extra_specs` is used, and  any customisation would be lost.

### DIFF
--- a/openstack/manila/templates/type-seeds.yaml
+++ b/openstack/manila/templates/type-seeds.yaml
@@ -15,76 +15,22 @@ spec:
   requires:
   - monsoon3/manila-seed
   share_types:
-  - name: default
-    is_public: true
-    description: "High Performance"
-    specs:
-    {{- include "manila_type_seed.specs" . | indent 6 }}
-    extra_specs:
-      share_backend_name: "netapp-multi"
-      provisioning:max_share_size: "32768"
-      provisioning:max_share_extend_size: "32768" # keep this in sync with value "max_asset_sizes" of "- asset_type: 'nfs-shares.*'" in openstack/castellum/templates/configmap.yaml
-    {{- include "manila_type_seed.extra_specs" . | indent 6 }}
-  - name: integration
-    is_public: false
-    specs:
-    {{- include "manila_type_seed.specs" . | indent 6 }}
-    extra_specs:
-      share_backend_name: "integration"
-      provisioning:max_share_size: "32768"
-      provisioning:max_share_extend_size: "32768" # keep this in sync with value "max_asset_sizes" of "- asset_type: 'nfs-shares.*'" in openstack/castellum/templates/configmap.yaml
-    {{- include "manila_type_seed.extra_specs" . | indent 6 }}
   {{- range $shareTypeName, $shareTypeValues := .Values.share_types }}
-    {{- $enabled := false }}
-    {{- $old_style := false }}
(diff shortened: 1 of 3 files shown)


openstack/manila/templates/type-seeds.yaml
```diff
@@ -15,77 +15,23 @@ spec:
...
-          {{- else }}
-      provisioning:max_share_size: "65536"
-      provisioning:max_share_extend_size: "65536" # keep this in sync with value "max_asset_sizes" of "- asset_type: 'nfs-shares.*'" in openstack/castellum/templates/configmap.yaml
-          {{- end }}
-        {{- else }}
-      provisioning:max_share_size: "65536"
-      provisioning:max_share_extend_size: "65536" # keep this in sync with value "max_asset_sizes" of "- asset_type: 'nfs-shares.*'" in openstack/castellum/templates/configmap.yaml
-        {{- end }}
-      {{- end }}
+    {{ $extraSpecs :=  $.Values.share_type_extra_specs | merge (dict "share_backend_name" $shareTypeName) | merge (default (dict) $shareTypeValues.extra_specs) }}
```
kpawar-sap, line 29: this will take 32768 instead of 65536, correct ?

openstack/manila/values.yaml
```diff
@@ -687,13 +687,46 @@ seeds:
...
+    extra_specs:
+      share_backend_name: "netapp-multi"
+  integration:
+    enabled: true
+  standard:
+    enabled: false
+    description: "Standard"
+    extra_specs:
+      provisioning:max_share_size: "65536"
+      provisioning:max_share_extend_size: "65536"
```
kpawar-sap, line 719: can you add comment similar to line  703

openstack/manila/templates/type-seeds.yaml
chuan137: first argument takes precedence in `merge` function, so values defined in `.Values.share_types.*.extra_specs` will be rendered

kpawar-sap (approved):

chuan137:
we get warning with current implementation
> coalesce.go:286: warning: cannot overwrite table with non table for manila.share_types.hypervisor_storage (map[enabled:false extra_specs:map[provisioning:max_share_size:40960 provisioning:min_share_size:2048]])

Carthaca:
> we get warning with current implementation
> 
> > coalesce.go:286: warning: cannot overwrite table with non table for manila.share_types.hypervisor_storage (map[enabled:false extra_specs:map[provisioning:max_share_size:40960 provisioning:min_share_size:2048]])

Do you mean that the PR here will fix this or is this meant as a hint that the PR needs some more work?

chuan137:
> > we get warning with current implementation
> > > coalesce.go:286: warning: cannot overwrite table with non table for manila.share_types.hypervisor_storage (map[enabled:false extra_specs:map[provisioning:max_share_size:40960 provisioning:min_share_size:2048]])
> 
> Do you mean that the PR here will fix this or is this meant as a hint that the PR needs some more work?

The messages are copied from concourse pipeline, PR here will fix it

PR #7983, opened 2025-02-19
[Barbican] adapt sql_connection for openstack dalmatian

rajivmucheli:
https://docs.openstack.org/releasenotes/barbican/2024.1.html

### DIFF
--- a/openstack/barbican/templates/etc/_secrets.conf.tpl
+++ b/openstack/barbican/templates/etc/_secrets.conf.tpl
@@ -1,16 +1,15 @@
 [DEFAULT]
-sql_connection = {{ include "db_url_mysql" . }}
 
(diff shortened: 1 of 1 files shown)


openstack/barbican/templates/etc/_secrets.conf.tpl
```diff
@@ -1,17 +1,16 @@
 [DEFAULT]
-sql_connection = {{ include "db_url_mysql" . }}
 
 {{ include "ini_sections.default_transport_url" . }}
 
+[database]
+sql_connection = {{ include "db_url_mysql" . }}
```
s10, line 6: According to the release notes, the key should be `connection`, not `sql_connection`

s10 (changes requested):

s10 (approved):

PR #7984, opened 2025-02-19
(chore) bump image version

artherd42:


prometheus-exporters/cloudprober/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: cloudprober
-version: 0.13.0
+version: 0.13.9
```
viennaa, line 3: Since this chart is doing more than just pulling and deploying the image and there are other things, which could be subject to change in the future, I don't want to pin the chart version to the cloudprober version, since this will differ inevitably.

```suggestion
version: 0.13.1
```

viennaa (changes requested):
Thanks for housekeeping, just a small note to the chart version below

viennaa (approved):

PR #7986, opened 2025-02-19
[mariadb] fix maintenance cronjob resource names

s10:
* fix names of the maintenance cronjob resources
  * make names unique
* chart version bumped

### DIFF
--- a/common/mariadb/templates/configmap-mariadb-maintenance.yaml
+++ b/common/mariadb/templates/configmap-mariadb-maintenance.yaml
@@ -4,7 +4,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   namespace: {{ $.Release.Namespace }}
-  name: mariadb-cronjob-maintenance
+  name: mariadb-{{ .Values.name }}-cronjob-maintenance
   labels:
     {{- include "mariadb.labels" (list $ "version" "mariadb" "cm" "maintenance") | indent 4 }}
 data:
--- a/common/mariadb/CHANGELOG.md
+++ b/common/mariadb/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## v0.16.3 - 2025/02/19
+* fix names of the maintenance cronjob resources
+  * make cm names unique
+* chart version bumped
+
 ## v0.16.2 - 2025/02/14
 - maintenance job added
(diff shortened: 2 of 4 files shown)


businessbean:
I opt against that change because of https://github.com/kubernetes/kubernetes/issues/89657

common/mariadb/templates/configmap-mariadb-maintenance.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   namespace: {{ $.Release.Namespace }}
-  name: mariadb-cronjob-maintenance
+  name: mariadb-{{ .Values.name }}-cronjob-maintenance
```
businessbean, line 7:
```
{{ include "fullName" . }}-maint
```
instead of
```
mariadb-{{ .Values.name }}-cronjob-maintenance
```

s10:
> I opt against that change because of [kubernetes/kubernetes#89657](https://github.com/kubernetes/kubernetes/issues/89657)

This relates to `Job`, not a `CronJob`

businessbean (changes requested):
The cronjob name has to be randomized. The configmap name should be updated as described

common/mariadb/templates/configmap-mariadb-maintenance.yaml
s10: I'd prefer to keep consistency here. We have the following configmaps and secrets:
`mariadb-{{.Values.name}}-etc` cm
`mariadb-{{.Values.name}}-client` secret
`mariadb-{{.Values.name}}` secret

And temporary ones:
`{{ .Values.name }}-initdb` secret (hook)
`{{ .Values.name }}-configuredb-bin` cm (hook)

businessbean: Ok, then back to `mariadb-{{ .Values.name }}-cronjob-maintenance`

s10:
> The cronjob name has to be randomized. The configmap name should be updated as described

No, technically they do not have to be randomized, and we don't randomize their name in any other helm chart, because this is only required for `Jobs`, which are immutable.
Regarding the configmap name, I've left a comment about naming consistency.

businessbean (approved):
lgtm

PR #7989, opened 2025-02-19
[mariadb] use custom entrypoint script for mariadb deployment

s10:
* use custom entrypoint script for mariadb deployment
  * remove non-optional `healthcheck` user creation
* chart version bumped

### DIFF
--- a/common/mariadb/CHANGELOG.md
+++ b/common/mariadb/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.17.0 - 2025/02/27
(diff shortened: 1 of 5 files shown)


galkindmitrii (approved):
just please keep in mind that this might need attention as we update mariadb version

businessbean:
> just please keep in mind that this might need attention as we update mariadb version

That's the drawback of that workaround. The other option would be to build our own images that has it's own drawbacks. Let's use the custom entrypoint.sh for now and reevaluate it while we are upgrading to the next major version.

businessbean (approved):
lgtm

PR #8008, opened 2025-02-20
[nova] don't use rabbitmq._transport_url function for cell1

s10:
* don't use rabbitmq._transport_url function for cell1
* add simple cell1_db_path and cell1_transport_url functions to make cell configuration templating more consistent


joker-at-work:
Can you give any reason for your change, please? I do not know why we have to do this.

s10:
> Can you give any reason for your change, please? I do not know why we have to do this.
> 
> Edit: Best put it in the commit-message, too.

`nova` chart currently uses `utils.rabbitmq_url` for cell2 transport url templating, but `rabbitmq._transport_url` for cell1, which looks inconsistent.

We'd prefer to remove the `rabbitmq._transport_url` and `rabbitmq.transport_url` functions from the rabbitmq helm chart, and the `nova` chart is the last remaining place this function is being used. As mentioned in https://github.com/sapcc/helm-charts/pull/6361/commits/2b36cbbe0ab21d8ac7d7ec943cf2a7a9fc61de76, it would be better not to have unnecessary duplication of this code.

openstack/nova/templates/cells-secrets.yaml
```diff
@@ -1,7 +1,6 @@
 {{- $envAll := . }}
-{{- $transport_url := tuple . .Values.rabbitmq | include "rabbitmq._transport_url" }}
-{{- $database_connection := tuple . .Values.dbName .Values.dbUser (default .Values.dbPassword .Values.global.dbPassword) | include "db_url_mysql" }}
-{{ tuple $envAll "cell1" $transport_url $database_connection | include "nova.cell_secret" }}
+---
+{{ tuple $envAll .Values.cell1.name (include "cell1_transport_url" .) (include "cell1_db_path" .) | include "nova.cell_secret" }}
```
joker-at-work, line 3: Just put back the `cell1` string here instead of `.Values.cell1.name`.

joker-at-work (changes requested):
Thank your for pointing out the motivation for this change.

joker-at-work (approved):

fwiesel (approved):

PR #8016, opened 2025-02-21
(feat) federate netapp metrics with domain_id!=""

artherd42:
as requested by Chuan


viennaa (approved):
Technically this is ok, but we need to clarify the reasoning first, before merging.

viennaa:
clarified with @notque @chuan137 
This federation is not needed.

There will be a dedicated netapp exporter in the source region which takes the global metrics and enriches them as needed to provide the necessary customer metrics.

PR #8022, opened 2025-02-21
[mariadb] Fix maria_backup_status alerts

s10:
Fix backup-v2 maria_backup_status alerts.

There never were `release` labels on a backup-v2 deployment.

`app_kubernetes_io_instance` should be used, same as for mysqld-exporter alerts.

### DIFF
--- a/common/mariadb/templates/alerts/_backup-v2.alerts.tpl
+++ b/common/mariadb/templates/alerts/_backup-v2.alerts.tpl
@@ -1,7 +1,7 @@
 - name: backup-v2.alerts
   rules:
   - alert: {{ include "alerts.service" . | title }}MariaDatabaseFullBackupMissing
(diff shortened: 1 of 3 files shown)


common/mariadb/templates/alerts/_backup-v2.alerts.tpl
```diff
@@ -1,7 +1,7 @@
 - name: backup-v2.alerts
   rules:
   - alert: {{ include "alerts.service" . | title }}MariaDatabaseFullBackupMissing
-    expr: maria_backup_status{kind="full_backup", release={{ include "alerts.service" . | quote }} } != 1
+    expr: maria_backup_status{kind="full_backup", app_kubernetes_io_instance=~"{{ include "fullName" . }}" } != 1
```
s10, line 4: currently we upload these backups only to swift, without aws, so maybe we should add swift label filter here

businessbean (approved):
lgtm

PR #8027, opened 2025-02-21
[mariadb] Use init-db job with mysql client instead of kubectl exec

s10:
Replace kubectl exec job with mysql cli job applying init.sql when needed

Both new and old jobs in fact depend on my.cnf being correct, so kubectl exec doesn't prevent potential lockout situations, when the root password is being changed alongside database users.

Additionally, the mariadb chart originally didn't support root password update.

This change also sets `initdb_job` default value to true, because it's set to this value in all dependent helm charts anyway.

Support for `custom_initdb_secret` has been removed, because there were no registered use of this option.


fwiesel:
I think, there is a small problem: You need for that working credentials as the job runs on another host, or I am missing something?

The idea behind `kubectl exec` was, that you can use a locally trusted user without credentials.

As I read it, right now it would be using the new password after the [helm-upgrade/installation](https://github.com/sapcc/helm-charts/pull/8027/files#diff-e6978166b8b2657dee1ff4d680d247d0d209aea5db492a5d48079fffbabc98a4R13), and therefore not be able to change the password of the root user, as it is not valid yet.

s10:
> I think, there is a small problem: You need for that working credentials as the job runs on another host, or I am missing something?
> 
> The idea behind `kubectl exec` was, that you can use a locally trusted user without credentials.
> 
> As I read it, right now it would be using the new password after the [helm-upgrade/installation](https://github.com/sapcc/helm-charts/pull/8027/files#diff-e6978166b8b2657dee1ff4d680d247d0d209aea5db492a5d48079fffbabc98a4R13), and therefore not be able to change the password of the root user, as it is not valid yet.

This change indeed depends on the working root user password, which should be constantly reconciled and not 100% depend on the fact of a helm run, I address that in https://github.com/sapcc/helm-charts/pull/8051

Currently, an in-place solution has several flaws:
1) We can't change and are not changing the root user password at all
2) It anyway depends on the root user password, but not as is provided via env in the job, but as it's written in `/root/.my.cnf` in the pod, with the difference that it uses a unix socket connection instead of a tcp socket.

PR #8028, opened 2025-02-21
[mariadb] set default ignore_db_dirs value

s10:
Add "lost+found" directory to the ignored list to avoid it being displayed as a database.


businessbean (approved):
lgtm

PR #8032, opened 2025-02-24
[metal] removing MetalRedfishCertWarning for now to avoid alert flood…

BerndKue:
…ing until fixed


sandzwerg (approved):
We could have put them to info for now so it's easier to increase the priority later but that also works.

BerndKue:
We already have an info alert 10 days before expiration. That's why I deleted the warning alert. We can easily copy the info alert later and change two parameters and we are back in the game.

PR #8035, opened 2025-02-24
chore(secret): autogenerate smtp cred

ArtieReus:


majewsky:
@VoigtS Can you please align with Arturo? We should be using Campfire here to avoid double work.

PR #8041, opened 2025-02-24
[rabbitmq] container probes enhanced

businessbean:
- simple network checks for the liveness probe
- advanced API status checks for the readiness probe
- chart version bumped


common/rabbitmq/scripts/readiness.sh
```diff
@@ -0,0 +1,18 @@
...
+checkrabbitmq 'check_port_listener {{ $.Values.ports.public | int }}'
+checkrabbitmq 'check_port_listener {{ $.Values.ports.management | int }}'
+# Health check that exits with a non-zero code if target node does not have an active listener for given protocol
+checkrabbitmq 'check_protocol_listener http'
+checkrabbitmq 'check_protocol_listener clustering'
+checkrabbitmq 'check_protocol_listener amqp'
+# Health check that checks if all vhosts are running in the target node
+checkrabbitmq check_virtual_hosts
+# query RabbitMQ version
+checkrabbitmq server_version
```
bashar-alkhateeb, line 18: I would add a check to be sure that the users are configures and RabbitMQ is accessible, before routing traffic to rabbitmq, because if i am not mistaking openstack services will tolerant connection timeout but not wrong authentication 
and because our start script configures the usernames and sets the password while start up it could may take some time between the Rabbitmq is up and the usernames are there and we need to avoid getting such a situation where the rabbitmq reports bad credentials 
you can check credential using :
`rabbitmqctl authenticate_user`

bashar-alkhateeb (changes requested):

common/rabbitmq/scripts/readiness.sh
businessbean: Hi Bashar, i have thought that `ping` was enough to prove that.

`rabbitmqctl ping`
`Checks that the node OS process is up, registered with EPMD and CLI tools can authenticate with it`

businessbean: If we implement the password rotation `rabbitmqctl authenticate_user` can lead to service outages. I would skip that check for now and think about it after we have the rotation feature implemented

bashar-alkhateeb: If i am not mistaking it checks that the handschake works but it will not check that the authentication for a specific user is set, which we need for our openstack services, so we need to direct traffic only when defined users in the wrapper chart are configured and ready

bashar-alkhateeb: why should `rabbitmqctl authenticate_user` cause problems for passowrd rotation ? 
it must use the users from the same secret as the pod ,, and if the pod is not configured with the new secrets it must be set 'not ready'

businessbean: - the pod starts and the [credentials will be applied](https://github.com/sapcc/helm-charts/blob/e17f0a41c2e164284f16311ac3abd0053722cd79/common/rabbitmq/templates/bin/_rabbitmq-start.tpl#L35)
- the readiness check will succeed because the credentials are matching
- later the `secrets-injector` will update the secrets and that will cause the readiness check to use the new credentials
- the service will be down because the pod will not be ready because the auth check will fail

bashar-alkhateeb: is it possible to use the secret of the service/wrapper-chart for the readiness check ?

businessbean: I have added to admin credential check after we have clarified that the deadlock cannot happen with the current code because the credentials are only set at startup

PR #8046, opened 2025-02-25
[billing] fix billing_test group assignment (srs #485)

wagnerd3:


openstack/billing/templates/seeds.yaml
```diff
@@ -66,7 +66,7 @@ spec:
       description: 'project for used for testing the billing service'
       # Role provisioned from CAM to enable takeover of seeded project by billing admins
       role_assignments:
-      - group: CCADMIN_BILLING_ADMIN
+      - group: CCADMIN_BILLING_ADMIN@ccadmin
```
SuperSandro2000, line 69: Do we need to do this in line 91, too?

wagnerd3: no, because we are in `ccadmin` domain there, to which the group belongs. Stefan explained me that he omits the `@domain` in these cases for fast reading.

majewsky (approved):

PR #8065, opened 2025-02-26
[limes] Add commitment mail template

VoigtS:
Adds the commitment mail template to expiring and confirmed commitments.

### DIFF
--- a/openstack/limes/files/commitment_mail.tpl
+++ b/openstack/limes/files/commitment_mail.tpl
@@ -1,0 +1,63 @@
+<!doctype html>
+<html lang="en" xmlns="http://www.w3.org/1999/xhtml">
+    <head>
+        <meta xmlns:v="urn:schemas-microsoft-com:vml">
+        <meta xmlns:o="urn:schemas-microsoft-com:office:office">
+        <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1">
+        <style>
+            table {
+              border-collapse: collapse;
+              width: 100%;
+            }
+            
+            th, td {
+              border: 1px solid;
+              text-align: left;
+              padding: 8px;
+            }
+            </style>
+    </head>
+    <body>
+        <p><b>Dear SAP Cloud Infrastructure user</b>,</p>
+        <p>the following commitments are about to {{ .condition }} for the
+            following project:</p>
+        <p>
+            <ul>
+                <li>Region: {{ .region }}</li>
+                <li>Domain: {{"{{"}}  .DomainName {{"}}"}}</li>
+                <li>Project: {{"{{"}} .ProjectName {{"}}"}}</li>
+            </ul>
+        </p>
+
+        <div style="overflow-x: auto;">
+            <table>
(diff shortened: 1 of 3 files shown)


openstack/limes/files/commitment_mail.tpl
```diff
@@ -0,0 +1,63 @@
...
+            th, td {
+              border: 1px solid;
+              text-align: left;
+              padding: 8px;
+            }
+            </style>
+    </head>
+    <body>
+        <p><b>Dear SAP Cloud Infrastructure user</b>,</p>
+        <p>the following commitments are about to {{ .condition }} for the
```
majewsky, line 23: If condition is "be confirmed", then this will say that commitments are about to be confirmed, which is not accurate. They are already confirmed.
```suggestion
        <p>the following commitments are {{ .condition }} for the
```

And then say either "about to expire" or "confirmed and ready to use".

openstack/limes/templates/configmap.yaml
```diff
@@ -11,6 +11,15 @@ data:
...
     {{- end }}
 {{ toYaml .Values.limes.clusters.ccloud | indent 4 }}
+    mail_notifications:
+      endpoint: {{ .Values.limes.clusters.ccloud.catalog_url | replace "limes-3" "limes-campfire" }}
+      templates:
+        confirmed_commitments:
+          subject: {{ .Values.campfire.mail_type.expire.subject }}
+          body: {{ tpl (.Files.Get "files/commitment_mail.tpl") (merge (dict "region" .Values.global.region "condition" .Values.campfire.mail_type.confirm.condition)) | nindent 4 }}
+        expiring_commitments:
+          subject: {{ .Values.campfire.mail_type.confirm.subject }}
```
majewsky, line 21: The two subject variables are swapped around.

openstack/limes/templates/configmap.yaml
```diff
@@ -11,6 +11,15 @@ data:
       - {{ . }}
     {{- end }}
 {{ toYaml .Values.limes.clusters.ccloud | indent 4 }}
+    mail_notifications:
+      endpoint: {{ .Values.limes.clusters.ccloud.catalog_url | replace "limes-3" "limes-campfire" }}
```
majewsky, line 15: Docs require that there is a trailing slash, so we should play it safe:
```suggestion
      endpoint: {{ .Values.limes.clusters.ccloud.catalog_url | replace "limes-3" "limes-campfire" | trimSuffix "/" }}/
```

The trimSuffix is to ensure there is no double trailing slash if the URL already had one before.

majewsky (changes requested):
I can't get this to evaluate. `h diff upgrade` against qa-de-1 gives:

> Error: template: limes/templates/deployment-data-metrics.yaml:25:31: executing "limes/templates/deployment-data-metrics.yaml" at \<include "limes/templates/configmap.yaml" .\>: error calling include: template: limes/templates/configmap.yaml:19:29: executing "limes/templates/configmap.yaml" at \<.Values.campfire.mail_type.expire.subject\>: nil pointer evaluating interface {}.expire

majewsky (approved):

PR #8072, opened 2025-02-26
feat(manila): add grants for user metis

IvoGoman:
the metis user requires 'replication replica' to access the binlog


Carthaca (approved):
lgtm

kpawar-sap (approved):

PR #8074, opened 2025-02-26
feat(nova): add grants for user metis

IvoGoman:
the metis user requires 'replication replica' to access the binlog
this adds the grants to nova, nova_api and nova_cell mariadbs

### DIFF
--- a/openstack/nova/Chart.yaml
+++ b/openstack/nova/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: A Helm chart for Kubernetes
 name: nova
 icon: https://www.openstack.org/themes/openstack/images/project-mascots/Nova/OpenStack_Project_Nova_mascot.png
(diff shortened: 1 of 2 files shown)


joker-at-work (changes requested):
You should also replicate the cell2 and thus need change `mariadb_cell2`

joker-at-work:
What does "feat(nova/nova-api)" mean? afaik, we only have a "nova" helm-chart and no "nova-api" chart.

IvoGoman:
> What does "feat(nova/nova-api)" mean? afaik, we only have a "nova" helm-chart and no "nova-api" chart.

I was thinking in terms of MariaDBs, but then I would need to add the cell as well for consistency. So yes, nova should be sufficient.

IvoGoman:
Closing this PR. Grants are moved to internal PR specifying the User Credentials.

PR #8089, opened 2025-02-27
[sap-seeds] Filter out flavors without ID

fwiesel:
Print them as a comment, so one can cross-check it.


joker-at-work (approved):
Would have wished for an explanation how that could happen, i.e. only mentioned in a values file updating flavors defined in another values file.

fwiesel:
Updated the commit message.

joker-at-work (approved):

PhilippMatthes (approved):

PR #8090, opened 2025-02-27
[logs] Opentelemetry Openstack log parsing

Kuckkuck:


system/logs/templates/opentelemetry-logs-collector.yaml
```diff
@@ -114,32 +117,15 @@ spec:
...
-              - resource.attributes["log.type"] == "files"
-            statements:
-              - merge_maps(attributes, ExtractGrokPatterns(body, "%{TIMESTAMP_ISO8601:timestamp}%{SPACE}%{GREEDYDATA:log}", true), "upsert")
-              - set(attributes["config.parsed"], "files") where attributes["log_level"] != nil
 
 {{- if .Values.openTelemetryPlugin.openTelemetry.logsCollector.journaldConfig.enabled }}
     {{- include "journald.transform" . | nindent 6 -}}
 {{- end }}
-{{- if .Values.openTelemetryPlugin.openTelemetry.logsCollector.containerdConfig.enabled }}
+{{- if and (.Values.openTelemetryPlugin.openTelemetry.logsCollector.containerdConfig.enabled) (.Values.openTelemetryPlugin.openTelemetry.logsCollector.openstackConfig.enabled) }}
```
olandr, line 124: What is this condition supposed to check on? Should there be anything in this block?

system/logs/templates/_containerd-config.tpl
```diff
@@ -7,6 +7,17 @@ transform/protocol:
         - resource.attributes["network.protocol.name"] != nil
       statements:
         - set(attributes["network.protocol.name"], ConvertCase(attributes["network.protocol.name"], "lower"))
+
+transform/ingress:
+  error_mode: ignore
+  log_statements:
```
olandr, line 13: This one is also defined in https://github.com/sapcc/helm-charts/blob/8bc802b3c899c31ac07836459715f85c502a5322/system/logs/templates/_openstack-config.tpl#L35-L37

Would this be a problem if both the containerd and openstack would be added at the same time?

system/logs/templates/_openstack-config.tpl
```diff
@@ -0,0 +1,214 @@
+{{- define "openstack.receiver" }}
+filelog/containerd:
```
olandr, line 2: Similar to above but defined here https://github.com/sapcc/helm-charts/blob/8bc802b3c899c31ac07836459715f85c502a5322/system/logs/templates/_containerd-config.tpl#L24

system/logs/templates/opentelemetry-logs-collector.yaml
Kuckkuck: we should have two configs, one for openstack and another one for containerd in general, both should not be active at the same time. That was my idea.

system/logs/templates/_containerd-config.tpl
Kuckkuck: yes, because we have a special use case for swift to put swift logs into an extra index and we don't want to rollout an openstack config to all the kubernikus clusters.

system/logs/templates/opentelemetry-logs-collector.yaml
Kuckkuck: it should exit the rendering, because both shouldn't be active at the same time.

timojohlo:
resolved the open conflicted, otherwise: lgtm!
I like the `kubectl.kubernetes.io/default-container: "otc-container"` 👍

timojohlo (approved):

PR #8095, opened 2025-02-27
[mariadb] use passwordless local connections for healthchecks and exporter

s10:
* use local unix_socket connection for all status checks and mysqld_upgrade
  * this would allow to change root password without intermittent check failure
  * this would allow to use `'` and `\` in root password in mysqld_upgrade
* use monitor user with limited privileges for metrics sidecar

### DIFF
--- a/common/mariadb/templates/initdb/_init.sql.tpl
+++ b/common/mariadb/templates/initdb/_init.sql.tpl
@@ -32,6 +32,15 @@ GRANT {{ . }} TO {{ include "mariadb.resolve_secret_squote" $username }};
     {{- end }}
 {{- end }}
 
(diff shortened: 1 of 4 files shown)


common/mariadb/templates/initdb/_init.sql.tpl
```diff
@@ -32,6 +32,15 @@ GRANT {{ . }} TO {{ include "mariadb.resolve_secret_squote" $username }};
     {{- end }}
 {{- end }}
 
+ALTER USER 'root'@'localhost' IDENTIFIED VIA unix_socket;
```
s10, line 35: in addition to this, `'root'@'` with password will be added after https://github.com/sapcc/helm-charts/pull/8096

fwiesel (approved):

PR #8096, opened 2025-02-27
[mariadb] set root password on every database start

s10:
* start database with init.sql to update root password
* chart version bumped

### DIFF
--- a/common/mariadb/scripts/mariadb-set-root-password.sh
+++ b/common/mariadb/scripts/mariadb-set-root-password.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+set -eo pipefail
+shopt -s nullglob
+
+escape_special() {
+	{ set +x; } 2>/dev/null
+	echo "$1" |
+		sed 's/\\/\\\\/g' |
+		sed 's/'\''/'\\\\\''/g' |
+		sed 's/"/\\\"/g'
+}
+
+mysql_pass=$(cat /etc/mysql/mysql-users-secret/root || :)
+MYSQL_PASSWORD="${mysql_pass:-$MYSQL_ROOT_PASSWORD}"
+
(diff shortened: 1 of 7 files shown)


common/mariadb/scripts/mariadb-set-root-password.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -eo pipefail
+shopt -s nullglob
+set -o xtrace
```
fwiesel, line 4: Shouldn't that be disabled? Otherwise the password will appear in the logs.

s10: Yes, it must be disabled, missed the `set +x` below.

common/mariadb/templates/deployment.yaml
```diff
@@ -53,6 +53,24 @@ spec:
...
+      initContainers:
+      - name: set-root-password
+        image: {{ required ".Values.global.dockerHubMirrorAlternateRegion is missing" .Values.global.dockerHubMirrorAlternateRegion }}/{{ .Values.image }}
+        imagePullPolicy: {{ default "IfNotPresent" .Values.imagePullPolicy | quote }}
+        command:
+          - /bin/bash
+          - -c
+          - |
+            /configurator/mariadb-set-root-password.sh
+        volumeMounts:
```
galkindmitrii, line 65: Shouldn't we mount the /var/lib/mysql here too?
```
- name: mariadb-persistent-storage
            mountPath: /var/lib/mysql
            readOnly: false
```

s10: This init container only needs one shared with the mariadb container directory, where it puts the generated root init.sql

galkindmitrii (approved):

PR #8097, opened 2025-02-27
[mariadb] use `shared-app-images/alpine-kubectl` for pre-change job

s10:
* use `shared-app-images/alpine-kubectl` image for a pre-change job instead of the outdated `unified-kubernetes-toolbox`

### DIFF
--- a/common/mariadb/CHANGELOG.md
+++ b/common/mariadb/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v0.17.2 - 2025/03/05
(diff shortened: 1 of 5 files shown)


fwiesel (approved):

galkindmitrii (commented):
I think this might break when deployed to old qa labs on k8s `v1.25`

s10:
> We should proceed to use our `unified-kubernetes-toolbox` like the other Helm charts. Up to date images are available.

In almost all other charts, we use `shared-app-images/alpine-kubectl`. `unified-kubernetes-toolbox` is supposed to be used in CI, not in an application charts.

galkindmitrii (approved):

PR #8107, opened 2025-02-28
fix/otel node selector

olandr:
Adds nodeAffinity for otelcol, fluent, fluent-prometheus DS to prevent expected failed rollouts.


system/logs/templates/opentelemetry-logs-collector.yaml
```diff
@@ -13,6 +13,20 @@ spec:
   - operator: Exists
   podAnnotations:
     kubectl.kubernetes.io/default-container: "otc-container"
+    prometheus.io/scrape: "false"
```
timojohlo, line 16: I think this is a leftover and can be removed. The rest looks good 👍

timojohlo (approved):

PR #8108, opened 2025-03-03
[cinder] fix barbican_endpoint_type

leust:
The correct configuration property is barbican_endpoint_type


openstack/cinder/templates/etc/_cinder.conf.tpl
```diff
@@ -86,7 +86,7 @@ lock_path = /var/lib/cinder/tmp
 {{- include "ini_sections.cache" . }}
 
 [barbican]
-barbican_endpoint = internal
+barbican_endpoint_type = internal
```
joker-at-work, line 89: Why doesn't it use `auth_interface` like e.g. down in the `[nova]` section?

joker-at-work (approved):

fwiesel (approved):

PR #8115, opened 2025-03-03
fix(opensearch): use .netrc in opensearch scripts

joluc:
To prevent a [Command line leakage](https://everything.curl.dev/cmdline/passwords.html#command-line-leakage) replacing `curl -u` with netrc in current OpenSearch scripts.


olandr (approved):
Nice find, lg.

joluc:
Added in https://github.com/sapcc/helm-charts/pull/8128

PR #8116, opened 2025-03-03
postgresql, redis: support no alerts

stanislav-zaprudskiy:
The need to deploy without `PrometheusRule` comes from the cases where the corresponding CRDs aren't registered - typically, in local throw-away development clusters.


SuperSandro2000 (commented):
Fine for me but since we set the default value backwards compatible we can make this just a patch bump.

Let's see what @majewsky thinks about it.

SuperSandro2000 (approved):

PR #7268, opened 2024-10-17
[utils] fixing db_url_mariadb and _db_url_pxc_db double urlencoding

mutax:
The helpers _db_url_mariadb and _db_url_pxc_db are calling itself,
which leads to resolve_secret_urlquery being called twice, as the
helper db_credentials is also calling resolve_secret_urlquery.

This urlencodes the template string {{ resolve ... }} instead of the
actual value on the second pass through resolve_secret_urlquery.

PR #8117, opened 2025-03-03
[ironic] fix secrets-injector syntax

sandzwerg:
Add missing resolve_secret everywhere the validator complains about.
Found with the help of the secrets-inejector-validator:
helm-template-app ironic | secrets-injector-validator -kinds=OpenstackSeed

### DIFF
--- a/openstack/ironic/templates/etc/_inspector_secrets.conf.tpl
+++ b/openstack/ironic/templates/etc/_inspector_secrets.conf.tpl
@@ -2,8 +2,8 @@
 {{- include "ini_sections.default_transport_url" . }}
 
 [ironic]
-username = {{ .Values.global.ironicServiceUser }}
-password = {{ required ".Values.global.ironicServicePassword is missing" .Values.global.ironicServicePassword }}
+username = {{ .Values.global.ironicServiceUser | include "resolve_secret" }}
+password = {{ required ".Values.global.ironicServicePassword is missing" .Values.global.ironicServicePassword | include "resolve_secret" }}
 
 [database]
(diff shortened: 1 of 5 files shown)


BerndKue (approved):

fwiesel (approved):

openstack/ironic/templates/etc/_inspector_secrets.conf.tpl
```diff
@@ -2,20 +2,20 @@
...
 
 [ironic]
-username = {{ .Values.global.ironicServiceUser }}
-password = {{ required ".Values.global.ironicServicePassword is missing" .Values.global.ironicServicePassword }}
+username = {{ .Values.global.ironicServiceUser | include "resolve_secret" }}
+password = {{ required ".Values.global.ironicServicePassword is missing" .Values.global.ironicServicePassword | include "resolve_secret" }}
 
 [database]
 {{- if eq .Values.dbType "mariadb" }}
-connection = {{ tuple . .Values.mariadb.users.ironic_inspector.name .Values.mariadb.users.ironic_inspector.name .Values.mariadb.users.ironic_inspector.password | include "utils._db_url_mariadb" }}
```
s10, line 10: It looks like we have a problem with `utils._db_url_mariadb` `openstack/utils` function, because it is supposed to secrets resolving, but it doesn't since https://github.com/sapcc/helm-charts/pull/7268

sandzwerg: Ok, so what do you suggest? I merge this, you fix mariadb and afterward we change this back? Or is it not needed after all?

PR #8119, opened 2025-03-04
[cronus] add CRL config section for TLS

sapmli:


openstack/cronus/templates/cronus/_config.yaml.tpl
```diff
@@ -106,6 +106,18 @@ cronus:
...
+      clientCertWarningPeriod: {{ .Values.global.tls.clientCertWarningPeriod }}
+      {{- end }}
+      {{- if .Values.global.tls.checkRevocations }}
+      checkRevocations: true
+      revokedSerialNumbers:
+      {{- range $v := .Values.global.tls.revokedSerialNumbers }}
+        - {{ $v | quote }}
+      {{- end }}
+      {{- end }}
+      {{- end }}
```
kayrus, line 120:
```suggestion
      {{- if and .Values.global.tls .Values.global.tls.clientCertWarningPeriod }}
      {{- with .Values.global.tls }}
      clientCertWarningPeriod: {{ .clientCertWarningPeriod }}
      {{- end }}
      {{- if .checkRevocations }}
      checkRevocations: true
      revokedSerialNumbers:
      {{- range $v := .revokedSerialNumbers }}
        - {{ $v | quote }}
      {{- end }}
      {{- end }}
      {{- end }}
```

sapmli: I extracted the `.tls` check by purpose because it's used for both the included blocks.
We want to use the revocation check config independently of the value of `clientCertWarningPeriod`
@kayrus

dhalimi1975 (approved):
Looks okay

PR #8121, opened 2025-03-04
Masakari hypervisor agents

fwiesel:


fwiesel:
Dropping this in favor of our new ha service

PR #8128, opened 2025-03-04
[Opensearch-logs] remove users from cmd for all scripts, add permissions to dashboard to save and download csv files

Kuckkuck:


joluc (approved):
lgtm

PR #8131, opened 2025-03-04
[mariadb] add credential-updater sidecar

s10:
* add credential-updater sidecar, that reapplies init.sql on the detected change
* make init.sql not optional
* remove pre-change job, because it's no longer needed with the above sidecar - this part depends on depends on https://github.com/sapcc/helm-charts/pull/8096


fwiesel:
My understanding is, `shared-app-images/alpine-mariadb:python3.13-alpine3.21-20250305175417` is already a custom image to include python. Wouldn't it make sense to add then the `mariadb-credential-updater` script to it?

s10:
> My understanding is, `shared-app-images/alpine-mariadb:python3.13-alpine3.21-20250305175417` is already a custom image to include python. Wouldn't it make sense to add then the `mariadb-credential-updater` script to it?

I'm not sure that we should include the script in that image, because this image is supposed to be a simple custom base image, based on library/python:3.13-alpine3.21 with the only difference having installed mariadb client.

fwiesel (approved):

galkindmitrii (approved):

PR #8139, opened 2025-03-05
Settings for keystone federation via OIDC

bbobrov:
* Implemented support for Keystone federation using OpenID Connect (OIDC).
* Updated the Keystone Chart to version 0.10.0 reflecting the new feature.
* Added a new _federation-oidc.conf.tpl template to support OIDC configurations.
* Modified the wsgi-keystone.conf.tpl to add location blocks for OIDC endpoints.
* Created a new federation.yaml file to manage OIDC-related secrets within
  Kubernetes.
* Extended ingress configuration in ingress-api.yaml to handle session affinity
  required for browser-based federation workflows.
* Updated values.yaml to include federation configuration options with
  placeholders for provider, client_id, client_secret, and crypto_passphrase.
* Update keystone.conf to include the necessary federation bits

Right now the setup makes several assumptions:

* SAP IAS with OpenID Connect are used for the federation
* IDP name is sap-ias, protocol name is openid
* They are pre-created manually by the operator using the following
commands:

```
openstack identity provider create --remote-id <url-provided-in-secrets> --domain <any-domain> sap-ias
openstack mapping create --rules mapping.json --schema-version 2.0 sap-ias-mapping
openstack federation protocol create openid --mapping sap-ias-mapping --identity-provider sap-ias

```

### DIFF
--- a/openstack/keystone/templates/deployment-api.yaml
+++ b/openstack/keystone/templates/deployment-api.yaml
@@ -41,7 +41,9 @@ spec:
         configmap-bin-hash: {{ include (print $.Template.BasePath "/configmap-bin.yaml") . | sha256sum }}
         secrets-bin-hash: {{ include (print $.Template.BasePath "/secret-bin.yaml") . | sha256sum }}
         secrets-hash: {{ include (print $.Template.BasePath "/secrets.yaml") . | sha256sum }}
+        {{- if .Values.federation.oidc.enabled }}
         federation-hash: {{ include (print $.Template.BasePath "/federation.yaml") . | sha256sum }}
+        {{- end }}
         {{- if .Values.proxysql.mode }}
         prometheus.io/scrape: "true"
         prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
--- a/openstack/keystone/templates/deployment-api.yaml
+++ b/openstack/keystone/templates/deployment-api.yaml
@@ -41,6 +41,7 @@ spec:
         configmap-bin-hash: {{ include (print $.Template.BasePath "/configmap-bin.yaml") . | sha256sum }}
         secrets-bin-hash: {{ include (print $.Template.BasePath "/secret-bin.yaml") . | sha256sum }}
         secrets-hash: {{ include (print $.Template.BasePath "/secrets.yaml") . | sha256sum }}
+        federation-hash: {{ include (print $.Template.BasePath "/federation.yaml") . | sha256sum }}
         {{- if .Values.proxysql.mode }}
         prometheus.io/scrape: "true"
         prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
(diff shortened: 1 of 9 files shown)


openstack/keystone/templates/deployment-api.yaml
```diff
@@ -231,6 +236,10 @@ spec:
           secret:
             secretName: {{ .Release.Name }}-secrets
             defaultMode: 0444
+        - name: keystone-federation
+          secret:
+            secretName: keystone-federation
```
stanislav-zaprudskiy, line 241:
```suggestion
            secretName: {{ .Release.Name }}-federation
```

To make it match how Secret is named

stanislav-zaprudskiy (commented):
Volume, volumeMounts, Secret with oidc-federation.conf, and annotations are set unconditionally w/o checking `if .Values.federation.oidc.enabled`. Even though it would not harm probably (as empty files would be mounted), it would look confusing e.g. for a person trying to understand the configuration during an incident - I suggest you guard all related configuration with the similar conditionals.

openstack/keystone/templates/deployment-api.yaml
bbobrov: I don't understand why it matters, especially if we hardcode the names everywhere else. Maybe we could remove the `{{ .Release.Name }}` for the secrets above to make it be the same way as everything else?

bbobrov: ah, but their name is generated this way... ok then!

openstack/keystone/templates/deployment-api.yaml
```diff
@@ -41,6 +41,7 @@ spec:
         configmap-bin-hash: {{ include (print $.Template.BasePath "/configmap-bin.yaml") . | sha256sum }}
         secrets-bin-hash: {{ include (print $.Template.BasePath "/secret-bin.yaml") . | sha256sum }}
         secrets-hash: {{ include (print $.Template.BasePath "/secrets.yaml") . | sha256sum }}
+        federation-hash: {{ include (print $.Template.BasePath "/federation.yaml") . | sha256sum }}
```
stanislav-zaprudskiy, line 44:
```suggestion
        {{- if .Values.federation.oidc.enabled }}
        federation-hash: {{ include (print $.Template.BasePath "/federation.yaml") . | sha256sum }}
        {{- end }}
```

Otherwise unrelated changes in the template even when `enabled=false` would be causing API restarts

openstack/keystone/templates/etc/_keystone.conf.tpl
```diff
@@ -29,7 +29,7 @@ default_tag = vc-{{ $az }}-0
 
 {{- if .Values.api.auth }}
 [auth]
-methods = {{ .Values.api.auth.methods | default "password,token,application_credential" }}
+methods = {{ if .Values.federation.oidc.enabled }}openid,{{ end }}{{ .Values.api.auth.methods | default "password,token,application_credential" }}
```
stanislav-zaprudskiy, line 32: This is very implicit IMHO, and might confuse a person reading `.Values.api.auth.methods` value and not realizing why the provided values are not in effect. I would suggest adding a comment next to the value declaration/definition, or otherwise make it explicit.

stanislav-zaprudskiy: The docs suggests to make sure that `external` is removed

[1] https://docs.openstack.org/keystone/latest/admin/external-authentication.html#configuration
> You should disable the external auth method if you are currently using federation. External auth and federation both use the REMOTE_USER variable. Since both the mapped and external plugin are being invoked to validate attributes in the request environment, it can cause conflicts.

[2] https://docs.openstack.org/keystone/latest/admin/federation/configure_federation.html#add-the-auth-method
> You should also remove external as an allowable method.

Explicit configuration and comment would best here, IMHO

bbobrov: > The docs suggests to make sure that external is removed

And we cannot do that. We have to go against the official documentation unfortunately :)

stanislav-zaprudskiy (approved):

BerndKue (approved):

PR #8156, opened 2025-03-10
[greenhouse] thanos preset ingress host clusteroptionoverride

trouaux:
to allow for the central cluster to connect to remote thanos query instances we are creating ingress grpc resources.

using clusterOptionsOverride we can specify unique domain name for each cluster


system/greenhouse-ccloud/templates/thanos-pluginpreset.yaml
```diff
@@ -5,9 +5,17 @@ metadata:
...
-  {{- if .Values.thanos.clusterOptionOverrides }}
   clusterOptionOverrides:
-    {{ toYaml .Values.thanos.clusterOptionOverrides | nindent 4 }}
+  {{- range $cluster, $ingressPlugin := .Values.ingressPlugins }}
+  {{- $ingressDomain := trimSuffix "." (trimPrefix "ingress." $ingressPlugin.recordName) }}
+  - clusterName: {{ $cluster }}
+    overrides:
+    - name: thanos.query.ingress.grpc.hosts
+      value:
+        - host: "thanos-grpc.{{ $ingressDomain }}"
```
richardtief, line 15: It's a range over a list. So this should work like this:
```suggestion
  {{- range $plugin := .Values.ingressPlugins }}
  {{- $ingressDomain := trimSuffix "." (trimPrefix "ingress." $plugin.recordName) }}
  - clusterName: {{ $plugin.cluster }}
    overrides:
    - name: thanos.query.ingress.grpc.hosts
      value:
        - host: "thanos-grpc.{{ $ingressDomain }}"
```

richardtief (changes requested):
Looks good. I left a suggestion. And please increase the Chart version as well.

system/greenhouse-ccloud/templates/thanos-pluginpreset.yaml
trouaux: indeed, thanks

richardtief (approved):
Looking good.

PR #8157, opened 2025-03-10
[nova] Allow GET on os-attach-interfaces for viewers

joker-at-work:
Upstream allows listing and showing os-attach-interface i.e. Nova-known virtual interfaces to readers. We didn't allow it for our equivalent (viewers) until know. I couldn't find a specific reason. It looks like we made them "inherit" what we had before when they got split up into multiple policies.


sandzwerg (approved):

fwiesel:
Because people will then use that API call and ask, why it isn't matching what neutron reports.
And that is, because the networking related information is only synced on a best-effort basis to nova.

PR #8162, opened 2025-03-10
[hermes] move logstash conf to config map, all secrets in secrets

notque:


openstack/hermes/README.md
```diff
@@ -4,7 +4,7 @@
 
 Important Note:
 
-Hermes utilizes Persistent Volumes for ElasticSearch. These PV's are set to RECYCLE, so when they are no longer used, they are deleted and available for other pods.
+Hermes utilizes Persistent Volumes for Opensearch. These PV's are set to RECYCLE, so when they are no longer used, they are deleted and available for other pods.
```
viennaa, line 7:
```suggestion
Hermes utilizes Persistent Volumes for OpenSearch. These PV's are set to RECYCLE, so when they are no longer used, they are deleted and available for other pods.
```

openstack/hermes/templates/etc/_logstash.conf.tpl
```diff
@@ -1,13 +1,12 @@
 input {
 
 {{ range $key, $value := .Values.hermes.rabbitmq.targets }}
```
viennaa, line 3: To me $key and $value are not explanatory. Can they be renamed? I've tried to understand what $key is actually holding but there is also no example in the [values.yaml](https://github.com/sapcc/helm-charts/blob/master/openstack/hermes/values.yaml#L23-L28)

Checking in the config repo it is the replica. So maybe this would be a better naming here:

```suggestion
{{ range $replica_name, $replica := .Values.hermes.rabbitmq.targets }}
```

Then the bottom part also would read easier:

```
    user => "${RABBITMQ_{{ upper $replica_name }}_USER}"
    password => "${RABBITMQ_{{ upper $replica_name }}_PASSWORD}"
```

viennaa (commented):
Thanks for secreting this chart. Like the idea of using environment vars here.

Two things from my pov inline.

viennaa (approved):
Thanks :)

PR #8175, opened 2025-03-12
[CODEOWNERS] add @s10 as shared services and utils codeowner

s10:
add @s10 as shared services (rabbitmq, mariadb, memcached) and utils codeowner


galkindmitrii (approved):
lgtm

businessbean (approved):
lgtm

PR #8191, opened 2025-03-13
[utils] Add support for different db types in `utils.db_url` helper

s10:
Add support for different db types in `utils.db_url` helper by passing an optional sixth item

This change is required for `nova`, `placement` and `ironic-inspector` database url configuration


businessbean (approved):
lgtm

galkindmitrii (approved):

PR #8193, opened 2025-03-13
[designate] seed generic domains using global.domains_seeds

wagnerd3:

The expressions are technically equal, only the `ora` domain was added. If this is on purpose, I can introduce an exception. I excepted `iaas-` (external) domains already, as I thought they are probably missing on purpose, but I could also add those if desired. 
I ran `h3 diff` against `qa-de-1`, you can have a look at the output here: [designatediff.txt](https://github.com/user-attachments/files/19233309/designatediff.txt)

### DIFF
--- a/openstack/designate/ci/test-values.yaml
+++ b/openstack/designate/ci/test-values.yaml
@@ -9,9 +9,9 @@ global:
   registry: registry
   dockerHubMirror: dockerHubMirror
   dockerHubMirrorAlternateRegion: dockerHubMirrorAlternateRegion
-  global:
   domain_seeds:
-    skip_hcm_domain: false
+    customer_domains: [ bar, foo, baz ]
+
 
 vice_president: true
 image_version_designate: v1
--- a/openstack/designate/templates/seeds.yaml
+++ b/openstack/designate/templates/seeds.yaml
@@ -1,3 +1,7 @@
+{{/* cdomains equals [ "global" ] in global region */}}
+{{- $cdomains := .Values.global.domain_seeds.customer_domains | required "missing value for .Values.global.domain_seeds.customer_domains" -}}
+{{- $domains  := concat (list "ccadmin" "Default" "cc3test") $cdomains -}}
+
 apiVersion: "openstack.stable.sap.cc/v1"
 kind: "OpenstackSeed"
 metadata:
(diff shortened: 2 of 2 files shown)


galkindmitrii:
Hi @wagnerd3, thanks for working on that.

The `.Values.global_setup` is a special case and separate region where customer domains should not be present at all (any).
Please adjust accordingly, that `{{- if .Values.global_setup }}` we only seed default domains (`ccadmin`, `cc3test` etc.) + `domain-global-seed`.

Thank you.

galkindmitrii (changes requested):
^^

wagnerd3:
@galkindmitrii I updated the PR and now implemented the global region also via the `domain_seeds.customer_domains` variable. In this region we now have `domain_seeds.customer_domains = list ("global")` coming from the values. 
Additionally, I added all domains (even `ccadmin, cc3test`) to the loop except `default, tempest` - because they are completely different from the rest. 

I generated some more `helm diff` outputs to check (this time in the `dyff` syntax for better overview):
[designateGlobalQA1diff.txt](https://github.com/user-attachments/files/19348707/designateGlobalQA1diff.txt)
[designateGlobalDE1diff.txt](https://github.com/user-attachments/files/19348710/designateGlobalDE1diff.txt)
[designateQA1diff.txt](https://github.com/user-attachments/files/19348711/designateQA1diff.txt)
[designateDE1diff.txt](https://github.com/user-attachments/files/19348716/designateDE1diff.txt)

You can see, that a) only the orders change and b) as already asked initially the `ora` domain gets some missing assignments in the non-global regions. Let me know if this gives enough confidence now to merge. :-)

wagnerd3:
@galkindmitrii I saw some errors from the designate nanny seed in `global` region where the `monsoon3global/domain-default-seed` is required. As designate is set up in some form of HA-setup, it is currently also seeded from all the involved regions. The seeding for global is not layed out in such distributed fashion though, because the seedings are one-time-actions and not required to work when one region is down. The domain seeds are just seeded from eu-de-2. I would include some lines in this PR to limit this for global, is this okay with you?

wagnerd3:
I updated the PR with regard to the above.

galkindmitrii (changes requested):
Unfortunately changing `.Values.global_setup` to `.Values.global.is_global_region ` introduces many issues:
1) All other chart [templates need to be updated](https://github.com/search?q=repo%3Asapcc%2Fhelm-charts%20global_setup&type=code)
2) Secrets config need to be updated, services redeployed
3) Keystone which uses the same variable/approach will then differ.

Therefore, I'd not touch that `global_setup` setting.

majewsky:
To clarify, the problem with `.Values.global_setup` is that it falls out of the standard pattern of "values shared by multiple services are declared in `$REGION/values/globals.yaml` in the secrets repo and are declared under `.Values.global`. The way that `.Values.global_setup` is used makes it hard for us to adopt in our own services, at least because its semantics are not properly documented, so we are proposing a standardized value that follows the usual semantics.

> All other chart templates need to be updated.

Yes, we plan to do that.

> Secrets config need to be updated

Already done (see `global/values/globals.yaml` in the secrets repo).

> services redeployed

Well yes, but that happens frequently anyway. We are not in a hurry with this, it does not block anything else if you do not do a deployment immediately.

> Keystone which uses the same variable/approach will then differ.

Can you explain this point further, please?

galkindmitrii:
> Can you explain this point further, please?

https://github.com/search?q=repo%3Asapcc%2Fhelm-charts+path%3A%2F%5Eopenstack%5C%2Fkeystone%5C%2F%2F+global_setup&type=code

wagnerd3:
> > Can you explain this point further, please?
> 
> https://github.com/search?q=repo%3Asapcc%2Fhelm-charts+path%3A%2F%5Eopenstack%5C%2Fkeystone%5C%2F%2F+global_setup&type=code

I would soon open PRs for the other references of the old variable. I think the cut-over time in the repo's itself will be 1-2 weeks only. I will then just wait for all deployments before I delete the old variable in secrets.

galkindmitrii:
So there are [55 files where global_setup](https://github.com/search?q=repo%3Asapcc%2Fhelm-charts%20global_setup&type=code) is currently used in `helm-charts` and I'm not feeling very comfortable having partial migration to the new `is_global_region`.

Maybe @braunmatt or @mutax sees it different...

majewsky:
So if we make a PR that adjusts Designate and Keystone and domain-seeds all at once, that would be fine?

galkindmitrii:
> So if we make a PR that adjusts Designate and Keystone and domain-seeds all at once, that would be fine?

Please include Designate and Designate Nanny into 1 PR. Keystone separately since there are other approvals.

Thx.

galkindmitrii (approved):
lgtm

PR #8209, opened 2025-03-14
[rabbitmq] don't create metrics user if not needed

s10:
Creation of the `metrics` user with `monitoring` tag no longer coupled with enabling prometheus metrics

Native Prometheus RabbitMQ metrics don't rely on any user

To create `metrics` user, if needed, set:
`.Values.rabbitmq.monitoring.addMetricsUser: true`


galkindmitrii (approved):

businessbean (approved):
lgtm

PR #8214, opened 2025-03-17
Add secret reload annotation to digicert-issuer

jknipper:


kayrus (approved):

kayrus:
hm, I don't see a reference to a secret in the deployment template. How does it work?

PR #8215, opened 2025-03-17
[elektra] seed generic domains using global.domains_seeds

wagnerd3:

The expressions are technically equal, only the ora domain was added. If this is on purpose, I can introduce an exception. I excepted iaas- (external) domains already, as I thought they are probably missing on purpose, but I could also add those if desired.
I ran h3 diff against qa-de-1 and deduplicated the things which just switched their order, you can have a look at the output here: [elektra.txt](https://github.com/user-attachments/files/19286192/elektra.txt)

### DIFF
--- a/openstack/elektra/ci/test-values.yaml
+++ b/openstack/elektra/ci/test-values.yaml
@@ -2,6 +2,8 @@ global:
   registry: "myImage"
   region: qa-de-1
   vaultBaseURL: vault.example.com/secrets
+  domain_seeds:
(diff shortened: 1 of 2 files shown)


wagnerd3:
@hgw77 After our call today I generated some more `helm diff`s with `dyff`, here you can check much more easily what changed:
[elektraQA1diff.txt](https://github.com/user-attachments/files/19348850/elektraQA1diff.txt)
[elektraDE1diff.txt](https://github.com/user-attachments/files/19348856/elektraDE1diff.txt)

PR #8224, opened 2025-03-18
[hermes+maia] seed generic domains using global.domains_seeds

wagnerd3:

The expressions are technically equal, only the ora domain was added for Maia and fsn, kyma, ora for Hermes. If they were excluded on purpose, let me know and I can change it. I excluded `iaas-` for now.
I ran `h3 diff` with `dyff` option against qa-de-1 and eu-de-1, you can have a look at the output here:
[maiaDE1diff.txt](https://github.com/user-attachments/files/19349459/maiaDE1diff.txt)
[maiaQA1diff.txt](https://github.com/user-attachments/files/19349461/maiaQA1diff.txt)
[hermesDE1diff.txt](https://github.com/user-attachments/files/19349463/hermesDE1diff.txt)
[hermesQA1diff.txt](https://github.com/user-attachments/files/19349464/hermesQA1diff.txt)

### DIFF
--- a/openstack/hermes/ci/test-values.yaml
+++ b/openstack/hermes/ci/test-values.yaml
@@ -9,6 +9,9 @@ global:
   user_suffix: ""
   dockerHubMirror: "another-region"
   dockerHubMirrorAlternateRegion: "other.dockerhub.mirror"
+  domain_seeds:
+    customer_domains: [bar, foo, baz]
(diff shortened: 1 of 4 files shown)


notque (approved):
LGTM, nice work.

majewsky (approved):
Here's a second approval for ya.

PR #8226, opened 2025-03-18
[ironic] seed generic domains using global.domains_seeds

wagnerd3:

The expressions are technically equal, only the ora domain was added. If they were excluded on purpose, let me know and I can change it. I excluded `iaas-` for now.
I ran `h3 diff` with `dyff` option against qa-de-1 and eu-de-1 you can have a look at the output here:
[ironicDE1diff.txt](https://github.com/user-attachments/files/19349338/ironicDE1diff.txt)
[ironicQA1diff.txt](https://github.com/user-attachments/files/19349339/ironicQA1diff.txt)

### DIFF
--- a/openstack/ironic/ci/test-values.yaml
+++ b/openstack/ironic/ci/test-values.yaml
@@ -10,7 +10,8 @@ global:
   ironic_service_password: topSecret
   ironicServicePassword: topSecret
   domain_seeds:
-    skip_hcm_domain: false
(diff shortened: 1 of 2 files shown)


sandzwerg (approved):

wagnerd3:
Thx for the review.

I just replaced `contains "iaas-" .` with `hasPrefix "iaas-" .` to make it more robust. will merge and watch out for errors in the next weeks.

PR #8230, opened 2025-03-18
[greenhouse] thanos preset fix missing tls config

trouaux:


system/greenhouse-ccloud/templates/thanos-pluginpreset.yaml
```diff
@@ -16,6 +16,11 @@ spec:
           paths:
             - path: /
               pathType: ImplementationSpecific
+    - name: thanos.query.ingress.grpc.tls
+      value:
+        - hosts:
+            - "thanos-grpc.{{ $ingressDomain }}"
+          secretName: {{ printf "tls-%s" (print "thanos-grpc." $ingressDomain | replace "." "-") }}
```
richardtief, line 23: Some suggestion to make it cleaner. 
```suggestion
          secretName: {{ printf "tls-thanos-grpc-%s" $ingressDomain | replace "." "-" }}
```

richardtief (commented):
Thanks. One small comment

system/greenhouse-ccloud/templates/thanos-pluginpreset.yaml
```diff
@@ -16,6 +16,11 @@ spec:
           paths:
             - path: /
               pathType: ImplementationSpecific
+    - name: thanos.query.ingress.grpc.tls
+      value:
+        - hosts:
+            - "thanos-grpc.{{ $ingressDomain }}"
+          secretName: {{ printf "tls-thanos-grpc-%s" $ingressDomain | replace "." "-" }}
```
viennaa, line 23: do we really need to explicitly overwrite this here? Can't we go with a default of the host with respective domain?

viennaa (commented):
Thanks for adding the overwrite, just some question for my understanding.

system/greenhouse-ccloud/templates/thanos-pluginpreset.yaml
trouaux: it is part of clustetOptionOveride to be able to configure it with its domain name. what do you mean by default of the host ?

PR #8245, opened 2025-03-19
[nats] calico prep

ronchi-oss:
Needed for the migration to Calico.

### DIFF
--- a/common/nats/templates/service.yaml
+++ b/common/nats/templates/service.yaml
@@ -14,6 +14,7 @@ spec:
   externalIPs:
     - {{ .Values.service.externalIP }}
   {{- end }}
+  externalTrafficPolicy: {{ default "Local" .Values.service.externalTrafficPolicy }}
   ports:
     - port: 8222
       targetPort: 8222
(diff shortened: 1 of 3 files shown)


defo89 (approved):

common/nats/templates/service.yaml
```diff
@@ -14,6 +14,7 @@ spec:
   externalIPs:
     - {{ .Values.service.externalIP }}
   {{- end }}
+  externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy }}
```
notandy, line 17: do we want some default value in values.yaml? (I would say yes). Else this chart could be rendered without a value but failed to deploy since it needs one.

E.g., for passwords which obviously we don't have a default, we use the `required` function to fail rendering if no password was provided. But in this case a default should be fine

notandy (changes requested):

common/nats/templates/service.yaml
ronchi-oss: since Andromeda is the only chart using it, I was going to skip a default value but I can add it.

ronchi-oss: It should be good to go now, please check again @notandy

notandy (approved):

PR #8257, opened 2025-03-19
[cinder] Add new lockfile cinder nanny cronjob

jagoleni:
A cronjob template that executes the newly added `cinder-manage sap clean_old_lock_files`command daily. See https://github.wdf.sap.corp/cc/secrets/pull/16725 also.

### DIFF
--- a/openstack/cinder/templates/cinder-nanny-lockfiles.yaml
+++ b/openstack/cinder/templates/cinder-nanny-lockfiles.yaml
@@ -1,0 +1,73 @@
+{{- if .Values.nanny.lockfiles.enabled }}
+kind: CronJob
+apiVersion: batch/v1
+metadata:
+  name: cinder-nanny-lock-files
+  labels:
+    system: openstack
+    service: cinder-nanny
+spec:
+  schedule: "{{ .Values.nanny.lockfiles.crontab_schedule }}"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    metadata:
+      annotations:
+      labels:
+        component: cinder-nanny
+    spec:
+      template:
+        metadata:
+          annotations:
+            {{- include "utils.linkerd.pod_and_service_annotation" . | indent 12 }}
+          labels:
+            component: cinder-nanny
+        spec:
+          {{- include "utils.proxysql.pod_settings" . | nindent 10 }}
+          shareProcessNamespace: true
+          restartPolicy: Never
+          volumes:
+          - name: etccinder
+            emptyDir: {}
+          - name: cinder-etc
+            configMap:
+              name: cinder-etc
+          - name: cinder-etc-confd
+            secret:
+              secretName: {{ .Release.Name }}-secrets
+          {{- include "utils.proxysql.volumes" . | indent 10 }}
+          {{- include "utils.coordination.volumes" . | indent 10 }} 
+          containers:
+            - name: lock-files
+              image: {{required ".Values.global.registry is missing" .Values.global.registry}}/loci-cinder:{{required ".Values.imageVersion is missing" .Values.imageVersion}}
+              imagePullPolicy: IfNotPresent
+              command:
+                - /bin/sh
+                - -c
+                - |
+                  set -e
+                  cinder-manage sap clean_old_lock_files --batch-size {{ .Values.nanny.lockfiles.batch_size }} {{- if .Values.nanny.lockfiles.dry_run}} --dry-run {{- end}}
+                  {{- include "utils.script.job_finished_hook" . | nindent 18 }}
+              volumeMounts:
+                - name: etccinder
+                  mountPath: /etc/cinder
+                - name: cinder-etc
+                  mountPath: /etc/cinder/cinder.conf
+                  subPath: cinder.conf
+                  readOnly: true
+                - name: cinder-etc-confd
+                  mountPath: /etc/cinder/cinder.conf.d
+                - name: cinder-etc
+                  mountPath: /etc/cinder/logging.ini
+                  subPath: logging.ini
+                  readOnly: true
+                  {{- include "utils.proxysql.volume_mount" . | indent 16 }}
+                  {{- include "utils.coordination.volume_mount" . | indent 16 }}
+              resources:
+                requests:
+                  memory: "1000Mi"
+                  cpu: "100m"
+                limits:
+                  memory: "2000Mi"
+                  cpu: "200m"
+          {{- include "utils.proxysql.container" . | indent 12 }}
+{{- end }}
--- a/openstack/cinder/values.yaml
+++ b/openstack/cinder/values.yaml
@@ -110,6 +110,15 @@ sap_allow_independent_clone: True
 
 rpc_ping_enabled: true
 
+nanny:
+  lockfiles:
+    enabled: false
+    # daily at 22:00' 
+    crontab_schedule: '0 22 * * *' 
+    verbose: false
+    batch_size: 10000
+    dry_run: true
+
 proxysql:
   mode: null # Disabled by default
 


openstack/cinder/templates/cinder-nanny-lockfiles.yaml
```diff
@@ -0,0 +1,100 @@
+{{- if .Values.nanny.lockfiles.enabled }}
+kind: CronJob
+apiVersion: batch/v1
+metadata:
+  name: cinder-nanny-lock-files
+  namespace: monsoon3
```
joker-at-work, line 6: Is this explicit namespace necessary? Do we do that in the other Deployments and do not use some variable at least or let helm decide by where we deploy the helm-chart? Reason is, that this will change in the near future.

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
```diff
@@ -0,0 +1,100 @@
+{{- if .Values.nanny.lockfiles.enabled }}
```
joker-at-work, line 1: This commit doesn't add this value. Did you forget to include the `values.yaml`?

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
```diff
@@ -0,0 +1,100 @@
...
+apiVersion: batch/v1
+metadata:
+  name: cinder-nanny-lock-files
+  namespace: monsoon3
+  labels:
+    system: openstack
+    service: cinder-nanny
+
+spec:
+  schedule: "{{ .Values.nanny.lockfiles.crontab_schedule }}"
```
joker-at-work, line 12: This commit doesn't add this value. Did you forget to include the `values.yaml`?

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
```diff
@@ -0,0 +1,100 @@
...
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    metadata:
+      annotations:
+        {{- if and $.Values.global.linkerd_enabled $.Values.global.linkerd_requested }}
+        linkerd.io/inject: enabled
+        {{- end }}
```
joker-at-work, line 19: Why don't you use the linkerd snippet from the `utils` chart here? Does Cinder not use that in the other templates?

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
```diff
@@ -0,0 +1,100 @@
...
+        component: cinder-nanny
+    spec:
+      template:
+        metadata:
+          annotations:
+          labels:
+            component: cinder-nanny
+        spec:
+          nodeSelector:
+            zone: farm
```
joker-at-work, line 30: What's this?

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
```diff
@@ -0,0 +1,100 @@
...
+              livenessProbe:
+                exec:
+                  command:
+                  - test
+                  - -f
+                  - /shutdown/keepalive
+              volumeMounts:
+                {{- include "utils.proxysql.volume_mount" . | indent 16 }}
+                - mountPath: /shutdown
+                  name: shutdown-signal
```
joker-at-work, line 66: There should be a helper in the `utils` chart for the proxysql container. Please take a look at other templates.

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
```diff
@@ -0,0 +1,100 @@
...
+            - name: lock-files
+              image: {{required ".Values.global.registry is missing" .Values.global.registry}}/loci-cinder:{{required ".Values.imageVersion is missing" .Values.imageVersion}}
+              imagePullPolicy: IfNotPresent
+              command:
+                - /bin/sh
+                - -c
+                - 'cinder-manage sap clean_old_lock_files --batch-size 10 --dry-run && rm /shutdown/keepalive'
```
joker-at-work, line 73: `--batch-size` should be configurable imho. adding `--dry-run` should be configurable. shutdown should be unconditional, because we otherwise wouldn't shut down the pod.

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
```diff
@@ -0,0 +1,100 @@
...
+              env:
+                - name: NAMESPACE
+                  valueFrom: {fieldRef: {fieldPath: metadata.namespace}}
```
joker-at-work, line 76: What is this used for? I don't see it used.

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
```diff
@@ -0,0 +1,100 @@
...
+              volumeMounts:
+                - name: cinder-etc
+                  mountPath: /etc/cinder/cinder.conf
+                  subPath: cinder.conf
+                  readOnly: true
+                - name: etccinder
+                  mountPath: /etc/cinder
```
joker-at-work, line 83: Does order not matter here? I would have expected having to first mount the emptyDir to `/etc/cinder` before being able to fill it with other files.

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
jagoleni: No, you're right - I will remove it. The helm commands in the CI pipeline use the `--namespace` parameter. It seems to be hardcoded in the CI pipeline though: https://github.wdf.sap.corp/cc/secrets/blob/55b316d33aeb5eb73e32874a9e12d96454467f4c/ci/shared/pipeline.yaml.erb#L80C1-L82C4

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
jagoleni: There is already a branch open (No PR) in the secrets repo, I added the values here: https://github.wdf.sap.corp/cc/secrets/compare/master...add_lockfile_cinder_nanny

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
jagoleni: There is already a branch open (No PR) in the secrets repo, I added the values here: https://github.wdf.sap.corp/cc/secrets/compare/master...add_lockfile_cinder_nanny

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
jagoleni: I don't know, that is a leftover from the original cinder nanny chart (it's also in the nova nanny chart). Will remove

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
jagoleni: This is how the old cinder nanny add the annotations, cinder uses the `utils.linkerd.pod_and_service_annotation` template. Will switch to the linkerd snippet used by cinder

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
jagoleni: Yes, I'm using now the helper. I modified the helper from the utils chart to make it shutdown, with the existing shutdown helper the modification is no longer needed

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
jagoleni: Shutdown is no longer needed, batchsize and dry run is now parametrized.

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
jagoleni: Removed

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
jagoleni: It doesn't seem to matter, just tested it. I changed the order nonetheless for better readability.

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
joker-at-work: We usually have default values in the helm-charts' `values.yaml` - at least if they make sense. If no default value makes sense, we still want to document the supported values this way and either add a comment with `# some_key: DEFINED-IN-SECRETS` or set the key to `nil` i.e. `some_key: nil`.

jagoleni: Added default values to `values.yaml`

openstack/cinder/values.yaml
```diff
@@ -110,6 +110,15 @@ sap_allow_independent_clone: True
...
 rpc_ping_enabled: true
 
+nanny:
+  lockfiles:
+    enabled: False
+    # daily at 22:00' 
+    crontab_schedule: '0 22 * * *' 
+    verbose: False
+    batch_size: 10000
+    dry_run: True
```
joker-at-work, line 120: minor: in yaml, I think `true` and `false` are lowercase

openstack/cinder/templates/cinder-nanny-lockfiles.yaml
```diff
@@ -0,0 +1,72 @@
...
+                  {{- include "utils.proxysql.volume_mount" . | indent 16 }}
+                  {{- include "utils.coordination.volume_mount" . | indent 16 }}
+          {{- include "utils.proxysql.container" . | indent 12 }}
+              resources:
+                requests:
+                  memory: "500Mi"
+                  cpu: "25m"
+                limits:
+                  memory: "500Mi"
+                  cpu: "100m"
```
joker-at-work, line 71: Did you mean this to be set for the proxysql container? If you meant this resources to be part of the `lock-files` container, it would have to go before the include.

With that limit, hard-coded, we will be OOM-killed at 500 MiB. Did you test that? IIRC, I've seen 1 GiB use regularly in multiple regions. Please take numbers that can work with 10000 files found and all the volumes in the DB i.e. run it in a big region and see what it consumes. Until then, don't set a limit.

joker-at-work (approved):

PR #8278, opened 2025-03-20
greenhouse-ccloud: add thanos option values for compute clusters

richardtief:


viennaa (commented):
Thanks. Can we get an example of `kubeMonitoring.cephClusters` in the `values.yaml` to immediately see what is expected?

richardtief:
It is already. 

https://github.com/sapcc/helm-charts/blob/4fb761018632c3652b4c0984d317cfc221890f2c/system/greenhouse-ccloud/values.yaml#L113-L117

richardtief:
I've added the expected structure

viennaa (approved):
Thanks :)

PR #8284, opened 2025-03-20
[kubernikus] remove mail sending for quota alert

wagnerd3:
As discussed with @majewsky it does not make sense to alert customers, because they cannot raise quota themselves anymore. I will update the playbook to link the support staff how to create a quota override.


majewsky (approved):

majewsky:
> I never did this, when I touched the seeds. Is this only to be done for everything != seeds?

Bumping the chart version is necessary when the chart in question is not deployed directly, but only as part of another chart (see `grep -A2 kubernikus-monitoring system/*/Chart.lock`). You can run the `helm-dep-up` jobs in the pipelines of the respective top-level charts to pull in this change.

PR #8288, opened 2025-03-20
[barbican][shared] bump mariadb chart dependency and remove rabbitmq

s10:
* bump mariadb chart dependency

adds support for root password update and user-credential-updater sidecar

* remove rabbitmq dependency

it seems to be not being used


s10:
@rajivmucheli @galkindmitrii , please take a look at this PR

rajivmucheli (approved):

galkindmitrii (approved):
lgtm, looks like barbican worker is not used and thus rabbit not required.

PR #8290, opened 2025-03-20
[manila][shared] update mariadb and rabbitmq chart dependencies

s10:
* adds user-credential-updater sidecars to both services
* allows to update root user password in mariadb
* allows to use secrets-injector for backup-v2 secrets
* bumps mariadb and rabbitmq to latest bugfix releases


Carthaca (approved):
❤️

kpawar-sap (approved):

PR #8211, opened 2025-03-17
Update VMPropertiesCollector to collect the net ip address

zjeoff:
Add collecting Virtual Machine network Ip address. 
     - metric_suffix: "net_ip_address"
        key: "net|ip_address"

PR #8301, opened 2025-03-22
Update Chart.yaml

zjeoff:
this is related to https://github.com/sapcc/helm-charts/pull/8211


viennaa:
Please bump it on the original branch, not on a new one.

[zjeoff:patch-2](https://github.com/zjeoff/helm-charts/tree/patch-2)

PR #8302, opened 2025-03-24
[Sentry] add alias: sentry-redis

rajivmucheli:
the redis exporter is unable to find sentry-redis


BerndKue (approved):

majewsky:
You should not have both in there; from what I can see in the Helm manifest, they interfere with each other in weird ways.

rajivmucheli:
works when the redis pwd from the secrets is removed.

PR #8310, opened 2025-03-24
[nova] seed generic domains using global.domains_seeds

wagnerd3:

When changing the `sap-seeds/templates/nova-seed` I compared it to the other seed and did not find anything which would not be included in the other seed - if this is incorrect I can also apply the pattern to this file.
Otherwise the expressions are technically equal, only the `ora` domain was added. If it was excluded on purpose, let me know and I can change it. I excluded iaas- for now.
I ran h3 diff with dyff option against qa-de-1 and eu-de-1 you can have a look at the output here:
[novaDE1diff.txt](https://github.com/user-attachments/files/19427958/novaDE1diff.txt)
[novaQA1diff.txt](https://github.com/user-attachments/files/19427959/novaQA1diff.txt)


openstack/nova/templates/seed.yaml
```diff
@@ -101,579 +98,93 @@ spec:
       - project: admin
         role: cloud_compute_admin
 
-  - name: ccadmin
+  {{- range $domains}}
+  {{- if and (not (hasPrefix "iaas-" .)) (ne . "Default") }}
```
joker-at-work, line 102: Why do we have all those `hasPrefix "iaas-"` if we never enter the loop-body with a `iaas-*` domain?

wagnerd3: When I started this task, I thought there might be individual handling per group for the `iaas-` domains at some point. From what I heard in the last weeks, this is unlikely, but all service seeds now use this pattern. Do you want me to change it anyways?

openstack/nova/templates/seed.yaml
```diff
@@ -1,26 +1,18 @@
+{{- $cdomains := .Values.global.domain_seeds.customer_domains | required "missing value for .Values.global.domain_seeds.customer_domains" -}}
+{{- $domains  := concat (list "Default" "ccadmin") $cdomains -}}
```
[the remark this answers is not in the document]
wagnerd3: This was required, but not used anywhere - therefore I removed it.

openstack/nova/templates/seed.yaml
joker-at-work: Don't you plan to upgrade the other seeds? I don't think it makes sense to keep what we don't need, especially since it makes things less readable :)

Btw. if we all use the same pattern, would it make sense to put this into an `openstack/utils` chart function and include it for the necessary parameters? I think only the role and group names change, right? So we'd have to pass in `compute` and would get all the compute_support and compute_admin stuff generated?

wagnerd3: Theoretically, I would agree to tidy up further, use a common chart function where applicable (there are many exceptions/ slight deviations) etc.. In practice, I need to open 15 PRs and chase the code owners for reviews, which I am very tired of. ^^ Someone with better overview/ knowledge than me would have done a better job upfront for sure. The distributed nature of the seeds in the charts makes it hard for me to iteratively improve them.

I would even trigger discussions to improve the overall situation, but the future of the seeder itself is unclear. There is v1 used for service seeds. v2 is used for domain-seeds. Both did not work on the `global` region (I have a PR for that too). v3 is theoretically needed from feature side (e.g. rotation for application credentials), but it's not a priority.

Long story short: I would prefer to merge this version now if you can agree it's semantically correct. I might do another iteration on everything in a few weeks, but I need to shift my focus to other action items for now. (cc @majewsky )

majewsky: Yes please. This was supposed to be a quick Friday-afternoon cleanup, which has ballooned into a several-months-long tarpit. We want to move on to other tasks.

joker-at-work: I guess we have clashing expectations here. I was asked to review - but you just wanted a stamp of approval. My bad, I guess.

joker-at-work (approved):

openstack/nova/templates/seed.yaml
majewsky: I'm not saying that this is a bad suggestion for a further cleanup, but that this should be owned by the teams owning the seed.

majewsky (approved):

PR #8311, opened 2025-03-24
[opensearch-hermes] prometheus plugin new repo for 2.19.1

Kuckkuck:


system/opensearch-hermes/values.yaml
```diff
@@ -131,7 +131,7 @@ opensearch_hermes_master:
   ca:
     crt: DEFINED-IN-SECRETS
   image:
-    tag: 2.19.1
+    tag: 2.18.0
```
joluc, line 134: Back to 2.18?

Kuckkuck: no. fixed it.

joluc (approved):

PR #8327, opened 2025-03-24
[utils] add an option to select default user for rabbitmq and db

s10:
This change introduces an option to select the default user for RabbitMQ and MariaDB template function:
* `ini_sections.default_transport_url`
* `utils.db_url` with sub-functions

Currently, these functions select the default user the following way:
* RabbitMQ selects the 'default' user.
* MariaDB selects the `first` defined database and then chooses the key from users with the same name as the database.

After this change, `defaultUser` option (`.Values.rabbitmq.defaultUser` or `.Values.mariadb.defaultUser`) could be used to select another user when required.

After this change, this user could be selected two ways:
* `defaultUser` option (`.Values.rabbitmq.defaultUser`, `.Values.mariadb.defaultUser`) could be used in service values to select another user when required
* `.Values.global.db.defaultUser` for database and `.Values.global.rabbitmq.defaultUser` for rabbitmq
  * this options should be set in `region/values/openstack-globals.yaml` and used to select default user for all OpenStack services in region

### DIFF
--- a/openstack/utils/Chart.yaml
+++ b/openstack/utils/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: utils
-version: 0.25.0
+version: 0.26.0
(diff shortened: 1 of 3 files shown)


businessbean (approved):
lgtm

galkindmitrii (approved):

PR #8358, opened 2025-03-25
[maia] enforce roles in maia

notque:


viennaa (approved):
Went through it in the call and it not disruptive. Changes make sense to me.

artherd42 (approved):
lgtm

PR #8359, opened 2025-03-26
Update VMPropertiesCollector to collect the VM net ip address

jeoffzheng:
Add below VM properties:
  - metric_suffix: "net_4000_ip" key: "net:4000|ip_address"
  - metric_suffix: "net_4001_ip" key: "net:4001|ip_address"
  - metric_suffix: "net_4002_ip" key: "net:4002|ip_address"

### DIFF
--- a/system/vmware-monitoring/Chart.yaml
+++ b/system/vmware-monitoring/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vmware-monitoring
-version: 2.1.4
+version: 2.1.5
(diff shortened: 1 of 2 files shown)


jeoffzheng:
The previous PR: https://github.com/sapcc/helm-charts/pull/8211 does not work, it did not collect the properties. After check and investigate, modify the collector as the new request...

jeoffzheng (commented):
bump to the chart.yaml with version 2.1.5

viennaa (approved):
Looks reasonable to me.

PR #8364, opened 2025-03-26
[glance] seed generic domains using global.domain_seeds

wagnerd3:


rajivmucheli (approved):

wagnerd3:
Thx. Will watch for errors in QA and prod after rollouts, let me know if you spot anything.

PR #8365, opened 2025-03-26
[arc] seed generic domains using global.domain_seeds

wagnerd3:

The expressions are technically equal, only the `ora` domain was added. If it was excluded on purpose, let me know and I can change it. I excluded iaas- for now.
I ran h3 diff with dyff option against qa-de-1 and eu-de-1 you can have a look at the output here:
[arcQA1diff.txt](https://github.com/user-attachments/files/19465110/arcQA1diff.txt)
[arcDE1diff.txt](https://github.com/user-attachments/files/19465111/arcDE1diff.txt)


ArtieReus:
The service will be shut down in 3 months. No need to updated anything for now. Thx!!

PR #8369, opened 2025-03-26
[nova][pxc-db] Add support for PXC galera cluster: part 1

s10:
* Add optional pxc-db chart dependencies and sample values
* Add required ci/test-values.yaml values
* Use `cell1_transport_url` and `cell1_db_path` helper for cell1.conf similar to the way how cell0 and cell2 are configured
* Replace `db_url_mysql` helper with `utils.db_url` and explicitly pass database name and type to this helper function


s10:
@joker-at-work @seb-kro , could you please take a look at this PR?

PR #8373, opened 2025-03-27
[manila] seed generic domains using global.domain_seeds

wagnerd3:

The expressions are technically equal, only the `ora` domain was added. If it was excluded on purpose, let me know and I can change it. I excluded iaas- for now.
I ran h3 diff with dyff option against eu-de-1 you can have a look at the output here:
[manilaDE1diff.txt](https://github.com/user-attachments/files/19483943/manilaDE1diff.txt)

### DIFF
--- a/openstack/manila/ci/test-values.yaml
+++ b/openstack/manila/ci/test-values.yaml
@@ -20,7 +20,8 @@ global:
         mtu: 9000
   registry: myRegistry
   domain_seeds:
-    skip_hcm_domain: false
+    customer_domains: [ bar, foo, baz ]
+    customer_domains_without_support_projects: [ baz ]
   availability_zones: []
 
 loci:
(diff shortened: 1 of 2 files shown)


Carthaca:
Please skip ORA, it is behind the `global_setup` flag and only used by keystone-global and designate-global afaics

wagnerd3:
@Carthaca are you sure that you are not mixing this up?
`ora` domain has some projects in e.g. eu-de-1, so I assume they also need the support roles there. Additionally, `ora` does not exist in global and I could not find any hardcoding about `ora` in the old designate/ keystone seeds.

galkindmitrii:
Yes, `ora` seem to be present in all regions..

Carthaca:
> @Carthaca are you sure that you are not mixing this up? `ora` domain has some projects in e.g. eu-de-1, so I assume they also need the support roles there. Additionally, `ora` does not exist in global and I could not find any hardcoding about `ora` in the old designate/ keystone seeds.

Thanks for correcting me!

Carthaca (approved):

PR #8375, opened 2025-03-27
[octavia] move osprofiler configuration to secrets

s10:
osprofiler configuration contains a secret hmac value, so it should be in the secret instead of a configmap

### DIFF
--- a/openstack/octavia/templates/etc/_secrets.conf.tpl
+++ b/openstack/octavia/templates/etc/_secrets.conf.tpl
@@ -22,3 +22,8 @@ record_payloads = {{ if .Values.audit.record_payloads -}}True{{- else -}}False{{
 metrics_enabled = {{ if .Values.audit.metrics_enabled -}}True{{- else -}}False{{- end }}
 {{- include "ini_sections.audit_middleware_notifications" . }}
 {{- end }}
+
+{{- if .Values.osprofiler.enabled }}
+# Tracing
+{{- include "osprofiler" . }}
+{{- end }}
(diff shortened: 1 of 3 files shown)


s10:
@notandy @benjamin-sap , please take a look at this PR

openstack/octavia/templates/secrets.yaml
```diff
@@ -17,3 +17,7 @@ type: Opaque
 data:
   secrets.conf: |
     {{ include (print .Template.BasePath "/etc/_secrets.conf.tpl") . | b64enc | indent 4 }}
+{{- if .Values.osprofiler.enabled }}
+  osprofiler.conf: |
```
velp, line 21: I would prefer to do [the same](https://github.com/search?q=repo%3Asapcc%2Fhelm-charts+%22%7B%7B-+if+.Values.osprofiler.enabled+%7D%7D%22&type=code) and keep this configuration as part of `_secrets.conf.tpl` file instead of create a new one.

velp (changes requested):

velp (approved):

openstack/octavia/templates/etc/_secrets.conf.tpl
```diff
@@ -22,3 +22,7 @@ record_payloads = {{ if .Values.audit.record_payloads -}}True{{- else -}}False{{
 metrics_enabled = {{ if .Values.audit.metrics_enabled -}}True{{- else -}}False{{- end }}
 {{- include "ini_sections.audit_middleware_notifications" . }}
 {{- end }}
+
+{{- if .Values.osprofiler.enabled }}
```
BenjaminLudwigSAP, line 26: Please keep the `# Tracing` comment (unless of course it's somehow not correct). Thanks!

BenjaminLudwigSAP (approved):
Just one little ask, but apart from that it looks good to me, thx 👍

PR #8378, opened 2025-03-27
[nova] move osprofiler configuration to secrets

s10:
osprofiler configuration contains a secret hmac value, so it should be in the secret instead of a configmap.

This change moves [osprofiler] configuration section to a separate `osprofiler.conf` file in `nova-etc-secret` and mounts it into containers.


joker-at-work (approved):
I don't really like having non-secrets in a Secret, because that doesn't make sense to me. Since we don't use `osprofiler`, I don't care in this case, though.

seb-kro (approved):

PR #8383, opened 2025-03-27
[domains] switch Values.global_setup to Values.global.is_global_region

wagnerd3:
When working on https://github.com/sapcc/helm-charts/pull/8193 I noticed that the `global_setup` variable is misplaced so I created `global.is_global_region` in the values.
This handles all usages of this new variable in all domain seeds.

The diff is in line with the current diff in the pipe for global + DE2. 

related:
https://github.com/sapcc/helm-charts/pull/8385
https://github.com/sapcc/helm-charts/pull/8384

### DIFF
--- a/openstack/domain-seeds/README.md
+++ b/openstack/domain-seeds/README.md
@@ -27,5 +27,5 @@ Only the following domains should be seeded:
 - ccadmin
 - global
 
-Set **global_setup** value to **true** for such scenario. All other domain seeds will be ignored.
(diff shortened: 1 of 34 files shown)


majewsky:
Given that this deployment appears to be abandoned in favor of the new domain-seeds deployment in the internal repo, we probably will not need this. I would rather have a PR that deletes this chart entirely to avoid future confusion.

wagnerd3:
I am on it to create a PR to remove the old charts, but I need a pipeline in place first to seed global from the new repo. For global, only a pipeline with the old charts exists right now.

PR #8384, opened 2025-03-27
[keystone] switch Values.global_setup to Values.global.is_global_region

wagnerd3:
When working on https://github.com/sapcc/helm-charts/pull/8193 I noticed that the `global_setup` variable is misplaced so I created `global.is_global_region` in the values.
This handles all usages of this new variable in keystone service.

The diff is in line with the current diff in the pipe for global + DE2. 

related:
https://github.com/sapcc/helm-charts/pull/8385
https://github.com/sapcc/helm-charts/pull/8383

### DIFF
--- a/openstack/keystone/ci/test-values.yaml
+++ b/openstack/keystone/ci/test-values.yaml
@@ -9,6 +9,7 @@ global:
   registryAlternateRegion: other.docker.registry
   dockerHubMirror: my.dockerhub.mirror
   dockerHubMirrorAlternateRegion: other.dockerhub.mirror
+  is_global_region: false
   osprofiler:
     jager:
       enabled: true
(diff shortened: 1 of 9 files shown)


openstack/keystone/ci/test-values.yaml
```diff
@@ -9,6 +9,7 @@ global:
   registryAlternateRegion: other.docker.registry
   dockerHubMirror: my.dockerhub.mirror
   dockerHubMirrorAlternateRegion: other.dockerhub.mirror
+  is_global_region: false
```
stanislav-zaprudskiy, line 12: Keystone is not using `global/values/globals.yaml`, so for it the value would appear as not defined. Would it be possible to not set it here explicitly too, to make it similar to the real use?

wagnerd3: @majewsky requested [here](https://github.com/sapcc/helm-charts/pull/8390#discussion_r2018889680) to set the default for services which use the `is_global_region` flag in their default `values.yaml`. I updated this now, is this solution satisfactory for you, too?

wagnerd3:
Ah, I also added here the fix that the seeds are not applied from global/eu-de-1 or global/eu-nl-1.

stanislav-zaprudskiy (approved):

PR #8389, opened 2025-03-28
[neutron] move osprofiler configuration to secrets

s10:
osprofiler configuration contains a secret hmac value, so it should be in the secret instead of a configmap.


mutax (approved):

s10:
@sven-rosenzweig @sebageek , could you please take a look at this PR?

PR #8390, opened 2025-03-28
[billing] seed new billing role and cleanup seeding to global

wagnerd3:
This does a few things:
* add new role for listing flavors
* use `is_global_region` from values (needs pipeline change)
* fix `missing project api_support@global` error in global region
* remove `masterdata_scanner` from global (probably not in use ATM)
* remove `iaas-` support seeding (not necessary according to @majewsky )

[billingDE1diff.txt](https://github.com/user-attachments/files/19503967/billingDE1diff.txt)
[billingGlobalDE2diff.txt](https://github.com/user-attachments/files/19503969/billingGlobalDE2diff.txt)

### DIFF
--- a/openstack/billing/templates/prometheus-alerts.yaml
+++ b/openstack/billing/templates/prometheus-alerts.yaml
@@ -8,7 +8,7 @@ metadata:
     type: alerting-rules
     prometheus: openstack
 
-{{- $keystone_namespace := .Values.is_global | ternary "monsoon3global" "monsoon3" }}
+{{- $keystone_namespace := .Values.global.is_global_region | ternary "monsoon3global" "monsoon3" }}
 
 spec:
   groups:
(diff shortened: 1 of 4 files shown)


openstack/billing/templates/prometheus-alerts.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     type: alerting-rules
     prometheus: openstack
 
-{{- $keystone_namespace := .Values.is_global | ternary "monsoon3global" "monsoon3" }}
+{{- $keystone_namespace := eq .Values.global.is_global_region true | ternary "monsoon3global" "monsoon3" }}
```
majewsky, line 11: `is_global_region` is already a bool, so we don't need to `eq/ne` against `true` here and in several other places:

```suggestion
{{- $keystone_namespace := .Values.global.is_global_region | ternary "monsoon3global" "monsoon3" }}
```

For this to be well-behaved, `values.yaml` should declare a default value:

```yaml
global:
  is_global_region: false
```

majewsky (changes requested):

openstack/billing/templates/prometheus-alerts.yaml
wagnerd3: Ah, I did not go the "default"-way and rather assumed the flag can be undefined in the values. Will change that, thanks!

majewsky (approved):

wagnerd3:
@majewsky I extended the scope here to make use of the new `globals.db_region` where necessary in the billing chart. Do you want to re-review? The 2nd commit holds all the related changes.

majewsky (approved):

PR #8394, opened 2025-03-28
[cronus] seed generic domains using global.domain_seeds

wagnerd3:

I think the change in the below seed is small enough, that looking at diffs of various regions is not necessary.

**Please review, but don't merge.** This depends on a pipeline change that was not `fly`d yet.


wagnerd3:
Pipeline change has been rolled out, PR is ready to review now.

dhalimi (approved):
LGTM

PR #8399, opened 2025-03-28
[Neutron] Run rabbitmq credential-helper as sidecar

sven-rosenzweig:
The RabbitMQ chart offers a credential updater helper, which is executed either as init or sidecar container. The first option is the default setting (based on the global values).

Acutally the rabbitmq container never starts as the credential-update initcontianer never terminates. Seems like it is constantly checking for credential updates.

### DIFF
--- a/openstack/neutron/values.yaml
+++ b/openstack/neutron/values.yaml
@@ -9,6 +9,9 @@ osprofiler:
   enabled: false
 
 global:
+  rabbitmq:
+    native_sidecar:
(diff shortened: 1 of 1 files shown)


mutax (approved):

s10:
>Acutally the rabbitmq container never starts as the credential-update initcontianer never terminates.

This is the expected behaviour of the native sidecar containers, it doesn't prevent the main container from starting.

PR #8401, opened 2025-03-31
[persephone] added new roles to blocklist_roles

ronchi-oss:
[Persephone](https://github.wdf.sap.corp/cc/persephone/) is a new project/service meant to serve as an "API proxy" to the Gardener API, so that customers can easily set up new Shoots in SCI in a couple straight-forward steps.

cc @jknipper @SuperSandro2000

### DIFF
--- a/openstack/keystone/templates/etc/_policy.yaml.tpl
+++ b/openstack/keystone/templates/etc/_policy.yaml.tpl
@@ -56,7 +56,9 @@
   'cloud_identity_viewer':%(target.role.name)s or
   'cloud_support_tools_viewer':%(target.role.name)s or
   'cloud_email_admin':%(target.role.name)s or
-  'cloud_inventory_viewer':%(target.role.name)s"
+  'cloud_inventory_viewer':%(target.role.name)s" or
+  'cloud_persephone_admin':%(target.role.name)s or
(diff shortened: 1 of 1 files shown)


SuperSandro2000 (approved):

openstack/keystone/templates/etc/_policy.yaml.tpl
```diff
@@ -56,7 +56,9 @@
   'cloud_identity_viewer':%(target.role.name)s or
   'cloud_support_tools_viewer':%(target.role.name)s or
   'cloud_email_admin':%(target.role.name)s or
-  'cloud_inventory_viewer':%(target.role.name)s"
+  'cloud_inventory_viewer':%(target.role.name)s" or
+  'cloud_persephone_admin':%(target.role.name)s or
+  'cloud_persephone_viewer':%(target.role.name)s
```
stanislav-zaprudskiy, line 61:
```suggestion
  'cloud_inventory_viewer':%(target.role.name)s or
  'cloud_persephone_admin':%(target.role.name)s or
  'cloud_persephone_viewer':%(target.role.name)s"
```

ronchi-oss:
The project direction is unclear at the moment but it looks like this won't be needed.

PR #8404, opened 2025-03-31
[logs] fix alert text

timojohlo:


notque (approved):
looks reasonable to me.

PR #8405, opened 2025-03-31
Update cinder.alerts

himanip94:
CinderBackendStorageEmpty alerts will only fire if backend is Standard_HDD. Added filter to remove alerts for SSD and FCD.


hemna:
I'm not sure it's wise to only alert on empty storage for standard_hdd.   what about vmware?  vmware_fcd?

PR #8408, opened 2025-04-01
[Neutron] Support Password Ratelimit Backend

sven-rosenzweig:
[Neutron] Support Password Ratelimit Backend
    
    The Redis helm chart enforces the access via a password. If no
    password is set, a random pw is created by container start
    automatically.
    
    So far the openstack-ratelimit middleware does not support connecting to
    a Redis backend with a password set. Requires https://github.com/sapcc/openstack-rate-limit-middleware/pull/29
    The middleware loads the backend password from a file. With this PR we
    mount the Redis password into the neutron-server so it will be picked up
    by the ratelimit middleware.

### DIFF
--- a/openstack/neutron/templates/deployment-server.yaml
+++ b/openstack/neutron/templates/deployment-server.yaml
@@ -139,6 +139,11 @@ spec:
             - mountPath: /etc/neutron/secrets/neutron-arista-secrets.conf
               name: neutron-arista-secrets
               subPath: neutron-arista-secrets.conf
+{{- if .Values.rate_limit.enabled }}
+            - mountPath: {{ .Values.rate_limit.backend_secret_file }}
+              name: neutron-ratelimit-backend-secret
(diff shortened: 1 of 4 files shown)


openstack/neutron/templates/deployment-server.yaml
```diff
@@ -139,6 +139,11 @@ spec:
             - mountPath: /etc/neutron/secrets/neutron-arista-secrets.conf
               name: neutron-arista-secrets
               subPath: neutron-arista-secrets.conf
+{{- if .Values.rate_limit.enabled }}
+            - mountPath: /etc/neutron/ratelimit-backend-secret.conf
```
mutax, line 143: shouldn't the path be taken from the values here as well if we have `backend_secret_file` there as a variable to prevent inconsistencies?

sebageek (approved):

sebageek (approved):

PR #8416, opened 2025-04-02
[mariadb] always try to create root user on pod init

s10:
* `set-root-password` init container now always tries to create `'root'@'localhost'` and `'root'@'%'` user if it doesn't exist

This helps to avoid lock-out issue, if this user was previously deleted


galkindmitrii (approved):

businessbean (approved):
lgtm

PR #8418, opened 2025-04-02
[unbound] introducing the DnsUnbound{1,2}DisproportionateAmountOfTraffic

vssldmtrv:
Normally we expoect both unbounds to get close to 50% of the total traffic. IOW it should be evenly devided between the two unbounds.

The default tolerance is 10%, meaning that the alert will fire if the traffic drops below 40% or jumps above 60%.

We can change that by setting ```$.Values.unbound.tolerable_traffic_distribution_deviation```


SoenkeL (approved):

galkindmitrii (approved):
ok, but the playbook mentions nothing what to do in that case.

PR #8426, opened 2025-04-03
[Openstack-exporter] add values config for cinder

hemna:
This adds the ability to change the 2 following values for the cinderbackend collector:

expected_sharding_backends
allow_unexpected_backends

### DIFF
--- a/prometheus-exporters/openstack-exporter/templates/openstack-exporter-openstack-config.yaml
+++ b/prometheus-exporters/openstack-exporter/templates/openstack-exporter-openstack-config.yaml
@@ -17,6 +17,8 @@ data:
       cinderbackend:
         collector: openstack_exporter.collectors.cinderbackend.CinderBackendCollector
         enabled: True
+        expected_sharding_backends: {{- .Values.collectors.cinderbackend.expected_sharding_backends }}
+        allow_unexpected_backends: {{- .Values.collectors.cinderbackend.set_allow_unexpected_backends }}
       novaservice:
         collector: openstack_exporter.collectors.novaservice.NovaServiceCollector
         enabled: True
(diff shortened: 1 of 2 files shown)


prometheus-exporters/openstack-exporter/templates/openstack-exporter-openstack-config.yaml
```diff
@@ -17,6 +17,12 @@ data:
       cinderbackend:
         collector: openstack_exporter.collectors.cinderbackend.CinderBackendCollector
         enabled: True
+        {{- if .Values.collectors.cinderbackend.expected_sharding_backends }}
```
joker-at-work, line 20: This prohibits us to set this to an empty list. The default, that would be taken from the `openstack-exporter`, is also not an empty list.

prometheus-exporters/openstack-exporter/templates/openstack-exporter-openstack-config.yaml
```diff
@@ -17,6 +17,12 @@ data:
...
+        expected_sharding_backends: {{- .Values.collectors.cinderbackend.expected_sharding_backends }}
+        {{- end }}
+        {{- if .Values.collectors.cinderbackend.allow_unexpected_backends }}
```
joker-at-work, line 23: This prohibits us from setting an explicit `False` value. It would only allow us to get the default from `openstack-exporter`, which is currently defaulting to `False`.

Renaming the variable to `set_allow_unexpected_backends` would transport that information better.

joker-at-work (approved):
Approving, because it works with the currently wanted setup, but it would need a follow-up to work properly.

prometheus-exporters/openstack-exporter/templates/openstack-exporter-openstack-config.yaml
hemna: So, would you prefer getting rid of the if ?

prometheus-exporters/openstack-exporter/templates/openstack-exporter-openstack-config.yaml
hemna: updated

joker-at-work (approved):

PR #8430, opened 2025-04-04
[designate] add reloader annotation to all services deployments

s10:
This PR adds stakater/Reloader annotations to all Designate deployments.


galkindmitrii (approved):
lgtm!

mutax (approved):
I don't want to stand in the way but would feel more comfortable with a different approach.

PR #8432, opened 2025-04-04
keppel: move reloader annotation to correct annotation

SuperSandro2000:
Why are there two annotations 💀


wagnerd3 (approved):

majewsky (approved):
> Why are there two annotations

Because one object contains a spec for another object.

SuperSandro2000:
I refuse to accept that this is a good design.

PR #8436, opened 2025-04-04
[limes] use global.is_global_region in helm-charts

wagnerd3:
* use `global.is_global_region` instead of deriving this info from the `limes_url`
* use `customer_domains = ["global"]` from the values for the global region (less special logic)
* replace `contains "iaas-"` with `hasPrefix "iaas-"` for more robustness

### DIFF
--- a/openstack/limes/templates/seed.yaml
+++ b/openstack/limes/templates/seed.yaml
@@ -1,12 +1,8 @@
 {{- $vbase  := .Values.global.vaultBaseURL | required "missing value for .Values.global.vaultBaseURL" -}}
 {{- $region := .Values.global.region       | required "missing value for .Values.global.region"       -}}
-{{- $is_global := $.Values.limes.clusters.ccloud.catalog_url | contains "global" -}}
+{{/* cdomains equals [ "global" ] in global region */}}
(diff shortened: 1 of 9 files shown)


openstack/limes/templates/seed.yaml
```diff
@@ -1,13 +1,9 @@
 {{- $vbase  := .Values.global.vaultBaseURL | required "missing value for .Values.global.vaultBaseURL" -}}
 {{- $region := .Values.global.region       | required "missing value for .Values.global.region"       -}}
-{{- $is_global := $.Values.limes.clusters.ccloud.catalog_url | contains "global" -}}
+{{/* cdomains equals [ "global" ] in global region */}}
```
SuperSandro2000, line 3:
```suggestion
```

SuperSandro2000 (approved):

majewsky (approved):

majewsky:
I presume you already ran this against global-qa? If not, please check the diff on CI.

wagnerd3:
I did not, else I would have seen that like for billing chart in global, we are missing the `global:globals` values. I will revert for now and check with you tomorrow whether the diff is fine, I found something odd there.

PR #8447, opened 2025-04-07
Revert "[limes] use global.is_global_region in helm-charts"

wagnerd3:
Reverts sapcc/helm-charts#8436


wagnerd3:
Just to unblock prod deployments until I found out about the weird setup of pg-backup in global.

PR #8452, opened 2025-04-08
[elektra] remove SOV_SUPPORT_TOOL_VIEWERS

wagnerd3:
I have no idea how this group was generated for the other domains, no seed has a reference for this, I assume via CAM. Might revert after @rajivmucheli returns.


wagnerd3:
@edda mentioned that this access level is vital to do certain admin things on the domain, so probably we cannot do this "fix".

wagnerd3:
Role added, we don't need this.

PR #8455, opened 2025-04-08
[mariadb] delete unmanaged localhost users

s10:
Previously, init.sql script was creating a pair of users: 'username'@'%' and 'username'@'localhost'.

The second one stopped being managed since the introduction of the granular user configuration.
See https://github.com/sapcc/helm-charts/pull/3598

These @'localhost' users will be removed now.


galkindmitrii (approved):

businessbean (approved):
lgtm

PR #8457, opened 2025-04-08
[mariadb] remove version label from mariadb deployment pod template

s10:
Remove version label from mariadb deployment pod template and related configmaps to avoid unnecessary database restarts on simple chart version update.


businessbean (approved):
lgtm

PR #8462, opened 2025-04-08
Nfs datastore exclusion

frederic-boulianne:
pull request to add nfs datastore type from vrops exporter and also a temp fix of nfs in datastore name until it's pushed to the vrops exported in production
as required by multiple false positive alert of datastore space usage on NFS datastore, we are excluding them in the alerting for datastore space usage
also removed most of vvol reference since there are no more vvol in production

### DIFF
--- a/prometheus-rules/prometheus-vmware-rules/alerts/datastore.alerts
+++ b/prometheus-rules/prometheus-vmware-rules/alerts/datastore.alerts
@@ -1,56 +1,6 @@
 groups:
 - name: datastore.alerts
   rules:
-  - alert: VVolDatastoreNotAccessible
-    expr: |
-      vrops_datastore_summary_datastore_accessible{type='vVOL'} == 0
-    for: 5m
-    labels:
-      severity: critical
-      tier: vmware
-      service: storage
-      support_group: compute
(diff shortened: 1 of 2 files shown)


prometheus-rules/prometheus-vmware-rules/alerts/datastore.alerts
```diff
@@ -138,7 +54,7 @@ groups:
 
   - alert: DataStoreCapacity
     expr: >
-      vrops_datastore_diskspace_total_usage_gigabytes{type!~"ephemeral|vVOL|vmfs.+", datastore!~".+swap"} / vrops_datastore_diskspace_capacity_gigabytes >=  0.8
+      vrops_datastore_diskspace_total_usage_gigabytes{type!~"ephemeral|vmfs.+|nfs", datastore!~".+swap", datastore!~"nfs.*"} / vrops_datastore_diskspace_capacity_gigabytes >=  0.8
```
himanip94, line 57: Can remove the second check for datastore, only excluding the type nfs should take care of it. Please change in warning and critical expressions both

himanip94 (changes requested):
Please change as suggested

himanip94 (approved):

PR #8465, opened 2025-04-09
[andromeda] reverted Oslo policy hardening.

ronchi-oss:
Unfortunately the [previous change](https://github.com/sapcc/helm-charts/pull/8381) makes the Andromeda Liquid Server
unable to perform requests on behalf of Limes. A different rule will
have to be applied.


ronchi-oss:
I'll merge it because it's blocking everything related to Andromeda. I've tested the change on QA and it's effective.

PR #8483, opened 2025-04-09
[nova][mariadb] bump database dependency chart version

s10:
Bump database dependency chart version 0.18.2 -> 0.23.0:
* removes unused and unmanaged username@localhost user
* fixes unnecessary restarts on every chart version update
* adds an option to run a job to rename CHECK constraints to unique names
* adds reloader annotation for backup-v2 deployment

Bump mysql-metrics chart to 0.4.4: adds an option to set vpa main container and reloader annotation

Bump utils chart to 0.26.0:
* adds optional defaultUser option for RabbitMQ and MariaDB configuration
    (though will not be used in Nova)

### DIFF
--- a/openstack/nova/Chart.lock
+++ b/openstack/nova/Chart.lock
@@ -1,13 +1,13 @@
 dependencies:
 - name: mariadb
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-  version: 0.18.2
+  version: 0.23.0
 - name: mariadb
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-  version: 0.18.2
+  version: 0.23.0
 - name: mysql_metrics
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-  version: 0.4.2
+  version: 0.4.4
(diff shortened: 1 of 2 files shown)


s10:
> Lgtm so far.
> 
> * `mariadb.root_password` is still used by `db_url_mysql` from `utils`. But we always pass a tuple and/or have `global.dbPassword` defined in the secrets so that `mariadb.root_password` is never required.

This code path will be partially changed in https://github.com/sapcc/helm-charts/pull/8369

> 
> I guess for us it would then make sense to:
> 
> * Enable `mysql_metrics.vpa.set_main_container`, although I guess it does not make a difference right now since the deployment's pod only has a single container?

This could be done separately, like in Manila in https://github.com/sapcc/helm-charts/pull/8519

> * Enable `mariadb.job.renameCheckConstraints.enabled` to allow for data migration to Percona / MySQL

This job with  `mariadb.job.renameCheckConstraints.enabled` will be running on every helm upgrade, so I think it would make sense to enable it temporarily later, closer to migration.

joker-at-work (approved):

seb-kro (approved):

PR #8488, opened 2025-04-09
[nova] update hermes audit map with unmapped resources

notque:
The unmapped resources are deprecated in the nova api, but they are still being called in production environments and require mapping.

### DIFF
--- a/openstack/nova/templates/etc/_nova_audit_map.yaml.tpl
+++ b/openstack/nova/templates/etc/_nova_audit_map.yaml.tpl
@@ -58,16 +58,28 @@ resources:
     migrations_legacy_api:
         api_name: os-migrations
         type_uri: compute/migrations
+    # os-floating-ips is deprecated, but it's still being called so it needs to be mapped.
+    os-floating-ips:
+        type_uri: compute/floating-ips
+        el_type_uri: compute/floating-ip
     os-server-external-events:
         type_uri: compute/servers/external-events
         type_name: events
         custom_id: tag
         custom_attributes:
           server_uuid: compute/server
+    # os-security-group-rules is deprecated, but it's still being called, must be mapped.
+    os-security-group-rules
+        type_uri: compute/security-group-rules
+        el_type_uri: compute/security-group-rule
     os-volumes_boot:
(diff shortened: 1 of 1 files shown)


openstack/nova/templates/etc/_nova_audit_map.yaml.tpl
```diff
@@ -58,16 +58,28 @@ resources:
...
     os-server-external-events:
         type_uri: compute/servers/external-events
         type_name: events
         custom_id: tag
         custom_attributes:
           server_uuid: compute/server
+    # os-security-group-rules is deprecated, but it's still being called, must be mapped.
+    os-security-group-rules
+        type_uri: compute/security-group-rules
+        el_type_uri: compute/security-group-rule
```
joker-at-work, line 74: If I see it right, setting `el_type_uri` is ignored, because it's always computed from `type_uri` https://github.com/sapcc/openstack-audit-middleware/blob/e1d3491d73bc87a5ee56a6e6600fbbc78e296368/auditmiddleware/_api.py#L229-L240

notque:
> How do I verify this? Does any of this have to be defined in the middleware itself or does this define a urlpath and what the object should be "named" in hermes?
> 
> > # NOTE: proxy-kind API are ignored
> 
> It looks to me like we're now adding proxy-kind APIs.

These define a typeURI. What this is fixing is that when a typeURI is not defined, it maps the resource automatically and adds an X. This is to make sure that they get mapped appropriately. 

The changes were determined by looking in production for nova typeuris that contain an X in the front.

This was found, because events with an X in the typeURI were making it into the documentation for customers.

I added endpoints where there were events in our live productive systems with an X infront of them.

openstack/nova/templates/etc/_nova_audit_map.yaml.tpl
notque: I've removed the el_type_uris as they will only be required if they are different

joker-at-work (approved):
> The changes were determined by looking in production for nova typeuris that contain an X in the front.

That helps. Thank you.

seb-kro (approved):

PR #8491, opened 2025-04-10
Exclude cert-manager owned secrets from secrets-injector webhook

Nuckal777:
This avoids issues with certificates
during chart installation.


system/secrets-injector/templates/webhook.yaml
```diff
@@ -26,6 +26,8 @@ webhooks:
   matchConditions: # webhook is called, when all conditions are true
```
uwe-mayer, line 26: Would it be good to have a way to template custom `matchConditions` as values?

uwe-mayer (approved):
Nice, thank you!
One comment, no blocker

PR #8500, opened 2025-04-10
[postgres + redis] fix type of reloader annotation

wagnerd3:
Should fix:
```
json: cannot unmarshal bool into Go struct field EnvVar.spec.template.spec.initContainers.env.value of type string
```


SuperSandro2000 (approved):

stanislav-zaprudskiy (approved):
thanks! 🩹

PR #8502, opened 2025-04-10
[limes] Add dashboard link to expiring commitment mail

VoigtS:
Mails for expiring commitments can include a link to the dashboard.

### DIFF
--- a/openstack/limes/templates/configmap.yaml
+++ b/openstack/limes/templates/configmap.yaml
@@ -22,7 +22,7 @@ data:
          body: {{ tpl (.Files.Get "files/commitment_mail.tpl") (dict "region" .Values.global.region "condition" .Values.campfire.mail_type.confirmed.condition) | toJson }}
         expiring_commitments:
          subject: {{ .Values.campfire.mail_type.expire.subject }}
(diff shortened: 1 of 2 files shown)


openstack/limes/templates/configmap.yaml
```diff
@@ -1,3 +1,6 @@
+{{- $dashboardURL := printf "https://dashboard.%s.%s/{{ .DomainName }}/{{ .ProjectName }}/resources/v2/project#/Renewal" $.Values.global.region $.Values.global.tld -}}
+{{- $dashboardInfo := printf "<p>Commitments can be renewed through the resource management <a href=%s>dashboard</a></p>." $dashboardURL -}}
```
majewsky, line 2: Is it safe to not use quotes around the `href` attribute here?

majewsky (approved):
Just one thing:

openstack/limes/templates/configmap.yaml
VoigtS: I tested the behavior locally.
If I add quotes `''` the stringification gets messed up and gets rejected by campfire.
If I add quotes ```` the anchor will not be displayed within the mail.
I found only if I leave the quotes out, the link will work correctly.

majewsky (approved):

PR #8517, opened 2025-04-14
[manila] fix audit for subnets

Carthaca:
child cannot be a string, but must be a dict

Since we are touching it anyway, noticed that we can also add metadata.


Carthaca:
@notque fyi

chuan137 (approved):

kpawar-sap (approved):

PR #8520, opened 2025-04-14
[placement][mariadb] bump database dependency chart version

s10:
Bump database dependency chart version 0.18.2 -> 0.23.0:
* removes unused and unmanaged username@localhost user
* fixes unnecessary restarts on every chart version update
* adds an option to run a job to rename CHECK constraints to unique names
* adds reloader annotation for backup-v2 deployment

Bump utils chart to 0.26.0:
* adds optional defaultUser option for RabbitMQ and MariaDB configuration

### DIFF
--- a/openstack/placement/Chart.lock
+++ b/openstack/placement/Chart.lock
@@ -1,7 +1,7 @@
 dependencies:
 - name: mariadb
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-  version: 0.18.2
+  version: 0.23.0
 - name: pxc-db
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
   version: 0.3.1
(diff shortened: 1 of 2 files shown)


seb-kro:
`mariadb.root_password` is still indirectly used by `utils.db_url` via `utils._db_url_mariadb`. But since we do not set `.Values.dbType` we are only calling that function in `_secrets.conf.tpl` with a tuple so that `mariadb.root_password` is never required.

I guess for us it would then make sense to:
- Enable `mariadb.job.renameCheckConstraints.enabled` to allow for data migration to Percona / MySQL

s10:
@fwiesel @joker-at-work @bashar-alkhateeb , could you please review and merge this PR?

joker-at-work (approved):

seb-kro (approved):

PR #8544, opened 2025-04-17
[gatekeeper] Enable GkRestrictMonsoon3Namespace in prod regions

Varsius:


majewsky (approved):
Confirmed that there are no active violations in prod, so this will not break anything.

PR #8550, opened 2025-04-17
common/prometheus-alertmanager: allow full url for playbook

timojohlo:


common/prometheus-alertmanager-base/notification-templates/slack.tmpl
```diff
@@ -24,7 +24,13 @@
...
 
 {{ if .CommonLabels.dashboard }}*<https://plutono.{{ .CommonLabels.region }}.cloud.sap/d/{{ .CommonLabels.dashboard }}|Plutono>* {{ end -}}
-{{ if .CommonLabels.playbook }}*<https://operations.global.cloud.sap/{{ .CommonLabels.playbook }}|Playbook>* {{ end -}}
+{{- if .CommonLabels.playbook }}
+  {{- if hasPrefix "http" .CommonLabels.playbook }}
+    *<{{ .CommonLabels.playbook }}|Playbook>*
+  {{- else }}
+    *<https://operations.global.cloud.sap/{{ .CommonLabels.playbook }}|Playbook>*
+  {{- end }}
+{{- end }}
```
richardtief, line 33: Those templates are executed inside the [alertmanager](https://prometheus.io/docs/alerting/latest/notifications/). I fear we can't use `hasPrefix` in this case.

richardtief (changes requested):
Good idea, but we need to tweak it a bit more.

common/prometheus-alertmanager-base/notification-templates/slack.tmpl
timojohlo: Thanks for the hint! I have adapted it to use regex instead

richardtief (changes requested):

richardtief:
+ ✅

PR #8557, opened 2025-04-22
[manila] don't use rabbitmqadmin for readiness probes

s10:
* Use curl with RabbitMQ Management API instead of rabbitmqadmin python script for readiness checks, because 'rabbitmqadmin' is no longer actively maintained.
* Use curl netrc configuration file instead of passing these values directly
* Use `metrics` user instead of `admin` for readiness checks
* Remove fetch-rabbitmqadmin init container and add create-guru-file when needed


Carthaca (changes requested):
This is a very nice improvement.

Please see my inline comments, the path where the netrc file will end up needs to be slightly adjusted, I think.

Carthaca (approved):

chuan137 (approved):

PR #8562, opened 2025-04-22
[octobus-query-exporter] create a new query: query_elasticsearch_octo…

jeoffzheng:
…bus_Mellanox15mins_issue

create a new mellanox config with 15mins interval to eliminate the double/triple mellanox alerts on slack & the phone call to oncall person. it used for testing purpose, once it works, we will change this production config "vmware_esxi_nic_mellanox_issue.cfg" with 15 mins


kevin-fischer:
@jeoffzheng, pls also bump the version of the octobus-query-exporter chart to **1.0.18** in the vendor directory.
Then update the dependency versions as follows:

- octobus-query-exporter → **1.0.18**
- octobus-query-exporter-global → **1.0.13**

After that, run:

```shell
helm dependency update prometheus-exporters/octobus-query-exporter
``` 

Thx!

Kuckkuck (approved):

max-len:
To test it, you may add an approbiate alert e.g. copy https://github.com/sapcc/helm-charts/blob/master/prometheus-exporters/octobus-query-exporter/vendor/octobus-query-exporter/alerts/scaleout/events.alerts#L20 with info severity. Then we can compare the alerts, if it resolves the issue but also reports reliably.
We need to be conservative here, as missing the event led to MIMs in the past.

kevin-fischer (approved):
+1

PR #8577, opened 2025-04-24
Add greenhouse controlplane kube-monitoring plugin template

videlov:


viennaa (commented):
A new pluginpreset sounds reasonable to me.

My question would be: Are we assuming the type controlplane cluster is always providing the gardener prometheus?

videlov:
> A new pluginpreset sounds reasonable to me.
> 
> My question would be: Are we assuming the type controlplane cluster is always providing the gardener prometheus?
> 
> Edit: Chart bump pls

tbh `clusterType=controlplane` is very general and I do not know if we assume is always the case as for the gardener prometheus, I would propose to specify something more specific like e.g. `gardenerControlPlane`

@SchwarzM do we have controlplane clusters that we want to onboard to greenhouse that are not running Gardener ? what do you think?

SchwarzM:
the runtime and management clusters have vastly different roles and qualify both as control plane. please do not start a preset yet we might need to evolve the cluster and it's role dramatically as we hit the vs-nfd requirements.

the infrastructure gardener will only exist one time worldwide in eu-de-1.

system/greenhouse-ccloud/templates/controlplane-plugin.yaml
```diff
@@ -0,0 +1,115 @@
+{{- if and .Values.kubeMonitoring.enabled .Values.pluginPreset.enabled .Values.kubeMonitoring.controlplaneCluster -}}
```
viennaa, line 1: Why pluginPreset?

viennaa (commented):
And a chart bump again pls :)

richardtief:
I suggest not creating another plugin, as most of the values ​​are the same as the default kube-monitoring. I'll propose the necessary changes in the Slack thread.

richardtief:
superseded with https://github.com/sapcc/helm-charts/pull/8583

PR #8578, opened 2025-04-24
greenhouse - change kvm monitoring namespace

frederic-boulianne:
change the namespace for the KVM monitoring plugin to use it's own namespace instead of using the kube-monitoring namespace


IvoGoman (commented):
The `releaseNamespace` is immutable, this means we need to delete and then recreate the PluginPreset, which will result in a fresh install. Just let me know if this is what is desired.

frederic-boulianne:
Hello Ivo
Yes it is what we desire, we need to have a separate namespace and we want to test it before we go in production

IvoGoman (approved):

IvoGoman:
PluginPreset is recreated with the new releaseNamespace

PR #8591, opened 2025-04-28
[win-exporter] aatpsensor takes longer to come up after reboot.

BerndKue:
Extending alert time for this service


viennaa (changes requested):
Chart bump please, then we're good to go.

viennaa (approved):

PR #8622, opened 2025-05-06
fix(disco): use external image repository

IvoGoman:
this chart is used externally as well as internally the default repository should be the one externally available.

follow-up PR will set the replica repository in the kube-system charts.


kengou:
Can we wait a little bit with this? The chart is in a move out from `kube-system` chart, and I would do the repo update once everything is done

PR #8635, opened 2025-05-07
Switch to gardener-operator extensions in cc-gardener

Nuckal777:


defo89:
should we use `ociRepository.repository=keppel.global.cloud.sap/ccloud-helm/` in all cases?

defo89 (approved):

Nuckal777:
Ideally, we can use the upstream helm charts, but some of them have HA assumptions backed in, which we need to patch out.

PR #8641, opened 2025-05-08
[mariadb] version bumped to 10.6.22

businessbean:
- InnoDB race condition has been fixed
- CVEs fixed:
- CVE-2025-30722
- CVE-2025-30693
- CVE-2023-52970
- CVE-2023-52969
- chart version bumped


businessbean:
The image is not yet available https://hub.docker.com/_/mariadb/tags?name=10.6.22

common/mariadb/values.yaml
```diff
@@ -4,7 +4,7 @@
 # name: value
 
 name: null
-image: library/mariadb:10.6.21
+image: library/mariadb:10.6.22
```
s10, line 7: We should update version in the entrypoint script https://github.com/sapcc/helm-charts/blob/master/common/mariadb/scripts/docker-entrypoint.sh#L223

s10 (changes requested):

common/mariadb/values.yaml
businessbean: fixed, by adding a regex lookup

s10 (changes requested):

businessbean:
Meanwhile the image is available on the Docker hub.

bashar-alkhateeb (approved):

PR #8644, opened 2025-05-08
postgresql-ng: allow creating extra DBs (second attempt)

majewsky:
I took a jab at completing the work in #8543. To make the values interface more coherent, I merged `postgresDatabase` and `extraDatabases` together into `.Values.databases`. This means that this is a major version bump, and I also indulged myself and cleaned up a bunch more stuff. Since this has caused the PR to balloon significantly, I'm putting it as a separate PR, just so that the old PR is preserved in case this changeset proves to be too involved.

**Test status**: The pipeline is now green when building from this branch, and also there is a secrets repo PR changing the postgresql-test chart to use the new values structure.

### DIFF
--- a/common/postgresql-ng/templates/configmap.yaml
+++ b/common/postgresql-ng/templates/configmap.yaml
@@ -33,7 +33,7 @@ metadata:
   name: {{ template "postgres.fullname" . }}-sql-on-startup
 data:
   # first phase of SQL on startup: statements that are executed in the "postgres" database
-  phase1.sql: |
+  phase1-system.sql: |
     {{- if hasKey .Values.extensions "pg_stat_statements" }}
     CREATE EXTENSION IF NOT EXISTS pg_stat_statements;
     {{- end }}
--- a/common/postgresql-ng/docker/entrypoint.sh
+++ b/common/postgresql-ng/docker/entrypoint.sh
@@ -210,9 +210,9 @@ PGDATABASE='' process_sql --dbname postgres --set user="$PGUSER" --set password_
   ALTER USER :user WITH PASSWORD :'password';
 EOSQL
 
-if [ -f /sql-on-startup.d/phase1.sql ]; then
-  echo "Processing /sql-on-startup.d/phase1.sql..."
-  PGDATABASE='postgres' process_sql -f <(substituteSqlEnvs /sql-on-startup.d/phase1.sql)
+if [ -f /sql-on-startup.d/phase1-system.sql ]; then
+  echo "Processing /sql-on-startup.d/phase1-system.sql..."
+  PGDATABASE='postgres' process_sql -f <(substituteSqlEnvs /sql-on-startup.d/phase1-system.sql)
 fi
 for FILE in /sql-on-startup.d/phase2-*.sql; do
   DB="$(basename "$FILE" | sed 's/^phase2-\(.*\)\.sql$/\1/')"


common/postgresql-ng/templates/_helpers.tpl
```diff
@@ -38,21 +34,44 @@ some space for the name suffixes on replicasets and pods.
...
+    {{- fail (printf "postgres-ng: Missing owner name for database %q" $db) }}
+  {{- else if not (hasKey $.Values.users $cfg.owner) }}
+    {{- fail (printf "postgres-ng: Invalid owner name for database %q (user %q is not declared in .Values.users)" $db $cfg.owner) }}
+  {{- end }}
 {{- end }}
 
-{{- if (contains "_" $db) }}
-{{- fail "postgres-ng: extraDatabases entry cannot contain underscores!" }}
+{{- if eq (len .Values.users) 0 }}
+  {{- fail "postgres-ng: need at least one entry in .Values.users" }}
```
SuperSandro2000, line 61:
```suggestion
  {{- fail "postgres-ng: needs at least one entry in .Values.users" }}
```

common/postgresql-ng/templates/configmap.yaml
```diff
@@ -18,69 +18,76 @@ data:
...
-    {{- /* duplicated with the code in rbac because the templating is to limited in its features */}}
-    {{- $users := .Values.users }}
-    {{- if .Values.persistence.createBackupUser }}
-      {{- $users := merge $users (dict "backup" (dict)) }}
-    {{- end }}
-    {{- if .Values.createMetricsUser }}
-      {{- $users := merge $users (dict "metrics" (dict)) }}
-    {{- end }}
+  # first phase of SQL on startup: statements that are executed in the "postgres" database
+  phase1.sql: |
```
SuperSandro2000, line 37:
```suggestion
```

SuperSandro2000:
> [ERROR] templates/: parse error at (postgresql-ng/templates/rbac.yaml:25): undefined variable "$users"

common/pgmetrics/templates/_helpers.tpl
```diff
@@ -0,0 +1,7 @@
+{{- if .Values.db_name }}
+{{- fail "pgmetrics: db_name is no longer supported! Please use .Values.databases instead." }}
+{{- end }}
+
+{{- if eq (len .Values.databases) 0 }}
+  {{- fail "pgmetrics: needs at least one entry in .Values.databases" }}
```
wagnerd3, line 6: Considering that we default to the hostname-DB in `alerts.yaml`, should this error message rather have a hint like `omit .Values.databases if you want the default (hostname-DB) to be used`? Else, this might read like you need `.Values.databases` in any case.

common/pgmetrics/templates/_helpers.tpl
```diff
@@ -0,0 +1,7 @@
...
+{{- end }}
```
wagnerd3, line 7: Do we need to assert the used properties of `.Values.databases` here? (`.name`)

common/postgresql-ng/values.yaml
```diff
@@ -41,23 +34,28 @@ config:
...
-    # Ensure the user has those grant lines run on each start.
-    # TO $USER with the username is added per line.
-    # %PGDATABASE% is replaced with the actual database name.
-    # For example, a read-only application user would need `GRANT CONNECT,SELECT ON DATABASE "%PGDATABASE%"`
-    #grant:
-    #- '"GRANT CONNECT,SELECT ON DATABASE "%PGDATABASE%"'
-    # If auto generating secrets doesn't work for you, you can also create them manually and refence them here:
+# Databases to create.
+databases: {}
+  # example:
```
wagnerd3, line 39: Somehow the ordering of the comment lines here does not make sense to me. The example refers to the last paragraph, right? can we move it down there?

common/postgresql-ng/templates/configmap.yaml
```diff
@@ -17,86 +17,77 @@ data:
...
-    {{- $users := .Values.users }}
-    {{- if .Values.persistence.createBackupUser }}
-      {{- $users := merge $users (dict "backup" (dict)) }}
-    {{- end }}
-    {{- if .Values.createMetricsUser }}
-      {{- $users := merge $users (dict "metrics" (dict)) }}
-    {{- end }}
-
+  # first phase of SQL on startup: statements that are executed in the "postgres" database
+  phase1.sql: |
```
SuperSandro2000, line 36:
```suggestion
  postgres-db.sql: |
```

common/postgresql-ng/templates/configmap.yaml
```diff
@@ -17,86 +17,77 @@ data:
...
-    {{- end }}
     {{- end }}
 
-    ALTER SCHEMA public OWNER TO "{{ .Values.tableOwner | required ".Values.tableOwner must be set to one user!" }}";
-    ALTER DATABASE "%PGDATABASE%" OWNER TO "{{ .Values.tableOwner }}";
-    SELECT format('ALTER TABLE %I.%I.%I OWNER TO %I;', table_catalog, table_schema, table_name, '{{ .Values.tableOwner }}')
+  # second phase of SQL on startup: statements that are executed in the respective payload databases
+  {{- range $dbName := keys .Values.databases | sortAlpha }}
+  {{- $settings := index $.Values.databases $dbName | default (dict) }}
+  phase2-{{ $dbName }}.sql: |
```
SuperSandro2000, line 78:
```suggestion
  {{ $dbName }}-db.sql: |
```

common/pgmetrics/templates/_helpers.tpl
majewsky: There are no required settings inside `.Values.databases[name]` at the moment, so I don't know what else to assert here.

common/pgmetrics/templates/_helpers.tpl
majewsky: The problem is that I forgot to update `alerts.yaml`. Fixed now.

common/postgresql-ng/templates/configmap.yaml
majewsky: See below.

common/pgmetrics/templates/_helpers.tpl
wagnerd3: Now it's gone due, in the old code `.name` was used. :-)

wagnerd3 (approved):

common/postgresql-ng/templates/configmap.yaml
SuperSandro2000: We could phase1 system-db.sql maybe?

SuperSandro2000 (approved):

PR #8663, opened 2025-05-12
hermes: adjust user sidecar for rabbitmq

notque:
Trying to fix the user credential sidecar, it shouldn't be in init containers, and is failing.


notque:
Deploy from branch validates this fixed the user credential sidecar no longer being in init containers, and instead is a sidecar container as expected.

PR #8678, opened 2025-05-13
[nova][vmware-rebalancer] introduce alert for (almost) full vCenters

anokfireball:
Consuming the newly introduced metrics from VMware Rebalancer, this change set will alert somebody to take action when VCs run critically low on resources.

### DIFF
--- a/openstack/nova/alerts/openstack/vmware-rebalancer.alerts
+++ b/openstack/nova/alerts/openstack/vmware-rebalancer.alerts
@@ -10,8 +10,7 @@ groups:
           support_group: compute
           tier: vmware
           context: "vCenter Capacity {{ $labels.vcenter }}"
-          meta: "The vCenter `{{ $labels.vcenter }}` is (almost) full and only allows scheduling less than 4 VMs."
-          playbook: docs/support/playbook/nova/alerts/alert-vcenter-full.md
+          meta: "The vCenter `{{ $labels.vcenter }}` is (almost) full and only allows scheduling less than 4 VMs of flavor {{ $labels.flavor }}."
         annotations:
-          description: "The vCenter `{{ $labels.vcenter }}` is (almost) full and only allows scheduling less than 4 VMs."
-          summary: VMware Rebalancer Metric to check if the vCenter has enough remaining capacity.
+          description: "The vCenter `{{ $labels.vcenter }}` is (almost) full and only allows scheduling less than 4 VMs of flavor {{ $labels.flavor }}."
+          summary: "VMware Rebalancer Metric to check if the vCenter has enough remaining capacity."
--- a/openstack/nova/alerts/openstack/vmware-rebalancer.alerts
+++ b/openstack/nova/alerts/openstack/vmware-rebalancer.alerts
@@ -2,7 +2,7 @@ groups:
   - name: vmware-rebalancer.alerts
     rules:
       - alert: VcenterLowNumberOfPlaceableVms
-        expr: wm_vmware_rebalancer_placeable_vms_per_vcenter{has_bbs_with_custom_hana_exclusive_host="False",has_bbs_without_custom_hana_exclusive_host="True",flavor="g_c128_m512"} <= 3
+        expr: wm_vmware_rebalancer_placeable_vms_per_vcenter{has_bbs_without_custom_hana_exclusive_host="True",flavor="g_c128_m512"} <= 3
         for: 30m
         labels:
           severity: info
--- a/openstack/nova/alerts/openstack/vmware-rebalancer.alerts
+++ b/openstack/nova/alerts/openstack/vmware-rebalancer.alerts
@@ -1,0 +1,17 @@
+groups:
+  - name: vmware-rebalancer.alerts
+    rules:
+      - alert: VmwareRebalancerLowPlacableVMsPerVcenter
+        expr: wm_vmware_rebalancer_placeable_vms_per_vcenter{has_bbs_with_custom_hana_exclusive_host="False",has_bbs_without_custom_hana_exclusive_host="True",flavor="g_c128_m512"} <= 3
+        for: 30m
+        labels:
+          severity: warning
+          service: compute
+          support_group: compute
+          tier: vmware
+          context: "vCenter Capacity {{ $labels.vcenter }}"
+          meta: "The vCenter `{{ $labels.vcenter }}` is (almost) full and only allows scheduling less than 4 VMs."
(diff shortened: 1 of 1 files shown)


openstack/nova/alerts/vmware-rebalancer/vmware-rebalancer.alerts
```diff
@@ -0,0 +1,17 @@
...
+        expr: wm_vmware_rebalancer_placeable_vms_per_vcenter{has_bbs_with_custom_hana_exclusive_host="False",has_bbs_without_custom_hana_exclusive_host="True",flavor="g_c128_m512"} <= 3
+        for: 30m
+        labels:
+          severity: warning
+          service: compute
+          support_group: compute
+          tier: vmware
+          context: "vCenter Capacity {{ $labels.vcenter }}"
+          meta: "The vCenter `{{ $labels.vcenter }}` is (almost) full and only allows scheduling less than 4 VMs."
+          playbook: docs/support/playbook/nova/alerts/cc3test-alert-vcenter-full.md
```
anokfireball, line 14: I followed naming convention of the other files, but am unsure why they have prefix `cc3test`.

- [ ] Also, this playbook does not exist yet.

joker-at-work: cc3test prefix comes from the cc3tests that we're running - our smoketests. Since this alert doesn't come from cc3test, it should have that prefix.

joker-at-work:
Some years ago, we were told to create any new alert as info-level alert, so we could get some experience with it and see if it works until some on-duty person gets it into their channel to work on it.
At the same time, starting with getting the level warning, a usable playbook needs to exist. Since we're not the ones working on this alert, some cross-team collaboration will be necessary to create that playbook. Therefore, I suggest to start with info-level alerts and present the alert in our call with the compute operations colleagues.

openstack/nova/templates/prometheus-alerts.yaml
```diff
@@ -1,6 +1,6 @@
 {{- $values := .Values }}
 {{- if $values.alerts.enabled }}
-{{- range $target := list "kubernetes" "openstack" "metrics-metal" }}
+{{- range $target := list  "metrics-metal" "openstack" }}
```
joker-at-work, line 3: additional space.

why do you see a need for that change? commit-message of [[nova] update Prometheus Alert target folders](https://github.com/sapcc/helm-charts/pull/8678/commits/4b923c83307cc330a210d30b46e71010cc83166b) seems empty

openstack/nova/alerts/openstack/vmware-rebalancer.alerts
```diff
@@ -0,0 +1,17 @@
...
+        labels:
+          severity: info
+          service: compute
+          support_group: compute
+          tier: vmware
+          context: "vCenter Capacity {{ $labels.vcenter }}"
+          meta: "The vCenter `{{ $labels.vcenter }}` is (almost) full and only allows scheduling less than 4 VMs."
+          playbook: docs/support/playbook/nova/alerts/alert-vcenter-full.md
+        annotations:
+          description: "The vCenter `{{ $labels.vcenter }}` is (almost) full and only allows scheduling less than 4 VMs."
```
joker-at-work, line 16: Should we add the flavor i.e. `less then 4 VMs of flavor g_c128_m512`?

openstack/nova/alerts/openstack/vmware-rebalancer.alerts
```diff
@@ -0,0 +1,17 @@
...
+        expr: wm_vmware_rebalancer_placeable_vms_per_vcenter{has_bbs_with_custom_hana_exclusive_host="False",has_bbs_without_custom_hana_exclusive_host="True",flavor="g_c128_m512"} <= 3
+        for: 30m
+        labels:
+          severity: info
+          service: compute
+          support_group: compute
+          tier: vmware
+          context: "vCenter Capacity {{ $labels.vcenter }}"
+          meta: "The vCenter `{{ $labels.vcenter }}` is (almost) full and only allows scheduling less than 4 VMs."
+          playbook: docs/support/playbook/nova/alerts/alert-vcenter-full.md
```
joker-at-work, line 14: if the playbook doesn't exist, imho don't add one. compute colleagues might want to have it in their namespace in the docs..

openstack/nova/alerts/openstack/vmware-rebalancer.alerts
```diff
@@ -0,0 +1,16 @@
...
+        labels:
+          severity: info
+          service: compute
+          support_group: compute
+          tier: vmware
+          context: "vCenter Capacity {{ $labels.vcenter }}"
+          meta: "The vCenter `{{ $labels.vcenter }}` is (almost) full and only allows scheduling less than 4 VMs of flavor {{ $labels.flavor }}."
+        annotations:
+          description: "The vCenter `{{ $labels.vcenter }}` is (almost) full and only allows scheduling less than 4 VMs of flavor {{ $labels.flavor }}."
+          summary: VMware Rebalancer Metric to check if the vCenter has enough remaining capacity.
```
seb-kro, line 16: Maybe also quote summary for consistency / precaution against future updates?

anokfireball:
Dropped the commit that removes the `kubernetes` target.

seb-kro (commented):
I believe this will create PrometheusRule `alerts-openstack-vmware-rebalancer.alerts`. Not sure about our naming conventions, but I think most (or all?) Nova resources carry `nova` in their name which is also quite helpful IMO. So maybe it makes sense to e.g. move the alert to the existing `openstack/nova.alerts` or prefix the file with `nova-`? Or perhaps better, add a prefix for all PrometheusRules directly in the template.

openstack/nova/alerts/openstack/vmware-rebalancer.alerts
```diff
@@ -0,0 +1,16 @@
+groups:
+  - name: vmware-rebalancer.alerts
+    rules:
+      - alert: VcenterLowNumberOfPlaceableVms
+        expr: wm_vmware_rebalancer_placeable_vms_per_vcenter{has_bbs_with_custom_hana_exclusive_host="False",has_bbs_without_custom_hana_exclusive_host="True",flavor="g_c128_m512"} <= 3
```
anokfireball, line 5: `has_bbs_with_custom_hana_exclusive_host="False"` excludes vCenters that should not be excluded (false negatives).

seb-kro (approved):
I assume it is intended to publish this with `warning` severity despite no proper playbook.

Phalut (approved):
Approved.

PR #8715, opened 2025-05-16
[neutron] update database dependencies

s10:
* update mariadb chart from 0.24.0 to 0.24.1

Sets max_connection_errors options to avoid prevent blocking database connections because of the failed monitor checks or other network issues.

* update mysql-metrics chart to 0.4.4

Adds reloader annotation to sql-exporter deployment


mutax (approved):
please merge, we had another downtime because the DB got unreachable

PR #8717, opened 2025-05-16
[placement] bump shared dependencies

s10:
* update mariadb to 0.24.1

Updates mariadb to 10.6.21 and set max_connection_errors options to avoid prevent blocking database connections because of the failed monitor checks or other network issues

* update pxc-db to 0.4.0

Updates pxc custom resource to 1.18.0 version

* update memcached to 0.6.10

### DIFF
--- a/openstack/placement/Chart.lock
+++ b/openstack/placement/Chart.lock
@@ -1,13 +1,13 @@
 dependencies:
 - name: mariadb
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-  version: 0.23.0
+  version: 0.24.1
 - name: pxc-db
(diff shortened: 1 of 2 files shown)


fwiesel (approved):

bashar-alkhateeb:
Tthis change is disruptive i guess, the diff contains a DB update correct ?
@seb-kro  would you please check this PR ?

fwiesel:
Yes, it would be disruptive. But not changing the version includes the risk of the database not accepting any connections due to too many connection errors.

seb-kro (approved):
Don't see any issues. For nova we also already deployed these mariadb updates to QA without experiencing any problems.

PR #8718, opened 2025-05-16
[nova] bump mariadb dependency

s10:
* update mariadb to 0.24.1

Updates mariadb to 10.6.21 and set max_connection_errors options to avoid prevent blocking database connections because of the failed monitor checks or other network issues


s10:
Merged https://github.com/sapcc/helm-charts/pull/8725

PR #8736, opened 2025-05-20
[prometheus-rules] - prometheus-vmware-rules | Updated template for bedrock labeling

christophrichtersap:
The current code just templates the expression we need to add also the label for bedrock.

Testing done:
- [x] local helm template
- [x] lint code
- [x] updated template manual in `qa-de-1`
- [x] check labels are assigned correctly in alertmanager on `qa-de-1` after manual changes

When `bedrock: nil`, the label simply doesn't appear in Alertmanager, which is acceptable.
Can you confirm that the same behavior is implemented across all environments except `qa-de-1`?

### DIFF
--- a/prometheus-rules/prometheus-vmware-rules/Chart.yaml
+++ b/prometheus-rules/prometheus-vmware-rules/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A collection of Prometheus alert rules.
 name: prometheus-vmware-rules
-version: 1.3.0
+version: 1.4.0
 dependencies:
   - name: owner-info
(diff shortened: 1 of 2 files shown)


prometheus-rules/prometheus-vmware-rules/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A collection of Prometheus alert rules.
 name: prometheus-vmware-rules
-version: 1.3.0
+version: 1.4.0
```
viennaa, line 4:
```suggestion
version: 1.3.1
```

viennaa (approved):
Thanks!

PR #8746, opened 2025-05-20
gk: add pod-security-v2 rule

Varsius:


system/gatekeeper/templates/constrainttemplate-pod-security-v2.yaml
```diff
@@ -0,0 +1,211 @@
...
+              # helper functions
+
+              # Begin at after first "/" and end at ":"
+              # i.e. keppel.REGION.cloud.sap/ccloud-dockerhub-mirror/cloudprober/cloudprober:v0.13.9
+              # becomes ccloud-dockerhub-mirror/cloudprober/cloudprober
+              extract_repository(image) := result if {
+                  start := indexof(image, "/") + 1
+                  end := indexof(image, ":")
+                  result := substring(image, start, end - start)
+              }
```
majewsky, line 202: `hostname/repo:tag` is not the only form of image reference that we will see in practice. Please ensure that the following inputs also parse correctly:

```
# existing testcase
foo.image-registry.com/test-repo/dummy:20250519150208 -> test-repo/dummy

# new testcase: digest only, no tag
foo.image-registry.com/test-repo/dummy@sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 -> test-repo/dummy

# new testcase: both tag and digest
foo.image-registry.com/test-repo/dummy:20250519150208@sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 -> test-repo/dummy

# new testcase: neither tag nor digest ("latest" tag is implied)
foo.image-registry.com/test-repo/dummy -> test-repo/dummy

# new testcase: no hostname (please repeat this with each combination of tag-or-not and digest-or-not)
test-repo/dummy:20250519150208 -> test-repo/dummy
```

These can all be put into a single testcase by listing several containers that all use a privileged feature, and checking that the allowlist matches as expected.

system/gatekeeper/tests/gatekeeper-policies/fixtures/pod-security-v2/host-path-volumes-allow.yaml
```diff
@@ -0,0 +1,16 @@
...
+  containers:
+  - name: foo-container
+    image: foo.image-registry.com/test-repo/dummy:20250519150208
+    volumeMounts:
+    - mountPath: /hostfs
+      name: hostfs
+  volumes:
+  - name: hostfs
+    hostPath:
+      path: /
```
majewsky, line 16: This should also demonstrate a second container that uses an image not on the allowlist, but is allowed because it does not mount the volume in question.

system/gatekeeper-config/templates/constraint-pod-security-v2.yaml
```diff
@@ -0,0 +1,45 @@
...
+  match: {{ include "match_pods_and_pod_owners" . | indent 4 }}
+  parameters:
+    allowlist:
+      {{- if eq .Values.cluster_type "test" }}
+      - matchNamespace: foo
+        matchRepository: foo-repo/dummy
+        mayUseCapabilities:
+          - CAP_FOO
+      - matchNamespace: test-allow-all
+        matchRepository: test-repo/dummy
```
majewsky, line 25: To make the testcases easier to follow, I suggest using more speaking repository names, e.g. `test-app/highly-privileged` or `test-app/unprivileged`.

majewsky (changes requested):
This is looking very good, but I would like to request some more polish:

majewsky (approved):
I'm giving my approval for the commits submitted so far, except for a few nitpicks. Next reviewer, please review my commits.

VoigtS (approved):

PR #8748, opened 2025-05-20
[greenhouse] kube-monitoring pluginpreset structure refinement

trouaux:
improvement to kube-monitoring pluginpreset to avoid repetition of settings

https://github.wdf.sap.corp/cc/secrets/pull/17213


system/greenhouse-ccloud/templates/kube-monitoring-pluginpreset.yaml
```diff
@@ -15,21 +15,24 @@ metadata:
...
   - clusterName: {{ $cluster }}
     overrides:
     - name: global.commonLabels
       value:
-        service: ceph
-        support_group: storage
+        service: {{ $serviceName }}
+  {{- if index $serviceValues "support_group" }}
+        support_group: {{ $serviceValues "support_group" }}
+  {{- end }}
```
viennaa, line 27: We want to enforce `service`, `support_group`, don't we? Apparently there is no `support_group` for kvm, but this needs to be added. So, I would not check if `support_group` is available, we map it and let it fail if it is not set.

system/greenhouse-ccloud/templates/kube-monitoring-pluginpreset.yaml
```diff
@@ -55,115 +58,9 @@ spec:
     - name: kubeMonitoring.prometheus.prometheusSpec.thanos.objectStorageConfig.existingSecret.key
       value: thanos.yaml
   {{- end }}
+  {{- if index $serviceValues special }}
+  {{- index $serviceValues special | nindent 4}}
```
viennaa, line 62: Why do we name it `special`?

trouaux: renamed to extraSpecs, not sure if we should add `toYaml`

helm template is a bit tricky using secrets with all vault values, any workaround ?

trouaux: for ref
`h template . -f ./ci/test-values.yaml -s templates/kube-monitoring-pluginpreset.yaml`

viennaa (approved):
Good to go!

viennaa (approved):

PR #8751, opened 2025-05-20
[maia] global seed support

notque:
Add Maia global seed support, should have done this at the same time as hermes. My mistake. Appreciate it.

### DIFF
--- a/openstack/maia/templates/seed.yaml
+++ b/openstack/maia/templates/seed.yaml
@@ -11,9 +11,10 @@ metadata:
     component: maia
 spec:
   requires:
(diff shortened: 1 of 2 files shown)


openstack/maia/templates/seed.yaml
```diff
@@ -11,9 +11,10 @@ metadata:
     component: maia
 spec:
   requires:
+  - {{ .Values.global.keystoneNamespace | default "monsoon3" }}/{{ .Values.global.is_global_region | ternary "keystone-global-seed" "keystone-seed" }}
```
wagnerd3, line 14: `default "monsoon3"` could be omitted due to the default in the `values.yaml` - semantically, it's fine either way :-)

wagnerd3 (approved):

PR #8785, opened 2025-05-23
elektra: revert cloud_resource_admin assignment

majewsky:
Campfire does not look at this role anymore since #8770.


ArtieReus (approved):
thx

PR #8793, opened 2025-05-26
controlplane-operations-pluginpreset.yaml

videlov:


richardtief (commented):
Thanks. One question below

PR #8806, opened 2025-05-27
[neutron] remove redundant whitespace in templates

felix-kaestner:
This patch ensures a more consistent format of the neutron manifests, removing redundant whitespace.

### DIFF
--- a/openstack/neutron/templates/configmap-bin.yaml
+++ b/openstack/neutron/templates/configmap-bin.yaml
@@ -11,5 +11,3 @@ data:
 {{ include (print .Template.BasePath "/bin/_neutron-server-start.tpl") . | indent 4 }}
   common.sh: |
 {{ include "common.sh" .| indent 4 }}
-
-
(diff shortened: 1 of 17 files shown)


notandy:
this pr doesn't remove whitespaces at all, it removes newlines. I personally like the visual separation of metadata and payload via an extra newline.

felix-kaestner:
> this pr doesn't remove whitespaces at all, it removes newlines.

Don't newlines fall under the category of ["whitespace characters"](https://learn.microsoft.com/en-us/cpp/c-language/white-space-characters?view=msvc-170)? 

> I personally like the visual separation of metadata and payload via an extra newline.

I'm fine to keep it. However, in this case I would prefer to have it at least consistent, i.e. following a certain rule. Currently it's a bit of a mess, with newlines sometimes placed before, sometimes after the `metadata`. And then there are templates like `configmap-bin.yaml` or  `configmap-etc-apod.yaml` with multiple consecutive newlines.

fwiesel (approved):

PR #8808, opened 2025-05-27
[utils][proxysql] Add restartPolicy to support native sidecar mode

s10:
Add restartPolicy option, which is enabled when the `proxysql.native_sidecar` option is set.

This would allow us to move the proxysql sidecar from the `containers` deployment section to `initContainers` and run it as a native sidecar.


fwiesel (approved):

galkindmitrii:
Wouldn't that break in `qa-de-2`?

galkindmitrii (approved):

PR #8816, opened 2025-05-28
postgresql, redis: fix annotateGeneratedSecrets

stanislav-zaprudskiy:
The feature originally introduced in #6396 was not working properly because of the conditional always mistakenly evaluating to false.

```console
$ docker run --rm -it keppel.eu-de-1.cloud.sap/ccloud/shared-app-images/alpine-kubectl:latest-latest ash -c '[[ "true" -eq "true" ]] && echo "yes" || echo "no"'
ash: true: out of range
no
```

The issue appears to be `ash` specific.
```console
$ bash -c '[[ "true" -eq "true" ]] && echo "yes" || echo "no"'
yes
```

### DIFF
--- a/common/postgresql-ng/Chart.yaml
+++ b/common/postgresql-ng/Chart.yaml
@@ -43,4 +43,7 @@ description: Chart for PostgreSQL
 #
 # 2.0.3
 # - Fix DB name double quoting issues in startup sql script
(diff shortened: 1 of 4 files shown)


SuperSandro2000 (approved):
Lgtm

wagnerd3 (approved):

PR #8826, opened 2025-05-30
[neutron] remove unused neutron-ironic-agent deployment

felix-kaestner:
With this patch we are removing the Deployment of the neutron-ironic-agent which is currently not used.


fwiesel (approved):

dhoeller (approved):
LGTM

PR #8834, opened 2025-05-30
[reloader][octavia] add reloader annotations

s10:
Add reloader annotations to the service deployments.

This would ensure that services are restarted on the respective DB or RabbitMQ credentials update.

Annotations added:
* `secret.reloader.stakater.com/reload` - list of secrets to track
* `deployment.reloader.stakater.com/pause-period` - pause interval before deployment rollout. E.g., if we have different credentials changed within one minute, then only one restart would be triggered. This will also help with kubelet sync interval, on which the time to apply vault change depends


BenjaminLudwigSAP (approved):
LGTM
Discussed in #octavia channel

velp (approved):

PR #8835, opened 2025-05-30
[reloader][placement] add reloader annotations

s10:
Add reloader annotations to the service deployment.

This would ensure that services are restarted on the respective DB or RabbitMQ credentials update.

Annotations added:
* `secret.reloader.stakater.com/reload` - list of secrets to track
* `deployment.reloader.stakater.com/pause-period` - pause interval before deployment rollout. E.g., if we have different credentials changed within one minute, then only one restart would be triggered. This will also help with kubelet sync interval, on which the time to apply vault change depends

### DIFF
--- a/openstack/placement/Chart.yaml
+++ b/openstack/placement/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v2
 description: A Helm chart for the Openstack Placement Service
 name: placement
 icon: https://object-storage-ca-ymq-1.vexxhost.net/swift/v1/6e4619c416ff4bd19e1c087f27a43eea/www-images-prod/project-mascots/Placement/OpenStack_Project_Placement_horizontal.jpg
-version: 0.2.2
+version: 0.2.3
 appVersion: "dalmatian"
 dependencies:
   - condition: mariadb.enabled
--- a/openstack/placement/templates/placement-api-deployment.yaml
+++ b/openstack/placement/templates/placement-api-deployment.yaml
@@ -7,6 +7,9 @@ metadata:
     system: openstack
     type: api
     component: placement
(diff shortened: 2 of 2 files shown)


joker-at-work (changes requested):
I have to think about the consequences and don't want this merged right away.

joker-at-work:
My problem here is: `placement-etc` contains multiple secrets (no rabbitmq): db connection and keystone authentication (username + password).

1. When we restart `placement-api`, it doesn't only read the new secret but all mounted volumes i.e. also configmaps and thus other settings. This is most likely fine.
2. We start restarting for Keystone username and password change. With replication of both not necessarily happening at the same time, `placement-api` could read the new username and the old password or vice versa. This would lead to an outage *until* the other secret was also replicated in Vault and synced by secrets-injector. This, of course, could also happen with single Pod restarts triggered by k8s itself. But reloader would, as I understand it, restart all pods and thus break them all at once.
3. Keystone username/password are supposed to change to something else. I do not know the concept for the rotation of these, yet, and if it would be fine to just restart and read new values at any time.

Therefore, I would feel more comfortable if the rotation needed a deployment of placement like we do for Nova. I'm open to re-visiting this, once we know how all other secrets get rotated.

joker-at-work:
> Should the password not be there yet, the new pods shouldn't start yet, and the rollout should be wait until the password is there.

With all those shoulds - who checks that inside Placement? Does the API not become ready i.e. the pod unhealthy if the credentials are invalid?

fwiesel:
Okay, it works with the database, but indeed not with the service user.
If these credentials are not working yet, then we replace all working api servers with broken ones.

PR #8848, opened 2025-06-02
[unbound] introducing DnsUnbound{1,2}DisproportionateAmountOfSERVFAILs

vssldmtrv:
The amount of SERVFAIL errors both unbounds are getting should be distributed evenly.

If one of the unbound deployments is getting a disproportionate amount of SERVFAIL errors then that could be an indicator for a network issue in that particular AZ.


mbrowny (approved):

galkindmitrii (approved):
lgtm, but imo 5m is too short

PR #8895, opened 2025-06-05
[ironic] add TZ to certrobot

sandzwerg:
Add the timezone to the cronjob, so it runs everywhere at the same time.

### DIFF
--- a/openstack/ironic/templates/redfish-certrobot-cronjob.yaml
+++ b/openstack/ironic/templates/redfish-certrobot-cronjob.yaml
@@ -28,7 +28,7 @@ metadata:
   name: {{ .Release.Name }}-redfish-certrobot-0
 spec:
   schedule: {{ .Values.cert_robot.schedule | quote }}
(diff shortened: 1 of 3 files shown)


openstack/ironic/templates/redfish-certrobot-cronjob.yaml
```diff
@@ -28,6 +28,7 @@ metadata:
   name: {{ .Release.Name }}-redfish-certrobot-0
 spec:
   schedule: {{ .Values.cert_robot.schedule | quote }}
+  timeZone:  {{ .Values.cert_robot.timezone | default "Etc/UTC" | quote }}
```
majewsky, line 31: The `default` is redundant because you already have the same default in `values.yaml`:
```suggestion
  timeZone:  {{ .Values.cert_robot.timezone | quote }}
```

businessbean (approved):
lgtm

fwiesel (approved):

PR #8898, opened 2025-06-05
Thanos host alert maintenance mode 10 days

frederic-boulianne:
Moved the prometheus host alert to Thanos scaleout so it gets data that contains more than the 25h of the local instance


himanip94 (approved):
Thank you, lgtm

frederic-boulianne:
it was in the vmware prometheus rules but is not deployed on the right prometheus node 
so I took the alerts on this node https://metrics.scaleout.qa-de-1.cloud.sap/ruler/alerts and found where it was in the repo to add it there

PR #8921, opened 2025-06-11
Use ghcr for metal operator remote

SzymonSAP:


system/kustomize/metal-operator-managedresources/rbac.yaml
```diff
@@ -1,145 +0,0 @@
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
-metadata:
-  name: cc:oidc-ias-admin
```
Nuckal777, line 5: All the `cc:oidc*` clusterrolebindings are missing in the managed resources right now, no?

Nuckal777 (commented):
Thanks.

PR #8926, opened 2025-06-11
[cortex/rebalancer] Grant scheduler technical user access to Manila

PhilippMatthes:
This change is needed to fully access Manila resources in Cortex.

### DIFF
--- a/openstack/cortex/templates/seed.yaml
+++ b/openstack/cortex/templates/seed.yaml
@@ -33,3 +33,5 @@ spec:
         role: cloud_volume_admin
       - user: scheduler_service@Default
         role: cloud_network_admin
+      - user: scheduler_service@Default
+        role: cloud_sharedfilesystem_admin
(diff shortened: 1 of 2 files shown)


PhilippMatthes:
FYI the validation fails due to 
```
[ERROR] templates/: template: scheduler/templates/seed.yaml:1:22: executing "scheduler/templates/seed.yaml" at <.Values.global.vaultBaseURL>: nil pointer evaluating interface {}.vaultBaseURL
```
which is unrelated to this change.

openstack/cortex/templates/seed.yaml
```diff
@@ -33,3 +33,5 @@ spec:
         role: cloud_volume_admin
       - user: scheduler_service@Default
         role: cloud_network_admin
+      - user: scheduler_service@Default
+        role: cloud_sharedfilesystem_admin
```
Carthaca, line 37:
```suggestion
        role: cloud_sharedfilesystem_viewer
```

is sufficient for initial placement

Carthaca (changes requested):

PR #8936, opened 2025-06-12
to remove iaas-* domains from customer_domains

sumitarora2786:
in preparation for: https://github.wdf.sap.corp/cc/secrets/pull/17181

Remove iaas-* domains from the list of customer_domains


wagnerd3 (approved):
I am not sure for what the `service` role is in there, but probably cloud_admin works for that too.

crenduchinta88 (approved):

PR #8942, opened 2025-06-13
Added changes for redfish certrobot project

Pooja-Sangle:


openstack/ironic/templates/redfish-certrobot-cronjob.yaml
```diff
@@ -43,6 +45,26 @@ spec:
...
+                  key: ipmi_user_passwd
+            - name: VAULT_REGION
+              value: "{{ .Values.vault.region }}"
+            - name: VAULT_ADDR
+              value: "{{ .Values.vault.addr }}"
+            - name: VAULT_PROJECT
+              value: "{{ .Values.vault.project }}"
+            - name: NETBOX_URL
+              value: "{{ .Values.netbox.url }}"
+            - name: VAULT_ENV
```
fwiesel, line 66: We do not need any vault variables anymore. That should be covered by the secrets-injector.

openstack/ironic/values.yaml
```diff
@@ -33,6 +33,12 @@ debug: "True"
 
 kos_conductor: true
 
+vault:
+  region: qa-de-1
+
+netbox:
+  url: https://netbox.global.cloud.sap
```
fwiesel, line 40: Can you remove that please? I'd rather avoid having our internal urls in public repositories.
It isn't really a major issue, just stylistic.

openstack/ironic/values.yaml
```diff
@@ -33,6 +33,12 @@ debug: "True"
 
 kos_conductor: true
 
+vault:
+  region: qa-de-1
```
fwiesel, line 37: Could you remove that default? It is only valid in one region, and if left out will result in the wrong credentials being used.

openstack/ironic/templates/redfish-certrobot-secret.yaml
```diff
@@ -2,7 +2,12 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ .Release.Name }}-redfish-certrobot
+  annotations:
+    cloud.sap/inject-secrets: "true"
 type: Opaque
+data:
+  ipmi_user_name: {{ printf "vault+kvv2:///secrets/%s/ironic/impi-user/ironic/username" .Values.vault.region | b64enc }}
```
fwiesel, line 9: Can you add here a `required` for `.Values.vault.region` please`

openstack/ironic/templates/redfish-certrobot-cronjob.yaml
```diff
@@ -43,6 +45,18 @@ spec:
...
+                secretKeyRef:
+                  name: {{ .Release.Name }}-redfish-certrobot
+                  key: ipmi_user_name
+            - name: BMC_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Release.Name }}-redfish-certrobot
+                  key: ipmi_user_passwd
+            - name: NETBOX_URL
+              value: "{{ .Values.netbox.url }}"
```
fwiesel, line 59: Here another `required` check please.

fwiesel (changes requested):
I think, we are almost there.
Could you just move the two values out and check their existence?
You'll have to add then some values to `ci/test-values.yaml` for the helm linter to work.

Also, could you make a single commit out of it, please?

fwiesel (approved):

sandzwerg (approved):

PR #8960, opened 2025-06-18
[hermes] Enable SSL for rabbitmq

fwiesel:
This only enables the option for clients to connect to rabbitmq
using AMQPS, so with TLS encryption. As other services might
want to switch to SSL, and we only can enable it for all AMQP
connections at once, we need this in place before the services
switch TLS on for AMQP.

### DIFF
--- a/openstack/hermes/Chart.lock
+++ b/openstack/hermes/Chart.lock
@@ -4,9 +4,9 @@ dependencies:
   version: 0.1.3
 - name: rabbitmq
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-  version: 0.18.5
+  version: 0.18.8
 - name: owner-info
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
   version: 0.2.0
-digest: sha256:683c560f093fa5a00c54049679e6eb24d7d7c6596f16563ec906bcec151aa14d
-generated: "2025-07-28T10:39:25.057311-07:00"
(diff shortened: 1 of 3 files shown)


fwiesel:
I haven't tested it, as I am missing some values to test it manually.

But I have tested it with the rabbitmq chart itself, so I am fairly optimistic that it should work.

notque:
Fails due to "spec.commonName: Too long: must have at most 64 bytes" from this

https://github.com/sapcc/helm-charts/blob/master/common/rabbitmq/templates/certificate.yaml#L13

Can't merge until this is sorted.

notque:
> I removed the `CN` and also fixed the "group" for the Certificate provider. We have our own fork with custom-resource for digicert.

Sounds good, I will deploy as a branch today into qa to validate.

notque:
"MountVolume.SetUp failed for volume "ssl" : secret "tls-hermes-rabbitmq-notifications" not found"

I haven't spent any time looking through it yet, so will add detail, but that's the current error trying to deploy

notque:
Just a reminder this is still an issue. I wasn't able to sort it @fwiesel

fwiesel:
@notque It should be fixed now.
The `commonName` has been taken from the DNS name (subject-alternative-name), if not set. I now check the length of the DNS name and set a shorter `commonName`.

notque:
> @notque It should be fixed now. The `commonName` has been taken from the DNS name (subject-alternative-name), if not set. I now check the length of the DNS name and set a shorter `commonName`.

sorry, missed this update. i will test it now and see. thank you.

PR #8968, opened 2025-06-18
[neutron] Enable ssl on rabbitmq

fwiesel:
This change is mostly preparatory.

Internal use requires another commit, which has a dependency on a roll out of hermes enabling AMQPS, but also when rolled out before we won't have a race between the two changes.

Main purpose is to enable external use of the rabbitmq for KVM.


fwiesel:
Duplicate of #9092

PR #8985, opened 2025-06-19
Use ghcr for boot operator remote

SzymonSAP:


system/boot-operator-remote/values.yaml
```diff
@@ -1,40 +1,68 @@
...
+          readOnly: true
+          source:
+            secret:
+              secretName: boot-operator-remote-kubeconfig
+              items:
+              - key: token
+                path: token
+              - key: bundle.crt
+                path: ca.crt
+      serviceAccountName: controller-manager
```
defo89, line 60: I think it needs to be `default` in remote use cases
```suggestion
      serviceAccountName: default
```

system/boot-operator-remote/values-managed-resources.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+prometheus:
+  enable: false
+certmanager:
+  enable: false
+networkPolicy:
+  enable: false
+controllerManager:
+  enable: false
+  serviceAccountName: boot-operator-controller-manager
```
defo89, line 19: is `serviceAccountName` needed here?

SzymonSAP: it depends what we need `ServiceAccount` name to be. It is in `kube-system` namespace so `controller-manager` might be already present

system/boot-operator-remote/values.yaml
SzymonSAP: done

PR #8995, opened 2025-06-20
add swift_caretaker as admin in cloud_admin

sumitarora2786:
add swift_caretaker as admin in cloud_admin

### DIFF
--- a/openstack/swift-utils/templates/seed.yaml
+++ b/openstack/swift-utils/templates/seed.yaml
@@ -33,8 +33,8 @@ spec:
         description: 'Swift Account Caretaker'
         password: {{ $.Values.swift_account_caretaker.password }}
         role_assignments:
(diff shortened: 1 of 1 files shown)


majewsky:
What is the change here? The red section of this diff already has this role assignment. Do you need the role assignment in project `service@Default`?

sumitarora2786:
perhaps we don't need service@Default related assignments, I was not sure so still kept it.
if we are certain that its not required and we ok just with 

   - name: cloud_admin
          role_assignments:
            # permission to enumerate all projects and domains
            - user: swift_caretaker@Default
              role: admin
              
              
            if you confirm and feel the same, I will remove the role assignment for 
            
          role_assignments:
          - project: service
            role: service

majewsky:
Is this what you are looking for? #9080

PR #9005, opened 2025-06-23
feat(opensearch): add admin user credentials

joluc:
Added support for setting admin user credentials in operator-managed OpenSearch deployments


joluc:
Get admin user credetials from Vault via pipeline used here: https://github.com/cloudoperators/greenhouse-extensions/pull/975

IvoGoman (approved):

PR #9024, opened 2025-06-24
[nova][proxysql] use native sidecars for proxysql

s10:

Update openstack/utils helm chart to 0.27.0 version with native sidecar support for proxysql


seb-kro (commented):
Please also see the more general comment regarding container termination and documentation in https://github.com/sapcc/helm-charts/pull/9033

seb-kro (approved):

PR #9025, opened 2025-06-24
[neutron][proxysql] use native sidecars for proxysql

s10:


s10:
@notandy @fwiesel , could you please take a look at this PR?

PR #9031, opened 2025-06-24
[barbican][proxysql] use native sidecars for proxysql

s10:


s10:
@fwiesel @rajivmucheli , please take a look at thir PR

PR #9032, opened 2025-06-24
[octavia][proxysql] use native sidecars for proxysql

s10:


BenjaminLudwigSAP (approved):
Ah, I missed that. LGTM :)

velp (approved):

PR #9033, opened 2025-06-24
[placement][proxysql] use native sidecars for proxysql

s10:

### DIFF
--- a/openstack/placement/Chart.lock
+++ b/openstack/placement/Chart.lock
@@ -10,12 +10,12 @@ dependencies:
   version: 0.6.10
 - name: utils
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-  version: 0.26.0
+  version: 0.27.2
 - name: owner-info
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
   version: 1.0.0
 - name: linkerd-support
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
   version: 1.1.0
-digest: sha256:31ce18904522853f8187ed36bc1f152fc700ea2dcae95b23568404195181371b
-generated: "2025-05-16T16:14:01.219378+03:00"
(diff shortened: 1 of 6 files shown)


seb-kro:
The way I understand [native sidecars](https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/), we no longer need to terminate the proxysql container if `native_sidecar: true`. So should we adjust `utils.script.job_finished_hook` and/or `utils.proxysql.proxysql_signal_stop_script` accordingly? As well as `utils.proxysql.job_pod_settings`? Or is it intended to stay this way?

s10:
> The way I understand [native sidecars](https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/), we no longer need to terminate the proxysql container if `native_sidecar: true`. So should we adjust
> 
> * `utils.proxysql.proxysql_signal_stop_script` (or maybe `utils.script.job_finished_hook`)
> * `utils.proxysql.job_pod_settings`
> 
> accordingly? Or is it intended to stay this way?
> 
> As a side note: it might also make sense to update `openstack/utils/proxysql_sidecar.md`.

`utils.script.job_finished_hook` includes `utils.linkerd.signal_stop_script`, and we can't disable it, because for linkerd we, for now, in general, don't use native sidecars. So this change (removal of proxysql pkill) is nice to have, but not mandatory.

seb-kro:
Agreed that we shouldn't completely disable `utils.script.job_finished_hook`. I meant sth like adding `{{- if .Values.proxysql.native_sidecar }}` as an additional condition to `utils.proxysql.proxysql_signal_stop_script``
```
{{- define "utils.proxysql.proxysql_signal_stop_script" }}
  {{- if and .Values.proxysql .Values.proxysql.mode .Values.proxysql.native_sidecar -}}
pkill proxysql || true
  {{- end }}
{{- end }}
```
Or is that an issue?

seb-kro (approved):

joker-at-work (approved):

PR #9058, opened 2025-06-26
Add monitoring plugin to customer runtime

jknipper:


viennaa (changes requested):
Thanks for adding this cluster, could you bump the chart pls?

PR #9059, opened 2025-06-26
[nova][mysql-metrics] remove cell2 db path helper function

s10:
Remove cell2 db path helper function, which is being used in mysql-metrics chart.

This PR could be merged after the related secrets PR is merged, in which we define `connections` configuration.


joker-at-work (changes requested):
Chart bump is already part of Nova. We can reduce this PR to removing the function.

businessbean (approved):
lgtm

PR #9065, opened 2025-06-26
feat(opensearch): add opensearch plugin preset

joluc:
- Add OpenSearch plugin preset for `obs` clusters
- Add dashboard user


timojohlo (approved):
LGTM!

PR #9072, opened 2025-06-27
feat(greenhouse) kube-monitoring cluster specific extra spec

trouaux:
to allow cluster specific extraspecs for kube-monitoring plugin preset


system/greenhouse-ccloud/templates/kube-monitoring-pluginpreset.yaml
```diff
@@ -52,6 +52,9 @@ spec:
     - name: kubeMonitoring.prometheus.prometheusSpec.thanos.objectStorageConfig.existingSecret.key
       value: thanos.yaml
   {{- end }}
+  {{- if index $additionalSettings "extraSpecs" }}
+  {{- index $additionalSettings "extraSpecs" | toYaml | nindent 4}}
+  {{- end }}
```
viennaa, line 57: I'd like to explain the additional extraSpecs here, because at a first glance it reads the same.

```suggestion
  {{/* nested extraSpecs for specific clusters */}}
  {{- if index $additionalSettings "extraSpecs" }}
  {{- index $additionalSettings "extraSpecs" | toYaml | nindent 4}}
  {{- end }}
```

viennaa (changes requested):
Thanks for adding the specific specs, just some little explanation for future me and we're good.

PR #9078, opened 2025-06-27
[absent-metrics-operator] update registry to ghcr

trouaux:
updating registry to ghcr to deprecate concourse build pipeline for absent metrics operator


majewsky (approved):
Do we have any images newer than 6 weeks old?

PR #9082, opened 2025-06-27
[swift] Create proxy-external-service.yaml

crenduchinta88:
eu-de-1 has additional external proxy IP
this is to create new service with this IP


crenduchinta88:
PR for updating the secret
https://github.wdf.sap.corp/cc/secrets/pull/17575

sumitarora2786 (approved):
looks good

PR #9083, opened 2025-06-28
[nova] Enable rate limiting.

misamoylov:
Enables traffic shaping for OpenStack Nova APIs.

### DIFF
--- a/openstack/nova/ci/test-values.yaml
+++ b/openstack/nova/ci/test-values.yaml
@@ -106,3 +106,51 @@ cors:
 utils:
   cors:
     allowed_origin: 'https://test.domain'
+
+# openstack-rate-limit
+rate_limit:
+  enabled: false
+  rate_limit_by: target_project_id
+  max_sleep_time_seconds: 0
+  log_sleep_time_seconds: 10
+  backend_timeout_seconds: 1
+  whitelist:
+    - Default/service
+    - Default/admin
+    - tempest/performance-testing
+    - ccadmin/cloud_admin
+    - ccadmin/master
+  whitelist_users: null
+  blacklist: null
+  blacklist_users: null
+  groups:
+    write:
+      - delete
+      - update
+  rates:
+    default:
+      servers:
+        - action: read/list
+          limit: 100r/m
+        - action: create
+          limit: 100r/m
+      servers/server:
+        - action: read/list
+          limit: 100r/m
+        - action: write
+          limit: 100r/m
+      servers/server/migrations:
+        - action: read/list
+          limit: 100r/m
+      servers/server/migrations/migration:
+        - action: read/list
+          limit: 100r/m
+      servers/server/action:
+        - action: create
+          limit: 100r/m
+      flavors:
+        - action: read/list
+          limit: 100r/m
+      servers/server/remote-consoles:
+        - action: create
+          limit: 100r/m
--- a/openstack/nova/templates/etc/_api-paste.ini.tpl
+++ b/openstack/nova/templates/etc/_api-paste.ini.tpl
@@ -5,6 +5,11 @@
 {{- define "watcher_pipe" -}}
 {{- if .Values.watcher.enabled }} watcher{{- end -}}
 {{- end -}}
+
+{{- define "rate_limit_pipe" -}}
+{{- if .Values.rate_limit.enabled }} rate_limit{{- end -}}
+{{- end }}
+
 ############
 # Metadata #
 ############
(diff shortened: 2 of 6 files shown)


openstack/nova/templates/etc/_api-paste.ini.tpl
```diff
@@ -40,8 +40,11 @@ use = call:nova.api.openstack.urlmap:urlmap_factory
 /healthcheck: healthcheck
 
 [composite:openstack_compute_api_v21]
+{{- define "rate_limit_pipe" -}}
+{{- if .Values.rate_limit.enabled }} rate_limit{{- end -}}
+{{- end }}
 use = call:nova.api.auth:pipeline_factory_v21
-keystone = cors http_proxy_to_wsgi compute_req_id faultwrap request_log sizelimit {{- include "osprofiler_pipe" . }} authtoken keystonecontext {{- include "watcher_pipe" . }} sentry {{- include "audit_pipe" . }} osapi_compute_app_v21
+keystone = cors http_proxy_to_wsgi compute_req_id faultwrap request_log sizelimit {{- include "osprofiler_pipe" . }} authtoken keystonecontext {{- include "watcher_pipe" . }} {{ if .Values.rate_limit.enabled }}{{- include "rate_limit_pipe" . }}{{ end }} sentry {{- include "audit_pipe" . }} osapi_compute_app_v21
```
grandchild, line 47: why the same if again here? `rate_limit_pipe` already has the same?

openstack/nova/templates/etc/_api-paste.ini.tpl
```diff
@@ -40,8 +40,11 @@ use = call:nova.api.openstack.urlmap:urlmap_factory
 /healthcheck: healthcheck
 
 [composite:openstack_compute_api_v21]
+{{- define "rate_limit_pipe" -}}
+{{- if .Values.rate_limit.enabled }} rate_limit{{- end -}}
+{{- end }}
```
grandchild, line 45: i'd rather see this defined at the top with the other `_pipe`s.

openstack/nova/values.yaml
```diff
@@ -1013,3 +1013,84 @@ vpa:
...
+        - action: read/list
+          limit: 100r/m
+      os-server-groups:
+        - action: read/list
+          limit: 100r/m
+        - action: write
+          limit: 100r/m
+      os-server-groups/os-server-group:
+        - action: read/list
+          limit: 100r/m
```
grandchild, line 1096: These are _all_ 100r/m. Did you check which numbers may make sense (perhaps going by current peak usage), or is this left for the future?

openstack/nova/values.yaml
```diff
@@ -1013,3 +1013,84 @@ vpa:
...
+      os-aggregates:
+        - action: read/list
+          limit: 100r/m
+        - action: create
+          limit: 100r/m
+      os-aggregates/os-aggregate:
+        - action: read/list
+          limit: 100r/m
+        - action: write
+          limit: 100r/m
```
grandchild, line 1082: Why are we limiting these endpoints? They are admin-only, and so are in the whitelist (sic) anyway.
There are others (e.g. flavors write) -- maybe we don't need entries for admin APIs at all? Or are we protecting us from ourselves as well?

openstack/nova/ci/test-values.yaml
```diff
@@ -65,3 +65,55 @@ cors:
...
+        - action: read/list
+          limit: 100r/m
+      servers/*/action:
+        - action: create
+          limit: 100r/m
+      flavors:
+        - action: read/list
+          limit: 100r/m
+        - action: create
+          limit: 100r/m
```
grandchild, line 114: This can go.

openstack/nova/ci/test-values.yaml
```diff
@@ -65,3 +65,55 @@ cors:
...
+      flavors/flavor:
+        - action: read/list
+          limit: 100r/m
+        - action: write
+          limit: 100r/m
```
grandchild, line 119: This can go.

grandchild (changes requested):
Sorry for not following up after the first review.

Some admin APIs left in the rate_limit section.

Also we may want to limit remote console creation: https://docs.openstack.org/api-ref/compute/#create-console
It's unlikely, but conceivable to maliciously overload either Nova or VCenter that way:

```yaml
    servers/*/remote-consoles:
      - action: create
        limit: 100r/m
```

grandchild (approved):

seb-kro (approved):

PR #9088, opened 2025-06-30
[nova] Add nanny failure alert

seb-kro:
We add a new Prometheus alert that fires when Nova nanny cronjobs are continuously failing or not completing.


joker-at-work (approved):

seb-kro:
I still made a small adjustment after testing. The previous implementation only triggered if the job terminated successfully at least once. I.e., it did not trigger if a job always failed right from the beginning.

openstack/nova/alerts/kubernetes/nova.alerts
```diff
@@ -0,0 +1,24 @@
+groups:
+- name: kubernetes-nova.alerts
+  rules:
+  - alert: OpenstackNovaNannyCronjobNotCompleting
+    # NOTE: `time() - kube_cronjob_status_last_successful_time` does not exist
+    # for a cronjob that has never terminated successfully. We use
+    # `kube_cronjob_status_last_schedule_time` as a default in this case. This
+    # exists for all cronjobs that ever started and should always be >= 24h.
```
joker-at-work, line 8: This comment doesn't make sense to me. Why would `kube_cronjob_status_last_schedule_time` always be >= 24h? We use it one line after with `< 24h` - then we can get rid of that?

openstack/nova/alerts/kubernetes/nova.alerts
```diff
@@ -0,0 +1,24 @@
...
+    expr: >
+      (time() - kube_cronjob_status_last_successful_time{cronjob=~"nova-nanny-.*"}
+      or on(cronjob) kube_cronjob_status_last_schedule_time{cronjob=~"nova-nanny-.*"}) >= 24*60*60
+      and on(cronjob) (time() - kube_cronjob_status_last_schedule_time{cronjob=~"nova-nanny-.*"} < 24*60*60
+      or on(cronjob) kube_cronjob_status_active{cronjob=~"nova-nanny-.*"} >= 1)
```
joker-at-work, line 13: With this `and` do we really catch anything that never ran? I would assume these don't have (at least one) active Job and they definitely weren't scheduled < 24h ago.

openstack/nova/alerts/kubernetes/nova.alerts
seb-kro: We use `time() - kube_cronjob_status_last_schedule_time{cronjob=~"nova-nanny-.*"}` <= 24h, i.e., the time since `kube_cronjob_status_last_schedule_time`. 

The timestamp itself that we use in the first brackets should be >= 24h. Note that `kube_cronjob_status_last_schedule_time{cronjob=~"nova-nanny-.*"}` replaces the entire `time() - kube_cronjob_status_last_successful_time{cronjob=~"nova-nanny-.*"}` here since "-" has precedence over "or".

openstack/nova/alerts/kubernetes/nova.alerts
seb-kro: Ah, maybe you mean since the job is running for the very first time, it was never scheduled? But even then, it must have been scheduled for this first run. Plus, for the scenario of consistent errors, where this becomes relevant, the job has already been scheduled many times within the last 24h, it just didn't successfully run to completion.

joker-at-work (approved):

PR #9099, opened 2025-07-01
[nova] add vcsu and edit vct, remove host username

PaulPickhardt:


joker-at-work (commented):
I don't like that the commit-message explains nothing. How are people not directly involved in this change going to learn what this is about and how it's supposed to work?

PR #9101, opened 2025-07-01
make the parameter osd-scrub-auto-repair global

senolcolak:
apply scrub_auto_repair fix parameter globally on all production environments.


kayrus:
@senolcolak can you bump the chart version?

sumitarora2786 (approved):
looks good

PR #9114, opened 2025-07-02
Revert "[neutron] Enable ssl for rabbitmq"

fwiesel:
This reverts commit 2ed196849157c84cd69276b3ca042563498a2de3.


sebageek (approved):
We could also keep the chart and enable this on a per-region basis.

fwiesel:
Sure, but personally, I am a fan from rolling a change through all the regions, so one doesn't get a surprise in the largest productive region for something that might have failed earlier.
QA is unfortunately not similar enough to production.

PR #9146, opened 2025-07-07
[common/rabbitmq] No externalIPs in certificates

fwiesel:
Our certificate provider doesn't support these IPs, so that part never worked. If you want to add them still, you can do that explicitly by adding them via `certificate.ipAddresses`


galkindmitrii (approved):

s10 (approved):

s10 (changes requested):
Could you please bump the chart version to `0.18.4` in `Chart.yaml`

PR #9188, opened 2025-07-10
[keystone] don't fail db migration on doctor check

s10:
Don't fail db migration job on keystone-manage doctor check, because it prevents keystone-api from starting and is expected to fail in some environments (eg, debug is enabled in qa or memcached check is failing in keystone-global)

### DIFF
--- a/openstack/keystone/Chart.yaml
+++ b/openstack/keystone/Chart.yaml
@@ -9,7 +9,7 @@ maintainers:
 name: keystone
 sources:
   - https://github.com/sapcc/keystone
-version: 0.10.5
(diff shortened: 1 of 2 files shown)


galkindmitrii (approved):

bbobrov (changes requested):
Sorry, but this should not be merged. We should not generally ignore warnings from the doctor, it is there for a reason. Is it possible to individually exclude the problems that you see? What code is being returned? What makes you think that these are not valid issues?

s10:
db-migrate is disabled everywhere since yesterday, so this change isn't needed for now.

PR #9207, opened 2025-07-14
[utils][proxysql] add support for unlimited transactions time in jobs

s10:
Add support for setting max_transaction_time and default_query_timeout variables to default large values, which can be used in Jobs, in which some transactions or queries might run longer than 60-120s.

For example, it might be useful for some ALTER TABLE queries.


galkindmitrii (approved):

businessbean (approved):
lgtm

PR #9235, opened 2025-07-16
nova: add secret hashes to pod annotations

leust:
Add secret-etc-hash annotations to all Nova deployment templates to ensure pods restart when secrets are updated during helm rollouts. This follows the same pattern as existing configmap-etc-hash annotations.

### DIFF
--- a/openstack/nova/templates/_console_deployment.yaml.tpl
+++ b/openstack/nova/templates/_console_deployment.yaml.tpl
@@ -39,6 +39,7 @@ spec:
         {{- include "utils.topology.pod_label" . | indent 8 }}
       annotations:
         configmap-etc-hash: {{ include (print .Template.BasePath "/etc-configmap.yaml") . | sha256sum }}
+        secret-etc-hash: {{ include (print .Template.BasePath "/etc-secret.yaml") . | sha256sum }}
         {{- if .Values.proxysql.mode }}
         prometheus.io/scrape: "true"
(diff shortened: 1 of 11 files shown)


joker-at-work:
We still wouldn't restart pods if we add new keys/remove keys from the Secret. That's what this is meant for - not the actual values.

joker-at-work (changes requested):
As stated above: mks and shellinabox consoles need a hash of their secret, too.

fwiesel:
> We still wouldn't restart pods if we add new keys/remove keys from the Secret. That's what this is meant for - not the actual values.

But the reloader also restarts any pod referencing a secret/configmap whenever they change.

joker-at-work:
We have reloader annotations on the console pods? I think I wasn't aware.

joker-at-work (approved):

joker-at-work:
What does reloader solve for us:
* restart pods, when a secret value changes

With dual-user secret rotation, the process is as follows:
1) change the password of the unused user (mariadb/rabbitmq will pick up on that, because they have reloader on their secrets)
2) switch Nova to using that ^ user (reloader could pick up on that, but helm could, too)
3) switch the password for the now-unused user (mariadb/rabbitmq …)

So for the rabbitmq/mariadb dual-user password rotation, we don't need reloader. Hence, I'd only like to introduce reloader when we need it.

PR #9236, opened 2025-07-16
nova: switch Secret templates from stringData to data with base64

leust:
Switch nova Secret templates from stringData to data attribute to resolve Vault secrets-injector compatibility issues. Using stringData is broken by design as removing keys from stringData does not remove them from data.


fwiesel:
Could you please add the terminal newlines to the files? Thanks

joker-at-work (approved):

bashar-alkhateeb (approved):

PR #9238, opened 2025-07-17
fix(perses): Switch to Perses Greenhouse Plugin as chart dependency

ibakshay:
I'm switching from the upstream chart to using the Perses Greenhouse Plugin as a dependency. This makes way more sense for two key reasons:

Eliminates duplicate work - No point maintaining two separate charts that do the same thing
Keeps us current with Greenhouse innovations - We automatically get new features and improvements happening in Greenhouse

Thank you @timojohlo @viennaa  for your valuable insights and feedback. 

Signed-off-by: Akshay Iyyadurai Balasundaram <akshay.iyyadurai.balasundaram@sap.com>


timojohlo (approved):
no remarks, LGTM!

PR #9249, opened 2025-07-18
Delete openstack/cerebro directory

PhilippMatthes:
This old seed is no longer needed and should be deleted.


PhilippMatthes:
Note to anyone picking this up: We should also delete the path from CODEOWNERS

PR #9251, opened 2025-07-18
[cc-runtime-cluster] deploy calico via helm provider

defo89:


system/cc-runtime-cluster/templates/helmchartproxy-calico.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: addons.cluster.x-k8s.io/v1alpha1
+kind: HelmChartProxy
+metadata:
+  name: nginx-ingress
```
Nuckal777, line 4: We're deploying calico.

Nuckal777 (changes requested):

system/cc-runtime-cluster/templates/helmchartproxy-calico.yaml
defo89: good catch fixed

Nuckal777 (approved):

PR #9256, opened 2025-07-21
[netapp-exporter] remove redundant rules

chuan137:
Catellum rules migrated to netapp-harvest-exporter.


chuan137:
removing entire netapp-cap-exporter

PR #9263, opened 2025-07-21
[placement] use mariadb values for database connection

s10:
* Use `mariadb.users` and `mariadb.defaultUser` values for the Placement DB connection template.

* Stop using empty `dbType` option to force the use of `nova-api-mariadb` service, because `utils.db_host` uses `mariadb.name` for service name, which for regions with `nova-api-mariadb` already points to the right database.

### DIFF
--- a/openstack/placement/templates/_helpers.tpl
+++ b/openstack/placement/templates/_helpers.tpl
@@ -9,10 +9,5 @@
 
 
 {{- define "placement.db_name" }}
-  {{- if or (eq .Values.dbType "mariadb") (eq .Values.dbType "pxc-db") }}
 {{- include "utils.db_host" . -}}
-{{/* Use nova-api-mariadb database, if no database is selected explicitly */}}
-  {{- else -}}
-"nova-api-mariadb"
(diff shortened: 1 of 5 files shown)


openstack/placement/templates/_helpers.tpl
```diff
@@ -9,10 +9,5 @@
 
 
 {{- define "placement.db_name" }}
-  {{- if or (eq .Values.dbType "mariadb") (eq .Values.dbType "pxc-db") }}
 {{- include "utils.db_host" . -}}
-{{/* Use nova-api-mariadb database, if no database is selected explicitly */}}
```
s10, line 14: This logic was added to allow for the following configuration:
`mariadb.enabled` set to `true`, `mariadb.name` set to `placement` - new empty database is started, and dbType="", so placement continues to use `nova-api-mariadb` database until explicitly migrated

s10:
@fwiesel @seb-kro , please take a look at this PR

seb-kro (approved):

PR #9265, opened 2025-07-21
[prometheus-vmware-alerting] - improvement of bedrock alerting 

christophrichtersap:
Switch from vSphere tags to VMware hostgroups for Bedrock host identification

Replaced the use of `vrops_hostsystem_vsphere_tags ` with `vrops_hostsystem_hostgroups ` as the primary identifier for Bedrock hosts, aligning with affinity rule usage in Nova. Tags were previously managed manually, which posed a risk of incomplete or inconsistent host coverage.


kevin-fischer (approved):
LGTM

PR #9276, opened 2025-07-22
fix(audit-logs): custom audit-logs otel receiver

olandr:
This PR replaces the auditbeat setup for audit-log shipping with an OpenTelemetryCollector.

To make this work we have had to implement custom receivers to replace the filewatch (file_integrity) and the auditd modules from auditbeats.

---------

on-behalf-of: @SAP Simon Olander (simon.olander@sap.com)


olandr:
Superseded by #9892

PR #9282, opened 2025-07-22
[cinder] Add nanny failure alert

jagoleni:
We add a new Prometheus alert that fires when Nova nanny cronjobs are continuously failing or not completing.


hemna (approved):

joker-at-work (approved):
Oh, I see, you added it in the comment. Not sure we ever do this, commit-message would have been fine for me.

PR #9313, opened 2025-07-24
[rabbitmq] add and option to disable a user

s10:
- Added an option to disable a user

Usage example:
```yaml
rabbitmq:
  users:
    default:
      user: rabbitmq
      password: secret
      enabled: false
    user1:
      user: user1
      password: secret
    user2:
      user: user2
      password: secret
```
This will result in the `default` user being removed from the RabbitMQ users secret.


joker-at-work:
Could you explain the use-case for this? What's the difference to removing `default` from that list? Why would I want to keep it in but disabled?

s10:
@bashar-alkhateeb @fwiesel , could you please take a look at this PR?

fwiesel (approved):

PR #9321, opened 2025-07-25
[utils][proxysql] skip disabled database users and null password

s10:
Skip database users in ProxySQL configuration:
* with `password` value set to `null`
* with `enabled` value set to `false`

Examples:

```yaml
mariadb:
  users:
    test:
      password: null
```

```yaml
mariadb:
  users:
    test:
      password: secret
      enabled: false
```

Currently, in MariaDB chart we don't create users with `null` password, but utils/proxysql chart doesn't support this and causes template failure.

This change fixes this problem and adds two ways to remove a user from ProxySQL config.

### DIFF
--- a/openstack/utils/Chart.yaml
+++ b/openstack/utils/Chart.yaml
@@ -2,4 +2,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: utils
-version: 0.28.0
+version: 0.29.0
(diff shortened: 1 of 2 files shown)


businessbean (approved):
lgtm. Hopefully ProxySQL is removing the [backend user config](https://proxysql.com/documentation/main-runtime/#mysql_users) in the future.
```
Note, currently all users need both “frontend” and “backend” set to 1. Future versions of ProxySQL will separate the crendentials between frontend and backend. In this way frontend will never know the credential to connect directly to the backend, forcing all the connections through ProxySQL and increasing the security of the system.
```

joker-at-work (approved):

PR #9335, opened 2025-07-28
[openstack/nova] Nanny for orphaned migrations

fwiesel:
Since we should only have few migrations ongoing, the query should be quite fast. We trigger it often as it will be a blocker to re-installing kvm hosts, as they get decomissioned each time, but have to wait for no migrations being associated with the compute-host.


fwiesel:
Draft, as it depends on another [pr](https://github.com/sapcc/nova/pull/567)

PR #9352, opened 2025-07-29
[nova] Filter VcenterLowNumberOfPlaceableVms to exclude VCs with small HVs

anokfireball:
The flavor used for verification (`g_c128_m512`) can not be spawned on every BB or sometimes at all in a VC.
Introduce a label filter to exclude VCs from the alert where it is not possible to spawn this flavor, irrespective of the usage.


openstack/nova/alerts/openstack/vmware-rebalancer.alerts
```diff
@@ -2,7 +2,17 @@ groups:
...
-        expr: sum(wm_vmware_rebalancer_placeable_vms_per_vcenter{flavor="g_c128_m512",has_bbs_without_custom_hana_exclusive_host="True"}) by (flavor, vcenter, az) < 3
+        expr: |
+          sum(
+            wm_vmware_rebalancer_placeable_vms_per_vcenter{
+              flavor="g_c128_m512",
+              has_bbs_without_custom_hana_exclusive_host="True",
+              # >=64 (counted without mutlithreading while the flavor is multithreaded)
+              largest_non_hana_host_system_cpu_cores=~"(6[4-9]|[7-9][0-9]|[1-9][0-9]{2,})",
+              # >=512 GiB = 512 * 1024 * 1024 KiB = 536 870 912 KiB
+              largest_non_hana_host_system_memory_kilobytes=~"(5[3-9][6-9][8-9][7-9][0-9]9[1-9][2-9]|[6-9][0-9]{8}|[1-9][0-9]{9,})(.[0-9]+)?"
```
joker-at-work, line 13: I don't think that's maintainable. I'd rather work with silences that we keep in for long periods of time (half a year) than with this.

In case you disagree: did you factor in 5% reserved memory already? How would I check that?

Are you maybe missing support for 12 TiB HVs?

We would match for our small HVs here (< 1 TiB), but they don't allow this big flavor on them, so would need to be excluded.

joker-at-work (changes requested):

openstack/nova/alerts/openstack/vmware-rebalancer.alerts
anokfireball: Regarding 12TB HV support: Why would you think it's missing? The last regex `[1-9][0-9]{9,})` [should allow](https://www.debuggex.com/r/uyRmIuyiNHIj3YkS) all HVs with more than 953(-ish) GiB.

anokfireball: Regarding the small HVs concern: This label-based approach is limiting our options. Right now, the filter excludes entire VCs based on the _largest_ HV in that VC. While we can absolutely bump the threshold to 1TB, that is still only best-effort and will not cover heterogeneous VCs.

Sidenote: Should those VCs not be excluded via the CPU filter, anyway?

joker-at-work: > Regarding memory reservation
We only have `512 GiB * 0.95` available in Placement :)

> Regarding 12TB HV support: Why would you think it's missing? The last regex [1-9][0-9]{9,}) [should allow](https://www.debuggex.com/r/uyRmIuyiNHIj3YkS) all HVs with more than 953(-ish) GiB.
ah, because we have `{9,}` so more than ... proves my point of maintainability ;)

PR #9357, opened 2025-07-29
Prometheus host alert maintenance mode 10 days

frederic-boulianne:
Cleanup following move of alert to thanos ruler instead of prometheus
bump of chart version


viennaa (changes requested):
update your fork and submit a clean PR with one commit please

PR #9362, opened 2025-07-29
[system/awx-api-crawler] - Moved from cron to deployment

christophrichtersap:
**Updates:**
- `secrets.yaml` used instead of env-vars for security
- moved from cronjob to deployment
- added interval variable to define how often the code itself is executed

TLDR: We moved from cronjob usage to long running pod to avoid volume claim errors we had in past


kevin-fischer (approved):
+1

PR #9363, opened 2025-07-30
Inventory updater change

Pooja-Sangle:


sandzwerg (changes requested):
So far some minor issues, but this should be a draft until https://github.com/sapcc/inventory-updater/pull/7 is merged or we break something.

Pooja-Sangle:
Thank you for the detailed review. Just to clarify, this code was originally written by Frank (in mac_serial_ng.py) and Bernd (inventory updater), I only integrated the two into this branch. I’ll go through the comments and address what I can, and for questions, I’ll sync with Frank to make sure we handle them correctly.

PR #9380, opened 2025-07-31
[rabbitmq] add an option to skip creation of default user

s10:
- Added an option to skip creation of default user

Usage example:
```yaml
rabbitmq:
  addDefaultUser: false
  users:
    default:
      user: test
      password: secret
    user1:
      user: user1
      password: secret
    user2:
      user: user2
      password: secret
```
This will result in the `default` user not being added to the RabbitMQ users secret.

### DIFF
--- a/common/rabbitmq/CHANGELOG.md
+++ b/common/rabbitmq/CHANGELOG.md
@@ -2,6 +2,28 @@
 
 This file is used to list changes made in each version of the common chart rabbitmq.
 
+## 0.19.0 - 2025/07/31
+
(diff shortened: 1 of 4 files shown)


businessbean:
I prefer the structure
```yaml
rabbitmq:
  users:
    default:
      enabled: false
    user1:
      enabled: true
    user2:
      enabled: true
```
proposed in https://github.com/sapcc/helm-charts/pull/9313. We can start with assuming `enabled: true` and later switch to `false` to avoid introducing too many changes at a time.

PR #9388, opened 2025-08-01
[octavia] add default-container annotation to all pods

felix-kaestner:
The value of the `kubectl.kubernetes.io/default-container` annotation is the container name that is default for the Pod. It will be picked up by commands such as `kubectl logs` or `kubectl exec` if no `-c/--container` flag is specified.

Ref: https://kubernetes.io/docs/reference/labels-annotations-taints/#kubectl-kubernetes-io-default-container


BenjaminLudwigSAP (approved):
lgtm

notandy (approved):
🙇

PR #9397, opened 2025-08-03
[ironic] Enable rate-limiting

misamoylov:


openstack/ironic/templates/etc/_ironic.conf.tpl
```diff
@@ -137,3 +137,18 @@ disable_deep_image_inspection = {{ .Values.conductor.defaults.conductor.disable_
...
+{{ if .Values.rate_limit.enabled }}
+[rate_limit]
+enabled = true
+service_type = baremetal
+config_file = /etc/ironic/ratelimit.yaml
+rate_limit_by = {{ .Values.rate_limit.rate_limit_by }}
+max_sleep_time_seconds: {{ .Values.rate_limit.max_sleep_time_seconds }}
+clock_accuracy = 1ns
+log_sleep_time_seconds: {{ .Values.rate_limit.log_sleep_time_seconds }}
+backend_host = {{ .Release.Name }}-api-ratelimit-redis
```
sandzwerg, line 150: Is that name autogenerated? Because above it's only configured as
`alias: api-ratelimit-redis` which doesn't container service or releasename so I'm not sure if that might not clash with the redis of other services.

PR #9427, opened 2025-08-06
[secrets-injector] add client truststore

pkr4711:
Add additional k8 objects with helm deployment.


uwe-mayer:
Hi @pkr4711 ,
Thx for the pull request. Could you explain the use case for these extra deployments? 
TY

Nuckal777:
The clientCA value is for providing the secrets-injector with CA it can use to authenticate the kube-apiserver. It is completely unrelated to outgoing calls to vault.

PR #9452, opened 2025-08-08
fix(secrets-injector): failsafe deployment

pkr4711:
* failsafe deployment if kube-monitoring crds are not present
* add option to disable deployment of alerts

### DIFF
--- a/system/secrets-injector/templates/alerts.yaml
+++ b/system/secrets-injector/templates/alerts.yaml
@@ -1,3 +1,4 @@
+{{- if and .Values.alerts.enabled (lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" "prometheusrules.monitoring.coreos.com") }}
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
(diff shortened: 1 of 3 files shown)


system/secrets-injector/templates/alerts.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if and .Values.alerts.enabled (lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" "prometheusrules.monitoring.coreos.com") }}
```
Nuckal777, line 1: Consider using `.Capabilities.APIVersions` over `lookup`. [Ref](https://helm.sh/docs/chart_template_guide/builtin_objects/#:~:text=Capabilities.APIVersions.Has%20%24version%20indicates%20whether%20a%20version%20(e.g.%2C%20batch/v1)%20or%20resource%20(e.g.%2C%20apps/v1/Deployment)%20is%20available%20on%20the%20cluster.)

pkr4711: done

Nuckal777 (approved):

PR #9456, opened 2025-08-08
[cc3test] updated alert expression with last_over_time label and also updated cronjob schedule to run only once in a day

ashifnihalb:


openstack/cc3test/templates/cronjob-certificate.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: cc3test-certificate
   namespace: {{ .Values.cc3test.namespace }}
 spec:
-  schedule: "*/30 8-15 * * *"
+  schedule: "0 1 * * *"
```
viennaa, line 7: Do something odd here, not exactly on top of the hour.

```suggestion
  schedule: "27 1 * * *"
```

viennaa (commented):
What happens, if cc3test fails for some reason, then the one required dataset for this day would be missing and the alert would not trigger either.

Not sure if this is actually a case, but might it be useful, to have it running 2 times a day just in case it fails for some reason?

openstack/cc3test/templates/cronjob-certificate.yaml
ashifnihalb: @viennaa updated cronjob schedule to run twice in a day (1:27AM and 13:27 PM) in new commit request

viennaa (changes requested):
Just bump the chart and we're good :)

PR #9473, opened 2025-08-12
Fix default rate-limits for Octavia API

velp:
This PR fixes default Octavia's rate-limits:
- use `read` instead of `read/limit` for sibgle-obejct requests
- add rate-limit for `create` to correct level of API
- add `stats`, `status` and `failover` rate-limits
- rename `write` group to `delete_or_update` for better debugging


BenjaminLudwigSAP (approved):
lgtm

PR #9479, opened 2025-08-12
[neutron-netns-prober] provide initial chart

nikatza:
* We introduce this chart as a replacement for ns-exporter (for now it is essentially a copy of this chart). The new version of the "ns-exporter" binary introduces breaking changes affecting our monitoring and alerting. For example: we use a different naming scheme for some alerts and as a consequence some of these alerts would not reach the incident management system.
* We will use this new chart to deploy and validate newer versions of the ns-exporter without interfering with the current deployment.
* We opt for a new chart to make the relation to neutron more explicitly in the name.  There will we a follow up to remove all artifacts related to ns-exporter once the shift is completed.

### DIFF
--- a/prometheus-exporters/neutron-netns-prober/Chart.lock
+++ b/prometheus-exporters/neutron-netns-prober/Chart.lock
@@ -1,0 +1,6 @@
+dependencies:
+- name: owner-info
+  repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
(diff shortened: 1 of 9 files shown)


nikatza:
After merging mark following PR as ready: https://github.com/sapcc/helm-charts/pull/9480

mutax (approved):

PR #9480, opened 2025-08-12
Update gatekeeper policies to allow neutron-netns-prober

nikatza:
The neutron-netns-prober is a replacement for the ns-exporter. It needs to access network namespaces created by the neutron dhcp agents and run probes from there, e.g., pings and DNS queries. Thus, it requires 1) to be deployed on the same nodes where these agents run, and 2) run with privileged access to realize its function.


SuperSandro2000 (approved):
LGTM

PR #9505, opened 2025-08-14
disable rate limit for moonson3/ml-benchmarks to barbican whitelist

senolcolak:
we need to disable the rate limit to test SSE Barbican cache on Ceph RGW side


senolcolak:
PR not needed, changes done manually..

PR #9523, opened 2025-08-18
adds helm-chart for metal-api settings

stefanhipfel:
creates biosSettings for our metal-api automation.


system/metal-settings/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+name: metal-automation
```
defo89, line 2:
```suggestion
name: metal-settings
```

PR #9524, opened 2025-08-18
remove consistently firing alert

viennaa:


viennaa:
I pulled in @ashifnihalb to investigate the other alert you mentioned. Though this is not related.

Talking about this particular alert it is constantly having ~500 firing alerts which are ignored. This means we can also turn it off. This type of information should go into a dashboard if it is needed and have maybe a single alert pointing to this dashboard if it hits a certain limit.

Looking at possible results I think `600` would be a good limit to indicate, where the difference of these host paths is really big

```
vrops_hostsystem_storage_number_of_path < on(vccluster) group_left() (max(vrops_hostsystem_storage_number_of_path) by(vccluster)) > 600
```
24 alerts in 2 regions (average 12)

vs original
515 in 13 regions (average 39 alerts)

But still, if no one is looking at these INFO alerts anyway, what's the point of keeping them?

viennaa:
Ok, apparently there is some cc3test problem which is resulting in not firing `DatastoreMissingHost`. @ashifnihalb and Nishant need to debug this.

PR #9525, opened 2025-08-19
attempt new snmp f5 rules to attach project_id

notque:
much more complex. slower to run. likely not the ideal solution. may not handle our load. 

but it takes information we have available, and attached project_id to the metrics. 

Provided it works :)


notque:
closing, the hack is unneeded, the real problem was in the snmp exporter.

PR #9526, opened 2025-08-19
[designate] add new role cloud_dns_backup

mutax:
This new role allows the backup process to retrieve tsig keys via the openstack api to perform zone transfers from mdns.

In addition the new role has all cloud_dns_viewer permissions to retrieve a list of all zones.


proddi (approved):
schaut für mich funktionierend aus. qa kann es zeigen

PR #9527, opened 2025-08-19
Add validating admission policy chart

jknipper:
adds annotation check before CRD deletion

### DIFF
--- a/system/validating-admission-policy/templates/protect-crds.yaml
+++ b/system/validating-admission-policy/templates/protect-crds.yaml
@@ -1,0 +1,25 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: protect-crds.cloud.sap
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   ["*"]
+      apiVersions: ["*"]
+      operations:  ["DELETE"]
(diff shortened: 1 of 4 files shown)


system/validating-admission-policy/templates/protect-crds.yaml
```diff
@@ -0,0 +1,25 @@
...
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: protect-crds.cloud.sap
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   ["*"]
+      apiVersions: ["*"]
+      operations:  ["DELETE"]
```
Nuckal777, line 11: `"DELETECOLLECTION"` is missing.

Nuckal777 (changes requested):

system/validating-admission-policy/templates/protect-crds.yaml
Nuckal777: Ok, so `"DELETECOLLECTION"` [does not exist](https://github.com/kubernetes/kubernetes/blob/1bcfd5cee7905ee7c2b17cec0ecd88609693f44c/pkg/apis/admissionregistration/types.go#L1068) on `ValidatingAdmissionPolicies`, but it exists for the `verbs` field on `(Cluster)Roles`. 🙄

Nuckal777 (approved):

databus23 (approved):

PR #9538, opened 2025-08-20
Add BIOS version YAML files

stefanhipfel:
for Dell PowerEdge R640, R660, R860 and HPE ProLiant DL380 Gen11


defo89:
version bump is needed

PR #9540, opened 2025-08-21
keppel/swift: do not autoscale keep-image-pulled daemonsets

majewsky:
While we were chatting about the resource crunch in qa-de-1, I noticed that our keep-image-pulled daemonsets allow autoscaling, which in this case means autowasting resources. This is not a problem in QA because we do not deploy those there at all, but let's try not wasting prod resources, too.


SchwarzM (approved):
lgtm

PR #9544, opened 2025-08-21
feat(greenhouse-ccloud): rm runtime cluster from preset

timojohlo:


system/greenhouse-ccloud/templates/logs-rt-pluginpreset.yaml
```diff
@@ -0,0 +1,114 @@
...
+---
+apiVersion: greenhouse.sap/v1alpha1
+kind: PluginPreset
+metadata:
+  name: logs
+  namespace: {{ $.Release.Namespace }}
+spec:
+  clusterSelector:
+    matchLabels:
+      greenhouse.sap/cluster-presets-enabled: "true"
```
viennaa, line 26: This is the default now, please use this matcher from now on.

```suggestion
      greenhouse.sap/pluginpreset: "true"
```

system/greenhouse-ccloud/templates/logs-rt-pluginpreset.yaml
```diff
@@ -0,0 +1,110 @@
+{{- if .Values.openTelemetry.enabled -}}
```
uwe-mayer, line 1: @timojohlo and @viennaa 
The PluginPresets for the **new** runtime cluster are expected in [this repository](https://github.wdf.sap.corp/PlusOne/greenhouse/tree/main/p-eu-de-1/sci/support-groups/observability). I hope that came across...

Is the intention here to refactor the old one?

uwe-mayer (changes requested):

system/greenhouse-ccloud/templates/logs-rt-pluginpreset.yaml
timojohlo: The idea was to comply with having dedicated PluginPresets for cluster-type: `sci-k8s-runtime`. So your comment is valuable. I will go on and move the pluginpreset for the runtime cluster from this repository to the the one mentioned above.

timojohlo:
We will still need the preset as only the new clusters will be onboarded to the new org. Will close this for now.

PR #9577, opened 2025-08-27
[cinder] Add vcenter-operator pw rotation

PaulPickhardt:


openstack/cinder/templates/backup_deployment.yaml
```diff
@@ -8,6 +8,9 @@ options:
   jinja2_options:
     variable_start_string: '{='
     variable_end_string: '=}'
+    {{- if .Values.manage_service_user_passwords }}
+    uses-service-user: cinder
```
fwiesel, line 12: It is probably too late for that, but I do not think that adding it to "jinja2_options" was the right choice. These options are for the jinja2 rendering engine itself.

There was also the field `options`, which I think, would have been more appropriate.

I'll still approve it, as the relevant change already has been merged.

fwiesel (approved):

PR #9581, opened 2025-08-27
Thanos ruler vmware alert update to the query

frederic-boulianne:
update to the query for alert host in maintenance mode for more than 10 days to read correctly the openstack decommissioning tag that is set at the cluster level


kevin-fischer (approved):
+1

PR #9584, opened 2025-08-29
Updating NAT port allocation alert

loujith1985:
This alert is applicable to firmware versions prior to 17.15.x. In version 17.6.3, the metric is used to verify the NAT port chunks downloaded to the data plane. At router bootup, 45K ports were immediately downloaded from the Port-Manager.

From version 17.15.x onward, NAT requests all 45K ports from the Port-Manager, but the ports are downloaded to the data plane only when traffic requires them.

This alert applies only to Neutron routers running firmware version 17.6.3.


loujith1985:
@Kuckkuck @swagner-de @geisslet , Can you please review/approve this PR?

PR #9592, opened 2025-09-01
[oomkill-exporter] remove prometheus.io/targets annotations

trouaux:
metrics are scraped via pod scrape config

those values are not used and can be confusing, additionally they could accidentally be moved to labels and create duplicate scraping


viennaa (approved):
Good from my side. Rollout belongs to @jknipper / team services. We need another approval and rollout :)

jknipper (approved):

PR #9602, opened 2025-09-03
[andromeda] f5 agents deployment

ronchi-oss:


global/andromeda/templates/etc/_andromeda.yaml.tpl
```diff
@@ -43,4 +43,8 @@ audit_middleware_notifications:
 {{- end }}
 
 house_keeping:
-  enabled: true
\ No newline at end of file
+  enabled: true
+
+{{- if .Values.f5.enabled }}
```
notandy, line 48:
```
Error: Failed to render chart: exit status 1: Error: template: andromeda/templates/deployment.yaml:21:25: executing "andromeda/templates/deployment.yaml" at <include (print $.Template.BasePath "/configmap.yaml") .>: error calling include: template: andromeda/templates/configmap.yaml:11:3: executing "andromeda/templates/configmap.yaml" at <include (print .Template.BasePath "/etc/_andromeda.yaml.tpl") .>: error calling include: template: andromeda/templates/etc/_andromeda.yaml.tpl:48:14: executing "andromeda/templates/etc/_andromeda.yaml.tpl" at <.Values.f5.enabled>: nil pointer evaluating interface {}.enabled
```

could you add f5.enabled as a default value in values.yaml?

notandy (changes requested):

global/andromeda/templates/etc/_andromeda.yaml.tpl
ronchi-oss: Sure, I missed that in the end.

PR #9621, opened 2025-09-08
[cinder] Make cinder nanny alerts more explicit

jagoleni:
Alerts adjusted to only alert on db-purge, lock-files and quota-sync nannies. Consistency nanny is excluded as it runs on a different schedule (weekly). There is no extra alert required for the consistency nanny because it mostly logs stuff, is not important for platform operation and it may get deprecated.

### DIFF
--- a/openstack/cinder/alerts/kubernetes/cinder.alerts
+++ b/openstack/cinder/alerts/kubernetes/cinder.alerts
@@ -27,11 +27,11 @@ groups:
     # does not terminate, then new jobs were no longer scheduled. So we cover
     # non-terminating jobs by the second part.
     expr: >
-      time() - kube_cronjob_created{cronjob=~"cinder-nanny-.*"} >= 24*60*60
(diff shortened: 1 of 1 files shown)


joker-at-work (commented):
I'd feel better with excluding the nanny we don't want if possible. Reason: We will forget to add new nannies here, when we add them, and then don't get alerted.
Alternative could be a big note for attention in a place we have to touch for new nannies. Still, I'd prefer excluding here.

jagoleni:
> I'd feel better with excluding the nanny we don't want if possible. Reason: We will forget to add new nannies here, when we add them, and then don't get alerted. Alternative could be a big note for attention in a place we have to touch for new nannies. Still, I'd prefer excluding here.

PromQL/Go regex don't support negative lookahead (`cinder-nanny-(?!consistency).*`). I couldn't come up with an alternative that is still easy to read :/ I added an attention note in the values.yaml.

PR #9633, opened 2025-09-09
Fixing linting errors, Copy Bedrock alerts and rewrite support group …

SqiSch:
Fixing most of the linting errors of our alerts
Copy and modify a alert rule inside the template, if it is a Bedrock alert. Filtering for iaas Hosts and override the support-group to iaas-support


christophrichtersap (commented):
Thank you for the changes.

I succesfully `pint` your changes and check with `helm template` looks good to me.
Please additional increase the chart version.

PR #9665, opened 2025-09-17
nova serialproxy enable web

notandy:
Depends on a pr that includes serialproxy-web in the nova-image.

- **[nova] Serve of serialproxy-web static webpage for nova-serialproxy**
- **[nova-hypervisor-agents] Configure nova-compute to use nova-serialproxy**


fwiesel (approved):

openstack/nova-hypervisor-agents/templates/_helpers.tpl
```diff
@@ -10,10 +10,25 @@ https://identity-3.{{.Values.global.region}}.{{.Values.global.tld}}/v3
 enabled = {{ $config.enabled }}
 {{- if $config.enabled }}
 novncproxy_base_url = https://{{include "nova_console_endpoint_host_public" .}}:{{ .Values.global.novaConsolePortPublic }}/{{ $cell_name }}/novnc/vnc_auto.html?path=/{{ $cell_name }}/novnc/websockify
+server_listen = $my_ip
+server_proxyclient_address = $my_ip
```
joker-at-work, line 14: This seems unrelated to serialproxy

openstack/nova-hypervisor-agents/templates/_kvm-configmap.yaml.tpl
```diff
@@ -19,4 +19,5 @@ data:
     log_level = 3
     log_outputs = "3:stderr"
     listen_addr = "127.0.0.1"
+    live_migration_inbound_addr = $my_ip
```
joker-at-work, line 22: Sounds unrelated to serial - does "Utilizes $my_ip (host ip) to simplify configuration." mean "unrelated to serial, I also utilized $my_ip" everywhere it fits? If yes, can you do that in its own commit, please, and explain why the previous setting should be changed i.e. what the problems were we encountered?

openstack/nova/templates/console-configmap.yaml
```diff
@@ -2,10 +2,15 @@
...
+  {{- if $config.enabled }}
 [serial_console]
+# nova-serialproxy
+serialproxy_host = 0.0.0.0
+serialproxy_port = {{ $config.portInternal }}
+# nova-compute
 enabled = {{ $config.enabled }}
-  {{- if $config.enabled }}
 base_url = https://{{include "nova_console_endpoint_host_public" .}}:{{ .Values.global.novaConsolePortPublic }}/{{ $cell_name }}/serial
+serialproxy_address = $my_ip
```
joker-at-work, line 13: I cannot find that option in Nova. Can you point me to it?

Shouldn't it be up in the `#nova-serialproxy` section of the `[serial_console]` group?

Maybe you mean `proxyclient_address`? Then it looks like it belongs here, yes.

openstack/nova/templates/console-configmap.yaml
```diff
@@ -2,10 +2,15 @@
 {{- $cell_name := index . 1 }}
 {{- $config := index . 2 }}
 {{- with index . 0 }}
+  {{- if $config.enabled }}
 [serial_console]
+# nova-serialproxy
+serialproxy_host = 0.0.0.0
```
joker-at-work, line 8: This seems to be the default and we have no way to change it with this change. What's the reason we put it here? Visibility?

openstack/nova-hypervisor-agents/templates/hypervisors-kvm-daemonset.yaml
```diff
@@ -111,23 +111,7 @@ spec:
...
-            - name: OS_VNC__SERVER_PROXYCLIENT_ADDRESS
-              valueFrom:
-                fieldRef:
-                  fieldPath: status.hostIP
-            - name: OS_SPICE__SERVER_LISTEN
-              valueFrom:
-                fieldRef:
-                  fieldPath: status.hostIP
-            - name: OS_SPICE__SERVER_PROXYCLIENT_ADDRESS
+            - name: OS_DEFAULT__MY_IP
```
joker-at-work, line 114: Also unrelated to serial.

openstack/nova-hypervisor-agents/templates/_helpers.tpl
notandy: nope, it's crutial: https://docs.openstack.org/nova/latest/configuration/config.html#serial_console.proxyclient_address

This is the IP that the compute agent provides to the serialproxy service to connect to the Serial Port TCP port.

openstack/nova-hypervisor-agents/templates/_kvm-configmap.yaml.tpl
notandy: it is unrelated, true. Before that, https://github.com/sapcc/helm-charts/pull/9665/commits/26a684625fc163dd05a6e0f3f8b9482f5865f698#diff-150da47e3e75f3df9c7a003ee26ce9bb25e33382641bbcd5f9138eff501a30b5L114 injected the host ip for the configuration value `live_migration_inbound_addr`, alongside of other configuration flags that need the same value. 

I've refactored the nova-hypervisor-agents a little bit to introduce $my_ip, which is a variable that will hold he ip address of the host but can be reused anytime, thus reduces the amount of injected environment variables to only one.

openstack/nova/templates/console-configmap.yaml
notandy: you're right, good catch!

openstack/nova-hypervisor-agents/templates/hypervisors-kvm-daemonset.yaml
notandy: It is related, since `proxyclient_address` is supposed to use this value.

notandy:
> > Also includes correct configuration templating for nova-compute.
> 
> Could you please explain what this means? What was wrong and how did we fix it?
> 
> Could you explain how this now differs from the already-deployed nova-serialproxy? Do we need to inform our customers - if there are any - that the serialproxy is different now? Did you test it with Ironic, too? Afaik currently serial is only usable with Ironic nodes.

This configuration enables the use of serialproxy with kvm compute-agents.

As I see it, nova-serialproxy in it's current configuration is not usable at all. E.g. the base address was using https although there was never a web-gui provides by nova-serialproxy and the `proxyclient_address` was never correctly configured which is required for nova api to validate that the serial console is available. I don't even know why it's deployed.
Ironic seems to use shellinabox.

notandy:
I've moved the $my_ip refactoring into it's own commit so it's clear it's impacting, but not directly related, to the serial-console enablement.

joker-at-work (approved):
Talked to Jens, who provided some baremetal hosts that support `serial` console - but that's currently broken, because it ends up trying to use an endpoint providing shellinabox as a proxy .... so I don't think we break anything here.

PR #9679, opened 2025-09-22
Add multi region cloudprofile to gardener operator chart

jknipper:


Nuckal777 (changes requested):
Please don't flip any enabled switches by default, due to how this chart needs to be installed, ref: https://github.com/sapcc/helm-charts/tree/master/global/cc-gardener#readme

jknipper:
> Please don't flip any enabled switches by default, due to how this chart needs to be installed, ref: https://github.com/sapcc/helm-charts/tree/master/global/cc-gardener#readme

@Nuckal777 This is only for the chart test but okay.

PR #9696, opened 2025-09-24
[mariadb] add optional `owner-info` chart dependency

s10:
* Add an option to enable `owner-info` dependency chart to be able to install `mariadb` as a stand-alone chart
* chart version bumped


businessbean (approved):
lgtm

PR #9723, opened 2025-09-26
[Cinder] Split up cinder-api into external and internal

joker-at-work:

This requires a [change in the openstack/utils chart](https://github.com/sapcc/helm-charts/pull/9724) to change the endpoint host for the seed to Keystone.

Rolling this out also depends on the secrets getting updated.


fwiesel (dismissed):

hemna (dismissed):
Just have to fix the conflicts

joker-at-work:
> Error: the lock file (Chart.lock) is out of sync with the dependencies file (Chart.yaml). Please update the dependencies

is what the CI says

PR #9742, opened 2025-09-30
[percona-cluster] make force recovery option to work for primary

s10:
Set `safe_to_bootstrap: 1` even when cluster member is already a primary
when PXC_FORCE_BOOTSTRAP is set


galkindmitrii (approved):

businessbean (approved):
lgtm

PR #9765, opened 2025-10-01
[ironic] remove kos node update

sandzwerg:
We no longer need the node update. All nodes are already switched. But
we might now want to switch settings for some nodes for example for
secure boot use cases or in the future to switch all nodes to the
generic redfish driver and we don't need the kos to change nodes back
the whole time.


stefanhipfel:
if all those settings are already applied, there is no need for it anymore. 
In the past consoles got disabled and this script made sure consoles were always enabled.

PR #9794, opened 2025-10-02
[infra-mon] remove scrapeconif for windows-exporter on WSUS servers

BerndKue:
They will be shut down.


ibakshay:
Please bump the version in Chart.yaml. @BerndKue

PR #9829, opened 2025-10-07
chore(thanos): unify thanos settings

viennaa:
distributed query should be turned on everywhere
turn off thanos query engine because it would only show last 6h of data


ibakshay (approved):
🔥

PR #9859, opened 2025-10-08
[thanos-global] add distributed-with-overlapping-interval arg

trouaux:


ibakshay (approved):

viennaa (approved):
Nice, that's exactly what we need 👍

PR #9866, opened 2025-10-08
bump all images to latest

VoigtS:


majewsky (approved):
The `common/postgresql-ng` part was already done by the pipeline now. In general, please do not update `postgresql-ng` manually. (And we should work on getting the other ones autoupgraded as well.)

PR #10327, opened 2025-12-04
[ironic] Drop dumb-init thanks to subreaper in containerd

fwiesel:
Since begining of 2017, the kernel offers an option for a subreaper (torvalds/linux@c6c70f4455d1eda91065e93cc4f7eddf4499b105), and a little bit later (containerd/containerd@df4898) made use of it, rendering dumb-init superfluous.

Time to drop it.


sandzwerg (approved):
I'm fine with testing this. Having less dependencies for sure would be an improvement.

fwiesel:
We need to investigate, if the processes install a SIGTERM handler, otherwise the process won't get killed, as it will be PID 1.

PR #10478, opened 2026-01-06
[Nova] Seed common traits

joker-at-work:
We seed the traits that we documented in the operations documentation and that are not automatically created by Nova (e.g. not CUSTOM_EXTERNAL_CUSTOMER_*).

We skip adding CUSTOM_NUMASIZE_* traits, which we currently still create manually, because they _should_ be managed by `nova-compute`.

---

Created as draft, because `traits` are only supported in the Python part of the seeder and not the Go part and thus this does nothing at the moment.

### DIFF
--- a/openstack/nova/templates/seed.yaml
+++ b/openstack/nova/templates/seed.yaml
@@ -188,3 +188,15 @@ spec:
         inherited: true
   {{- end }}
   {{- end }}
+
+  traits:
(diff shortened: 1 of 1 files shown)


grandchild:
https://github.com/sapcc/kubernetes-operators/pull/124

grandchild (approved):

seb-kro (approved):

PR #10588, opened 2026-01-21
[secrets-injector] MutatingWebhookConfiguration namespace scope

trouaux:


uwe-mayer (changes requested):
Currently blocking this until further discussion.

sjentzschsap:
Would also need to update the `ClusterRole` name in the `ClusterRoleBnding` with the Release namespace suffix.

sjentzschsap (approved):

PR #10967, opened 2026-03-09
[designate] Add shared pools.

misamoylov:


kpawar-sap (dismissed):

openstack/designate/templates/etc/_designate-policy.yaml.tpl
```diff
@@ -164,7 +164,10 @@ update_pool: rule:admin
 delete_pool: rule:admin
 zone_create_forced_pool: rule:admin or role:cloud_dns_support or role:cloud_dns_ops or role:dns_zonemaster
 iaas_zone_create_forced_pool: rule:admin or rule:context_is_master
-
+share_pool: rule:admin
+unshare_pool: rule:admin
+get_pool_share: rule:admin or rule:context_is_dns_ops
```
galkindmitrii: The `rule:admin` part is actually redundant. It is already a part of `rule:context_is_dns_ops` so we don't need it here. I know it is also duplicated above in the find_pools, etc.

openstack/designate/templates/etc/_designate-policy.yaml.tpl
```diff
@@ -164,7 +164,10 @@ update_pool: rule:admin
...
+find_pool_shares: rule:admin or rule:context_is_dns_ops
```
galkindmitrii: Same as above comment.

openstack/designate/templates/etc/_pools.yaml.tpl
```diff
@@ -1,6 +1,9 @@
 {{- range $pool := .Values.bind_pools }}
 - name: {{ $pool.name }}
   description: Bind9 Pool
+  {{- with $pool.domain_id }}
+  domain_id: {{ . }}
```
galkindmitrii, line 5: That requires that `domain_id` is always set. As discussed, we only want to enable shared pools in a few regions where it is needed. Therefore the `domain_id` should only be added if it is present.

galkindmitrii (changes requested):

openstack/designate/templates/etc/_pools.yaml.tpl
misamoylov: Hmm. From my understanding from documentation it's not require domain_id field to be set. If domain_id not set it returns NIL, https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with and I guess it's also golang behaviour. I tested it and looks fine

```
helm template test_template --debug  
install.go:214: [debug] Original chart version: ""
install.go:231: [debug] CHART PATH: /Users/xxxxx/test_template

---
# Source: test_template/templates/test.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  pools: |
    - name: with-domain
      domain_id: 12345-1234-121324
    - name: without-domain

cat test_template/templates/test.yaml 
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  pools: |
{{- range $pool := .Values.bind_pools }}
    - name: {{ $pool.name }}
      {{- with $pool.domain_id }}
      domain_id: {{ . }}
      {{- end }}
{{- end }}


cat test_template/values.yaml         
bind_pools:
  - name: with-domain
    domain_id: 12345-1234-121324
  - name: without-domain
```

openstack/designate/values.yaml
```diff
@@ -75,6 +75,9 @@ rndc_keys:
 db_name: designate
 pool_id: 794ccc2c-d751-44fe-b57f-8894c9f5c842
 scheduler_filters: attribute, pool_id_attribute, in_doubt_default_pool, fallback
+scheduler_filters_shared_pools: attribute, pool_id_attribute, pool_domains_attribute_filter, domain_id
+shared_pools: false
```
galkindmitrii: please name it `shared_pools_enabled`

openstack/designate/templates/etc/_pools.yaml.tpl
galkindmitrii: Ok

galkindmitrii (changes requested):

openstack/designate/templates/etc/_designate-policy.yaml.tpl
misamoylov: Fixed

openstack/designate/templates/etc/_designate-policy.yaml.tpl
misamoylov: Fixed

openstack/designate/values.yaml
misamoylov: Done

galkindmitrii (approved):
Thx!

PR #11217, opened 2026-03-30
[Barbican] TLS Implementation

rajivmucheli:
## What this does

- Migrate Barbican API from paste.httpserver to Apache+mod_wsgi (keystone-style startup script)
- TLS termination at Apache (port 443) for BSI TR-02102-2 compliance
- NGINX ingress TLS passthrough mode when `tls.enabled=true`
- TLS Bootstrap Job (Helm pre-install/pre-upgrade hook, fail-closed)
- TLS Rotation CronJob (weekly, configurable renewal threshold)
- BSI-compliant TLS protocol hardening (TLS 1.2/1.3, cipher suites, ECDHE curves)
- Linkerd mTLS skip for HTTPS inbound port
- Chart version bump: 0.8.6 -> 0.9.0

## Review fixes

- **imageTag**: removed `latest`, now a required field -- fails fast at render time if unset
- **hook-delete-policy**: added `before-hook-creation` so failed bootstrap jobs don't block subsequent helm upgrade
- **RBAC**: fixed `create`+`resourceNames` rule (Kubernetes silently ignores `create` with `resourceNames`); split into two clean rules
- **Dual WSGI daemon groups**: collapsed `barbican-api-https` + `barbican-api-http` into a single `barbican-api` process group -- was doubling the mod_wsgi worker pool
- **LimitRequestBody 114688**: added to both VirtualHosts (matches keystone)
- **Startup script**: aligned with keystone -- removed `a2enmod` calls and fallback path lookup that masked broken images; fail fast if binary missing
- **Trailing newlines**: fixed in 3 generated files
- **Inline comments**: removed verbose WP-xxxx task references from all templates

## Merge from master

- Added `rbac.enabled: true` from master
- Added `barbican-tmp` and `barbican-run` emptyDir volumes from master
- Set `readOnlyRootFilesystem: false` -- Apache writes to /var/run/apache2, /var/log/apache2 at runtime
- Added `tls: enabled: false` to CI test values to prevent nil pointer in linter


BerndKue (dismissed):

rajivmucheli:
fixed in https://github.com/sapcc/helm-charts/pull/12641

PR #11230, opened 2026-03-30
 SAML 2.0 Service Provider Support for Keystone

bbobrov:
Adds Shibboleth SP 3 (mod_shib) integration to the Keystone Helm chart, enabling SAML 2.0 federated authentication with per-tenant cryptographic isolation.

### Shibboleth SP Configuration
- mod_shib integration with Apache (WSGIScriptAliasMatch, VirtualHost handler, handlerSSL=false for TLS-terminated ingress)
- Per-tenant ApplicationOverride with dedicated Sessions block, handlerURL, MetadataProvider, CredentialResolver
- RequestMapper with nested Path elements for multi-tenant routing
- NameIDAttributeDecoder for unspecified format (SAP IAS sends P-numbers)
- SAP IAS plain-name attribute mapping (mail, first_name, last_name, user_uuid)
- Chaining CredentialResolver for SP key rotation support

### Runtime Configuration
- Python script generates shibboleth2.xml and federation-saml.conf at startup from tenant-list ConfigMap
- Zero per-tenant values in the Keystone chart — all tenant data lives in the federation chart
- Stakater Reloader triggers rolling restarts on federation ConfigMap/Secret changes

### SP Key Management
- Per-tenant, per-region RSA-4096 key pairs stored in Vault
- Distroless Go init container (keystone-saml-sp-init) copies SP keys from Vault-injected Secret to tmpfs
- Secret volume defaultMode 0444 for nonroot container compatibility

### Monitoring
- Prometheus alert for SAML assertion validation failures


BerndKue (dismissed):

openstack/keystone/templates/bin/_keystone_api.sh.tpl
```diff
@@ -41,6 +41,19 @@ function start () {
...
+  {{- if .Values.federation.saml.enabled }}
+  # The Shibboleth SP config (shibboleth2.xml) and per-tenant Apache
+  # configuration (federation-saml.d/*.conf) are mounted read-only from
+  # ConfigMaps shipped by the keystone-saml-federation chart. No runtime
+  # generation. Just make sure the writable runtime directories exist
+  # before shibd starts.
+  mkdir -p /var/run/shibboleth /var/cache/shibboleth
+  if command -v shibd &> /dev/null; then
+    echo "Starting shibd daemon for SAML federation..."
+    shibd -t && shibd -f || echo "WARN: shibd config validation failed or daemon not started, check logs above"
```
BerndKue: You don't exit in case of an error.
Wouldn't it be better to add something like
shibd -t || { echo "ERR: shibd config test failed"; exit 1; }
shibd -f || echo "WARN: shibd daemon not started, check logs above"

Maybe even also the second command with an exit and as an error. Because in case you enable SAML you would also want it to come up, wouldn't you?

BerndKue (dismissed):

openstack/keystone/templates/bin/_keystone_api.sh.tpl
bbobrov: Done, thank you

BerndKue (approved):

PR #11246, opened 2026-03-31
Add ssl-passthrough ingress-nginx controller

bbobrov:
Add a dedicated ingress-nginx controller for SSL passthrough, deployed as a separate instance with its own ingress class (nginx-passthrough).

Services that need the nginx.ingress.kubernetes.io/ssl-passthrough annotation require --enable-ssl-passthrough on the ingress-nginx controller. However, enabling this flag on the primary controller intercepts all HTTPS traffic through a Go TCP proxy for SNI inspection, introducing per-connection overhead for every HTTPS connection — not just passthrough ones.

To avoid this, we deploy a separate ingress-nginx controller instance (ingress-nginx-passthrough) using the same alias pattern as ingress-nginx-external. Services that need SSL passthrough can opt in by setting ingressClassName: nginx-passthrough on their Ingress resources. All other HTTPS traffic continues through the primary controller unaffected.

The new controller is enabled by default.

### DIFF
--- a/system/kube-system-metal/Chart.lock
+++ b/system/kube-system-metal/Chart.lock
@@ -32,6 +32,9 @@ dependencies:
 - name: ingress-nginx
   repository: https://kubernetes.github.io/ingress-nginx
   version: 4.15.1
+- name: ingress-nginx
+  repository: https://kubernetes.github.io/ingress-nginx
+  version: 4.15.1
 - name: wormhole
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
   version: 3.1.8
@@ -77,5 +80,5 @@ dependencies:
 - name: validating-admission-policy
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
   version: 0.0.2
-digest: sha256:57e2b121a9d4b3d4ef874a32f9dbe8a7cce8607cc5d7cbbd00905c4a88ebd2bf
-generated: "2026-04-10T12:54:48.034012+03:00"
+digest: sha256:76ee6045de2e20d1549ab5d5c567966b35c910f19ee2a70a28f11ef4574643c8
+generated: "2026-04-21T18:47:48.507313157+02:00"
--- a/system/kube-system-metal/values.yaml
+++ b/system/kube-system-metal/values.yaml
@@ -82,10 +82,6 @@ ingress-nginx:
     # See https://github.com/kubernetes/ingress-nginx/issues/7837 .
     allowSnippetAnnotations: false
 
-    extraArgs:
-      # Required for ssl-passthrough ingress annotations to take effect.
-      enable-ssl-passthrough: true
-
     config:
       custom-http-errors: "495"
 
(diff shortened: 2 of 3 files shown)


BerndKue (dismissed):

system/kube-system-metal/values.yaml
```diff
@@ -82,6 +82,10 @@ ingress-nginx:
     # See https://github.com/kubernetes/ingress-nginx/issues/7837 .
     allowSnippetAnnotations: false
 
+    extraArgs:
+      # Required for ssl-passthrough ingress annotations to take effect.
+      enable-ssl-passthrough: true
```
tz3: is enabling this globally intentional, or should we start with a per-region override (e.g. qa-de-x) until the Keystone/Barbican passthrough ingresses are ready?

tz3 (commented):
thanks for the PR, only one thing i know this is a prerequisite for ADR, not a standalone change but left one comment worth considering.

system/kube-system-metal/values.yaml
bbobrov: I don't see a problem in enabling this globally.
This change does not change existing behavior - it allows developers to use it when they need it. After this change is merged, keystone, barbican and other services still will have TLS terminated on nginx. To enable passthrough there separate PRs are needed to their helm-charts.

s10: This option may have an impact on all ingresses, even those not using ssl passthrough

https://kubernetes.github.io/ingress-nginx/user-guide/tls/#ssl-passthrough
>This feature is implemented by intercepting all traffic on the configured HTTPS port (default: 443) and handing it over to a local TCP proxy. This bypasses NGINX completely and introduces a non-negligible performance penalty.

So it might be useful to introduce another ingress-nginx controller with a different ingress class and use that if ssl passthrough is needed.

tz3: Thanks @s10 — this is one of things what I was worried about. The performance penalty applies to all ingresses, not just passthrough ones. Given that, the separate ingress-nginx controller with its own ingress class seems like the cleaner approach,isolates the impact to only the services that need it.

bbobrov: OK, done. There is now a new ingress.

tz3 (approved):
Thanks for the update. LGTM!

JoJoPuppe (approved):

system/kube-system-metal/Chart.lock
```diff
@@ -32,6 +32,9 @@ dependencies:
 - name: ingress-nginx
   repository: https://kubernetes.github.io/ingress-nginx
   version: 4.15.1
+- name: ingress-nginx
```
kayrus, line 35: this chart is already defined above

bbobrov: Yes. And i was wondering if i should add it or not.
Should i delete it now, or should it stay?

kayrus (approved):

system/kube-system-metal/values.yaml
```diff
@@ -230,6 +230,30 @@ ingress-nginx-external:
...
+
+  controller:
+    ingressClass: nginx-passthrough
+    electionID: passthrough-ingress-controller-leader
+
+    extraArgs:
+      # Required for ssl-passthrough ingress annotations to take effect.
+      # Deployed as a separate controller to avoid the performance penalty
+      # on all HTTPS traffic that --enable-ssl-passthrough introduces.
+      enable-ssl-passthrough: true
```
abhijith-darshan, line 252:
```suggestion
      enable-ssl-passthrough: “”
```

It is just a flag on controller

bbobrov:
For 007 WP-4 it was decided not to use ingress-nginx passthrough feature, and instead use direct service exposure. So this patch is not needed for keystone anymore.

PR #11292, opened 2026-04-06
[barbican]: Return 404 for favicon.ico requests

rajivmucheli:
Browsers automatically request /favicon.ico which causes Pecan framework to generate content-type errors:
'Controller on_get does not support content_type image/vnd.microsoft.icon'

This is because Pecan guesses content type from file extensions and Barbican's VersionsController only allows application/json and application/json-home content types.

Adding nginx server-snippet to return 404 for /favicon.ico requests prevents these errors from cluttering the logs.

Change-Id: I1624F460B8A7408D87F5998416D0E62E

### DIFF
--- a/openstack/barbican/templates/ingress.yaml
+++ b/openstack/barbican/templates/ingress.yaml
@@ -18,6 +18,15 @@ metadata:
     disco: "true"
   {{- end }}
   {{- include "utils.linkerd.ingress_annotation" . | indent 4 }}
+    # Return 404 for favicon.ico requests to prevent Pecan content-type errors
+    # Browsers automatically request /favicon.ico which triggers:
+    # "Controller 'on_get' does not support content_type 'image/vnd.microsoft.icon'"
+    nginx.ingress.kubernetes.io/server-snippet: |
+      location = /favicon.ico {
+        return 404;
+        access_log off;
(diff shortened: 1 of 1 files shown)


BerndKue (approved):

openstack/barbican/templates/ingress.yaml
```diff
@@ -18,6 +18,15 @@ metadata:
     disco: "true"
   {{- end }}
   {{- include "utils.linkerd.ingress_annotation" . | indent 4 }}
+    # Return 404 for favicon.ico requests to prevent Pecan content-type errors
+    # Browsers automatically request /favicon.ico which triggers:
+    # "Controller 'on_get' does not support content_type 'image/vnd.microsoft.icon'"
+    nginx.ingress.kubernetes.io/server-snippet: |
```
s10: This would break barbican ingress, as ingress-nginx rejects annotations with `Critical` risk level, see https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations-risk/

s10 (changes requested):

rajivmucheli:
fixed in https://github.com/sapcc/barbican/pull/46

PR #11330, opened 2026-04-08
[hermes] Add log-router for data plane audit events

notque:
## Summary
- Adds log-router as a new opt-in component (`logRouter.enabled: false` by default)
  to the hermes chart for routing data plane audit events to customer Ceph/S3 buckets
- Consumes from a dedicated `dataplane.audit` RabbitMQ queue (separate from
  the existing `notifications.info` control plane queue)
- PostgreSQL (postgresql-ng) dependency for dynamic per-tenant routing configuration
- New `audit_admin` Keystone role 
- Phase 1: QA/testing only — validates end-to-end event flow

## New Files
- `templates/log-router-statefulset.yaml` — StatefulSet with WAL PVC
- `templates/log-router-configmap.yaml` — Non-secret config + policy.json
- `templates/log-router-secret.yaml` — RabbitMQ, S3, Keystone credentials
- `templates/log-router-service.yaml` — ClusterIP + headless service
- `templates/log-router-pdb.yaml` — PodDisruptionBudget
- `templates/_log-router-utils.tpl` — Image helper + common env vars

## Modified Files
- `Chart.yaml` — postgresql-ng dependency, version bump to 0.2.0
- `values.yaml` — logRouter + log-router-postgresql sections
- `keystone-seed.yaml` — audit_admin role (conditional on logRouter.enabled)
- `ci/test-values.yaml` — log-router test overrides

## Test plan
- [ ] `helm template` renders cleanly with `logRouter.enabled: false` (no log-router resources)
- [ ] `helm template` renders cleanly with `logRouter.enabled: true` and test values
- [ ] Verify postgresql-ng secret/service naming matches _utils.tpl references
- [ ] Deploy to QA and validate end-to-end event flow
- [ ] Confirm existing hermes components are unaffected


joluc (dismissed):
LGTM

openstack/hermes/templates/keystone-seed.yaml
```diff
@@ -79,6 +80,14 @@ spec:
           role: audit_viewer
         - group: CCADMIN_MONITORING_USERS
           role: audit_viewer
+      {{- if $.Values.logRouter.enabled }}
+      - name: cloud_admin
+        role_assignments:
+        - user: hermes@Default
+          role: audit_admin
+        - group: CCADMIN_CLOUD_ADMINS
```
bbobrov: I decided to comment on this line and got into the rabbit hole of issues.

1. Nothing technically prevents an external customer from creating a group with this name. So it might be, that group CCADMIN_CLOUD_ADMINS exists in both ccadmin domain and `iaas-tstoic` domain. Which will be chosen by seeder?
2. We have many groups mentioned like this already. Why it doesn't cause issues!? How do they work?
3. It does not. Quote:
>  After a thorough analysis of the codebase, the short answer is: **it wouldn't do anything at all** — group role assignments are defined in the spec/schema but the reconciliation logic to process them is not implemented.

Could you please try this out in qa first and make sure that the assignment is getting created?

Regardless of the outcome of the experiment, please make sure that all names have corresponding domain names.

openstack/hermes/templates/keystone-seed.yaml
```diff
@@ -20,6 +20,7 @@ spec:
 
   roles:
     - name: audit_viewer
+    - name: audit_admin
```
bbobrov, line 23: What happens if a colleague from another domain a customer coming through federation assigns this role? Should they be able to?

If they should be able to, then good.
If they should not be able to, then this role is breaking our convention on role naming. Roles specific to cloud administrators start with `cloud_` and must also be defined in blocklisted roles in keystone's policy.yaml.

openstack/hermes/templates/keystone-seed.yaml
```diff
@@ -79,6 +80,14 @@ spec:
           role: audit_viewer
         - group: CCADMIN_MONITORING_USERS
           role: audit_viewer
+      {{- if $.Values.logRouter.enabled }}
+      - name: cloud_admin
```
bbobrov: If you look above, you will notice that before the assignments were created on project `master`. Now they are created on `cloud_admin`. Why?

bbobrov (changes requested):
> New audit_admin Keystone role scoped to ccadmin/cloud_admin project

Sorry, but this makes no sense. A role is not scoped to anything - it exists by itself. It can be assigned to a user and project, and then the user can get a bearer token on the project with this role, and go to hermes with it.

What you do is probably a rule that utilizes this combination... but to that i have comments inline.

openstack/hermes/templates/keystone-seed.yaml
notque: CCADMIN_CLOUD_ADMINS@CCADMIN will be required from Boris

bbobrov:
```suggestion
        - group: CCADMIN_CLOUD_ADMINS@ccadmin
```

openstack/hermes/templates/log-router-configmap.yaml
```diff
@@ -0,0 +1,40 @@
...
+  LOG_ROUTER_RABBITMQ_ROUTING_KEY: {{ .Values.logRouter.rabbitmq.routing_key | quote }}
+  LOG_ROUTER_RABBITMQ_PREFETCH: {{ .Values.logRouter.rabbitmq.prefetch | quote }}
+  LOG_ROUTER_S3_ENDPOINT: {{ required ".Values.logRouter.s3.endpoint must be set" .Values.logRouter.s3.endpoint | quote }}
+  LOG_ROUTER_S3_REGION: {{ .Values.logRouter.s3.region | quote }}
+  LOG_ROUTER_S3_BUCKET: {{ required ".Values.logRouter.s3.bucket must be set" .Values.logRouter.s3.bucket | quote }}
+  LOG_ROUTER_S3_PREFIX: {{ .Values.logRouter.s3.prefix | quote }}
+  policy.json: |
+    {
+      "cluster_viewer": "project_domain_name:ccadmin and project_name:cloud_admin",
+      "cluster_admin": "project_domain_name:ccadmin and project_name:cloud_admin and (role:admin or role:audit_admin)",
```
bbobrov: We just had a call with Nathan about this. The intention is that actual customers and our LoB colleagues can get these roles.
In this case, this needs to be changed to allow this - the rules mentioned here now limit viewing to tokens scoped to project cloud_admin@ccadmin.

notque:
@bbobrov I've attempted to change this to what I've actually intended as we discussed on the call. Tomorrow if you have time, I'd appreciate your consideration on it, and again thank you for giving me a proper check.

bbobrov (dismissed):
from authorization pov this looks now better, thanks

LeoZhangZXZ (dismissed):

LeoZhangZXZ (dismissed):

LeoZhangZXZ (dismissed):

openstack/hermes/ci/test-values.yaml
```diff
@@ -89,3 +89,50 @@ rabbitmq_notifications:
...
+    limits:
+      memory: "128Mi"
+      cpu: "100m"
+
+log-router-postgresql:
+  postgresVersion: 18
+  databases:
+    log_router: {}
+  users:
+    log_router: {}
```
viennaa: according to postgresql-ng they shall not include `_`. will fail

```suggestion
    log-router: {}
  users:
    log-router: {}
```

openstack/hermes/templates/_log-router-utils.tpl
```diff
@@ -0,0 +1,114 @@
...
+      key: RABBITMQ_PASSWORD
+- name: RABBITMQ_URL
+  value: "amqp://$(RABBITMQ_USER):$(RABBITMQ_PASSWORD)@{{ $.Release.Name }}-rabbitmq-notifications.{{ $.Release.Namespace }}.svc:{{ $.Values.logRouter.rabbitmq.port }}/"
+- name: LOG_ROUTER_DB_PASSWORD
+  valueFrom:
+    secretKeyRef:
+      name: '{{ $.Release.Name }}-pguser-log-router'
+      key: postgres-password
+- name: LOG_ROUTER_DB_URL
+  value: "postgres://log-router:$(LOG_ROUTER_DB_PASSWORD)@{{ $.Release.Name }}-postgresql.{{ $.Release.Namespace }}.svc:5432/log-router?sslmode=disable"
```
viennaa: ideally the `log-router` occurrences of these strings go into `values.yaml`. if `log-router-postgresql` values change, this fails silently.

```
values.yaml:
logRouter:
  db:
    user: log-router
    name: log-router
```

openstack/hermes/values.yaml
```diff
@@ -182,3 +182,59 @@ elasticsearch_hermes_kibana:
...
+  resources:
+    requests:
+      memory: "256Mi"
+      cpu: "100m"
+    limits:
+      memory: "512Mi"
+      cpu: "500m"
+
+log-router-postgresql:
+  postgresVersion: 17
```
viennaa, line 232: ci test uses `18`, adjust either this or ci.

openstack/hermes/templates/log-router-statefulset.yaml
```diff
@@ -0,0 +1,110 @@
...
+        runAsGroup: 4200
+        fsGroup: 4200
+        fsGroupChangePolicy: "OnRootMismatch"
+        supplementalGroups: [4200]
+      volumes:
+        - name: log-router-secrets
+          emptyDir: {}
+      initContainers:
+        - name: fix-permissions
+          image: {{ required ".Values.init_image is missing" .Values.logRouter.init_image }}
```
viennaa: Different values are checked here, is this on purpose?

openstack/hermes/values.yaml
```diff
@@ -182,3 +182,59 @@ elasticsearch_hermes_kibana:
 
 opensearch_hermes:
   enabled: false
+
+logRouter:
+  enabled: false
+  image_tag: DEFINED_BY_PIPELINE
```
viennaa: I'd leave this empty to not end up with an image tag `DEFINED_BY_PIPELINE` in case it is not defined by the pipeline :D 
```suggestion
  image_tag:
```

+ add to ci values for linting

openstack/hermes/values.yaml
```diff
@@ -182,3 +182,59 @@ elasticsearch_hermes_kibana:
...
+    log-router: {}
+  users:
+    log-router: {}
+  config:
+    max_connections: 64
+  persistence:
+    enabled: true
+    size: 10Gi
+  alerts:
+    support_group: observability
```
viennaa, line 243: Not used, is it?

Also it is in ci values with the same value.

viennaa (commented):
Thanks for providing the log-router PR. Couple of remarks before it can go live.

openstack/hermes/templates/log-router-statefulset.yaml
LeoZhangZXZ: typo

LeoZhangZXZ:
@viennaa thanks for the review, good catch

openstack/hermes/templates/_log-router-utils.tpl
```diff
@@ -0,0 +1,114 @@
...
+  valueFrom:
+    secretKeyRef:
+      name: log-router-secret
+      key: RABBITMQ_PASSWORD
+- name: RABBITMQ_URL
+  value: "amqp://$(RABBITMQ_USER):$(RABBITMQ_PASSWORD)@{{ $.Release.Name }}-rabbitmq-notifications.{{ $.Release.Namespace }}.svc:{{ $.Values.logRouter.rabbitmq.port }}/"
+- name: LOG_ROUTER_DB_PASSWORD
+  valueFrom:
+    secretKeyRef:
+      name: '{{ $.Release.Name }}-pguser-log-router'
```
viennaa: this user also needs to be changed. no?

```suggestion
      name: '{{ $.Release.Name }}-{{$.Values.logRouter.db.user}}'
```

openstack/hermes/templates/keystone-seed.yaml
viennaa, line 24: The PR description says:

```
keystone-seed.yaml — audit_admin role (conditional on logRouter.enabled)
```

This setting does not depend on logRouter enabled. Not sure if we always want this, but I think it should be gated.

viennaa, line 24: 2nd question: where is this actually being used? I don't see it being referred in the `domains` section below.

openstack/hermes/templates/_log-router-utils.tpl
LeoZhangZXZ: it is not called hermes-log-router, but hermes-pguser-log-router, which is determined by the postgrepsql-ng helm chart, but i think we can write it as` {{ $.Release.Name }}-pguser-{{$.Values.logRouter.db.user}}`

openstack/hermes/templates/keystone-seed.yaml
LeoZhangZXZ: I think admin is needed for some operations (e.g. purging audit events), not very sure

openstack/hermes/values.yaml
LeoZhangZXZ: this is used in postng helm chart

joluc (approved):

PR #11350, opened 2026-04-10
bump cert manager

adziauho:
## Breaking changes

**1.14**
The startupapicheck job uses a new OCI image called "startupapicheck", instead of the ctl image. If you run in an environment in which images cannot be pulled, be sure to include the new image.
The KeyUsage and BasicConstraints extensions will now be encoded as critical in the CertificateRequest's CSR blob
  **1.16**
Helm schema validation may reject your existing Helm values files if they contain typos or unrecognized fields.
Venafi Issuer may fail to renew certificates if the requested duration conflicts with the CA’s minimum or maximum policy settings in Venafi.
Venafi Issuer may fail to renew Certificates if the issuer has been configured for TPP with username-password authentication.

### DIFF
--- a/system/kube-system-admin-k3s/Chart.lock
+++ b/system/kube-system-admin-k3s/Chart.lock
@@ -16,7 +16,7 @@ dependencies:
   version: 2.0.13
 - name: cert-manager
   repository: https://charts.jetstack.io
-  version: v1.13.3
+  version: v1.20.1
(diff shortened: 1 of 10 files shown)


kayrus (approved):

defo89 (changes requested):
as agreed, lets reduce the scope to one cluster first

PR #11352, opened 2026-04-10
fix regex pattern and linting errors

thecodingshrimp:
fixing:
- IsMatch only supports regex pattern, but grok patterns were used 
- Linter interprets `[...] user:` in line 282 as an object declaration instead of grok pattern.


system/audit-logs-otel/templates/audit-logs-collector.yaml
```diff
@@ -273,13 +273,14 @@ spec:
               - set(log.attributes["log.kind"], "cadf_action") where IsMatch(log.attributes["msg"], "^\\s*cadf action for '")
               - set(log.attributes["log.kind"], "target_type_uri") where IsMatch(log.attributes["msg"], "^\\s*target type URI of requests '")
               - set(log.attributes["log.kind"], "authentication") where IsMatch(log.attributes["msg"], "^\\s*Authenticating")
-              - set(log.attributes["log.kind"], "failed_authentication") where IsMatch(log.attributes["msg"], "^\\s*Authorization failed. The request you have made requires authentication. from %{IP:ip}%{GREEDYDATA}")
+              - set(log.attributes["log.kind"], "failed_authentication") where IsMatch(log.attributes["msg"], "^\\s*Authorization failed. The request you have made requires authentication. from [0-9a-fA-F\\.:]+")
```
Kuckkuck, line 276: doesn't it make sense to just do the search for `Authorization failed. The request you have made requires authentication.` if the fields are not used?

PR #11353, opened 2026-04-10
run ct link against whole repo

defo89:
Running ct link against whole repo should be fine since we would return 0 if chart is not owned by needed team.


s10 (dismissed):

kayrus (changes requested):
ct is not needed here

s10 (approved):

kayrus (approved):

PR #11362, opened 2026-04-13
nova-hypervisor-agents: Wait for certificate

fwiesel:
We have a potential race between the compute agent trying to connect to the libvirt socket, which in turn would trigger the start of libvirt and the provisioing of the TLS certificates.

Without this check, either we prematurely start the service without having TLS certificates installed, or we crashloop because libvirt could not be started due to the missing TLS certificates (depending on how libvirt is configured).

By waiting for the certificate to be deployed, we model that dependency explicitly hopefully making it easier to identify the problem.

### DIFF
--- a/openstack/nova-hypervisor-agents/templates/hypervisors-kvm-daemonset.yaml
+++ b/openstack/nova-hypervisor-agents/templates/hypervisors-kvm-daemonset.yaml
@@ -26,6 +26,7 @@ spec:
         configmap-etc-hash: {{ include (print .Template.BasePath "/etc-configmap.yaml") . | sha256sum }}
         configmap-kvm-etc-hash: {{ include "kvm_configmap" . | sha256sum }}
     spec:
+      serviceAccountName: {{ .Values.hypervisorAgentsServiceAccountName | default .Release.name }}
       terminationGracePeriodSeconds: {{ $hypervisor.default.graceful_shutdown_timeout | default .Values.defaults.default.graceful_shutdown_timeout | add 5 }}
       hostNetwork: true
       hostPID: true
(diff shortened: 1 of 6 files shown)


openstack/nova-hypervisor-agents/templates/hypervisors-kvm-daemonset.yaml
```diff
@@ -55,6 +56,21 @@ spec:
...
+        - name: COMMAND
+          value: "true"
+        - name: POD_NAME
+          valueFrom: {fieldRef: {fieldPath: metadata.name}}
+        - name: NAMESPACE
+          valueFrom: {fieldRef: {fieldPath: metadata.namespace}}
+        - name: NODE_NAME
+          valueFrom: {fieldRef: {fieldPath: spec.nodeName}}
+        - name: DEPENDENCY_CUSTOM_RESOURCE
+          value: '[{"apiVersion":"kvm.cloud.sap/v1","kind":"Hypervisor","name":"$(NODE_NAME)","fields":[{"key":".status.conditions[?(@.type==\"TLSCertificateInstalled\")].status","value":"True"}]}]'
```
joker-at-work, line 73: How does it work that the NODE_NAME env variable ends up in the string here? `$()` does not seem to be shell variable syntax.

fwiesel: Exactly... that is a variable interpolation that [K8S does](https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/#define-an-environment-dependent-variable-for-a-container).

joker-at-work (dismissed):

notandy (changes requested):
Why you're adding a rolebinding and a clusterrolebinding?
There only differ being namespace vs clusterwide.

openstack/nova-hypervisor-agents/templates/hypervisors-kvm-daemonset.yaml
notandy, line 73: This will likely bring more problems than it fixes.
A known problem is if deleting a hypervisor CRO, the CRO will be regenerated but will missing this specific condition.
Maybe a script that check for an working libvirt socket would be sufficient?

fwiesel:
> Why you're adding a rolebinding and a clusterrolebinding?
There only differ being namespace vs clusterwide.

Yes, and hypervisor is clusterwide. I had a rolebinding for the hypervisor crd before and it didn't work.

openstack/nova-hypervisor-agents/templates/hypervisors-kvm-daemonset.yaml
fwiesel: > Maybe a script that check for an working libvirt socket would be sufficient?

That would trigger socket activation. Something I wanted to avoid.

notandy:
> > Why you're adding a rolebinding and a clusterrolebinding?
> > There only differ being namespace vs clusterwide.
> 
> Yes, and hypervisor is clusterwide. I had a rolebinding for the hypervisor crd before and it didn't work.

make sense, then you can remove the regular rolebinding since it has no effect?

fwiesel:
> make sense, then you can remove the regular rolebinding since it has no effect?

The role & rolebinding gives access to the pods in the namespace.
The clusterrole & clusterrolebinding gives access to the clusterwide hypervisor resource.

notandy:
> > make sense, then you can remove the regular rolebinding since it has no effect?
> 
> The role & rolebinding gives access to the pods in the namespace. The clusterrole & clusterrolebinding gives access to the clusterwide hypervisor resource.

I see now. In QA, you're deploying two releases of the same chart. How do you want to ensure that the roles don't get overwritten by each other? Do you want to configure their name manually, why not using the release name?

fwiesel:
> In QA, you're deploying two releases of the same chart. How do you want to ensure that the roles don't get overwritten by each other? Do you want to configure their name manually,  why not using the release name?

Good point. I will change that.

fwiesel:
Done, I have left the "serviceAccountName" variable, but set it to `null`, defaulting to `.Release.Name`.

notandy (approved):

joker-at-work (approved):

PR #11369, opened 2026-04-13
Add IAS configuration to edith

darpan92:
Add IAS vars that will used inside the container during app runtime.


barath-source (approved):

martinpink (approved):
lgmt

PR #11373, opened 2026-04-14
[cc-shoots-compute] replace hard-coded channel logic with generic imageVersions map

anokfireball:
Replace the two hard-coded channels (release/preRelease) with a generic imageVersions map. Channel/version resolution now looks up the channel name directly in `.Values.imageVersions`. Adding a new channel requires only a `values.yaml` entry.

`-dev` channels (e.g. `release-dev`) are derived automatically from their base channel by replacing `usi` with `usidev` in the version string; no explicit entry is needed, but an explicit entry always takes precedence.

The prerelease channel retains its existing special ignition/systemd path.

### DIFF
--- a/system/cc-shoots-compute/templates/kvm-shoot.yaml
+++ b/system/cc-shoots-compute/templates/kvm-shoot.yaml
@@ -155,13 +155,18 @@ spec:
     workers:
     {{- range $workerkey, $worker := $cluster.workers }}
       {{- $workerChannel := coalesce $worker.channel $cluster.channel "release" }}
-      {{- $imageVersion:= "unset" }}
-      {{- if eq $workerChannel "preRelease" }}
-      {{- $imageVersion = coalesce $worker.preReleaseImageVersion $cluster.preReleaseImageVersion $.Values.preReleaseImageVersion }}
-      {{- else if eq $workerChannel "release" }}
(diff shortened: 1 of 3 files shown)


system/cc-shoots-compute/templates/kvm-shoot.yaml
```diff
@@ -155,13 +155,18 @@ spec:
...
-      {{- $imageVersion = coalesce $worker.imageVersion $cluster.imageVersion $.Values.imageVersion }}
+      {{- $imageVersion := index $.Values.imageVersions $workerChannel }}
+      {{- if not $imageVersion }}
+      {{- if hasSuffix "-dev" $workerChannel }}
+      {{- $baseChannel := trimSuffix "-dev" $workerChannel }}
+      {{- $baseVersion := index $.Values.imageVersions $baseChannel }}
+      {{- if not $baseVersion }}
+      {{- fail (printf "worker %s: channel '%s' and base channel '%s' not found in .Values.imageVersions" $workerkey $workerChannel $baseChannel) }}
+      {{- end }}
+      {{- $imageVersion = replace "usi" "usidev" $baseVersion }}
```
anokfireball, line 166: Not sure if we want to have this done automagically... sounded like a nice way to avoid a long list of almost identical channel/version mappings

PR #11381, opened 2026-04-14
[manila] Bump dependencies to switch kubernetes-entrypoint

fwiesel:
Also remove kubernetes-entrypoint access inside the container, so we can drop it from the image.


fwiesel:
I think, [this](https://github.com/sapcc/helm-charts/pull/11381/changes#diff-ee75ca5b1706c805cdd2f3fe8eeb9cd09ba7ddb5a7fa04538697fffa1a7b631dL86) should work, but I haven't tested it. So, please keep that in mind when reviewing.

PR #11384, opened 2026-04-14
[Nova] Remove deprecated image values

seb-kro:
~`helm lint` requires comments to start with a space so we add a missing space.~

Removes Neutron and Open vSwitch image values since they are no longer used and additionally the related comment is causing CI pipeline failures for `helm lint`.

### DIFF
--- a/openstack/nova/values.yaml
+++ b/openstack/nova/values.yaml
@@ -161,13 +161,6 @@ loci:
   nova: false
   # neutron images use loci by default
 
-#TODO we need to move to global or find another way to share image versions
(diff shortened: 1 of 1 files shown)


fwiesel (dismissed):

joker-at-work (changes requested):
Can we instead just remove this comment and the imageNameNeutron and imageNameOpenvswitch? Both are unused.

joker-at-work (approved):

fwiesel (approved):

PR #11398, opened 2026-04-15
[manila-nanny] Switch to sapcc/kubernetes-entrypoint fork & initConta…

fwiesel:
…iner

Standardize on our own fork, so we can fix issues more quickly Use the snippet for initcontainers so we do not need to maintain the kubernetes-entrypoint inside loci images.

### DIFF
--- a/openstack/manila-nanny/Chart.lock
+++ b/openstack/manila-nanny/Chart.lock
@@ -1,12 +1,12 @@
 dependencies:
 - name: utils
   repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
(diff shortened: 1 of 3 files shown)


Carthaca (approved):
Thanks!

for the next reviewer that wonders what "shellCommand" did:
```tpl
{{/* Start shell command or sleep */}}
{{- define "shellCommand" -}}
{{ if not .Values.debug }}/bin/bash /scripts/{{ .script }}{{ else }}sleep inf{{ end }}
{{- end -}}
```

kpawar-sap (approved):

PR #11399, opened 2026-04-15
[octavia] Switch to sapcc/kubernetes-entrypoint fork & initContainer

fwiesel:
Standardize on our own fork, so we can fix issues more quickly. Use the snippet for initContainer so we do not need the old kubernetes-entrypoint in the loci images.


BenjaminLudwigSAP (approved):
LGTM

PR #11402, opened 2026-04-15
[keystone] Switch to sapcc/kubernetes-entrypoint fork & initContainer

fwiesel:
Standardize on our own fork, so we can fix issues more quickly. Use the sapcc/kubernetes-entrypoint image from ghcrIoMirrorAlternateRegion so we do not need the old kubernetes-entrypoint in the loci images.

### DIFF
--- a/openstack/keystone/templates/_helpers.tpl
+++ b/openstack/keystone/templates/_helpers.tpl
@@ -63,6 +63,41 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
   {{- end }}
 {{- end }}
 
+{{- define "keystone.job_dependencies" }}
+  {{- if .Release.IsInstall }}
+    {{- printf "%s,%s" (tuple . "job-migration" | include "job_name") (tuple . "job-bootstrap" | include "job_name") }}
+  {{- else if .Values.run_db_migration }}
+    {{- tuple . "job-migration" | include "job_name" }}
+  {{- end }}
+{{- end }}
+
+{{/*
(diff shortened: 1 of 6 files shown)


openstack/keystone/templates/_helpers.tpl
```diff
@@ -63,6 +63,41 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
...
+{{/*
+kubernetes-entrypoint init container using the sapcc fork from ghcrIoMirrorAlternateRegion.
+Usage: tuple . (dict "SERVICE" "svc1,svc2" "JOBS" "job1") | include "keystone.kubernetes_entrypoint_init_container" | indent N
+Empty values are skipped so callers may pass "" for optional dependencies.
+*/}}
+{{- define "keystone.kubernetes_entrypoint_init_container" }}
+  {{- $envAll := index . 0 }}
+  {{- $params := index . 1 }}
+- name: kubernetes-entrypoint
+  image: {{ $envAll.Values.global.ghcrIoMirrorAlternateRegion }}/sapcc/kubernetes-entrypoint:{{ $envAll.Values.imageVersionKubernetesEntrypoint }}
```
bbobrov, line 83: would it be possible to `require` these variables since they are not defined in this helm chart?

fwiesel: Of course, I've added a `required` for both values.

bbobrov (approved):

tz3 (approved):

PR #11405, opened 2026-04-15
add credentialName

mikemichelsap:


ronchi-oss:
just checking whether you really want this or if you think this is required... it shouldn't be required by any system (id and secret are enough for authentication purposes).

mikemichelsap:
> just checking whether you really want this or if you think this is required... it shouldn't be required by any system (id and secret are enough for authentication purposes).

when i configured backups the deployed etcd (backup container) failed with "openstack swift credentials are not passed correctly" which lead me to https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapstore/swift_snapstore.go#L634 which from my understanding fails when ApplicationCredentialName is not present

ronchi-oss (approved):

ronchi-oss:
> > just checking whether you really want this or if you think this is required... it shouldn't be required by any system (id and secret are enough for authentication purposes).
> 
> when i configured backups the deployed etcd (backup container) failed with "openstack swift credentials are not passed correctly" which lead me to https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapstore/swift_snapstore.go#L634 which from my understanding fails when ApplicationCredentialName is not present

Ok, then you're right. It's a requirement not of OpenStack/Keystone but this etcd backup utility.

PR #11409, opened 2026-04-16
[ceph-seed] add ccadmin role assignments

s10:
* Add `ccadmin` domain role assignments
* Add objectstore roles
* Add `objectstore_admin` role for `db_backup@Default` user

These roles previously were seeded by Swift chart, see https://github.com/sapcc/helm-charts/blob/master/openstack/swift/templates/seeds.yaml#L56

### DIFF
--- a/openstack/ceph-seed/templates/seed.yaml
+++ b/openstack/ceph-seed/templates/seed.yaml
@@ -7,6 +7,7 @@ spec:
   requires:
   - monsoon3/domain-default-seed
   - monsoon3/domain-ccadmin-seed
+  - monsoon3/domain-monsoon3-seed
 
   roles:
   - name: cloud_objectstore_admin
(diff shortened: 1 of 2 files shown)


openstack/ceph-seed/templates/seed.yaml
```diff
@@ -38,3 +38,18 @@ spec:
...
+        - group: CCADMIN_CLOUD_ADMINS
+          role: cloud_objectstore_admin
+      - name: master
+        role_assignments:
+        - group: CCADMIN_CLOUD_ADMINS
+          role: objectstore_admin
+        - group: CCADMIN_DOMAIN_ADMINS
+          role: objectstore_admin
+        - user: db_backup@Default
+          role: objectstore_admin
```
majewsky, line 65: This will break in regions without Swift because no seed defines those roles. They should be defined as part of this seed like in:

https://github.com/sapcc/helm-charts/blob/a00cfab12feb296921469fd07ca169a8d83373a8/openstack/swift/templates/seeds.yaml#L19-L23

majewsky (changes requested):

openstack/ceph-seed/templates/seed.yaml
s10: I've added these roles and `spec.requires`

kayrus (approved):

majewsky (approved):

sumitarora2786 (approved):

PR #11425, opened 2026-04-17
[neutron] rbac service account

sebageek:
Needed for accessing mariadb and rabbitmq in eu-de-3 (and maybe also in other regions soonish) by kubernetes entrypoint used by the migration job, neutron-server and neutron-rpc-server.

---

Most of this is copied over from keystone. I am not sure if these are the right permissions needed.


fwiesel (approved):

toanju (approved):

s10:
Minimal required permissions for neutron kubernetes-entrypoint would be closer to something like this:
```yaml
rules:
- apiGroups:
  - "*"
  resources:
  - services
  - endpoints
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - batch
  resources:
  - jobs
  verbs:
  - get
  - list
  - watch
```

sebageek:
Closed in favor of #11466 which also contains `endpointslices` + makes it look like in the other services.

PR #11435, opened 2026-04-21
prometheus datapath tests absent alerts

ashifnihal:
- **[cc3test-prometehus-datapath-absent-alert] created an alert for absent cc3test datapath prometheus metrics**
- **[cc3test-prometheus-absent-alert] moved the alert to the global thanos and updated the expression**
- **fixed syntax in alert**
- **fixed syntax in alert 1**
- **bumped chart version**
- **updated service name and removed context**


viennaa (approved):
Enabling critical paging for this meta alert. Thanks @ashifnihalb

PR #11437, opened 2026-04-21
[VmWare-Alerts] Add no_alert_on_absence: true to iaas alerts

SqiSch:


prometheus-rules/prometheus-vmware-rules/templates/prometheus-alerts.yaml
```diff
@@ -38,6 +38,7 @@ spec:
         {{- $_ := set $copiedAlert.labels "support_group" "iaas-support" }}
         {{- $_ := set $copiedAlert.labels "bedrock" "true" }}
         {{- $_ := unset $copiedAlert.labels "tier" }}
+        {{- $_ := set $copiedAlert.labels "no_alert_on_absence" "true" }}
```
viennaa, line 41: Not needed, this is a deepCopy (L34). Since you add it to `iaas.alerts` it copies this label already. This would overwrite and I'd drop it for _simplicity_.

viennaa (commented):
Thanks, one thing to talk about inline.

prometheus-rules/prometheus-vmware-rules/templates/prometheus-alerts.yaml
SqiSch: You misunderstand the Template here. 

We copy some of the existent VmWare alerts which are important for us and overriding some details to match with our dedicated clusters (to prevent to duplication  of vmware alerts). 
The list of copied alerts is defined in the secrets

F.e.: 
HostHasLostConnectivityToPhysicalNetwork
is copied to: 
IaasHostHasLostConnectivityToPhysicalNetwork

viennaa: If the source alert has the `no_alert_on_absence`, it would be carried over here.

Do you really want to point blank disable it for all bedrock alerts? Looking over the list I can see 5 / 21 alerts have no label set, they would get it with this expression.

Generally setting it on the source would make it easier to understand later. Or is it a feature you never want?

SqiSch: i do not want to edit the source alerts.

I cannot expect, that on every vrops exporter the added metric vrops_cluster_summary_custom_tag_openstack_nova_traits_external_customer_exclusive
is present except on eu-de-1. So i want to add the label no_alert_on_absence on our copied alerts to prevent these absent alerts.

viennaa (approved):

PR #11449, opened 2026-04-21
[cinder]Use isImageTransportTemplating variable

hemna:
This variable is set by the shared pipeline for the image version transport and makes it possible to include templates into the pipeline, that would not be rendered for normal "helm template" runs because of conditionals checking the APIVersions.


joker-at-work (approved):
commit-message is missing to explain why we add the exclusion here.

Scsabiii (approved):

PR #11463, opened 2026-04-22
[netapp] extend snapmirror metrics to all setups

chuan137:
Previously snapmirror metrics were only available for EC2/HXM backup
targets (localbkp/remotebkp) and only in regions with federation enabled.

- Add :local recording rules for all snapmirror metrics (labels, lag_time,
  last_transfer_*, snapshot_count) so regions without federation also get
  full coverage — filtered to known Manila shares, no backup-only constraint
- Rename netapp_snapmirror_labels:enhanced → netapp_snapmirror_detailed_labels:enhanced
  to distinguish rich destination-filer data from endpoint-only labels
- Drop intermediate :enhanced recording rules for numeric metrics and
  inline label_replace directly in federation queries
- Remove stale go_build_info stub and region exclusion list from maia.yaml


Carthaca (approved):

kpawar89 (approved):

kpawar-sap (approved):

kpawar-sap:
just approved again from SAP profile instead of upstream github profile.

PR #11473, opened 2026-04-22
[nova] Add kubernetes-entrypoint RBAC

s10:


openstack/nova/values.yaml
```diff
@@ -25,6 +25,9 @@ global:
 
 osprofiler: {}
 
+rbac:
+  enabled: true
```
joker-at-work, line 29: Would there be a reason to disable it?

joker-at-work (changes requested):
Can you indent the ifs properly please? Otherwise my yaml will not get folded properly in the editor.

openstack/nova/values.yaml
s10: Not really. Here I just used the same option/pattern used previously in designate and keystone chart.

joker-at-work (approved):

Scsabiii (approved):

PR #11474, opened 2026-04-22
[placement] Add kubernetes-entrypoint RBAC

s10:


joker-at-work (changes requested):
Can you please indent the `if` and `end` like the surroundings? Otherwise folding doesn't work properly for my editor and the indentation also doesn't fit.

joker-at-work (approved):

Scsabiii (approved):

PR #11477, opened 2026-04-23
Revert "keystone: run cron daily at 03:00 UTC and enable dry-run mode"

bbobrov:
Reverts sapcc/helm-charts#11259. This is not needed any more.


bbobrov:
The failing validation is due to new rules that were added separately. Our change by itself does not introduce any of the errors mentioned in the report. I propose to merge this like it is, and fix the errors separately.

drochow (approved):

tz3 (approved):

PR #11482, opened 2026-04-23
[Nova] Prepare super-conductor split

joker-at-work:
These changes are basically cleanups for consistency to better allow thinking which services need which resources. They're also preparing for eu-de-3 having 3 properly named cells.

Please see the individual commits' message for more details.

### DIFF
--- a/openstack/nova/templates/_helpers.tpl
+++ b/openstack/nova/templates/_helpers.tpl
@@ -290,8 +290,7 @@ Params:
 
 {{- define "nova.helpers.cell_rabbitmq_service" }}
   {{- $envAll := index . 0 }}
-  {{- $rmqChartAlias := include "nova.helpers.cell_rabbitmq_chart_alias" . }}
-  {{- $rmqValues := get $envAll.Values $rmqChartAlias }}
+  {{- $rmqValues := include "nova.helpers.cell_rabbitmq_chart_alias" . | get $envAll.Values }}
   {{- $rmqName := default "rabbitmq" $rmqValues.nameOverride }}
   {{- printf "%s-%s" $envAll.Release.Name $rmqName | trunc 63 | replace "_" "-" | trimSuffix "-" }}
 {{- end }}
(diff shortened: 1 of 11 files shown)


grandchild (approved):

dhoeller (approved):
LGTM

openstack/nova/templates/_helpers.tpl
```diff
@@ -315,13 +314,9 @@ Params:
...
-  {{- $rmqHostInfix := "" }}
-  {{- if not (has $cellId (list "cell0" "cell1")) }}
-    {{- $rmqHostInfix = printf "%s-" (include "nova.helpers.cell_name" .) }}
-  {{- end }}
-  {{- $rmqHost := printf "%s-%srabbitmq" $envAll.Release.Name $rmqHostInfix}}
-  {{- $rmqChartAlias := include "nova.helpers.cell_rabbitmq_chart_alias" . }}
-  {{- $rmqValues := get $envAll.Values $rmqChartAlias }}
+  {{- $rmqValues := include "nova.helpers.cell_rabbitmq_chart_alias" . | get $envAll.Values }}
+  {{- $rmqName := default "rabbitmq" $rmqValues.nameOverride }}
+  {{- $rmqHost := printf "%s-%s" $envAll.Release.Name $rmqName }}
```
seb-kro, line 319: I believe this is missing the cell name now if `nameOverride` is not specified.

joker-at-work: If `nameOverride` is not specified, then the rabbitmq itself will come up as `nova-rabbitmq` afaik. So I think this is correct, don't you think? This would totally blow up, as there can't be two rabbitmq deployments with the same name.

seb-kro: Yes, should work fine. Just figured that we'd want to have a cell name component in there. But not too crucial and will probably need to be readjusted anyways once we create separate cell charts.

seb-kro (approved):

openstack/nova/templates/_helpers.tpl
joker-at-work: If we want the cell name component in there, then we have to put it into the nameOverride. The reason is, that this here doesn't matter at all for the `common/rabbitmq` helm-chart - it only cares about its values. Hence, `nameOverride` is the only way to get some cell component into the rabbitmq chart and thus into the hostname for this cell's rabbitmq.

PR #11483, opened 2026-04-23
Add timestamp alias to audit-ds

carmeroa:
Adds the alias `timestamp` to the index field `@timestamp`, which is the field name required by the Security Analytics plugin.


carmeroa:
Wrong branch for this PR, correct branch is https://github.com/sapcc/helm-charts/tree/opensearch-logs.
Commited there: https://github.com/sapcc/helm-charts/commit/d5473285e881e179e819b4279278241f50d2d565

PR #11484, opened 2026-04-23
metal-operator-remote: fixes for webhook enablement

videlov:
* Fix namespace used by webhook-injector
* Fix port collision between sidecar and controller-manager
* Add missing egress network policy for kube-apiserver-> metal-operator-webhook server
* fix remote-kubeconfig configmap
* fix rbac to allow webhook modification in remote cluster
* use securityContext to have both containers use the same uid which fixes permission problems when webhook server tries to read certificates
* deploy web hook-injector as native sidecar


databus23:
The use of downward api was not needed, removing the `--namespace=kube-system` flag already fixed the problem.

I also added some more changes to fix the deployment

databus23 (dismissed):

databus23 (approved):

PR #11487, opened 2026-04-24
[Nova] Move defaults into helm-chart

joker-at-work:
There are a lot of regional values set in all or most regions. We're trying to reduce that looking especially at creating new regions with. This needs an accompanying values change when getting rolled out.


seb-kro (dismissed):

bashar-alkhateeb (dismissed):

joker-at-work:
Rebased onto latest master.

bashar-alkhateeb (approved):

seb-kro (approved):

PR #11490, opened 2026-04-24
[nova] Add endpointslices to kubernetes-entrypoint RBAC

s10:
sapcc/kubernetes-entrypoint watches endpointslices instead of endpoints, see https://github.com/sapcc/kubernetes-entrypoint/blob/main/dependencies/service/service.go#L54

### DIFF
--- a/openstack/nova/templates/rbac.yaml
+++ b/openstack/nova/templates/rbac.yaml
@@ -22,6 +22,13 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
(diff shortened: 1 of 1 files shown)


openstack/nova/templates/rbac.yaml
s10, line 8: This apiGroup could be removed after service switched to sapcc/kubernetes-entrypoint

joker-at-work: I think we're rolling that out for Nova next week.

joker-at-work (approved):

Scsabiii (approved):

PR #11491, opened 2026-04-24
[Nova] Set a timeout for Placement requests

joker-at-work:
We'll set a default timeout of 10min for requests to Placement. Previously, we didn't have a timeout and this made nova-conductor get stuck in a lock, not doing any new migrations, while waiting for a call to Keystone that's happening from inside the `openstacksdk`'s `Proxy` to get the project as a cache key. The Placement client in Nova uses `openstacksdk`.

### DIFF
--- a/openstack/nova/values.yaml
+++ b/openstack/nova/values.yaml
@@ -200,6 +200,7 @@ cross_az_attach: 'False'
 cinder_http_retries: 10
 neutron_http_retries: 10
 neutron_timeout: 600
+placement_timeout: 600
 
 api:
   config_file:
--- a/openstack/nova/templates/etc/_nova.conf.tpl
+++ b/openstack/nova/templates/etc/_nova.conf.tpl
@@ -116,6 +116,7 @@ project_name = "{{.Values.global.keystone_service_project | default "service" }}
 project_domain_name = "{{.Values.global.keystone_service_domain | default "Default" }}"
 valid_interfaces = internal
 region_name = {{.Values.global.region}}
+timeout = {{.Values.placement_timeout}}
 
 {{- include "ini_sections.cache" . }}
 


bbobrov:
> this made nova-conductor get stuck in a lock, not doing any new migrations, while waiting for a call to Keystone that's happening from inside the openstacksdk's Proxy to get the project as a cache key.

What's the mechanism of that? My expectation would be that the connection gets kicked out via the normal TCP timeout.

Why do you think that it gets stuck in a lock on the keystone call? At what stage of the call does it get stuck?

openstack/nova/values.yaml
```diff
@@ -200,6 +200,7 @@ cross_az_attach: 'False'
 cinder_http_retries: 10
 neutron_http_retries: 10
 neutron_timeout: 600
+placement_timeout: 600
```
bbobrov, line 203: decrease to 60 seconds maybe?

bbobrov: (i would even say that anything above 3 seconds is unacceptable and should be terminated)

joker-at-work: This is supposed to prevent the conductor getting stuck in locks. I don't want to introduce any breakage through this just because a request takes a little longer. Having a 10min timeout is on the safer side and thus looks more reasonable for me.

joker-at-work:
https://docs.astral.sh/ruff/rules/request-without-timeout/

According to that, requests might hang indefinitely. I'm not sure what makes the tcp-connection not close. Might be (or not) linkerd-related. I have not insights on that.

> Why do you think that it gets stuck in a lock on the keystone call? At what stage of the call does it get stuck?

Internal chat contains tracebacks that I interpret as suggesting exactly that. The traceback shows [openstacksdk](https://github.com/openstack/openstacksdk/blob/65dcdb2fac7b21b1e9e5f6432456e7bc3380d240/openstack/proxy.py#L184) using the keystoneauth Session doing [a request](https://opendev.org/openstack/keystoneauth/src/branch/master/keystoneauth1/session.py#L1052) and the Session has no `self.timeout` set according to a look into the eventlet backdoor.

All other greenthreads are hanging on a lock in [keystoneauth](https://opendev.org/openstack/keystoneauth/src/branch/master/keystoneauth1/identity/base.py#L137-L139). Placement client in Nova is a singleton i.e. just one per process.

seb-kro (approved):

bbobrov (approved):

PR #11503, opened 2026-04-27
#1102 Calico network alerts: bonds + node interfaces

natatchikhinashvili1:


natatchikhinashvili1:
Open questions

Routing scope: new alerts use `support_group: network-api`, existing ones still go to `containers`. Should the old ones re-route too? One-line change if yes.

BGP `Node_*` peers: found 4 sessions Established but exporting 0 routes, all `Node_*`. Mesh peers were fine, so I scoped the alert to Mesh only. Are `Node_*` peers expected zero-exporters (route reflectors?) or actually broken?

Policy enforcement: Felix only exposes inventory metrics here, nothing for denies or eval failures. Would need extra instrumentation (flow logs / BPF) to alert on this. Separate follow-up?

Still TODO
- Runbooks (one per new alert)
- Simulating failures in QA
- Grafana dashboards
- Inhibition rules to avoid alert storms

goerangudat:
I like to have the alert in network group as well as in containers (when our nodes -> runtime and mgmt-clusters)
storage cluster (st1-*) should be routed to storage team

goerangudat (approved):
lgtm

PR #11505, opened 2026-04-27
[keystone][logging]: add use_json toggle for structured JSON logging

Machariajane:
## Summary
Add a \`use_json\` values flag (default: false) that switches all keystone log sources to structured JSON output when enabled.

## Changes
- oslo.log: \`use_json = true\` in \`[DEFAULT]\` switches ContextFormatter → JSONFormatter at runtime
- Apache access logs: JSON LogFormat (json_combined/json_proxy)
- Apache error logs: JSON ErrorLogFormat with \`%{escape}M\` (safe for stack traces, requires Apache >= 2.4.46)

## Motivation
Better consumption to the  OTel Collector instead of grok pattern matching 

## Rollback
Set \`use_json: false\` in the region values file.

### DIFF
--- a/openstack/keystone/templates/etc/_keystone.conf.tpl
+++ b/openstack/keystone/templates/etc/_keystone.conf.tpl
@@ -1,6 +1,7 @@
 [DEFAULT]
 debug = {{.Values.debug}}
 insecure_debug = {{.Values.insecure_debug}}
+use_json = {{.Values.use_json | default false}}
 verbose = true
 
 {{- with .Values.max_db_limit }}
--- a/openstack/keystone/templates/etc/_keystone.conf.tpl
+++ b/openstack/keystone/templates/etc/_keystone.conf.tpl
@@ -1,7 +1,6 @@
 [DEFAULT]
 debug = {{.Values.debug}}
 insecure_debug = {{.Values.insecure_debug}}
-use_json = {{.Values.use_json | default false}}
 verbose = true
 
 {{- with .Values.max_db_limit }}
--- a/openstack/keystone/templates/etc/_wsgi-keystone.conf.tpl
+++ b/openstack/keystone/templates/etc/_wsgi-keystone.conf.tpl
@@ -16,7 +16,7 @@ Listen 0.0.0.0:5000
 
 {{- if .Values.use_json }}
 ErrorLog /dev/stdout
-ErrorLogFormat "{\"timestamp\":\"%{cu}t\",\"levelname\":\"ERROR\",\"name\":\"apache.error\",\"pid\":%P,\"message\":\"%{escape}M\"}"
+ErrorLogFormat "%M"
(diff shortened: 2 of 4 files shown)


openstack/keystone/templates/etc/_keystone.conf.tpl
```diff
@@ -1,6 +1,7 @@
 [DEFAULT]
 debug = {{.Values.debug}}
 insecure_debug = {{.Values.insecure_debug}}
+use_json = {{.Values.use_json | default false}}
```
bbobrov: I am looking at the docstring for `use_json` option on https://docs.openstack.org/keystone/latest/_static/keystone.conf.sample. There it says:
```
# Use JSON formatting for logging. This option is ignored if log_config_append
# is set. (boolean value)
#use_json = false
```
log_config_append is set below on line 13.
Does it work? Was it tested? Why does it work?

openstack/keystone/templates/etc/_wsgi-keystone.conf.tpl
```diff
@@ -14,14 +14,23 @@
 
 Listen 0.0.0.0:5000
 
+{{- if .Values.use_json }}
+ErrorLog /dev/stdout
+ErrorLogFormat "{\"timestamp\":\"%{cu}t\",\"levelname\":\"ERROR\",\"name\":\"apache.error\",\"pid\":%P,\"message\":\"%{escape}M\"}"
+LogFormat "{\"timestamp\":\"%{%Y-%m-%dT%H:%M:%S}t.%{msec_frac}t\",\"pid\":%{pid}P,\"levelname\":\"INFO\",\"name\":\"apache.access\",\"request_id\":\"%{X-Openstack-Request-ID}i\",\"client_ip\":\"%a\",\"method\":\"%m\",\"uri\":\"%U%q\",\"protocol\":\"%H\",\"status\":%>s,\"bytes_sent\":%B,\"duration_ms\":%{ms}T,\"referer\":\"%{Referer}i\",\"user_agent\":\"%{User-Agent}i\"}" json_combined
```
bbobrov, line 20: `referer` and `user_agent` here are user-controlled strings.
What will happen if a user sends a user agent like this:
```
hello", "client_ip": "fake ip", "levelname": "CRITICAL", "
```
Will it inject the json into our logs? Will it overwrite the keys supplied by us? @drochow @giobbe do you happen to know?

openstack/keystone/templates/etc/_keystone.conf.tpl
Machariajane: Good catch , I tested by importing the `JSONFormatter` directly and called it manually but it bypassed the config loading path.
I will move it to  ` _logging.conf.tpl`

openstack/keystone/templates/etc/_wsgi-keystone.conf.tpl
Machariajane: `{
  "referer": "-",
  
  "user_agent": "hello\", \"client_ip\": \"fake ip\", \"levelname\": \"CRITICAL\", \""
}
`

The. [mod_log_config.html](https://httpd.apache.org/docs/current/mod/mod_log_config.html) specifically address the concern of injection in user request headers 

> Format Notes
For security reasons, starting with version 2.0.46, non-printable and other special characters in format strings that may contain arbitrary string data are escaped using \xhh sequences, where hh stands for the hexadecimal representation of the raw byte. Exceptions from this rule are " and \, which are escaped by prepending a backslash, and all whitespace characters, which are written in their C-style notation (\n, \t, etc). In versions prior to 2.0.46, no escaping was performed on these strings so you had to be quite careful when dealing with raw log files.
%i	Request header	Yes

Machariajane: Test

`curl -H 'User-Agent: evil", "levelname": "CRITICAL", "client_ip": "6.6.6.6' http://localhost:5000/v3`

Output


`"evil\", \"levelname\": \"CRITICAL\", \"client_ip\": \"6.6.6.6"`

bbobrov (approved):

rajivmucheli (approved):

PR #11506, opened 2026-04-27
opensearch-logs: add custom vpa

ztomaszewska:
As opensearch-logs chart doesn't use the VPA mechanizm and set the requests by itself, we'll override the automatically created ones with custom. The custom ones will have the `Off` policy. To switch to automatically created ones set the `vpa.enabled` to `false`.


system/opensearch-logs/templates/vpa-opensearch-nodes.yaml
```diff
@@ -0,0 +1,80 @@
+{{- if .Values.vpa.enabled }}
+---
+apiVersion: autoscaling.k8s.io/v1
+kind: VerticalPodAutoscaler
+metadata:
+  name: off-opensearch-logs-client-statefulset
```
joluc: I would not add the `off` status to the name in case we will use it at some point

system/opensearch-logs/templates/vpa-opensearch-nodes.yaml
```diff
@@ -0,0 +1,80 @@
...
+    ccloud/support-group: observability
+spec:
+  targetRef:
+    apiVersion: apps/v1
+    kind: StatefulSet
+    name: opensearch-logs-snapshots
+  updatePolicy:
+    updateMode: "Off"
+{{- end }}
+{{- end }}
```
joluc: New line ```suggestion
{{- end }}

```

joluc (changes requested):

system/opensearch-logs/templates/vpa-opensearch-nodes.yaml
ztomaszewska: I've changed to `off-` prefix to `-custom` suffix. opensearch-logs-client-statefulset is a name that vpa_butler uses, helm upgrade fails if we use the same name.

ztomaszewska:
Already merged [here](https://github.com/sapcc/helm-charts/pull/11632/changes#diff-fc7c9e6e890d1852f0bbf4233db767992e983e438fcd498a2d485cb8e03cc60f)

PR #11513, opened 2026-04-28
limes: v2 cluster validation policy

VoigtS:


openstack/limes/files/policy.yaml
```diff
@@ -2,6 +2,7 @@ project_scope: project_domain_id:%(domain_id)s and project_id:%(project_id)s
 domain_scope: domain_id:%(domain_id)s
 
 cluster_editor: role:cloud_resource_admin
+cluster_resource_validator: role:resource_viewer and domain_id:Default and project_id:service
```
wagnerd3: I thought we wanted this to be a role assigned to the validation user, not a rule? 
I think we would want to add it [here](https://github.com/sapcc/helm-charts/blob/master/openstack/limes/templates/seed.yaml#L45) and somewhere [here](https://github.com/sapcc/helm-charts/blob/master/openstack/limes/templates/prometheus-alerts.yaml#L29) ?

wagnerd3 (dismissed):

openstack/limes/files/policy.yaml
VoigtS: You are right, this approach is better. I adjusted it accordingly.

openstack/limes/files/policy.yaml
```diff
@@ -2,6 +2,7 @@ project_scope: project_domain_id:%(domain_id)s and project_id:%(project_id)s
 domain_scope: domain_id:%(domain_id)s
 
 cluster_editor: role:cloud_resource_admin
+cluster_resource_validator: role:cloud_resource_validator
```
majewsky: Creating new cloud-admin roles requires changing the Keystone policy to block those roles from being assigned by regular project admins. It's ultimately easier (albeit a bit ugly) to match on the user and token scope directly:

```suggestion
cluster_resource_validator: project_name:service and project_domain_name:Default and user_name:limes-validation and user_domain_name:Default
```

majewsky (changes requested):

majewsky (approved):

PR #11529, opened 2026-04-29
Update metal-settings chart

### LINKED ISSUE #11521: feat(metal-settings): Add region-based filtering for BIOS/BMC settings and versions
### Problem

The current `metal-settings` helm chart deploys **all** BIOS settings, BIOS versions, and BMC versions to every cluster where it is installed. This "deploy everything everywhere" approach creates challenges:

1. **No staged rollout capability** - Cannot test new firmware versions or BIOS settings in a specific region before deploying globally
2. **No regional differentiation** - All regions receive identical settings, even when we want to trial configurations
3. **Risk of unintended changes** - Adding a new file automatically deploys it everywhere

### Current Behavior

The templates use `$.Files.Glob` to iterate over all YAML files in the data directories and render a Kubernetes resource for each:

    # bios_settings.yaml
    {{- range $path, $_ := $.Files.Glob "bios_settings/**" }}
    # Creates BIOSSettingsSet for every file found
    {{- end }}

Similar patterns exist for `bios_versions.yaml` and `bmc_versions.yaml`. The `values.yaml` is empty, providing no mechanism to control deployment.

### Proposed Solution

Add include/exclude list filtering in values:

    # values.yaml
    biosSettings:
      enabled: true
      # Include specific vendor/model/cluster-type combinations
      include:
        - vendor: dell
          model: poweredge-r660
          clusterType: k8s-controlplane
        - vendor: dell
          model: poweredge-r760
      # Or exclude specific combinations
      exclude: []
    
    biosVersions:
      enabled: true
      include: []  # Empty means all
      exclude:
        - vendor: dell
          model: poweredge-r640  # Exclude legacy model
    
    bmcVersions:
      enabled: true
      include: []
      exclude: []

When `include` is empty/not set, all resources are deployed (current behavior). When `include` has entries, only matching resources are deployed. `exclude` always takes precedence over `include`.

### Acceptance Criteria

- [ ] New BIOS/BMC settings can be tested in specific regions before global rollout
- [ ] Existing deployments are not affected (backward compatible defaults - empty include = deploy all)
- [ ] Clear documentation for operators on how to configure regional deployments
- [ ] Values schema validation (if using JSON schema)

### Additional Context

**Directory Structure Reference:**

    bios_settings/<vendor>/<model>/<cluster-type>/<version>.yaml
    bios_versions/<vendor>/<model>/<version>.yaml
    bmc_versions/<vendor>/<model>/<version>.yaml

**Current CRDs:**
- `BIOSSettingsSet` with `serverSelector.matchLabels` (type + cluster-type)
- `BIOSVersionSet` with `serverSelector.matchLabels` (type)
- `BMCVersionSet` with `bmcSelector.matchLabels` (type)

asergeant01:
fixes #11521

### DIFF
--- a/system/metal-settings/ci/filter-vendor-values.yaml
+++ b/system/metal-settings/ci/filter-vendor-values.yaml
@@ -1,0 +1,46 @@
+# Test vendor filtering - deploy only Dell resources
+# This tests the include filter functionality
+biosSettings:
+  enabled: true
+  filters:
+    vendors:
+      included:
+        - dell
+      excluded: []
+    models:
+      included: []
+      excluded: []
+    clusterTypes:
+      included: []
+      excluded: []
+    versions:
+      included: []
+      excluded: []
+
+biosVersions:
+  enabled: true
+  filters:
+    vendors:
+      included:
+        - dell
+      excluded: []
+    models:
+      included: []
+      excluded: []
+    versions:
+      included: []
+      excluded: []
+
+bmcVersions:
+  enabled: true
+  filters:
+    vendors:
+      included:
+        - dell
+      excluded: []
+    models:
+      included: []
+      excluded: []
+    versions:
+      included: []
+      excluded: []
--- a/system/metal-settings/templates/bios_versions.yaml
+++ b/system/metal-settings/templates/bios_versions.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.biosVersions.enabled }}
 {{- range $path, $_  := $.Files.Glob "bios_versions/**" }}
 {{- $name := base (dir $path) }}
 {{- if eq (len (splitList "/" $path)) 4 -}}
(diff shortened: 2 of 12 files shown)


system/metal-settings/templates/bios_versions.yaml
nagadeesh-nagaraja, line 31: not related to this PR, but I think we should not include the `version` in the name here. when the version is changed, we will leave behind the old CRD behind and create a new ones....
same goes for the BMCversion

system/metal-settings/ci/filter-vendor-values.yaml
```diff
@@ -0,0 +1,46 @@
...
+    vendors:
+      included:
+        - dell
+      excluded: []
+    models:
+      included: []
+      excluded: []
+    clusterTypes:
+      included: []
+      excluded: []
```
nagadeesh-nagaraja, line 15: Again not related directly to this PR. But May be we need to take it up after this PR, to add the dev cluster-type ( for m-qa-de-100 cluster) in here so that our QA metal-settings work as expected. 

this applies to all the other CRD as well.

nagadeesh-nagaraja (approved):

PR #11543, opened 2026-04-30
adding suffixies region wise for openstack credentials

pankajlmsap:
- Credentials for managed seed are coming from this helm chart and is hardcoded as openstack-infra-creds-secret
- using placeholders to give the names as per secrets file here in this PR https://github.wdf.sap.corp/cc/kube-secrets/pull/3750/files
- We need to get this merge first and then later the other one to make it work


jknipper (changes requested):
Please also bump the chart version so it gets pushed to the registry.

jknipper (approved):

PR #11591, opened 2026-05-07
[designate] disable SSL for rabbitmq

vssldmtrv:
All existing deployments don't have it enabled
but the new rabbitmq helm chart has it enabled by default.

Let's be backwards compatible with the existing deployments.

At some point we might want to start using SSL.

### DIFF
--- a/openstack/designate/Chart.yaml
+++ b/openstack/designate/Chart.yaml
@@ -2,7 +2,7 @@
 apiVersion: v2
 description: A Helm chart for Kubernetes to deploy Openstack Designate  (DNSaaS)
 name: designate
-version: 0.5.15
+version: 0.5.16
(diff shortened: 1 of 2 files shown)


fwiesel (commented):
Enabling SSL is backwards compatible. You can still connect without SSL.

vssldmtrv:
I would rather keep SSL explicitly disabled and when the time comes to turn it on, we:
1. flip that toggle above explicitly (per region or globally)
2. make designate actually connect to the SSL port

Just letting the new defaults implicitly take their course will introduce new k8s objects (certs, secrets), expose port 5671 just for all that to sit there and do nothing.

proddi (approved):

galkindmitrii (approved):

PR #11593, opened 2026-05-07
[tenso] add variables for region and az

wagnerd3:
See code PR: https://github.com/sapcc/tenso/pull/303

### DIFF
--- a/system/tenso-auto/templates/configmap-tenso-auto.yaml
+++ b/system/tenso-auto/templates/configmap-tenso-auto.yaml
@@ -6,3 +6,5 @@ metadata:
 
 data:
   helm_deployment_cluster_regex: {{ .Values.tenso.helm_deployment_cluster_regex | required "missing value for .Values.tenso.helm_deployment_cluster_regex" | quote }}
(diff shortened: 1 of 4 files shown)


system/tenso-auto/templates/configmap-tenso-auto.yaml
```diff
@@ -6,3 +6,5 @@ metadata:
 
 data:
   helm_deployment_cluster_regex: {{ .Values.tenso.helm_deployment_cluster_regex | required "missing value for .Values.tenso.helm_deployment_cluster_regex" | quote }}
+  region_regex: {{ .Values.tenso.region_regex | required "missing value for .Values.tenso.region_regex" | quote }}
+  awx_workflow_az_regex: {{ .Values.tenso.awx_workflow_az_regex | required "missing value for .Values.tenso.awx_workflow_az_regex" | quote }}
```
majewsky, line 10: These should be documented in `values.yaml` as a reference for which keys are recognized, preferably with a default value of `null` (like we already have for other values).

majewsky (changes requested):

majewsky (approved):

PR #11596, opened 2026-05-07
[ironic] drop custom policy

sandzwerg:
The current custom policy doesn't work with the new RBAC model. Instead
of trying to migrate it we switch to the upstream default, which is
already included in the images, meaning we don't need to provide a policy
file. We also remove our own custom "cloud_baremetal_admin" role and
replace it with baremetal_admin in the seed. But we probably want to
drop that as well and just set the ccadmin/master project as owner of
the nodes to grant baremetal admin permission. I didn't test the seed
change yet.


sandzwerg:
As the rollout of a new region is supposed to follow soon let's try to get this out before, so instead of seeding and deleting we try to just remove the role. I would still let this soak in QA for a week at least

BerndKue:
Deleting the role might cause issues for customers to use the console of the server.
As far as I remember only baremetal admins can view it.

PR #11603, opened 2026-05-08
fix: update cronjob for updating mtls cert manage resource with templating

videlov:


databus23 (approved):

databus23 (changes requested):
Chart test fails with:

```
[ERROR] templates/cronjob-auditing-certificate-updater.yaml: unable to parse YAML: error converting YAML to JSON: yaml: line 32: could not find expected ':'
```

databus23 (approved):

PR #11634, opened 2026-05-11
keystone: add credential_migrate pre-upgrade hook

tz3:
Adds a pre-upgrade Kubernetes Job that runs `keystone-manage credential_migrate` automatically before every helm upgrade. This re-encrypts any credentials not yet using the current primary key, ensuring no credentials break when old keys are removed.

Idempotent — no-op if all credentials are already on the current primary key. Controlled by `run_credential_migrate` in values (default: `false`).

### DIFF
--- a/openstack/keystone/templates/job-credential-migrate.yaml
+++ b/openstack/keystone/templates/job-credential-migrate.yaml
@@ -1,0 +1,115 @@
+{{- if .Values.run_credential_migrate }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ tuple . "job-credential-migrate" | include "job_name" }}"
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
+    "helm.sh/hook-weight": "0"
+  labels:
+    app: {{ template "fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+    system: openstack
+    component: keystone
+    type: job
+    ccloud/support-group: identity
+    ccloud/service: keystone
+spec:
+  template:
+    metadata:
+      labels:
+        name: "{{ tuple . "job-credential-migrate" | include "job_name" }}"
+        system: openstack
+        component: keystone
+        type: job
+        alert-tier: os
+        alert-service: keystone
+      annotations:
+        chart-version: {{.Chart.Version}}
+        configmap-bin-hash: {{ include (print $.Template.BasePath "/configmap-bin.yaml") . | sha256sum }}
+        configmap-etc-hash: {{ include (print $.Template.BasePath "/configmap-etc.yaml") . | sha256sum }}
+{{- include "utils.linkerd.pod_and_service_annotation" . | indent 8 }}
+    spec:
+{{- if .Values.rbac.enabled }}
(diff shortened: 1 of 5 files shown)


bbobrov:
Please confirm that you have tested this end-to-end in a qa environment

tz3:
@bbobrov Tested in qa-de-1 and qa-de-2. Verified by comparing encrypted_blob values in the credential table before and after — all blobs re-encrypted with the new primary key. Idempotent on re-run. Reverted after testing.

bbobrov (changes requested):
I am blocking this change until 2026-05-29 to prevent accidental merge; this is a technical -1, it will be lifter after the mentioned date

openstack/keystone/values.yaml
```diff
@@ -28,6 +28,9 @@ insecure_debug: false
 # run a db_sync to migrate the db schema ?
 run_db_migration: true
 
+# run credential_migrate as a pre-upgrade hook to re-encrypt credentials with the current primary key
+run_credential_migrate: false
+
```
bbobrov: Why do we need this?

tz3: This flag is a one-time migration control. It defaults to false and gets set to true per-region during the Phase 2 upgrade window. Once all regions are migrated, the true overrides are removed and the hook stops running.

bbobrov: I understand what it does, but don't understand why we need it. What if the credential_migrate runs on every deployment?

tz3: Once the migration is done, all credentials are already on the current primary key. So credential_migrate would just decrypt and re-encrypt with the same key every time — no actual changes, just unnecessary DB load on every upgrade. That's why the flag exists, so we can turn it off once we're done.

bbobrov: > So credential_migrate would just decrypt and re-encrypt with the same key every time

This is not what is happening. Keystone checks that the credential is encrypted with a different key and only then performs re-encryption.

Remove the flag please, let the job run on every deployment.

tz3: Gound it here, https://opendev.org/openstack/keystone/src/commit/18e6fd7d6/keystone/cmd/cli.py#L944. it has already a condition so no need for the flag. 
I removed it and updated the two repos.

openstack/keystone/templates/job-credential-migrate.yaml
```diff
@@ -0,0 +1,117 @@
...
+      {{- end }}
+      containers:
+        - name: keystone-credential-migrate
+          image: {{ required ".Values.global.registryAlternateRegion is missing" .Values.global.registryAlternateRegion }}/{{ .Values.api.image }}:{{ .Values.api.imageTag }}
+          imagePullPolicy: {{ default "IfNotPresent" .Values.api.imagePullPolicy | quote }}
+          command:
+            - kubernetes-entrypoint
+          env:
+            - name: COMMAND
+              value: "bash /scripts/credential_migrate"
```
bbobrov, line 52: I am wondering about this. How long will it take to run this? What will timeout? Anything in kubernetes? In the CI?

tz3: I don't think the runtime will exceeds 1 min, it only re-encrypts rows where `key_hash != primary_key_hash`, pure scan otherwise. And i checked the helm's default hook timeout is 5m0s, no risk of hitting it.
Do you think i should state some timeout there, what would be appropriate in that case ?

bbobrov (approved):

BerndKue (approved):

PR #11635, opened 2026-05-11
Remove PII enrichment from Hermes Logstash pipeline

notque:
## Summary

- Strip user name (I-number) and project name enrichments from jdbc_static filter
- Retain domain-name lookup required by Octobus SIEM forwarding
- Stop dropping Default domain events (was suppressing admin audit trail)

## Motivation

**GDPR**: The jdbc_static filter resolved user UUIDs to SAP I-numbers via `id_mapping`/`local_user` Keystone tables, embedding PII into immutable audit records where it cannot be deleted on request.

**Audit immutability**: PCI DSS 10.3.2 and BSI C5 LOG-04 require stored audit records to reflect what the originating service emitted. Fields like `initiator.name`, `initiator.project_name`, and `project_domain_name` were injected post-emission.

## What changed

| Removed | Reason |
|---------|--------|
| `initiator.name` population | PII (SAP I-numbers) |
| `initiator.project_name` / `target.project_name` | Not in source CADF event |
| `initiator.project_domain_name` / `target.project_domain_name` | Not in source CADF event |
| `project_domain_mapping` loader + local table | No longer needed |
| `id_mapping`/`local_user` joins in SQL | Source of PII |
| Drop of Default domain events | Was suppressing privileged admin audit trail |
| Placeholder "unavailable" fields for project_id lookups | No longer needed |

| Retained | Reason |
|----------|--------|
| `jdbc_static` domain lookup (user_id → domain_id + domain_name) | Octobus SIEM forwarding depends on `[initiator][domain]` |
| `initiator.domain_id` and `initiator.domain` population | SIEM condition uses domain name matching |
| METIS credentials and `logstash.jdbc` values | Feed the retained lookup |
| Octobus forwarding block | Unchanged (condition simplified) |

## Octobus condition change

The SIEM forwarding condition simplified from:
```
(domain=='Default' AND name=='admin') OR domain=='ccadmin'
  OR target.project_domain_name=='ccadmin'
  OR initiator.project_domain_name=='ccadmin'
```
to:
```
domain=='Default' OR domain=='ccadmin'
```

This broadens Default domain forwarding slightly (previously filtered to admin user only) but removes the PII dependency. The `project_domain_name` conditions are dropped since those fields are no longer enriched.

## Follow-up

- Remove `jdbc_static` entirely once SIEM replacement no longer requires domain name resolution at ingestion time.
- Remove Hermes API filters that reference fields no longer captured (e.g. `initiator.name`, `project_name`, `project_domain_name`).

## Test plan

- [ ] Deploy to QA region, verify events index without errors
- [ ] Confirm indexed events no longer contain `initiator.name`, `initiator.project_name`, `target.project_name`, `project_domain_name` fields
- [ ] Verify Octobus SIEM still receives Default/ccadmin domain events
- [ ] Check Logstash pod starts without errors (JDBC still connects to Metis)
- [ ] Remove or disable Hermes API filters that refer to fields we no longer capture (`initiator.name`, `project_name`, `project_domain_name`)

### DIFF
--- a/openstack/hermes/templates/etc/_logstash.conf.tpl
+++ b/openstack/hermes/templates/etc/_logstash.conf.tpl
@@ -28,12 +28,6 @@ filter {
     drop { }
   }
 
-  # Drop liveliness check events that serve no value
-  # Everything in audit-default adds no value, internal communication
-  if [initiator][domain_id] == "default" {
-    drop { }
-  }
-
   # unwrap messagingv2 envelope
   if [oslo.message] {
     json {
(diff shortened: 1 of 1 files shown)


notque:
Connected to this, we need to remove the filters from hermes for those fields as they no longer will exist to be used in new events.

LeoZhangZXZ (approved):

notque:
This is kind of a mess, because we gain value in one direction, and lose it in routing. I am going to kill this, and try to rework it. It also has an order of operations issue.

PR #11660, opened 2026-05-13
[Neutron] Introduce Flag for Agent Liveness Probe

sven-rosenzweig:
Add a flag to enable agent liveness probes.

These probes should be disabled on a per-region basis, as observed in the past that the status checks against the Neutron DB put significant pressure on the database, especially in larger regions with hundreds of agents.

This becomes particularly problematic when many agents restart at the same time, causing a large number of simultaneous requests against the DB.

### DIFF
--- a/openstack/neutron/templates/asr_config_agents/_deployment-asr1k.yaml.tpl
+++ b/openstack/neutron/templates/asr_config_agents/_deployment-asr1k.yaml.tpl
@@ -45,12 +45,14 @@ spec:
           imagePullPolicy: IfNotPresent
           command:
             - /container.init/neutron-asr1k-start
+          {{- if $context.Values.global.agent_liveness_probe_enabled }}
           livenessProbe:
             exec:
               command: ["neutron-agent-liveness", "--agent-type", "ASR1K L3 Agent", "--config-file", "/etc/neutron/neutron.conf", "--config-file", "/etc/neutron/secrets/neutron-common-secrets.conf"]
             initialDelaySeconds: 30
             periodSeconds: 30
             timeoutSeconds: 10
+          {{- end }}
           env:
             - name: DEBUG_CONTAINER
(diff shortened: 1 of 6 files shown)


openstack/neutron/templates/asr_config_agents/_deployment-asr1k.yaml.tpl
```diff
@@ -45,12 +45,14 @@ spec:
           imagePullPolicy: IfNotPresent
           command:
             - /container.init/neutron-asr1k-start
+          {{- if $context.Values.global.agent_liveness_probe }}
```
toanju:
```suggestion
          {{- if $context.Values.global.agent_liveness_probe_enabled }}
```
as usually there are enabled switches for this kind of behaviour, I'd recommend to add at least that suffix.

openstack/neutron/templates/asr_config_agents/_deployment-asr1k.yaml.tpl
```diff
@@ -90,12 +92,14 @@ spec:
           imagePullPolicy: IfNotPresent
           command:
             - /container.init/neutron-asr1k-ml2-start
+          {{ if $context.Values.global.agent_liveness_probe }}
           livenessProbe:
             exec:
               command: ["neutron-agent-liveness", "--agent-type", "ASR1K ML2 Agent", "--config-file", "/etc/neutron/neutron.conf", "--config-file", "/etc/neutron/secrets/neutron-common-secrets.conf"]
             initialDelaySeconds: 30
```
mutax, line 99: as we have seen it takes some time for linkerd to initialize, do we want to increase the initial delay for all probes (not just in this template)  to 60 seconds or more -- this would also limit the number of failing probes sent out when all pods restart and delay them a bit hoping to get some spread

mutax (approved):

toanju (approved):

PR #11671, opened 2026-05-13
Add BMCSettings Resource

nagadeesh-nagaraja:
add default settings for the BMCSettings resources.


xkonni (commented):
what happens on clusters where an older metal-operator is running? are these used? silently ignored? might they cause issues?

nagadeesh-nagaraja:
> what happens on clusters where an older metal-operator is running? are these used? silently ignored? might they cause issues?

Hence, this has been specifically "disabled" by default. I do not see any changes to any pipeline clusters. in the future we can enable them.

xkonni (approved):
ship it!

PR #11673, opened 2026-05-13
[srs] change playbook links to new structure

wagnerd3:
We moved this out of the old structure in the `documentation-operation` so that we can hand this over properly to the operations team.


majewsky (dismissed):

viennaa (changes requested):
Bump the chart on cc3test please

majewsky (approved):

viennaa (approved):

PR #11674, opened 2026-05-13
prometheus-server: add namespace to alerts prometheus label

trouaux:


viennaa (approved):
Very good catch, thank you!

PR #11683, opened 2026-05-15
campfire: allow cluster_editors to send mails

VoigtS:
`cluster_editors` should be able to send mails via campfire.


majewsky (changes requested):
> `cluster_editors` should be able to send mails via campfire.

Should they though?

If this is about testing whether mail delivery works at all, I would like to have that limited to just specific mails, not just mail sending as a whole. Compare how Tenso allows cloud admins to send certain pre-defined "synthetic events" for testing integrations, but it does not allow them to send real changes.

VoigtS:
You are right, we don't want that. I instead provided a test endpoint in campfire which sends a mail to the requesting cloud admin user. The rule `mail:send-test` is applied there.

majewsky (approved):

PR #11695, opened 2026-05-18
system/cc-runtime-cluster: set stable JWKS URI for apiserver

felix-kaestner:
Set --service-account-jwks-uri to the cluster's external control plane endpoint so the OIDC discovery document advertises a stable URI instead of the apiserver pod's IP address.

Without this, the /.well-known/openid-configuration endpoint returns a jwks_uri pointing to the current advertise-address (a pod IP). Consumers that cache the discovery response break when the seed apiserver endpoints change, because the cached pod IP becomes unreachable.

Using the external control plane hostname ensures the jwks_uri remains valid regardless of apiserver pod rescheduling and is reachable by both in-cluster and external consumers.

### DIFF
--- a/system/cc-runtime-cluster/templates/kubeadmcontrolplane.yaml
+++ b/system/cc-runtime-cluster/templates/kubeadmcontrolplane.yaml
@@ -89,6 +89,8 @@ spec:
             value: "10"
           - name: audit-log-maxsize
             value: "100"
+          - name: service-account-jwks-uri
+            value: "https://{{ .Values.controlplane.address }}:{{ .Values.controlplane.port }}/openid/v1/jwks"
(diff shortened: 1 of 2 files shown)


system/cc-runtime-cluster/templates/kubeadmcontrolplane.yaml
```diff
@@ -89,6 +89,8 @@ spec:
             value: "10"
           - name: audit-log-maxsize
             value: "100"
+          - name: service-account-jwks-uri
```
felix-kaestner, line 92: See https://github.com/kubernetes/kubernetes/blob/master/pkg/kubeapiserver/options/authentication.go#L452

PR #11697, opened 2026-05-18
Created the basic designate config and added cloud_dns_admin role

skook1:


openstack/manila/templates/etc/_secrets.conf.tpl
```diff
@@ -16,6 +16,12 @@ username = {{ .Values.global.manila_dns_username | default "manila-dns" | includ
 password = {{ .Values.global.manila_dns_password | default "" | include "resolve_secret" | replace "$" "$$" }}
 {{- end }}
 
+{{- if .Values.designate.enabled }}
```
Carthaca: this is duplicated, see the lines above

when you touch it, please adjust the section similar to neutron and keystone_authtoken, i.e. `printf "%s/%s/manila/keystone-user/dns/username` and so on

Carthaca (changes requested):

Carthaca (approved):

kpawar-sap (approved):

PR #11699, opened 2026-05-19
[bridge-attach] new chart to attach a host interface to a Linux bridge

SchwarzM:
New chart `system/bridge-attach`. Privileged DaemonSet that runs on every node and attaches a host interface (default `eth1`) to a Linux bridge (default `br0`) via `ip link set <iface> master <bridge>`. Script ships as a ConfigMap, waits for the bridge to appear, is idempotent (master is checked via `/sys/class/net/<iface>/master`, not `ip link show ... master ...` which silently ignores the filter when a device is named), then sleeps to keep the pod Running.

Image is templated, defaults to `keppel.global.cloud.sap/ccloud/bridge-attach:latest`. The Dockerfile is included in this PR — alpine 3.22 with `bash` and `iproute2` installed.

### DIFF
--- a/system/bridge-attach/Dockerfile
+++ b/system/bridge-attach/Dockerfile
@@ -1,0 +1,6 @@
+FROM keppel.eu-de-1.cloud.sap/ccloud-dockerhub-mirror/library/alpine:3.22
+
+RUN apk add --no-cache bash iproute2
+
+LABEL source_repository="https://github.com/sapcc/helm-charts/tree/master/system/bridge-attach"
+LABEL maintainer="Marian Schwarz"
--- a/system/bridge-attach/Dockerfile
+++ b/system/bridge-attach/Dockerfile
@@ -1,6 +1,0 @@
-FROM keppel.eu-de-1.cloud.sap/ccloud-dockerhub-mirror/library/alpine:3.22
-
-RUN apk add --no-cache bash iproute2
-
-LABEL source_repository="https://github.com/sapcc/helm-charts/tree/master/system/bridge-attach"
-LABEL maintainer="Marian Schwarz"
--- a/system/bridge-attach/templates/daemonset.yaml
+++ b/system/bridge-attach/templates/daemonset.yaml
@@ -12,7 +12,6 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: bridge-attach-{{ $apod }}
-  namespace: kube-system
   labels:
     app.kubernetes.io/name: bridge-attach
     apod: {{ $apod }}
--- a/system/bridge-attach/templates/daemonset.yaml
+++ b/system/bridge-attach/templates/daemonset.yaml
@@ -1,15 +1,26 @@
 {{- if .Values.enabled }}
+{{- if .Values.global.apods }}
+{{- $root := . }}
+{{- range $azLong, $apods := .Values.global.apods }}
+{{- range $apods }}
+{{- $apod := . }}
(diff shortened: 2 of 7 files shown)


system/bridge-attach/templates/daemonset.yaml
```diff
@@ -0,0 +1,76 @@
...
+    app.kubernetes.io/name: bridge-attach
+    apod: {{ $apod }}
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: bridge-attach
+      apod: {{ $apod }}
+  updateStrategy:
+    rollingUpdate:
+      maxUnavailable: '10%'
```
sebageek, line 25: [optional] In theory we could just roll them out all at the same time, right? On the other hand it's probably that quick that it doesn't make any difference at all.

system/bridge-attach/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM keppel.eu-de-1.cloud.sap/ccloud-dockerhub-mirror/library/alpine:3.22
+
+RUN apk add --no-cache bash iproute2
+
+LABEL source_repository="https://github.com/sapcc/helm-charts/tree/master/system/bridge-attach"
+LABEL maintainer="Marian Schwarz"
```
sebageek: Do we do single maintainers or would it make sense to put in the network-api team here?

system/bridge-attach/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM keppel.eu-de-1.cloud.sap/ccloud-dockerhub-mirror/library/alpine:3.22
```
sebageek: [question] What's the best practice on verisoned dependencies right now? Hardcode one or have this set to something dynamic like `latest` or something stable?

sebageek (commented):
I like it, added three questions. Also, do you already want to squash the thing before merge?

system/bridge-attach/Dockerfile
SchwarzM: I did not know who to put in so I put myself in ... open for other things

system/bridge-attach/templates/daemonset.yaml
SchwarzM: I would do this as it is a bit lighter on the api instead of rolling x nodes at the same time

system/bridge-attach/Dockerfile
SchwarzM: I think team SRS came up with some auto updating base images but I did not find them. so I would love to have something controlled by us that auto updates

system/bridge-attach/Dockerfile
sebageek: Looking at our repo it seems to be individuals in most cases. I think you can either leave it that way or put in "network-api", if that makes sense in the context - then at least people won't ping you but the responsible team? Both options would work for me.

system/bridge-attach/Dockerfile
sebageek: I think that's done via our values repo by adding a `# apline:latest` or something like that to the line, sooo nothing for the helm-charts. Other charts seem to just depend on `apline:3` should we choose that as a default for now?

SchwarzM: ok I got rid of the Dockerfile and used the cc/base-images

system/bridge-attach/Dockerfile
SchwarzM: Dockerfile is now gone

sebageek (approved):

rhalina (approved):

PR #11705, opened 2026-05-19
updated alert for nfs SSD datastore so it doesn't flag STD DS

frederic-boulianne:
updated alert for nfs SSD datastore so it doesn't flag std datastores


himanip94 (changes requested):
Could you please update the playbook as well? Also add SSD in the description and bump the chart version.

himanip94 (approved):

PR #11710, opened 2026-05-20
logs: align region resource attribute to cloud.region

q8isar-khan:
This PR aligns collector resource metadata with our infrastructure logging semantic conventions by replacing the non-standard region key with cloud.region


q8isar-khan:
This PR#11710 is stacked on top of PR#11663 and should be merged after PR#11663  .
It  contains only the region -> cloud.region semconv alignment in system/logs/templates/opentelemetry-logs-collector.yaml.

melvinsimmert (changes requested):
Can be closed. I created https://github.com/sapcc/helm-charts/pull/11767 instead so we only have the changes necessary in the PR.

@q8isar-khan you can delete this PR. I will resolve your story as "Completed".

PR #11721, opened 2026-05-21
[bind] get rid of `pvc_zone_selector`

vssldmtrv:
We're now scheduling the **pods** explicitly using `spec.nodeSelector.topology.kubernetes.io/zone`

Previously we were using `pvc_zone_selector` to conditionally schedule the **PVCs** to a zone and this was flag set to `false` by default. This means that it was possible to have two deployments, say bind1 and bind2, ending up in the same AZ.

All of that is now explicit and we're actually scheduling the **pods** and not the **PVCs**. The PVCs now follow the pods, as opposed to the other way around.

### DIFF
--- a/system/bind/Chart.yaml
+++ b/system/bind/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Bind DNS software. 
 name: bind
-version: 1.0.2
+version: 1.1.0
 appVersion: "9.18.24"
 dependencies:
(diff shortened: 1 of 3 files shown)


vssldmtrv:
This cannot be applied retroactively. All existing PVCs are immutable as far as `spec.selector.matchLabels.failure-domain.beta.kubernetes.io/zone` goes. We can neither add, nor remove it.

PR #11771, opened 2026-05-27
[trust-manager] Template cert exporter image registry, repo and version

jagoleni:


openstack/trust-manager/values.yaml
```diff
@@ -30,3 +30,7 @@ namespaces:
 cert_exporter:
   namespaces:
     - monsoon3
+  image:
+    registry: docker.io
```
joker-at-work: Having this default would allow us to  deploy with a non-keppel registry. Should we maybe put keppel here, too?

joker-at-work (dismissed):

openstack/trust-manager/values.yaml
jagoleni: I explicitly put the non-keppel registry here. @vssldmtrv mentioned that keppel should not be used in public repos: https://github.com/sapcc/helm-charts/pull/11572

joker-at-work: Little hard to do given that all our helm-charts' dependencies lie on keppel e.g. https://github.com/sapcc/helm-charts/blob/36c9f1a420c6f785dd3a7218c79e5a204121b191/openstack/trust-manager/Chart.yaml#L12

jagoleni: Yes... I'm using Keppel now

joker-at-work (approved):

Carthaca (approved):

PR #11770, opened 2026-05-27
upgrade prometheus-operator CRDs to v0.91.0

ibakshay:
Upgrades the prometheus-operator CRDs from v0.88.0 to v0.91.0.

## Changes

- Bumped `Chart.yaml` version: `8.0.0` → `9.0.0` (major bump consistent with sapcc/helm-charts#10897)
- Refreshed all 10 CRDs via `./get-crds v0.91.0`:
  - `controller-gen.kubebuilder.io/version`: `v0.19.0` → `v0.20.1`
  - `operator.prometheus.io/version`: `0.88.0` → `0.91.0`

## Some breaking changes that needs to be monitored

Per the v0.91.0 release notes, several `[CHANGE]` items add validations that may reject previously-valid CRs:
- `ScrapeConfig`: `basicAuth` / `authorization` / `oauth2` are now mutually exclusive; new minLength on string fields
- `AlertmanagerConfig`: stricter validations for VictorOps, OpsGenie, and Email receivers

PR #11772, opened 2026-05-27
bump prometheus-crds dependency to 9.0.0 in consumer charts

ibakshay:
Bumps the `prometheus-crds` dependency from `8.0.0` to `9.0.0` 
follow-up to #11770


viennaa (approved):
And fire 👍

PR #11778, opened 2026-05-27
[maia]configure if the seed needs to be skip

LeoZhangZXZ:
## Summary

Adds `seed.skip` and `seed_global.skip` boolean values to the Maia Helm chart to allow selective suppression of the `OpenstackSeed` resources without disabling Maia itself.

## Motivation

In deployments where the OpenStack seeder operator is not available in the local region (e.g. Swift is provisioned in a different region), the seed resources would fail to apply. This change allows operators to skip seed deployment in the local cluster and manage the `OpenstackSeed` resources independently (e.g. via `system/thanos-seeds` deployed in the target region).

## Changes

- `templates/seed.yaml` — wraps the existing template in `{{- if not .Values.seed.skip }}`
- `templates/seed_global.yaml` — wraps the existing template in `{{- if not .Values.seed_global.skip }}`
- `values.yaml` — adds `seed.skip: false` and `seed_global.skip: false` (default behaviour unchanged)

## Usage

To skip seed deployment in a region, set in the region values:

```yaml
seed:
  skip: true

seed_global:
  skip: true

### DIFF
--- a/openstack/maia/templates/seed.yaml
+++ b/openstack/maia/templates/seed.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.seed.skip }}
 {{- if not .Values.global.is_global_region }}
 {{- if or .Values.maia.enabled .Values.seed.enabled}}
 {{- $cdomains := .Values.global.domain_seeds.customer_domains | required "missing value for .Values.global.domain_seeds.customer_domains" -}}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by viennaa — COMMENTED** [review]
This PR should not be needed. 

Global can be disabled when we are not in a global region (this should be automatically done)
```
{{- if .Values.global.is_global_region }}
```

Thanos seed can be disabled by setting this:
https://github.com/sapcc/helm-charts/blob/master/common/prometheus-server-pre7/templates/thanos/thanos-seed.yaml#L1

LeoZhangZXZ:
then i will close this PR

PR #11789, opened 2026-05-28
adding remote configuration options to disco operator

asergeant01:
## Adding remote cluster capability to the disco operator

This PR adds optional support for running the disco operator against a **remote Kubernetes cluster** via Gardener's token-requestor mechanism, while keeping all existing behaviour as the default.

### What it does

When `remote.enabled: true`, the chart configures disco to connect to a remote cluster rather than its own in-cluster API server:

- Creates a `disco-remote-kubeconfig` Secret annotated for Gardener's token-requestor to inject a kubeconfig and CA bundle for the target shoot cluster
- Mounts that secret into the controller pod at `/var/run/secrets/kubernetes.io/serviceaccount`, overriding the default in-cluster credentials
- Sets `KUBERNETES_SERVICE_HOST` / `KUBERNETES_SERVICE_PORT` env vars to point the operator at the remote API server
- Adds the required Gardener network policy labels to the pod so egress to the shoot API server is permitted
- Pins `NAMESPACE` to `remote.leaderElectionNamespace` (default: `kube-system`) instead of deriving it from the pod's own namespace

RBAC resources gain namespace and subject overrides for cross-cluster scenarios where the ServiceAccount, Roles, and ClusterRoleBindings may need to reference subjects in a different namespace or under a different name than the release namespace.

### Backwards compatibility

`remote.enabled` defaults to `false`. With remote disabled, all rendered resources are functionally identical to the previous version of the chart — no new objects are created and the operator behaviour is unchanged.

### DIFF
--- a/system/disco/templates/disco-controller-manager.yaml
+++ b/system/disco/templates/disco-controller-manager.yaml
@@ -6,8 +6,10 @@ metadata:
     {{- with .Values.controllerLabels }}
     {{- toYaml . | nindent 4 }}
     {{- end }}
+  {{- if .Values.reloader.enabled }}
   annotations:
     secret.reloader.stakater.com/reload: "disco-config"
+  {{- end }}
   name: disco-controller-manager
 spec:
   replicas: 1
@@ -21,6 +23,10 @@ spec:
         checksum/config: {{ include (print $.Template.BasePath "/disco-config.yaml") . | sha256sum }}
       labels:
         app: disco
+        {{- if .Values.remote.enable }}
+        networking.gardener.cloud/to-shoot-apiserver: allowed
+        networking.resources.gardener.cloud/to-kube-apiserver-tcp-443: allowed
+        {{- end }}
         {{- with .Values.controllerLabels }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
@@ -30,9 +36,19 @@ spec:
         - /disco
         env:
         - name: NAMESPACE
+        {{- if .Values.remote.enable }}
+          value: {{ .Values.remote.leaderElectionNamespace | default "kube-system" | quote }}
+        {{- else }}
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        {{- end }}
+        {{- if .Values.remote.enable }}
+        - name: KUBERNETES_SERVICE_HOST
+          value: {{ required "remote.apiServerHost is required when remote.enable is true" .Values.remote.apiServerHost | quote }}
+        - name: KUBERNETES_SERVICE_PORT
+          value: {{ .Values.remote.apiServerPort | default "443" | quote }}
+        {{- end }}
         envFrom:
         - secretRef:
             name: disco-config
@@ -45,9 +61,11 @@ spec:
           periodSeconds: 20
         name: manager
         ports:
+        {{- if .Values.webhooks.enabled }}
         - containerPort: 9443
           name: webhook-server
           protocol: TCP
+        {{- end }}
         - containerPort: 8080
           name: https
           protocol: TCP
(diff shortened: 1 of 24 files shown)


stefanhipfel (commented):
lgtm

system/disco/values.yaml
```diff
@@ -2,6 +2,18 @@ image:
   repository: keppel.global.cloud.sap/ccloud-ghcr-io-mirror/sapcc/disco
   tag: v2.0.0
 
+# Remote Kubernetes cluster configuration (for Gardener token-requestor)
+remote:
+  # Enable remote kubeconfig mounting for connecting to a remote Kubernetes cluster
+  enable: false
```
IvoGoman: For consistency to the other `enabled` flags
```suggestion
  enabled: false
```

system/disco/templates/disco-leader-election-role.yaml
```diff
@@ -1,7 +1,9 @@
+{{- if .Values.rbac.create }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: disco-leader-election-role
+  namespace: {{ .Release.Namespace }}
```
IvoGoman, line 9: The namespace is defaulted by helm to this value. (same for the other  occurrences where this is added)
```suggestion
```

system/disco/values.yaml
```diff
@@ -13,8 +25,38 @@ seed:
...
+webhooks:
+  # Create webhook service and MutatingWebhookConfiguration
+  enabled: true
+
+# Metrics service for Prometheus scraping
+metrics:
+  # Create the metrics service
+  enabled: true
+  # Port to expose metrics on.
+  port: 9091
```
IvoGoman: This is not unused

system/disco/templates/disco-controller-manager.yaml
```diff
@@ -6,8 +6,10 @@ metadata:
     {{- with .Values.controllerLabels }}
     {{- toYaml . | nindent 4 }}
     {{- end }}
+  {{- if .Values.reloader.enabled }}
   annotations:
     secret.reloader.stakater.com/reload: "disco-config"
+  {{- end }}
```
IvoGoman: Instead of enabling/disabling the annotation block, how about making the annotations themselves configurable. With the default one being the reloader

system/disco/templates/remote-kubeconfig.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.remote.enable }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: disco-remote-kubeconfig
+  labels:
+    resources.gardener.cloud/purpose: token-requestor
+    resources.gardener.cloud/class: shoot
+  annotations:
+    serviceaccount.resources.gardener.cloud/name: disco-controller-manager
```
IvoGoman: This should not be hard-coded
```suggestion
serviceaccount.resources.gardener.cloud/name: {{ .Values.serviceAccount.name }}
```

system/disco/templates/remote-kubeconfig.yaml
```diff
@@ -0,0 +1,16 @@
...
+    serviceaccount.resources.gardener.cloud/namespace: kube-system
```
IvoGoman: The SA is deployed into the release namespace by default. I am not familiar with the token requestor API, but assume this needs to point to the release namespace?
```suggestion
    serviceaccount.resources.gardener.cloud/namespace: {{ .Release.Namespace }}
```

system/disco/templates/disco-controller-manager.yaml
```diff
@@ -30,9 +36,19 @@ spec:
         - /disco
         env:
         - name: NAMESPACE
+        {{- if .Values.remote.enable }}
+          value: {{ .Values.remote.leaderElectionNamespace | default "kube-system" | quote }}
```
IvoGoman: This is already defaulted by the values. Do we need to set the namespace explicitly or should this also be the `{{ .Release.Namespace }}`? 
```suggestion
          value: {{ .Values.remote.leaderElectionNamespace | quote }}
```

system/disco/templates/disco-controller-manager.yaml
```diff
@@ -30,9 +36,19 @@ spec:
...
+        {{- else }}
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        {{- end }}
+        {{- if .Values.remote.enable }}
+        - name: KUBERNETES_SERVICE_HOST
+          value: {{ required "remote.apiServerHost is required when remote.enable is true" .Values.remote.apiServerHost | quote }}
+        - name: KUBERNETES_SERVICE_PORT
+          value: {{ .Values.remote.apiServerPort | default "443" | quote }}
```
IvoGoman: This default is redundant to the default set in the `values.yaml`
```suggestion
          value: {{ .Values.remote.apiServerPort | quote }}
```

system/disco/tests/deployment_test.yaml
```diff
@@ -0,0 +1,251 @@
...
+      - template: disco-controller-manager.yaml
+        isKind:
+          of: Deployment
+
+  - it: should use the default image
+    asserts:
+      - template: disco-controller-manager.yaml
+        equal:
+          path: spec.template.spec.containers[0].image
+          value: keppel.global.cloud.sap/ccloud-ghcr-io-mirror/sapcc/disco:v2.0.0
```
IvoGoman: This requires manual updates on all succeeding PRs to set the correct disco version. Can we remove this check?

system/disco/templates/disco-leader-election-role.yaml
asergeant01: This was done due to kustomize manipulating the namespace downstream. However I will change this to make it configurable from the helm chart in the first place

system/disco/values.yaml
asergeant01: Will update template to use

IvoGoman (approved):

PR #11808, opened 2026-05-29
[maia]Update maia ingress annotation disc

LeoZhangZXZ:
add the annotation disc=true to enable the DNS registration for maia


viennaa:
We have ingresses for maia in every region, the explanation why we need this does not add up to me with the cc/secrets PR.

LeoZhangZXZ:
> We have ingresses for maia in every region, the explanation why we need this does not add up to me with the cc/secrets PR.

eu-de-3 is a newer region with no pre-existing DNS record for maia.eu-de-3.cloud.sap. The disco controller is responsible for registering DNS
  records for services, and it only acts on ingress that have disco: "true" annotation. Without it, no DNS record was created, and the hostname was unresolvable. Moreover, other services in eu-de-3 also have it, e.g. [elektra](https://github.com/sapcc/helm-charts/blob/master/openstack/elektra/templates/ingress.yaml#L38). in other region, we may have done the DNS record in another way. But disco is idempotent — if the DNS record already exists with the correct value, it does nothing, so it will not destroy other regions but fix eu-de-3.

viennaa (commented):
OK, sounds reasonable 👍 

Bump the Chart pls

viennaa (approved):

PR #11832, opened 2026-06-01
cinder: fix liveness probe path from / to /healthcheck

hemna:
Investigation findings (eu-de-2, 2026-06-01):

All cinder-api pods (8 replicas) showed 25-49 restarts each, with the kubelet logging 111,981 ProbeWarning events over 13 days:

  Warning  ProbeWarning  Liveness probe warning: Probe terminated
  redirects, Response body: {"versions": [{"id": "v3.0", ...}]}

Root cause: the liveness probe hits GET / on HTTP port 8776. The Cinder API root endpoint returns a 300-series redirect to the HTTPS versioned URL (https://volume-3.<region>.cloud.sap/v3/). Kubelet follows the redirect and gets a valid response body, but records every probe as a ProbeWarning. This generates ~8 warnings/minute continuously and buries real health signals in the event stream.

Additionally, GET / goes through the full WSGI routing pipeline. Under worker thread exhaustion (e.g. during ProxySQL connection pool pressure), even this lightweight probe request can queue behind in-flight handlers and miss the 5s timeout, triggering the 3-failure threshold and causing an unnecessary pod restart.

The readiness probe already uses GET /healthcheck correctly. The /healthcheck endpoint is handled by oslo_middleware.Healthcheck early in the WSGI pipeline (before auth and routing), returns 200 OK with body 'OK' directly, and produces no redirect.

Verified GET /healthcheck returns HTTP 200 OK on all three regions:
  eu-de-1: 200 OK
  na-us-2: 200 OK
  eu-de-2: 200 OK

Fix: align the liveness probe path with the readiness probe path.

### DIFF
--- a/openstack/cinder/Chart.yaml
+++ b/openstack/cinder/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v2
 description: A Helm chart for Kubernetes
 icon: https://www.openstack.org/themes/openstack/images/project-mascots/Cinder/OpenStack_Project_Cinder_mascot.png
 name: cinder
-version: 0.4.0
+version: 0.4.1
 dependencies:
   - name: utils
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
--- a/openstack/cinder/templates/api-deployment.yaml
+++ b/openstack/cinder/templates/api-deployment.yaml
@@ -82,7 +82,7 @@ spec:
               {{- include "utils.snippets.pre_stop_graceful_shutdown" . | indent 14 }}
           livenessProbe:
             httpGet:
-              path: /
+              path: /healthcheck
               port: cinder-api
             initialDelaySeconds: 15
             timeoutSeconds: 5


Scsabiii (approved):

joker-at-work:
Are you sure this is the right call? Not checking the actual application/API at all but instead relying on something that maybe not even need the DB and thus not really checking much? You can always increase the timeout.

I get the part about the spurious errors from redirect, but for that you could just change the URL to something else. I agree that checking `https://volume-3.*` doesn't make sense at that's effectively checking _any_ api pod is reachable via Ingress.

hemna:
Currently the call to / doesn't do much.  it doesn't check the db either.  It does return the api versions available and that's about it.  I saw this under load from our favorite Denial Of Service customer/App Gardener overloading cinder api again in eu-de-2.  

the /healthcheck url is also not overly useful looking deeper into it.  It's simply oslo_middleware call which runs when the cinder-api service process is running, and also doesn't touch the db. 

So neither urls are really useful in the sense that k8s is asking for "liveness", meaning that the service is running and should be able to service real customer requests.  neither tests really guarantees it.  

Since this is mostly a problem under load, adding a new check that goes to the DB, could also expose cinder to a DB DoS, even for a simple SELECT 1, sql query.


The real problem is the the exhaustion of the proxysql connections underneath that occupy all of the greenthreads and then new requests comming in time out, leading to the liveness probe timeout and bouncing the api.

Increasing the timeout might help some, but we really just need rate limiting enabled.

PR #11835, opened 2026-06-02
Add CinderTagMissingForHDDDatastoreNFS alert

ki-ste:
Adds a new alert for standard-tier NFS datastores (name contains "_std_") that should have the cinder tag "fcd-std".

Mirrors the structure of CinderTagMissingForSSDDatastoreNFS but matches the std datastores instead of excluding them, and expects "fcd-std" instead of "fcd".

Related JIRA: https://jira.tools.sap/browse/COMPUTEMONITORING-73


viennaa:
Fix failing tests pls.

himanip94 (changes requested):
Please fix the indentation.

himanip94 (approved):

PR #11836, opened 2026-06-02
[cc-gardener] scheduler region distance config

jknipper:


ronchi-oss (approved):

jknipper:
moving it to the persephone repo

PR #11849, opened 2026-06-03
add managed resource capabilities

asergeant01:
## Add ManagedResource support for Gardener remote deployments

When `remote.enabled: true`, the chart previously rendered RBAC and the ServiceAccount as plain
resources in the garden cluster — wrong cluster, wrong namespace. Deployers had to work around this
with external post-processing (kustomize wrap rules) to move these resources to the shoot cluster.

This change makes the chart self-contained for Gardener deployments.

### What changed

**When `remote.enabled: false` — no functional change.** Rendered output is byte-for-byte identical to 2.4.3.

**When `remote.enabled: true`:**

- RBAC templates (`ClusterRole`, `ClusterRoleBinding`, `Role`, `RoleBinding`) are suppressed from
  direct rendering — they no longer land on the garden cluster. The `ServiceAccount` continues to
  render locally (the pod needs it in the garden cluster) and is also included in the `disco-rbac`
  ManagedResource so it exists on the shoot cluster for RBAC and token-requestor purposes.
- Two `ManagedResource` + `Secret` pairs are emitted instead, causing Gardener's ResourceManager to
  apply them to the shoot cluster:
  - `disco-crd` — the CRD, read directly from `crds/` at render time (no duplication, stays in sync
    with kubebuilder-generated output)
  - `disco-rbac` — all RBAC resources and the `ServiceAccount`, namespaced to `remote.shootNamespace`
    (default: `kube-system`)
- The token-requestor annotation on `disco-remote-kubeconfig` now correctly uses `remote.shootNamespace`
  rather than defaulting to the release namespace (which caused empty token/CA injection)
- `remote.leaderElectionNamespace` removed — `remote.shootNamespace` covers both the ManagedResource
  namespacing and the `NAMESPACE` env var on the controller, which must always match
- `disco-crd` — the CRD, read directly from `crds/` at render time (no duplication, stays in sync
  with kubebuilder-generated output). Local CRD installation should be skipped when deploying in
  remote mode (e.g. `--skip-crds` / `includeCRDs: false`) since the CRD is managed on the shoot
  cluster via the ManagedResource.

**Refactor:** All RBAC resource definitions are centralised as named templates in `_helpers.tpl`. The
individual template files are thin wrappers that `include` them with the appropriate namespace context.
`disco.shootRBAC` reuses the same named templates with `shootNamespace` — no duplication between the
local and remote paths.


abhijith-darshan:
I gave it a glance and checking what needs to be adapted into **Greenhouse** - https://github.com/cloudoperators/greenhouse-extensions/blob/main/disco/plugindefinition.yaml

asergeant01:
@abhijith-darshan Greenhouse plugin is pinned to a particular version, therefore merging wouldn't affect it initially until the maintainers decide to update.

As stated, only when the new functionality is enabled with remote.enable: true, does anything new get rendered.

abhijith-darshan (approved):

PR #11875, opened 2026-06-05
[prometheus-server] fix playbook relabeling

trouaux:
`error parsing regexp: invalid or unsupported Perl syntax: `(?!`"`

### DIFF
--- a/common/prometheus-server/templates/_alertmanager_relabel.yaml.tpl
+++ b/common/prometheus-server/templates/_alertmanager_relabel.yaml.tpl
@@ -20,5 +20,5 @@
 - action: replace
   source_labels: [playbook]
   target_label: playbook
-  regex: ^(?!https?://)(.*)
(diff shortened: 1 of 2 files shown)


viennaa (approved):

common/prometheus-server/templates/_alertmanager_relabel.yaml.tpl
```diff
@@ -20,5 +20,5 @@
 - action: replace
   source_labels: [playbook]
   target_label: playbook
-  regex: ^(?!https?://)(.*)
-  replacement: {{ printf "http://%s/$1" ($root.Values.global.playbookDomain | default "https://operations.global.cloud.sap") }}
+  regex: '^(docs/.+)$'
+  replacement: {{ printf "%s/$1" ($root.Values.global.playbookDomain | default "https://operations.global.cloud.sap") }}
```
ibakshay, line 24: Is it fine to hard-code the Base-URL? Why can't we extract from the secrets repository?

PR #11889, opened 2026-06-08
Update opentelemetry-logs-collector.yaml

GouravL1296:
As per NETOBS-915 , need to update the opentelemetry-log-collector.yaml file


melvinsimmert (approved):
This is the version we tested sucessfully.

timojohlo (approved):

PR #11896, opened 2026-06-09
[kube-system-scaleout] remove ldap-named-user from umbrella

s10:
ldap-named-user chart is installed by the separate pipeline, so we need to remove this chart from kube-system-scaleout umbrella


goerangudat (approved):
lgtm

PR #11901, opened 2026-06-09
manila,barbican: fix migration job SA on fresh install

chuan137:
The lookup-gated pattern for `serviceAccountName` causes the migration job to fall back to the `default` SA on first install, since the named SA does not exist yet when the `post-install` hook fires. This aligns with the `rbac.enabled`-gated pattern already used by nova and cinder, which avoids the race condition entirely. The same bug was present in barbican and is fixed here as well.


Carthaca (approved):

sumitarora2786 (approved):

rajivmucheli (approved):

rajivmucheli:
@s10 FYR

PR #11914, opened 2026-06-10
[utils] use non-deprecated topology as default

Carthaca:


s10 (approved):

joker-at-work (approved):
I guess this means all clusters are new enough now?

galkindmitrii:
> I guess this means all clusters are new enough now?

probably not `qa-de-2`... but with 1 AZ that shouldn't matter.

s10:
`failure-domain.beta.kubernetes.io/zone` is deprecated since v1.17 (2019-12-09), see https://kubernetes.io/docs/reference/labels-annotations-taints/#failure-domainbetakubernetesiozone , so for this one all our clusters are new enough...

galkindmitrii (approved):

PR #11938, opened 2026-06-11
chore(argora-operator): bump argora to 916b022

defo89:
Bump argora dependency to `0.0.1-20260611084753-916b022` in both argora-operator and argora-operator-remote charts.


goerangudat (approved):
lgtm

PR #11942, opened 2026-06-11
[manila] add config options for 2025.1 upgrade

chuan137:


openstack/manila/templates/etc/_manila.conf.tpl
```diff
@@ -65,6 +65,13 @@ server_migration_driver_continue_update_interval = {{ .Values.server_migration_d
 server_migration_extend_neutron_network = {{ .Values.server_migration_extend_neutron_network | default true }}
 ensure_driver_resources_interval = {{ .Values.ensure_driver_resources_interval | default 14400 }}
 
+# Prevent shares from transitioning to 'ensuring' status (2025.1 default changed to True)
+update_shares_status_on_ensure = False
+
+# NetApp driver supports set_vlan and set_mtu as updatable subnet metadata keys.
+# Uncomment to allow overriding VLAN/MTU per share network subnet via metadata.
+# driver_updatable_subnet_metadata = set_vlan,set_mtu
+
```
Carthaca, line 73:
```suggestion

# manila by default would put "{project_id}_" as prefix, but since we anyhow don't have cross-project share servers, we can skip that.
# obviously having shares with the same mount point name in the same share server is not possible,
# but this cannot be solved by a prefix (for that case we have the share instance uuid, if mount point name is not used)
default_mount_point_prefix = {{ .Values.mount_point_prefix | default "" }}
```

openstack/manila/templates/etc/_manila.conf.tpl
```diff
@@ -65,6 +65,13 @@ server_migration_driver_continue_update_interval = {{ .Values.server_migration_d
 server_migration_extend_neutron_network = {{ .Values.server_migration_extend_neutron_network | default true }}
 ensure_driver_resources_interval = {{ .Values.ensure_driver_resources_interval | default 14400 }}
 
+# Prevent shares from transitioning to 'ensuring' status (2025.1 default changed to True)
+update_shares_status_on_ensure = False
+
+# NetApp driver supports set_vlan and set_mtu as updatable subnet metadata keys.
+# Uncomment to allow overriding VLAN/MTU per share network subnet via metadata.
+# driver_updatable_subnet_metadata = set_vlan,set_mtu
```
Carthaca: Those are not driver_updatable_subnet_metadata, but just metadata that would be applied on share server create when you set it on the subnet - updating the metadata will not cause an update in the driver for those.
Have a look at update_share_network_subnet_from_metadata() e.g. for showmount how that is handled differently

openstack/manila/templates/etc/_manila.conf.tpl
```diff
@@ -109,6 +116,9 @@ insecure = True
 
 [oslo_policy]
 policy_file = /etc/manila/policy.yaml
+# Set False to preserve old Keystone RBAC behavior (non-scoped); 2025.1 defaults changed to True
+enforce_new_defaults = False
```
Carthaca, line 124:
```suggestion
# look out for "using the intended scope is required" warnings in log
enforce_new_defaults = False
```

openstack/manila/templates/shares/_share-netapp.conf.tpl
```diff
@@ -96,6 +98,9 @@ netapp_volume_move_cutover_timeout = {{ $share.volume_move_cutover_timeout | def
 # state, that will be reported as pool property. Valid values are `in_build`, `live`, `in_decom` and `replacing_decom`
 netapp_hardware_state = {{ $share.hardware_state | default "live" }}
 
+# SAP default: 6 years. Upstream default is 1 year
```
Carthaca:
```suggestion
# SAP default: 6 years (we refresh the hardware earlier than that). Upstream default is 1 year
```

Carthaca (changes requested):
thanks, I was preparing nearly the same, have a look at my suggestions

openstack/manila/templates/etc/_manila.conf.tpl
chuan137: updated, should we keep them disabled?

Carthaca: we can enable those, I think, only `driver_updatable_metadata` depends on https://github.com/sapcc/manila/pull/323

Carthaca (dismissed):

kpawar-sap (dismissed):

Carthaca (approved):

PR #11956, opened 2026-06-12
[ironic] remove outdated sentry references

sandzwerg:
Sentry is connected via the templates/sentry.yaml snippet so the old way
via sentry_dsn is no longer needed. We remove all references to
sentry_dsn.

### DIFF
--- a/openstack/ironic/values.yaml
+++ b/openstack/ironic/values.yaml
@@ -455,22 +455,26 @@ logging:
       class: StreamHandler
       args: "(sys.stdout,)"
       formatter: context
-    sentry:
-      class: sapcc_sentrylogger.handler.SentryEventHandler
+    sentry_events:
+      class: "sapcc_sentrylogger.handler.EventHandler"
       level: ERROR
+      args: "()"
+    sentry_breadcrumbs:
+      class: "sapcc_sentrylogger.handler.BreadcrumbHandler"
+      level: INFO
       args: "()"
   loggers:
     root:
-      handlers: stdout, sentry
+      handlers: stdout, sentry_events, sentry_breadcrumbs
       level: WARNING
     ironic:
-      handlers: stdout, sentry
+      handlers: stdout, sentry_events, sentry_breadcrumbs
       level: DEBUG
     eventlet.wsgi.server:
-      handlers: stdout, sentry
+      handlers: stdout, sentry_events, sentry_breadcrumbs
       level: INFO
     auditmiddleware:
-      handlers: stdout, sentry
+      handlers: stdout, sentry_events, sentry_breadcrumbs
       level: INFO
 
 memcached:
--- a/openstack/ironic/values.yaml
+++ b/openstack/ironic/values.yaml
@@ -430,9 +430,6 @@ dbName: ironic
 max_pool_size: 1
 max_overflow: 50
 
-sentry:
-  enabled: true
-
 rabbitmq:
   name: ironic
   persistence:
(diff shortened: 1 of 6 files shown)


openstack/ironic/values.yaml
```diff
@@ -458,10 +455,18 @@ logging:
       class: StreamHandler
       args: "(sys.stdout,)"
       formatter: context
-    sentry:
-      class: sapcc_sentrylogger.handler.SentryEventHandler
+    "null":
```
BerndKue: Are the quotes really needed or rather a problem?

BerndKue (dismissed):

openstack/ironic/values.yaml
sandzwerg: I copied this 1:1 from neutron who wrote the replacement of raven and IIRC are already using this, so I'd expect that it needs to be this way. As it's the null handler I assume not quoting it will rather lead to an error as e.g. python is interpreting it instead of using it as a name.

davidgrun (dismissed):

joker-at-work (dismissed):
Are you sure that this reference to the secret created by the `SentryProject` from templates/sentry.yaml is not longer needed? I would assume the SENTRY_DSN environment variable is still in use. How else would it authenticate against sentry?

joker-at-work:
Neutron uses [utils.sentry_config](https://github.com/sapcc/helm-charts/blob/dfbc006abd0cd735eb5316c58929bc99ada75a78/openstack/neutron/templates/statefulset-network-agent-apod.yaml#L95) which still [sets SENTRY_DSN](https://github.com/sapcc/helm-charts/blob/dfbc006abd0cd735eb5316c58929bc99ada75a78/openstack/utils/templates/_helpers.tpl#L80-L92). So you would have to switch to that template, if you want to get rid of manually setting `SENTRY_DSN`.

sandzwerg:
No I'm not sure but I asked @rajivmucheli and he said we don't need the 
```
sentry:
  enabled: true
```
Or any reference to DSN anymore because we use a different [templates/sentry.yaml](https://github.com/sapcc/helm-charts/blob/master/openstack/ironic/templates/sentry.yaml) which apparently doesn't need any of this. He told me I could remove both from here as well as from the regional secrets. But maybe I misunderstood something? I stumbled over the whole sentry/DSN topic when working on the new region.

joker-at-work:
I suggest to try it out manually, if you know how to generate an internal server error (or deploy a custom image while trying it out).

rajivmucheli:
sentry.yaml guard {{- if .Values.logging.handlers.sentry }} is unchanged on the PR branch, and .Values.logging.handlers.sentry is nil after the PR. The SentryProject CRD will not be rendered.

What can be dismissed:

1. The concern about region overrides silently becoming no-ops via sentry.enabled — the secrets repo uses sentry_dsn:, not sentry.enabled:, so no region override breaks.
2. The inspector-conductor env: block concern — that block never had any other env vars, so the removal is clean.

sandzwerg:
So I adjusted the PR, the sentry.yaml now renders correctly. The sentry: enabled is fine to remove and with this logging handler config I see sentry events for ironic again in QA. I don't think there is a good way to test this without rolling it out so I'd like to roll this out to QA-DE-1 first. If something is broken I'll either add a new PR or revert this one.

joker-at-work:
I retracted my "changes requested". The PR still is different than my understanding of how things work, so I cannot approve - but I think you have enough approvers on your team :)
Please tell me if it works after your tests. If yes, I need to go and figure some things out.

BerndKue (approved):

PR #11963, opened 2026-06-12
[manila] update chart app version to epoxy

chuan137:


openstack/manila/Chart.yaml
```diff
@@ -10,7 +10,7 @@ maintainers:
 name: manila
 sources:
   - https://github.com/sapcc/manila
-version: 0.6.10
+version: 0.6.11
```
Carthaca: I'm inclined to require at least a minor version bump

Carthaca (dismissed):

openstack/manila/Chart.yaml
chuan137: yes, make sense

kpawar-sap (approved):

Carthaca (approved):

PR #11966, opened 2026-06-12
fix(manila): manila conf new lines malformat

chuan137:


Carthaca (changes requested):
no, we want to override it explicitly to be empty. otherwise manila would default the setting to `{project_id}_`

openstack/manila/templates/etc/_manila.conf.tpl
```diff
@@ -74,7 +74,11 @@ update_shares_status_on_ensure = False
 # manila by default would put "{project_id}_" as prefix, but since we anyhow don't have cross-project share servers, we can skip that.
 # obviously having shares with the same mount point name in the same share server is not possible,
 # but this cannot be solved by a prefix (for that case we have the share instance uuid, if mount point name is not used)
-default_mount_point_prefix = {{ .Values.mount_point_prefix | default "" }}
+{{- if .Values.mount_point_prefix }}
+default_mount_point_prefix = {{ .Values.mount_point_prefix }}
+{{- end }}
+# empty string value breaks the configmap format
+
```
Carthaca:
```suggestion
default_mount_point_prefix = {{ .Values.mount_point_prefix  | default "''" }}
```

tested with helm diff in qa that the format is working with this - we still need to see if the feature will work this way :)

Carthaca (approved):

kpawar-sap (approved):

PR #11968, opened 2026-06-15
[designate] Add domain_id field for catz pools

misamoylov:


openstack/designate/values.yaml
```diff
@@ -75,7 +75,7 @@ rndc_keys:
 db_name: designate
 pool_id: 794ccc2c-d751-44fe-b57f-8894c9f5c842
 scheduler_filters: attribute, pool_id_attribute, in_doubt_default_pool, fallback
-scheduler_filters_shared_pools: attribute, pool_id_attribute, pool_domains_attribute_filter, domain_id
```
galkindmitrii, line 78: I don't understand why the attribute filter is removed.

misamoylov: Done

galkindmitrii (approved):
oslo.context cannot be higher version than 5.6:
https://github.com/sapcc/requirements/blob/stable/2024.2-m3/upper-constraints.txt#L59

PR #11969, opened 2026-06-15
chore(argora-operator): bump argora to bc43f4d

stwilkes:
Bump argora dependency to 0.0.1-20260615072404-bc43f4d in both argora-operator and argora-operator-remote charts.


goerangudat (approved):
lgtm

PR #11981, opened 2026-06-16
NSX-T: do not set liveness probe to null

sven-rosenzweig:
We had issues with removing liveness probes. Trying to fix them by adding livenessProbe: null did not help.
There was another underlying problem.

The vcenter-operator executes a server-side apply to update k8s resources, which checks the ownership. Removing an element with a second owner, which we observed for very old deployments, did not remove the probe.

To overcome the issue, we cleanup the managed fields in the vcenter-operator.


joker-at-work (approved):

PaulPickhardt (approved):

rhalina (approved):
LGTM

PR #11990, opened 2026-06-16
[neutron] make swift-sftp bridge optional (default enabled)

notandy:
new regions don't have swift anymore, so we need to make it configurable

### DIFF
--- a/openstack/neutron/templates/sftp-deployment.yaml
+++ b/openstack/neutron/templates/sftp-deployment.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.sftp.swift.enabled }}
 kind: Deployment
 apiVersion: apps/v1
 metadata:
@@ -58,3 +59,5 @@ spec:
             secretName: neutron-sftp-backup
         - name: cache-volume
           emptyDir: {}
+{{- end -}}
+
--- a/openstack/neutron/values.yaml
+++ b/openstack/neutron/values.yaml
@@ -311,13 +311,14 @@ hypervisors_vmware: []
 
 # SFTP-swift bridge (e.g. for nsx-t backups in control-plane)
 sftp:
-  enabled: false
   server_key:
   externalIPs: []
   user: db_backup
   logins:
     db_backup: nsx-t-backup
     vcsa_backup: vcsa-backups
+  swift:
+    enabled: true
 
 drivers:
   nsxv3:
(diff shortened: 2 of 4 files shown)


openstack/neutron/values.yaml
```diff
@@ -311,13 +311,14 @@ hypervisors_vmware: []
...
 sftp:
-  enabled: false
   server_key:
   externalIPs: []
   user: db_backup
   logins:
     db_backup: nsx-t-backup
     vcsa_backup: vcsa-backups
+  swift:
+    enagbled: true
```
sebageek: Typo `enagbled` --> `enabled`

openstack/neutron/templates/sftp-deployment.yaml
```diff
@@ -58,3 +59,4 @@ spec:
             secretName: neutron-sftp-backup
         - name: cache-volume
           emptyDir: {}
+{{- end -}}
```
sebageek: Could you please re-add the newline at end of file?

openstack/neutron/values.yaml
notandy: 👍

openstack/neutron/templates/sftp-deployment.yaml
notandy: 👍

openstack/neutron/templates/sftp-deployment.yaml
```diff
@@ -58,3 +59,5 @@ spec:
...
+
```
sebageek, line 63: [optional] I was looking for a single newline at end of file ([POSIX  3.206](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206)), but I can also live with two.

sebageek (approved):
Optional newline thing, otherwise looks good to me

joker-at-work:
No double approval for changes going to production?

notandy:
> No double approval for changes going to production?

I honestly don't know - doesn't it depend on codeql and other checks? I assume if ruleset is not explictly enabled for double approval, then one is enough.

joker-at-work:
> I assume if ruleset is not explictly enabled for double approval, then one is enough.

I think it's the other way around: If you don't explicitly know of an exception, double-approval is the way to go.

notandy:
> > I assume if ruleset is not explictly enabled for double approval, then one is enough.
> 
> I think it's the other way around: If you don't explicitly know of an exception, double-approval is the way to go.

Thanks, taking it into consideration next time.

PR #11992, opened 2026-06-16
[neutron] Add neutron-agents library chart

notandy:
Helm `type: library` chart that exposes named templates for embedding neutron agents into a consumer's Pod spec, so charts using these agents (archer, neutron, kvm) can share one codebase across local/remote clusters and namespaces.

Structured as a generic utility library rather than a single-purpose one. The pieces every neutron agent shares live in reusable helpers — `neutron-agents.image`, `neutron-agents.configVolume`/`configVolumeMounts`, `neutron-agents.neutron-conf.default`/`.privsep`, and `neutron-agents.logging-conf`. The linuxbridge agent is the first concrete instance layered on top as `neutron-agents.linuxbridge.{container,volumes,configmap-etc,neutron-conf}`. A metadata or dhcp agent can be added as a sibling reusing the shared helpers without forking the config/volume plumbing; those agents are not implemented here.

### DIFF
--- a/openstack/neutron-linuxbridge-agent/templates/_configmap-etc.tpl
+++ b/openstack/neutron-linuxbridge-agent/templates/_configmap-etc.tpl
@@ -1,0 +1,17 @@
+{{/*
+neutron-linuxbridge-agent.configmap-etc renders a full ConfigMap resource.
+Call from a consumer template, e.g.:
+
+  {{- include "neutron-linuxbridge-agent.configmap-etc" (dict "root" . "values" .Values.linuxbridge_agent) }}
+*/}}
+{{- define "neutron-linuxbridge-agent.configmap-etc" -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .root.Release.Name }}-linuxbridge-agent-etc
+data:
+  neutron.conf: |
+    {{- include "neutron-linuxbridge-agent.neutron-conf" . | nindent 4 }}
+  logging.conf: |
+    {{- include "neutron-linuxbridge-agent.logging-conf" . | nindent 4 }}
+{{- end -}}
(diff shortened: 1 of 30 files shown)


openstack/neutron-linuxbridge-agent/templates/_configmap-etc.tpl
```diff
@@ -0,0 +1,17 @@
+{{/*
+neutron-linuxbridge-agent.configmap-etc renders a full ConfigMap resource.
+Call from a consumer template, e.g.:
+
+  {{- include "neutron-linuxbridge-agent.configmap-etc" (dict "root" . "values" .Values.linuxbridge_agent) }}
+*/}}
+{{- define "neutron-linuxbridge-agent.configmap-etc" -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
```
toanju: this currently omits the labels, should these be added?

openstack/neutron-linuxbridge-agent/templates/_configmap-etc.tpl
```diff
@@ -0,0 +1,17 @@
...
+  name: {{ .root.Release.Name }}-linuxbridge-agent-etc
+data:
+  neutron.conf: |
+    {{- include "neutron-linuxbridge-agent.neutron-conf" . | nindent 4 }}
+  logging.conf: |
+    {{- include "neutron-linuxbridge-agent.logging-conf" . | nindent 4 }}
```
toanju: for other services we will need additional config. Any thoughts how we add them (have all agents with all configs / some sort of flags here / ...)?

notandy: what other services?

openstack/neutron-linuxbridge-agent/templates/_configmap-etc.tpl
notandy: We can add them.

PR #12012, opened 2026-06-18
[octavia] gate proxysql native_sidecar init container on secret existence

m-kratochvil:
On first install the proxysql secret does not exist yet, so `$proxysql` is nil and the volumes block is skipped — but the native_sidecar init container is still rendered, referencing `etcproxysql/runproxysql` volumes that don't exist.

(taken from the same fix for manila in #11440)


BenjaminLudwigSAP (approved):
makes sense, thx

velp (approved):

fwiesel (approved):

PR #12023, opened 2026-06-19
fix: audit logs http cleartext tls

natatchikhinashvili1:
changing 
{{ if eq .Values.global.clusterType "metal" -}}
to 
{{ if ne .Values.global.clusterType "scaleout" -}}

TLS on the audit HTTP input was metal-only, so admin/virtual clusters exposed Basic Auth over plain HTTP (the pentest finding). ne scaleout enables TLS on every directly-exposed input and excludes only scaleout, which terminates TLS at its ingress. 

changing
 {{- if .Values.syslog.enabled }}
to
{{- if or .Values.http.enabled .Values.mtls.enabled }}

the cert was only mounted when syslog was enabled, so on syslog-disabled clusters the now-TLS input would have no cert to read and crashloop. Mounting it based on the http/mtls inputs ensures the cert is present wherever TLS is actually used.

### DIFF
--- a/system/audit-logs/Chart.yaml
+++ b/system/audit-logs/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Audit logging components
 name: audit-logs
-version: 0.3.27
+version: 0.3.28
 dependencies:
   - name: fluent-audit-container
     alias: fluent_audit_container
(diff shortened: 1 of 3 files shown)


stwilkes (approved):
lgtm

timojohlo (approved):

PR #12032, opened 2026-06-22
[cc-shoots-compute] Add coredump control for KVM shoots

toanju:
Make coredump configuration configurable instead of depending on region.


anokfireball (dismissed):

joker-at-work (commented):
As the CI points out: version bump missing.

anokfireball (approved):

joker-at-work (approved):

PR #12034, opened 2026-06-22
[neutron-hypervisor-agents] Add DHCP and metadata agent to linuxbridge daemonset

toanju:
This imports the DHCP and metadata agent based on the
[statefulset-network-agent-apod.yaml](https://github.com/sapcc/helm-charts/blob/master/openstack/neutron/templates/statefulset-network-agent-apod.yaml)
of the openstack/neutron helm chart. All required dependencies were
respectively added. In addition, the daemon set dependency to the
kubernetes.metal.cloud.sap/bb label was removed and is replaced by
reading the building block from the hostname.


notandy (commented):
can you please check out https://github.com/sapcc/helm-charts/pull/11992, it's supposed to provide a common ground for all kinds of (reusable) neutron network agent, so we don't start to copy/paste arbitrary files around different charts.

It would make sense to integrate the dhcp/metadata agents into it.

PR #12039, opened 2026-06-22
barbican: add NetworkPolicy to restrict exec and port-forward (SCIBPL-218)

rajivmucheli:
## Summary

- Adds a `NetworkPolicy` (`templates/networkpolicy.yaml`) that limits ingress on `barbican-api` pods to the declared API port (default `9311/TCP`).
- Adds `networkPolicy.enabled: true` and `networkPolicy.additionalIngressPorts: []` to `values.yaml`.
- Bumps chart version `0.8.5 → 0.8.6`.

## Motivation

Addresses **SCIBPL-218** (Restrict Exec and Port Forward Capabilities).

`kubectl exec` and `kubectl port-forward` both require the Kubernetes API server to open a SPDY/WebSocket connection to the target pod on an **arbitrary** ephemeral port. By restricting ingress to the API port only, this NetworkPolicy prevents those tunnels from being established — without requiring a cluster-level admission webhook or PSP.

The policy is enabled by default. Operators can add extra ports (e.g. a metrics scraper on a non-standard port) via `networkPolicy.additionalIngressPorts`.

## Test plan

- [ ] `helm template` renders NetworkPolicy with `networkPolicy.enabled: true` (default)
- [ ] No NetworkPolicy rendered with `networkPolicy.enabled: false`
- [ ] `kubectl exec` into a barbican-api pod is denied after policy is applied
- [ ] `kubectl port-forward` to barbican-api is denied after policy is applied
- [ ] Normal API traffic on port 9311 continues to work

### DIFF
--- a/openstack/barbican/Chart.yaml
+++ b/openstack/barbican/Chart.yaml
@@ -4,7 +4,7 @@ appVersion: flamingo
 description: A Helm chart for Openstack Barbican
 icon: https://www.openstack.org/themes/openstack/images/project-mascots/Barbican/OpenStack_Project_Barbican_vertical.png
 name: barbican
-version: 0.8.5
+version: 0.8.6
 dependencies:
   - condition: mariadb.enabled
     name: mariadb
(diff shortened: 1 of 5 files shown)


JoJoPuppe (dismissed):

drochow:
Can you explain me how this is preventing an K8s exec into the container? 

To my understanding the best way to archive this would be via an AdmissionController / Guardkeeper config.

rajivmucheli:
Closing — SCIBPL-218 follow-up in a separate PR.

PR #12047, opened 2026-06-22
[ironic] switch sentry to utils helper

sandzwerg:
The templates/sentry.yaml makes sure a sentry_dsn secret exists inside
kubernetes. To make use of that we either need to explcitly add a sentry
snippet as before, but would need to update that if something changes
there. Instead we use the utils helper function, which includes that
snippet for us. We just need to add the include at the right places and
get the indentation right.
Keyword for the utils is the "sentry: enabled" which was, back then used
different, but for the utils we need it so we add it again.

Thank you @joker-at-work for the support


sandzwerg:
Fixes the issues that came up as part of #11956

BerndKue (dismissed):

joker-at-work (approved):

BerndKue (approved):

PR #12051, opened 2026-06-22
Update cephcluster.yaml increase scrape interval for eu-de-2 & eu-de-1

sumitarora2786:


system/cc-ceph/values.yaml
```diff
@@ -75,6 +75,9 @@ pool:
 mgr:
   nodeRole: ceph-mon
 
+monitoring:
+  interval: 30
```
crenduchinta88: should be `interval: 30s`

crenduchinta88 (approved):

senolcolak (approved):

defo89 (approved):

PR #12052, opened 2026-06-23
[neutron] support new customdns config file

mutax:
customdns configuration is now handled by its own yaml config file.

This commit supports the old and new configuration to allow a
smooth migration. After updating values we can remove support for
the old config values and enforce non-empty matches when the feature
is enabled.

related: https://github.com/sapcc/neutron/pull/107

### DIFF
--- a/openstack/neutron/ci/test-values.yaml
+++ b/openstack/neutron/ci/test-values.yaml
@@ -118,3 +118,22 @@ imageVersionNetworkAgentDHCP: queens
 imageVersionNetworkAgentLinuxBridge: queens
 imageVersion: queens
 imageVersionEntanglement: queens
+
+customdns:
+  enabled: true
+  matches:
+    - domain_name_prefixes:
+        - external-abc
+      project_ids:
+        - 5dc81c6355ff478188f8fda11a971c41
+      upstream_dns_servers:
+        - 192.0.2.10
+        - 192.0.2.20
+      ednslogging: true
+    - domain_name_prefixes:
+        - external-abcd
+        - external-
+      upstream_dns_servers:
+        - 192.0.2.30
+        - 192.0.2.40
+      ednslogging: false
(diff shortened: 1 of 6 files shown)


openstack/neutron/templates/deployment-rpc-server.yaml
```diff
@@ -92,6 +92,10 @@ spec:
               name: neutron-etc
               subPath: neutron.conf
               readOnly: true
+            - mountPath: /etc/neutron/customdns.yaml
```
sven-rosenzweig, line 96: Are you missing a ``{{- if .Values.customdns.matches }}`` here?

openstack/neutron/ci/test-values.yaml
```diff
@@ -118,3 +118,22 @@ imageVersionNetworkAgentDHCP: queens
 imageVersionNetworkAgentLinuxBridge: queens
 imageVersion: queens
 imageVersionEntanglement: queens
+
+customdns:
+  enabled: true
```
sven-rosenzweig, line 123: Why do we need the enabled here? What I am missing?

mutax: this is still used in the neutron.conf template here to add the whole section
https://github.com/sapcc/helm-charts/blob/930b90fe1f88db7bbef9629ec48dbf847423a475/openstack/neutron/templates/etc/_neutron.conf.tpl#L181

the logic is: when matches are defined we create the config file, but only when enabled is set, we also configure neutron to load the file and enable the feature. 

I currently need that to support the old config -- would like to rethink that after migration.

e.g.:
- when customdns.enabled is true, make matches a mandatory field
- remove enabled and check if customdns: is present instead

openstack/neutron/templates/deployment-rpc-server.yaml
mutax: yes!

sven-rosenzweig (approved):

rhalina (approved):

PR #12053, opened 2026-06-23
[nova/Oho] add service user for oho smoke tests

bashar-alkhateeb:
Openstack hypervisor operator (oho) needs a service used to mange VMs while performing the smoke tests in cc3test/test project. because the openstack hypervisor operator is installed in the compute cluster where we don't have openstack seeder we decided to create the user in nova helm chart
`ccloud_compute_admin` was given to the service user, because the user needs to enable and disable the compute service

Change-Id: Ica0e6cc41b48dfdb62962f4130338e557f9ac5fe

### DIFF
--- a/openstack/nova/templates/seed.yaml
+++ b/openstack/nova/templates/seed.yaml
@@ -101,7 +101,7 @@ spec:
         role: compute_admin
       - project: test@cc3test
         role: member
-      - project: cloud_admin@ccadmin
+      - project: service
         role: cloud_compute_admin
     {{- end }}
     groups:
--- a/openstack/nova/templates/seed.yaml
+++ b/openstack/nova/templates/seed.yaml
@@ -92,6 +92,18 @@ spec:
         role: cloud_image_admin
       - project: service
         role: cloud_objectstore_admin
+    {{- if .Values.global.oho_service_password }}
+    - name: '{{ .Values.global.oho_service_user | default "openstack-hypervisor-operator" | include "resolve_secret" }}'
+      description: "OpenStack hypervisor operator service user"
+      password: '{{ .Values.global.oho_service_password | include "resolve_secret" }}'
+      role_assignments:
+      - project: test@cc3test
+        role: compute_admin
+      - project: test@cc3test
+        role: member
+      - project: cloud_admin@ccadmin
+        role: cloud_compute_admin
+    {{- end }}
     groups:
     - name: administrators
       role_assignments:


openstack/nova/templates/seed.yaml
```diff
@@ -92,6 +92,18 @@ spec:
         role: cloud_image_admin
       - project: service
         role: cloud_objectstore_admin
+    {{- if .Values.global.oho_service_password }}
+    - name: '{{ .Values.global.oho_service_user | default "oho_service_user" | include "resolve_secret" }}'
+      description: Nova Service
```
joker-at-work: Why "Nova Service" and not "OpenStack Hypervisor Operator Smoke tests"?

openstack/nova/templates/seed.yaml
```diff
@@ -92,6 +92,18 @@ spec:
         role: cloud_image_admin
       - project: service
         role: cloud_objectstore_admin
+    {{- if .Values.global.oho_service_password }}
+    - name: '{{ .Values.global.oho_service_user | default "oho_service_user" | include "resolve_secret" }}'
```
notandy: I would call it just `openstack-hypervisor-operator`, oho_service_user is confusing (and every user in the Default domain is a service user)

openstack/nova/templates/seed.yaml
```diff
@@ -92,6 +92,18 @@ spec:
         role: cloud_image_admin
       - project: service
         role: cloud_objectstore_admin
+    {{- if .Values.global.oho_service_password }}
```
notandy, line 95: Any reason to put this into a global scope? I would just render / deduct the vault path directly in helm to interpolate the region into it. And use a binary flag to enable it per region individually.

openstack/nova/templates/seed.yaml
```diff
@@ -92,6 +92,18 @@ spec:
...
+    - name: '{{ .Values.global.oho_service_user | default "oho_service_user" | include "resolve_secret" }}'
+      description: "OpenStack hypervisor operator service user"
+      password: '{{ .Values.global.oho_service_password | include "resolve_secret" }}'
+      role_assignments:
+      - project: test@cc3test
+        role: compute_admin
+      - project: test@cc3test
+        role: member
+      - project: cloud_admin@ccadmin
```
notandy: just noticed that nova just evaluates the role `cloud_compute_admin`, not the assigned project. In this case, I would just assign it to the service project instead.

notandy (changes requested):

openstack/nova/templates/seed.yaml
bashar-alkhateeb: I was trying to follow the same pattern used for defining the nova service user by using the full path in the values file `.Values.global.nova_service_user`.

Additionally, I used the presence of that value as a switch to create the user only in the regions where it is defined (the KVM regions).

On the other hand, we're using a tool that automates the creation of values and Vault secrets to speed up and simplify region deployments. The tool parses the values template to identify secret paths and creates the corresponding secrets in Vault. Because of that, if keeping the full path in the values file isn't a blocker for you or doesn't cause any issues that I'm not aware of, I'd prefer to keep it as it is ?

openstack/nova/templates/seed.yaml
bashar-alkhateeb: correct the operator is not scoped to `cloud_admin@ccadmin ` and it is enough to have this role on any project, but AFAIK we are trying to give `cloud_compute_admin` (cross-domain) role only on `cloud_admin@ccadmin`?

openstack/nova/templates/seed.yaml
bashar-alkhateeb: agree

bashar-alkhateeb: Fixed !

openstack/nova/templates/seed.yaml
notandy: `nova_service_user` might be used by subchart (e.g. utils), which is why it's needs to be in the global scope. `oho_service_password` will likely not used by anything except this seed in the nova chart.

> Additionally, I used the presence of that value as a switch to create the user only in the regions where it is defined (the KVM regions).
> 
> On the other hand, we're using a tool that automates the creation of values and Vault secrets to speed up and simplify region deployments. The tool parses the values template to identify secret paths and creates the corresponding secrets in Vault. Because of that, if keeping the full path in the values file isn't a blocker for you or doesn't cause any issues that I'm not aware of, I'd prefer to keep it as it is ?

I don't want to dictate how we handle vault secrets, but it's more a historic reason why vault paths are defined in value secrets instead of helm charts (because we migrated from cleartext secrets to vault paths). A lot of other charts won't work with this "tool": https://github.com/sapcc/helm-charts/blob/master/openstack/manila/templates/seeds.yaml#L88

But I am fine if handle these secrets.

openstack/nova/templates/seed.yaml
notandy: The operator is never scoping to cloud_admin@ccadmin - and nova doesn't care. So I would just keep it assigned to the service project of the Default domain.

notandy: also just tested it, since the operator is using the service project to scope against, it needs at least one role to be able to authenticate against keystone: 

```
2026-06-25T11:51:32.700-0400	ERROR	setup	unable to create controller	{"controller": "HypervisorMaintenance", "error": "Expected HTTP response code [201 202] when accessing [POST https://identity-3.qa-de-1.cloud.sap/v3/auth/tokens], but got 401 instead: {\"error\":{\"code\":401,\"message\":\"The request you have made requires authentication.\",\"title\":\"Unauthorized\"}}"}
```


```suggestion
      - project: service
```

notandy (changes requested):

notandy (approved):

joker-at-work (approved):

fwiesel (approved):

PR #12062, opened 2026-06-24
bump metal-operator to v0.5.2

xkonni:
contains fix for discovery to pull of mirrored registries


system/metal-operator-dhcp/values.yaml
```diff
@@ -1,4 +1,4 @@
-owner-info:
+-crdsowner-info:
```
defo89: is this intended?

xkonni: no. thanks!

defo89 (approved):

PR #12070, opened 2026-06-25
[manila] add seeds for integration test user

Carthaca:


kpawar-sap (approved):

chuan137 (approved):
@Carthaca Thanks for the PR. Just one reminder. We should disable the switch in qa, because there `shared/hammer/tech-user/qa` is used in integration test.

PR #12079, opened 2026-06-25
glance: nanny also purges the images table

rajivmucheli:
## Problem

`glance-manage db purge` (the only command the glance-nanny runs today) intentionally **skips the `images` table** to mitigate [OSSN-0075](https://wiki.openstack.org/wiki/OSSN/OSSN-0075). From `glance/db/sqlalchemy/api.py::purge_deleted_rows`:

```python
# NOTE(abhishekk): To mitigate OSSN-0075 images records should be
# purged with new ``purge-images-table`` command.
if tbl == 'images':
    continue
```

As a result, soft-deleted image rows accumulate forever and any subsequent

```
os image create … --id <previously-used-uuid> …
```

returns

```
409 Conflict: Image ID <uuid> already exists!
```

(raised at `glance/db/sqlalchemy/api.py:1074` on the PK collision).

### Observed in eu-de-1

```
total_soft_deleted               = 131,334
purgable_now (deleted_at < -14d) = 129,418
oldest deleted_at                = 2019-10-15 10:39:24
newest deleted_at                = 2026-06-25 17:13:28
```

i.e. no row has ever been purged from `images` on this region, even though the nanny has been logging successful hourly `purge` runs for years.

## Fix

Add a second invocation in the nanny loop that calls `glance-manage db purge_images_table` (the dedicated upstream command for this), gated by a new `glance_nanny.db_purge.images_table` value block. Defaults follow upstream's `purge_images_table` defaults (`older_than=180`).

## Changes

- `templates/scripts/_glance-db-consistency-and-purge.sh.tpl`: add a second `if` block invoking
  `glance-manage db purge_images_table --age_in_days $GLANCE_DB_PURGE_IMAGES_TABLE_OLDER_THAN --max_rows $GLANCE_DB_PURGE_IMAGES_TABLE_MAX_NUMBER`.
- `templates/glance-nanny-deployment.yaml`: plumb `GLANCE_DB_PURGE_IMAGES_TABLE_{ENABLED,OLDER_THAN,MAX_NUMBER}` env vars into the `db-consistency-and-purge` container.
- `values.yaml`: add
  ```yaml
  glance_nanny:
    db_purge:
      images_table:
        enabled: true
        older_than: 180
        max_number: 500
  ```

## Rollout

After this lands, regions with a backlog (e.g. eu-de-1 ~129k rows eligible) will start draining 500 rows/hour from `images`. To drain faster, temporarily bump `glance_nanny.db_purge.images_table.max_number`, or run a one-shot:

```bash
kubectl -n monsoon3 exec deploy/glance-nanny -c db-consistency-and-purge -- \
    /var/lib/openstack/bin/glance-manage db purge_images_table --age_in_days 180 --max_rows 10000
```

To keep the previous (no-images-purge) behavior, set `glance_nanny.db_purge.images_table.enabled: false`.

### DIFF
--- a/openstack/glance/templates/glance-nanny-deployment.yaml
+++ b/openstack/glance/templates/glance-nanny-deployment.yaml
@@ -72,6 +72,12 @@ spec:
           value: {{ .Values.glance_nanny.db_purge.older_than | quote }}
         - name: GLANCE_DB_PURGE_MAX_NUMBER
           value: {{ .Values.glance_nanny.db_purge.max_number | quote }}
+        - name: GLANCE_DB_PURGE_IMAGES_TABLE_ENABLED
+          value: {{ .Values.glance_nanny.db_purge.images_table.enabled | quote }}
+        - name: GLANCE_DB_PURGE_IMAGES_TABLE_OLDER_THAN
+          value: {{ .Values.glance_nanny.db_purge.images_table.older_than | quote }}
+        - name: GLANCE_DB_PURGE_IMAGES_TABLE_MAX_NUMBER
+          value: {{ .Values.glance_nanny.db_purge.images_table.max_number | quote }}
         - name: GLANCE_NANNY_INTERVAL
           value: {{ .Values.glance_nanny.interval | quote }}
         {{- if .Values.sentry.enabled }}
--- a/openstack/glance/templates/scripts/_glance-db-consistency-and-purge.sh.tpl
+++ b/openstack/glance/templates/scripts/_glance-db-consistency-and-purge.sh.tpl
@@ -31,6 +31,15 @@ while true; do
         date
         /var/lib/openstack/bin/glance-manage db purge --age_in_days $GLANCE_DB_PURGE_OLDER_THAN --max_rows $GLANCE_DB_PURGE_MAX_NUMBER
     fi
+
+    # `glance-manage db purge` skips the `images` table (OSSN-0075).
+    # Purge it separately so that soft-deleted image rows are removed
+    # and previously-used image UUIDs can be reused.
+    if [ "$GLANCE_DB_PURGE_IMAGES_TABLE_ENABLED" = "True" ] || [ "$GLANCE_DB_PURGE_IMAGES_TABLE_ENABLED" = "true" ]; then
+        echo -n "INFO: purging at max $GLANCE_DB_PURGE_IMAGES_TABLE_MAX_NUMBER deleted rows older than $GLANCE_DB_PURGE_IMAGES_TABLE_OLDER_THAN days from the glance images table - "
+        date
+        /var/lib/openstack/bin/glance-manage db purge_images_table --age_in_days $GLANCE_DB_PURGE_IMAGES_TABLE_OLDER_THAN --max_rows $GLANCE_DB_PURGE_IMAGES_TABLE_MAX_NUMBER
+    fi
     echo -n "INFO: waiting $GLANCE_NANNY_INTERVAL minutes before starting the next loop run - "
     date
     sleep $(( 60 * $GLANCE_NANNY_INTERVAL ))
--- a/openstack/glance/values.yaml
+++ b/openstack/glance/values.yaml
@@ -390,3 +390,9 @@ glance_nanny:
     older_than: 14
     # delete at max number of entries in one run
     max_number: 500
+    # also run `glance-manage db purge_images_table` to purge the `images`
+    # table itself (skipped by `glance-manage db purge` per OSSN-0075).
+    images_table:
+      enabled: true
+      older_than: 180
+      max_number: 500


seb-kro:
The code looks fine to me. 

I am just not sure in how far this change is recommendable, is there really any issue with not purging the image table? 

It is near impossible for anyone to randomly use the UUID of a previously soft-deleted image. So realistically this will only ever happen if someone makes use of an ID that they know may already have existed, whether they do so intentionally or not. And based on the OSSN it seems that even when this is not a malicious attack, it would still cause very surprising and unwanted side effects for any instance still referencing that image. So is this really an issue for users? And if so, will it change anything if all soft-deleted images from the last 6 months are still present?

Additionally, the current number of rows doesn't seem gigantic and there are other purged image tables with even more rows. E.g. for eu-de-1:
```
MariaDB [(none)]> select table_name as 'table',
    ->     table_rows as 'rows'
    -> from information_schema.tables
    -> order by table_rows desc;
+------------------------------------------------------+--------+
| table                                                | rows   |
+------------------------------------------------------+--------+
| image_members                                        | 296748 |
| image_properties                                     | 247033 |
| images                                               | 144388 |
| session_connect_attrs                                | 131072 |
| session_account_connect_attrs                        | 131072 |
```
Although this number may grow exponentially faster in the future. Is this a performance issue in any way?

rajivmucheli:
i had similar observations before creating this PR, i have asked Kai (raised this concern) to comment on this thread.

Urgol:
the use case for reusing an ID is the following
- image created and shared with one or more projects
- minor change to the image would result in a new upload of the image with a new ID
   - optional if an image need to be moved to another owner
- in minimize changes to integration and pipeline the ID should stay the same

This then also means, if someone follows the naming convetion they could create their own ID to manage their images.

My impression was that once the image is deleted the ID will be released, which is not the case. This is why i requested this. 

BR

Kai

seb-kro:
Thanks for the info @Urgol! I'm still not entirely sure whether I completely understand the motivation. Does _minor change_ include potential changes to the actual image data? Or is this only about changing image metadata and ownership? And I guess these changes are meant to be performed also by regular users, not just admin users?

Either way, upon initial investigation and contemplation, on top of the security concerns already mentioned, IMO there are numerous issues as to why we want to prevent the re-use of image IDs with adjusted image data altogether. And if this would be permitted also for regular (non-admin) users, then we cannot control whether and how the image data changes. 

Some reasoning for my objections is in the sections below. 

Presumably this is not sufficient for you, but just in case: depending on your exact use cases, there might already be other solutions:
- For admin users transferring ownership should be possible.
- If this is more about the ability to reference related images across re-uploads with different UUIDs, then aside from the image names, using image tags or image properties might be an option?

Also feel free to reach out to me if you'd like to discuss this further.

# Customer Trust and Experience
Customers will expect image UUIDs to reference immutable image data since this is how  Glance / OpenStack is supposed to behave. Even if we were somehow able to ensure that a _minor change_ of the image data does not break any customer VMs, this would result in a new checksum which affects customer workflows. Customers will need to account for the fact that checksums may change. And once new image data is uploaded, it will be impossible for them to verify the consistency of any image data they downloaded prior unless they persist the checksums themselves.

Additionally, is there even a way to ensure that a _minor change_ does not affect customers VMs? How do we guarantee that a VM created with the old image data still works the exact same way as a VM created with the updated image data? And does not in any way break certain behavior? Because that's what customers rely on e.g. when they reuse an image ID they tested in a dev environment in production. Or simply redeploy to production. 

Do we also adjust public images this way? Who defines what a _minor change_ is? How is this defined? Who ensures that image data is indeed only adjusted by minor changes?  How do we ensure this? 

Especially given that customer trust seems even more crucial now that the product is sold externally and even with a focus on security sensitive environments (sovereign cloud). 

# Other Technical Concerns
I didn't verify this in detail, so some concerns may be incorrect. But at the very least these  topics probably require additional investigation and then would need to be addressed accordingly. Also, this is not meant to be exhaustive, so there are likely further issues.

Each Nova compute service has an image cache that caches images based on the UUID. Depending on the instances running, images may be cached indefinitely without ever being refreshed. We might therefore need a way to update the image cache with changed image data. And even then, small delays might mean that customers cannot be entirely sure which version will be used shortly after a version update.

The image caches may also cause an existing VM to switch from a new image version to an old version (and vice versa) in case of e.g. live migration, cold migration and resize. And it seems that in case of live migrations, VMs with ephemeral boot disks could even mix both versions by loading some blocks before and other blocks after migration.

I believe that Cinder also has similar image caches, although these apparently auto-update on each use and consequently might not require any adjustment.

# Implementation
As already mentioned, I don't think we want to go down this road at all. But just to write it down, some additional thoughts wrt implementation.

Even when setting aside all of the issues above, as-is this PR doesn't solve the issue. Images will still be soft-deleted and with `older_than: 180` will remain in the images table and block re-upload for the same ID for 180 days. So with this approach we would have to 
1. switch to `older_than: 0` 
2. presumably also reduce `interval: 60` so the Nanny runs more frequently than just once every hour

I don't think that we want the Nanny to run that often. At the very least we would probably run the image table purging as a separate job so the remaining tables continue to be purged less frequently. More importantly, I don't like the idea of all the crucial image data being permanently gone immediately upon image deletion. That can make it much harder to troubleshoot and resolve onduty issues.

Hence, I'd probably look into alternative solutions and e.g. adjust the glance API to allow users  to do one of the following:
- overwrite the data of existing images without deletion
- allow deletion of images table data or at least reset of the UUID for deleted images via the image deletion endpoint or maybe a dedicated, separate endpoint

Depending on how exactly we would implement the image update, we might also need to take special care about some details. In case of updating the data without deletion, we should ensure to avoid collisions in the backend (Swift, Ceph) and to prevent data inconsistencies for simultaneously ongoing image downloads. In case of deletion and subsequent recreation, the image will not be available at all for at least a short moment.

Urgol:
@seb-kro Thank you for your reply and detailed summary. I agree that re-using an ID does violate immutablitiy of the ID and also security/compliance. 

My initial understanding was that the reuse of ID was possible and that the ID not being released was a bug. 

A good approach would be that customers perform changes on OS level and re-upload the image. They should implement some versioning and take this into account when building automation and pipelines.

@rajivmucheli I am ok with not being able to reuse the image ID.

PR #12096, opened 2026-06-26
feat(hermes): wire log-router Ceph RGW credentials and cloud-scoped seed

notque:
Grants `hermes@Default` `cloud_objectstore_admin` via `ccadmin/cloud_admin` so that log-router can write audit events to Ceph RGW using Keystone token auth against the Swift-compatible endpoint.

## Context

Logstash (the current audit event writer) uses its own service user with ec2/AWS HMAC credentials and continues to work unchanged. Log-router is the eventual replacement and takes a different approach: it uses the existing `hermes@Default` service account with direct Keystone token auth against Ceph RGW's Swift endpoint — no ec2 credentials needed.

The `cloud_objectstore_admin` role via `ccadmin/cloud_admin` covers both use cases with a single auth setup:
- **Admin path**: all events → `ccadmin/master` bucket (internal/support usage)
- **Customer path**: per-tenant events → customer-provided bucket in their own project (opt-in via the dataplane-config API)

This is the same pattern Keppel uses for cross-tenant Swift/RGW access.

## Changes

**`keystone-seed.yaml`**: when `logRouter.enabled`:
- Adds `monsoon3/ceph-seed` to `requires` (defines the `cloud_objectstore_admin` role)
- Grants `hermes@Default` `cloud_objectstore_admin` on `ccadmin/cloud_admin`

**`log-router-secret.yaml`**: removes `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` — no ec2 credentials with Keystone token auth

**`values.yaml`**: removes `access_key`/`secret_key` from `logRouter.s3` stanza

## Dependencies

Log-router requires a corresponding change to replace its AWS SDK S3 client with a Swift/Keystone-compatible RGW client.

### DIFF
--- a/openstack/hermes/templates/keystone-seed.yaml
+++ b/openstack/hermes/templates/keystone-seed.yaml
@@ -17,6 +17,9 @@ spec:
   - {{ $.Values.global.keystoneNamespace }}/domain-{{ replace "_" "-" . | lower }}-seed
   {{- end}}
   {{- end}}
+  {{- if .Values.logRouter.enabled }}
+  - monsoon3/ceph-seed
+  {{- end}}
 
   roles:
     - name: audit_viewer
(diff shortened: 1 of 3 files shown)


joluc (approved):

LeoZhangZXZ:
also need to remove the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY in the _log-router-utils.tpl

PR #12098, opened 2026-06-29
[ironic] make sure secret is created first

sandzwerg:
Some deployments depend on this secret so we need to create it first. We
do that with a pre-install annotation.

### DIFF
--- a/openstack/ironic/templates/secrets.yaml
+++ b/openstack/ironic/templates/secrets.yaml
@@ -9,6 +9,8 @@ metadata:
     # this secret is needed by the migration job, so it needs to be a
     # pre-upgrade hook with a lower weight than the migration job.
     "helm.sh/hook": "pre-upgrade"
+    # if we build a new region this needs to be there for the other deployments to work
+    "helm.sh/hook": "pre-install"
     "helm.sh/hook-weight": "-10"
     "helm.sh/hook-delete-policy": "before-hook-creation"
 type: Opaque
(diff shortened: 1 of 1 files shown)


openstack/ironic/templates/secrets.yaml
```diff
@@ -9,6 +9,8 @@ metadata:
     # this secret is needed by the migration job, so it needs to be a
     # pre-upgrade hook with a lower weight than the migration job.
     "helm.sh/hook": "pre-upgrade"
```
s10:
```suggestion
    "helm.sh/hook": "pre-install,pre-upgrade"
```

s10 (changes requested):

openstack/ironic/templates/secrets.yaml
```diff
@@ -8,7 +8,9 @@ metadata:
   annotations:
     # this secret is needed by the migration job, so it needs to be a
     # pre-upgrade hook with a lower weight than the migration job.
-    "helm.sh/hook": "pre-upgrade"
+    "helm.sh/hook": "pre-install,pre-upgrade"
+    # if we build a new region this needs to be there for the other deployments to work
+    "helm.sh/hook": "pre-install"
```
sandzwerg:
```suggestion
```

s10 (approved):

BerndKue (approved):

PR #12108, opened 2026-06-30
[nova-hypervisor-agents] Reduce max_queues to 8

joker-at-work:
In benchmarks it showed that using less queues provides better performance. Additionally, having too many queues through all of the interfaces currently blocks live-migration.


toanju (approved):

dhoeller (approved):
LGTM

PR #12138, opened 2026-06-30
creating new file for Dell r860 bios settings for kvm

piyushstva027:
creating new file for Dell r860 bios settings for kvm

### DIFF
--- a/system/metal-settings/Chart.yaml
+++ b/system/metal-settings/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.2.10
+version: 0.2.12
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/system/metal-settings/bios_settings/dell-r860-kvm.yaml
+++ b/system/metal-settings/bios_settings/dell-r860-kvm.yaml
@@ -8,6 +8,18 @@ settingsFlow:
   priority: 15
   settings:
     PxeDev1EnDis: Enabled
+- name: PXEBootWithNIC.Integrated.1.1
+  priority: 20
+  settings:
+    PxeDev1Interface: NIC.Integrated.1-1-1
+    PxeDev1Protocol: IPv4
+    PxeDev1VlanEnDis: Disabled
+    PxeDev2EnDis: Disabled
+    PxeDev3EnDis: Disabled
+    PxeDev4EnDis: Disabled
+    BootMode: Uefi
+    IscsiDev1EnDis: Disabled
+    MemOpMode: OptimizerMode
 - name: R860KVMBiosSettings
   priority: 20
   settings:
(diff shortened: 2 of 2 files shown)


system/metal-settings/bios_settings/dell-r860-kvm.yaml
```diff
@@ -0,0 +1,63 @@
+# BIOS configuration recommendations settings for Dell R860 KVM servers
+settingsFlow:
+- name: enableIntegratedNIC
+  priority: 10
+  settings:
+    IntegratedNetwork1: Enabled
+- name: enablePxeDev1
+  priority: 15
+  settings:
+    PxeDev1EnDis: Enabled
```
nagadeesh-nagaraja, line 10: could you add the pxe related stuff back?

```
- name: PXEBootWithNIC.Integrated.1.1
    PxeDev1Interface: NIC.Integrated.1-1-1
    PxeDev1Protocol: IPv4
    PxeDev1VlanEnDis: Disabled
    PxeDev2EnDis: Disabled
    PxeDev3EnDis: Disabled
    PxeDev4EnDis: Disabled
```

nagadeesh-nagaraja:
you need to update the version param in "sapcc/helm-charts/system/metal-settings/Chart.yaml" file.

system/metal-settings/bios_settings/dell-r860-kvm.yaml
nagadeesh-nagaraja, line 1: I would like to point out that the settings provided here are the only ones which will be tracked..
and other old settings which was removed will be left behind as is.
Hence, I recomend comparing the previous settings file and if any settings is missing in this one which need to be changed we should add it here. 

example:

if we already had a settings 
abc: 123
foo: bar

and now in this PR we are removing this abc
foo: car
xxx: 1234

the metal operator will leave abc:123

if you want it to be changed, you need to include that.  so the settings should be
foo: car
xxx: 1234
abc: 000

system/metal-settings/bios_settings/dell-r860-kvm.yaml
piyushstva027: ok

system/metal-settings/bios_settings/dell-r860-kvm.yaml
piyushstva027: Hi @nagadeesh-nagaraja for this "you need to update the version param in "sapcc/helm-charts/system/metal-settings/Chart.yaml" file." I have updated version from version: 0.2.10 to version: 0.2.11. Is it fine? Could you suggest exact version for version and appVersion?

system/metal-settings/Chart.yaml
nagadeesh-nagaraja, line 1: you need to rebase your PR, there are merge conflicts. 

but the version looks fine (as of now)

system/metal-settings/bios_settings/dell-r860-kvm.yaml
nagadeesh-nagaraja: @piyushstva027 the version looks fine, but needs a rebase. 

but did you double check the dell-default.yaml settings all settings are present in this list? because as above example explained, any missed settings key will left behind (untracked).

system/metal-settings/bios_settings/dell-r860-kvm.yaml
```diff
@@ -0,0 +1,75 @@
...
+    PxeDev1Protocol: IPv4
+    PxeDev1VlanEnDis: Disabled
+    PxeDev2EnDis: Disabled
+    PxeDev3EnDis: Disabled
+    PxeDev4EnDis: Disabled
+    BootMode: Uefi
+    IscsiDev1EnDis: Disabled
+    MemOpMode: OptimizerMode
+- name: R860KVMBiosSettings
+  priority: 20
```
nagadeesh-nagaraja: ah just noticed this... priority needs to be unique.

nagadeesh-nagaraja (approved):

PR #12139, opened 2026-06-30
[manila] opt hypervisor_storage_hana_qa out of capabilities:netapp_is_home

Carthaca:
hypervisor_storage_hana_qa is a flexgroup share type and does not report
the 'netapp_is_home' capability. Keep the cap in the global
share_type_extra_specs so every regional overlay (incl. bool-shorthand
share_types entries) inherits it, and let a per-share-type '__skip__'
sentinel filter the key out for the flexgroup type at render time.


kpawar-sap (approved):
Looks good. how did you determine its "netapp_is_home" issue (missing in reported capabilities) ?

chuan137 (approved):

PR #12148, opened 2026-07-01
[rabbitmq] Support external IPS on calico based clusters

joker-at-work:
In our newer Gardener-based cluster we have calico for networking that doesn't currently [0] support `.spec.externalIPs` for reading pre-defined external IPs. Instead, it requires an annotation to provide the IPs.

Therefore, we no set both `.spec.externalIPs` as well as the `projectcalico.org/loadBalancerIPs`, because having both should work in all clusters as the annotation has no effect in new clusters.

[0] https://github.com/projectcalico/calico/issues/11815

### DIFF
--- a/common/rabbitmq/CHANGELOG.md
+++ b/common/rabbitmq/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 This file is used to list changes made in each version of the common chart rabbitmq.
 
+## 0.25.3 - 2026/07/01
+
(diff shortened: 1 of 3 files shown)


sandzwerg (approved):

toanju (approved):

dhoeller (approved):
LGTM

s10 (approved):

PR #12169, opened 2026-07-02
Add prometheus kvm federate metrics

idafurjes:


openstack/maia/templates/etc/_maia_scrape_config.yaml.tpl
```diff
@@ -241,3 +241,25 @@
...
+  scrape_interval: "{{ $root.Values.prometheus_kvm.scrape_interval }}"
+  scrape_timeout: "{{ $root.Values.prometheus_kvm.scrape_timeout }}"
+  tls_config:
+    cert_file: /etc/prometheus/secrets/prometheus-auth-sso-cert/sso.crt
+    key_file: /etc/prometheus/secrets/prometheus-auth-sso-cert/sso.key
+  static_configs:
+    - targets:
+      {{- range .Values.global.availability_zones -}}
+      - compute-cc-{{ trimPrefix $root.Values.global.region . }}0.{{ $root.Values.global.region }}.{{ $root.Values.global.tld }}
+      {{- end -}}
```
viennaa:
```suggestion
      # currently only pulling in every az with 0, if there is compute-cc-$az1, it needs to be adapted
      {{- range .Values.global.availability_zones }}
      - compute-cc-{{ trimPrefix $root.Values.global.region . }}0-internal.{{ $root.Values.global.region }}.{{ $root.Values.global.tld }}
      {{- end }}
```

viennaa (changes requested):
Thanks a lot for adding the kvm federation. added a comment and removed the trailing `-` to render correctly.

openstack/maia/templates/etc/_maia_scrape_config.yaml.tpl
viennaa:
```suggestion
      # currently only pulling in every az with 0, if there is compute-cc-$az1, it needs to be adapted
      {{- range .Values.global.availability_zones }}
      - compute-cc-{{ trimPrefix $root.Values.global.region . }}0.{{ $root.Values.global.region }}.{{ $root.Values.global.tld }}
      {{- end }}
```

viennaa (approved):

PR #12182, opened 2026-07-03
Add kvm memory statistics metrics

idafurjes:


viennaa (changes requested):
Chart bump please :)

viennaa (approved):
Thanks for providing the new metrics :)

PR #12184, opened 2026-07-03
fix: add default cert for ingress-nginx to avoid nginx fake certificate

natatchikhinashvili1:
Pentest finding: SSL/TLS: Known Untrusted / Dangerous CA (Medium, CVSS 5) 
the virtual ingress-nginx controller has no default cert so requests hitting the NodePort by IP / without SNI get nginx's built-in self-signed "Fake Certificate". 
Real client traffic already gets a valid DigiCert cert via the edge.


goerangudat (approved):
lgtm

PR #12189, opened 2026-07-03
[mariadb] optionally cohost the mariadb and the mariadb-backup pods

vssldmtrv:
Toggle using the newly introduced `.Values.backup_v2.cohost_with_mariadb` flag. Defaulting to `false`.

If the mariadb and the backup pods share a RWO PVC, then the scheduler will refuse to start the second pod with an error like:
```
Warning  FailedAttachVolume  8m2s  attachdetach-controller  Multi-Attach
error for volume "pvc-***" Volume is
already used by pod(s) whatever-mariadb-***-***
```

The pod will bounce around until it ends up scheduled on the same node as the other one. And then it will start.

No need for all that if we can force the two pods to be scheduled on the same node.

NOTE: using the softer `preferred` and not the hard `required` affinity rule to make sure the main pod (mariadb) starts regardless of the state the auxiliary pod (mariadb-backup) finds itself in.

### DIFF
--- a/common/mariadb/CHANGELOG.md
+++ b/common/mariadb/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.40.0 - 2026/07/03
+* optionally cohost the mariadb and the mariadb-backup pods on the same node via the `.Values.backup_v2.cohost_with_mariadb` flag
+* chart version bumped
+
 ## v0.39.0 - 2026/06/22
 * support multiple Ceph S3 backup targets via `global.mariadb.backup_v2.ceph_s3.targets`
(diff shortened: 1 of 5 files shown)


s10:
See related PR https://github.com/sapcc/helm-charts/pull/11977 , the same issues apply

vssldmtrv:
My PR differs though.
- it establishes a symmetric dependency
- the option is entirely opt-in and has to be enabled explicitly
- co-hosting can be enabled for variety of reasons including - but not limited to - the PVC access modes

s10 (approved):

s10:
Indeed, this could be useful. We do mariadb backup with maridb-dump. So hosting mariadb pod near the mariadb-backup pod might be helpful if we want to make it a bit faster.

galkindmitrii (approved):

PR #12201, opened 2026-07-06
updating dell-r860-kvm.yaml file with very few recommended bios settings

piyushstva027:
updating dell-r860-kvm.yaml file with very few recommended bios settings

### DIFF
--- a/system/metal-settings/bios_settings/dell-r860-kvm.yaml
+++ b/system/metal-settings/bios_settings/dell-r860-kvm.yaml
@@ -19,59 +19,14 @@ settingsFlow:
     PxeDev4EnDis: Disabled
     BootMode: Uefi
     IscsiDev1EnDis: Disabled
-- name: SetSysProfile                                                                                                                                        
-  priority: 25                                                                                                                                               
-  settings:                                                                                                                                                  
-    SysProfile: PerfOptimized     # ← set this first to unlock other settings 
-- name: R860KVMBiosSettings
-  priority: 30
-  settings:
-    AcPwrRcvry: Last
-    AcPwrRcvryDelay: Random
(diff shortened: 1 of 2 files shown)


system/metal-settings/bios_settings/dell-r860-kvm.yaml
```diff
@@ -19,59 +19,14 @@ settingsFlow:
...
     TpmSecurity: "On"
-    UncoreFrequency: MaxUFS
-    UpiPrefetch: Enabled
-    WorkloadConfiguration: Balance
-    XptPrefetch: Enabled
+    ProcUncoreFreqRapl: Disabled
+    LlcPrefetch: Enabled
+    AdddcSetting: Disabled
+    MemFrequency: MaxPerf
+    SysProfile: PerfOptimized
```
nagadeesh-nagaraja, line 31: have we checked that sysProfile settings is not blocking any other settings we are using?

`SysProfile: PerfOptimized`

if you think so, then need to move it up to unblock other settings

nagadeesh-nagaraja (approved):

system/metal-settings/bios_settings/dell-r860-kvm.yaml
piyushstva027: HI @nagadeesh-nagaraja  As we discussed in yesterdays KVM Alignment call, we are going step by step, we have reduced many BIOS settings and included only the importnant ones as per document https://github.wdf.sap.corp/cc/baremetal-config/blob/main/dell/poweredge-r860/bios/2.9.4-settings.yml.  SysProfile is already set to PerfOptimized hence no action is required from Metal operator.

nagadeesh-nagaraja (approved):

PR #12220, opened 2026-07-07
dell-r6715: fix PxeDev1Interface

xkonni:
slot5 it is...


BerndKue (approved):

degricar:
actually this needs to be read dynamically. it can be a different port on the next machine of the same model.
take my ack for testing though

degricar (approved):

PR #12224, opened 2026-07-08
hermes: gate Octobus audit forwarding with logstash.octobus.enabled

notque:
## What

Adds a dedicated `logstash.octobus.enabled` flag (default `true`) to the hermes chart, gating only the Octobus HTTP audit output in `_logstash.conf.tpl`.

## Why

We need eu-de-1 and eu-de-3 to stop forwarding audit events to Octobus while continuing to ship those same events to Fortlogs SIEM (via the existing kafka output). Previously, the only lever was `logstash.audit`, which also gates the JDBC/Metis keystone-domain enrichment used by the OpenSearch output — turning it off region-wide would have regressed enrichment for all local audit consumers.

## Behavior

- **Default (`logstash.octobus.enabled: true`)**: chart renders identically to before. No behavior change for any region that doesn't override.
- **Overridden (`logstash.octobus.enabled: false`)**: the Octobus `http { ... }` output block is omitted. The `audit` clone still runs, so the kafka (Fortlogs) output and any future audit outputs continue to receive events.
- `logstash.audit` and `logstash.kafka.enabled` semantics unchanged.

## Verification

`helm template` against `ci/test-values.yaml`:

| Config | `output_octobus_audit` blocks rendered |
|---|---|
| Defaults | 1 ✅ |
| `--set logstash.octobus.enabled=false` | 0 ✅ |

Rendered logstash.conf remains structurally valid in both cases (empty conditional bodies are legal in Logstash config syntax).

## Follow-up (separate repos)

- `eu-de-1` region secrets: set `logstash.octobus.enabled: false`
- `eu-de-3` region secrets: set `logstash.octobus.enabled: false`
- All other regions: no change required (inherit `true` from `values.yaml`)

## Diff size

11 additions, 0 deletions across 2 files.

### DIFF
--- a/openstack/hermes/Chart.yaml
+++ b/openstack/hermes/Chart.yaml
@@ -4,7 +4,7 @@ description: Helm audit management for Openstack
 maintainers:
   - name: notque
 name: hermes
-version: 0.2.5
+version: 0.2.6
 dependencies:
   - name: linkerd-support
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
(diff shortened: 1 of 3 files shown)


ibakshay (commented):
Can you do the version bump please?

notque:
Done — bumped the chart version to `0.2.6` (master is at `0.2.5`). Pushed as 7ae9a8cc42.

PR #12225, opened 2026-07-08
metal-operator-dhcp: support per-pool url override in remote dhcpPools

xkonni:
Allow individual pools under remote.dhcpPools to specify a url that overrides the top-level remote.url for that pool's dhcp-boot entry. Pools without a url fall back to remote.url.


xkonni:
well. or use the local section.

PR #12229, opened 2026-07-08
disco ~ allow remote disco instances to have the mutating webhook con…

asergeant01:
…figuration scoped.


databus23:
Can you elaborate a bit on the intended setup? We are trying to get rid of disco midterm and I would like to understand who this is used in remote cluster scenarios?

asergeant01:
@databus23 Metal Operator creates records on the remote workerless cluster where all its other resources reside. I believe a Disco instance which watches for these records on the remote cluster is required?

PR #12238, opened 2026-07-08
feat(metal-operator-dhcp): bind dnsmasq interface dynamically and persist for stop hook

defo89:
## Summary

When dnsmasq.service=true, the dnsmasq config needs to bind to the node outgoing interface. 

## Changes

- initdnsmasq.sh: Detect outgoing interface, persist to /shared/interface, copy dnsmasq.conf to writable /shared/ volume, sed the __INTERFACE__ placeholder, and start dnsmasq from the modified copy
- dnsmasq.conf: Add interface=__INTERFACE__ placeholder (gated by .Values.dnsmasq.service)
- dhcp-stop.sh: Read interface from /shared/interface file instead of querying ip route; guard with non-empty check and || true to survive partial network teardown
- dhcp-init.sh: Gate ip addr add behind .Values.dnsmasq.service

### DIFF
--- a/system/metal-operator-dhcp/Chart.yaml
+++ b/system/metal-operator-dhcp/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: metal-operator-dhcp
 description: A Helm chart for the Ironcore metal-operator-dhcp.
 type: application
-version: 2.4.0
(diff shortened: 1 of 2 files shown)


xkonni (approved):
local render looks promising.

PR #12251, opened 2026-07-09
update gatekeeper to 3.23.0

VoigtS:
The update to 3.23.0 is low risk and should just work.


system/gatekeeper/vendor/gatekeeper-upstream/templates/namespace-post-upgrade.yaml
```diff
@@ -17,8 +17,6 @@ metadata:
 spec:
   template:
     metadata:
-      annotations:
```
SuperSandro2000, line 20: I think we want to keep this.

VoigtS: good catch. I reverted this change.

SuperSandro2000 (approved):

PR #12256, opened 2026-07-09
openstack/keppel: set OS_INTERFACE correct for ceph

SuperSandro2000:


wagnerd3 (commented):
Are we sure this isn't a temporary flaw where we should ask to change this to the proper ceph URL?

SuperSandro2000:
No, going to clarify that 😅

wagnerd3 (approved):

PR #12272, opened 2026-07-13
i503385-patch-1

piyushstva027:
i503385-patch-1


piyushstva027:
already done https://github.com/sapcc/helm-charts/pull/12239

PR #12284, opened 2026-07-13
fix volume-permissions for fresh PVs

viennaa:
If /var/lib/rabbitmq/.erlang.cookie does not exist the respective chmod command is failing causing the init container to fail.

This happens on empty PVs, does not affect installations where the erlang cookie is present already

### DIFF
--- a/common/rabbitmq/templates/statefulset.yaml
+++ b/common/rabbitmq/templates/statefulset.yaml
@@ -58,7 +58,7 @@ spec:
         - -c
         - |
           chown -R 999:999 /var/lib/rabbitmq
-          [ -f /var/lib/rabbitmq/.erlang.cookie ] && chmod 400 /var/lib/rabbitmq/.erlang.cookie
+          [ -f /var/lib/rabbitmq/.erlang.cookie ] && chmod 400 /var/lib/rabbitmq/.erlang.cookie; true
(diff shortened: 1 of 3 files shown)


common/rabbitmq/templates/statefulset.yaml
```diff
@@ -58,7 +58,7 @@ spec:
         - -c
         - |
           chown -R 999:999 /var/lib/rabbitmq
-          [ -f /var/lib/rabbitmq/.erlang.cookie ] && chmod 400 /var/lib/rabbitmq/.erlang.cookie
+          [ -f /var/lib/rabbitmq/.erlang.cookie ] && chmod 400 /var/lib/rabbitmq/.erlang.cookie; true
```
s10, line 61: Instead of `&&` I think we should've had here:

```
     if [ -f /var/lib/rabbitmq/.erlang.cookie ]; then
       chmod 400 /var/lib/rabbitmq/.erlang.cookie
     fi
```

viennaa: Agreed for better visibility. I was actually thinking about this or the short form and went for the shortcut

s10 (approved):

galkindmitrii (approved):

PR #12295, opened 2026-07-13
Ignore psk for Neutron VPNaaS IPsec Site Conn

sebageek:
We don't want to log the psk of an IPsec Site Connection for operations on this object in the scope of Neutron's VPNaaS plugin.


notque:
I am not an approver. I approve though, and we need this in. This is the correct change.

nikatza (approved):

notque (approved):

PR #12305, opened 2026-07-14
feat(keystone): add redis subchart for osprofiler backend

tz3:
## Summary

- Adds `redis` subchart (alias `osprofiler-redis`) to keystone chart, gated on `osprofiler.redis.enabled`
- Bumps `utils` from `0.32.0` → `0.38.0` to get dynamic `connection_string` support in the osprofiler helm template
- Adds `osprofiler.redis` defaults block to `values.yaml`

## Background

`jaegertracing/jaeger-agent` was removed in Jaeger v2.x (upstream, by the Jaeger maintainers). The last published image is `1.30.0` from 2022 which has accumulated CVEs making it rotten in Keppel. There is no clean version to upgrade to.

This PR enables osprofiler to use Redis as its tracing backend instead, which requires no sidecar injection and uses images already running in qa-de-1 (same `alpine-valkey` family as cinder's rate-limit redis).

## Test plan

- [x] `helm template` renders without errors, no `jaeger-agent` container in output
- [x] `connection_string = redis://keystone-osprofiler-redis:6379` in rendered `secrets.conf`
- [x] All redis subchart images verified present in Keppel (`alpine-valkey:9.1-20260611084111`, `alpine-kubectl:latest-20260616102342`, `redis-exporter:1.86.0-20260709012056`)
- [ ] Secrets PR: cc/secrets#21054

### DIFF
--- a/openstack/keystone/Chart.yaml
+++ b/openstack/keystone/Chart.yaml
@@ -36,7 +36,7 @@ dependencies:
     version: 1.3.5
   - name: utils
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-    version: 0.38.0
+    version: 0.32.0
   - name: redis
     alias: osprofiler-redis
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
--- a/openstack/keystone/Chart.yaml
+++ b/openstack/keystone/Chart.yaml
@@ -36,7 +36,12 @@ dependencies:
     version: 1.3.5
   - name: utils
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-    version: 0.32.0
+    version: 0.38.0
+  - name: redis
+    alias: osprofiler-redis
+    repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
+    version: 2.4.69
+    condition: osprofiler.redis.enabled
   - name: linkerd-support
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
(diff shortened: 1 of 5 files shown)


openstack/keystone/Chart.yaml
```diff
@@ -36,7 +36,12 @@ dependencies:
     version: 1.3.3
   - name: utils
     repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
-    version: 0.32.0
+    version: 0.38.0
```
bbobrov: Is update of utils to 0.38.0 required for redis? If yes, it should stay here. If no, it should be moved to a separate PR.

bbobrov (changes requested):
Please confirm that you have tested it in qa manually

openstack/keystone/Chart.yaml
tz3: got it revert, will test it in. QA manually and verify it.

tz3:
@bbobrov Tested manually in qa-de-1 — all pods up, all works.

bbobrov:
I have just talked to @tz3 about this, and we agreed that the password will get generated on the fly and will be temporary; we shall not store it in vault. The chart now needs to be changed accordingly.

bbobrov:
Please confirm that the new commits also were checked in qa

tz3:
Confirmed

bbobrov (dismissed):

rajivmucheli (dismissed):

rajivmucheli (approved):

PR #12306, opened 2026-07-14
increase limitas v2 scrape interval to 300s

VoigtS:
The metrics contain a few very expensive DB queries (f.e. towards project_az_resources which has 3mil. rows in a expensive region), this increases the scrape interval to reduce the DB load.


VoigtS:
60s are the default scrape interval, we keep it as that for now and look for alternatives.

PR #12322, opened 2026-07-16
Adding retry option for bios\bmc version and settings

rezash84:
Adding retry option for bios\bmc version and settings


nagadeesh-nagaraja (dismissed):

system/metal-settings/values.yaml
nagadeesh-nagaraja, line 76: do you want to add retry for biosversion as well?

rezash84: done

nagadeesh-nagaraja (approved):

PR #12327, opened 2026-07-17
[rabbitmq] skip linkerd for TLS listener

mutax:
Add linkerd skip-inbound-ports annotation to skip the TLS listener port.

    We do not need to push TLS traffic through linkerd. Also linkerd hides
    the real client IP from the rabbitmq server, which is essential, e.g.
    when debugging which clients are still using old credentials.

### DIFF
--- a/common/rabbitmq/CHANGELOG.md
+++ b/common/rabbitmq/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 This file is used to list changes made in each version of the common chart rabbitmq.
 
+## 0.25.7 - 2026/07/17
+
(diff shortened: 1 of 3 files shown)


fwiesel (dismissed):

galkindmitrii:
lgtm, but please update the changelog: https://github.com/sapcc/helm-charts/blob/master/common/rabbitmq/CHANGELOG.md

joker-at-work (changes requested):
CHANGELOG.md entry missing

galkindmitrii (approved):

joker-at-work (approved):

PR #12335, opened 2026-07-17
add chorus helm chart

arttor:
Based on https://github.com/cobaltcore-dev/cloud-storage/issues/517 - add chorus helm chart for Concourse pipelines

### DIFF
--- a/system/chorus/values.yaml
+++ b/system/chorus/values.yaml
@@ -1,0 +1,182 @@
+owner-info:
+  support-group: containers
+  service: ceph
+  helm-chart-url: https://github.com/sapcc/helm-charts/tree/master/system/chorus
+
+chorus:
+  dynamicCredentials:
+    enabled: false
+    disableEncryption: false
+    pollInterval: 3s
+    masterPassword: ""
+    existingSecret: ""
+    existingSecretKey: "master-password"
+
+  storage:
+    # Name of the main storage (must match a key in storages below)
+    main: ""
+
+    # Storage definitions
+    # For S3: same config works for both proxy and worker
+    # For Swift: worker uses authURL (Keystone), proxy uses storageURL
+    storages: {}
+      # Example S3 storage:
+      # main:
+      #   type: S3
+      #   address: s3.example.com
+      #   provider: Ceph  # Ceph, Minio, or Other
+      #   isSecure: true
+      #   defaultRegion: "us-east-1"
+      #   healthCheckInterval: 10s
+      #   httpTimeout: 1m
+      #   rateLimit:
+      #     enabled: false
+      #     rpm: 60
+      #
+      # Example Swift storage:
+      # swift-main:
+      #   type: SWIFT
+      #   # Worker config - uses Keystone for authentication
+      #   authURL: http://keystone:5000/v3
+      #   storageEndpointName: swift
+      #   storageEndpointType: object-store
+      #   region: RegionOne
+      #   # Proxy config - direct Swift endpoint (proxy cannot use Keystone)
+      #   storageURL: http://swift:8080/v1
+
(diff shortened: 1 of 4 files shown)


system/chorus/values.yaml
```diff
@@ -0,0 +1,182 @@
...
+    image:
+      repository: keppel.eu-de-1.cloud.sap/ccloud/cobaltcore-dev/cobaltcore-storage/chorus/proxy
+      tag: "" # Defaults to Chart.appVersion
+      pullPolicy: IfNotPresent
+
+  # =============================================================================
+  # WORKER
+  # =============================================================================
+  worker:
+    replicas: 2
```
jrse, line 124: are 2 replicas enough? is there information available how fast we sync with 2 replicas?

system/chorus/values.yaml
```diff
@@ -0,0 +1,182 @@
...
+    auth:
+      enabled: true
+      # Leave empty to auto-generate password, or specify existing secret name
+      existingSecret: ""
+      existingSecretPasswordKey: redis-password
+    # Persistence (recommended for production):
+    master:
+      persistence:
+        enabled: true
+        size: 64Gi
```
jrse, line 107: which storage class is used and how long do the messages stay in redis queue?

system/chorus/values.yaml
arttor: this is just example. we can scale it horisontally with numbers of replicas or vertically by increasing resources.
i think that concrete values should be set for every region based on expected load. maybe we can do some load tests and monitor cpu/memory metrics on QA.

when i did initial benchmarks for chorus, i was always constrained by network bandwidth, not memory or cpu.

system/chorus/values.yaml
arttor: i dont know which storage classes are used in regions.  tasks are removed from the queue as soon as they processed (object is replicated) - no task retention. So queue size depends on workers capacity.
we have some estimation for redis memory consumption based on unprocessed tasks: 
>memory: 1M migrated obj ~ 105MB, 1M tasks in the queue ~ 700MB

see: https://docs.clyso.com/docs/products/chorus/operations/#work-queue-redis

senolcolak:
I can't commit the changes here.. can you please add this file will be needed to inject the secrets

system/chorus/templates/secret.yaml
```
apiVersion: v1
kind: Secret
metadata:
  name: chorus-secrets
type: Opaque
data:
  master-password: {{ required "chorus.dynamicCredentials vault path required" .Values.chorus.dynamicCredentials.masterPassword | b64enc | quote }}
  redis-password: {{ required "chorus.redis vault path required" .Values.chorus.redis.auth.password | b64enc | quote }}
```

arttor:
> I can't commit the changes here.. can you please add this file will be needed to inject the secrets
> 
> system/chorus/templates/secret.yaml
> 
> ```
> apiVersion: v1
> kind: Secret
> metadata:
>   name: chorus-secrets
> type: Opaque
> data:
>   master-password: {{ required "chorus.dynamicCredentials vault path required" .Values.chorus.dynamicCredentials.masterPassword | b64enc | quote }}
>   redis-password: {{ required "chorus.redis vault path required" .Values.chorus.redis.auth.password | b64enc | quote }}
> ```

i will add but is can we also reuse secrets from chorus helm chart? if one of these params are set in values.yaml, chorus helm chart also stores it insecret:
- for dynamic creds password: https://github.com/clyso/chorus/blob/90d8890e81b5f44e54807fd50346b8aa8abd4a80/deploy/chorus/templates/secrets/secret.yaml#L27-L42
- and here for redis: https://github.com/clyso/chorus/blob/main/deploy/chorus/templates/secrets/redis-secret.yaml

PR #12339, opened 2026-07-20
[percona_cluster] try to fix IST

galkindmitrii:
When percona tries to do an IST on restart with the cluster it fails because the process cannot bind to K8s external service IP.

We need to set make it bind to 0.0.0.0 for IST specifically.


proddi (dismissed):

proddi (approved):

s10 (approved):

s10:
@bashar-alkhateeb , could you please approve this PR?

mbrowny (approved):

PR #12341, opened 2026-07-20
keppel: workaround broken Ceph API

majewsky:
Needs sapcc/keppel#754 merged first.

The new role assignment is already active in eu-de-3.


VoigtS (dismissed):

SuperSandro2000 (dismissed):

openstack/keppel/templates/seed.yaml
```diff
@@ -78,6 +78,8 @@ spec:
             - user: keppel@Default
```
SuperSandro2000, line 78: Can we delete the comment in line 69?

SuperSandro2000 (approved):

PR #12343, opened 2026-07-21
openstack/keystone: expand histogram buckets

tz3:
## Summary

Histogram buckets for `openstack_watcher_api_requests_duration_seconds` were capped at 2.5s, rendering `histogram_quantile` blind to slow requests above that threshold.

This is a prerequisite for future latency alerting — expanded buckets allow us to observe real latency distributions during the next outage and set grounded alert thresholds.

## Changes

**templates/configmap-etc.yaml**
- Expand statsd-exporter histogram buckets: `[.025, .1, .25, 1, 2.5]` → `[.025, .1, .25, 1, 2.5, 5, 10, 15, 30]`

## Test plan

- [x] helm template the keystone chart and confirm the configmap renders correctly
- [x] After merge, verify `openstack_watcher_api_requests_duration_seconds_bucket` has data in the new buckets in Prometheus


openstack/keystone/alerts/openstack/keystone.alerts
```diff
@@ -90,3 +90,43 @@ groups:
...
+      irate(openstack_watcher_api_requests_duration_seconds_sum{service="identity"}[2m])
+      /
+      irate(openstack_watcher_api_requests_duration_seconds_count{service="identity"}[2m])
+      > 10
+    for: 2m
+    labels:
+      context: latency
+      meta: Keystone API mean response time exceeded 10s
+      service: keystone
+      severity: critical
```
bbobrov: If we had this this metric, how often would the alert fire in the last 4 weeks?
We must understand this, because a critical alert leads to a call, and someone should only be called when things are really bad.

tz3 (commented):
I ran a 4-week backtest with the alert expression. The original version fired per-pod rather than at service level — individual pods showed brief spikes but that's not indicative of real degradation.

Fixed in fbb8f0ebcd: both alerts now aggregate with `sum()` across all pods so they reflect the service-level mean.

openstack/keystone/alerts/openstack/keystone.alerts
tz3: I ran a 4-week backtest with the alert expression. The original version fired per-pod rather than at service level — individual pods showed brief spikes but that's not indicative of real degradation.

Fixed in https://github.com/sapcc/helm-charts/commit/fbb8f0ebcd25bb237d0e6e4ead8e33cac8bd58c7: both alerts now aggregate with sum() across all pods so they reflect the service-level mean.

bbobrov: I still don't understand. How many times would the alert fire in the last 4 weeks? Please provide a number.

tz3: 0 times. Checked it again against prometheus-openstack.eu-de-1.cloud.sap over the last 4 weeks (Jun 27 – Jul 24). 
The service-level mean latency peaked at ~0.16s. The 10s threshold was never reached.

bbobrov: How can we then make sure that this metric works and that the alert will fire eventually?

tz3: Added promtool unit tests in the latest commit [3660bce](https://github.com/sapcc/helm-charts/pull/12343/commits/3660bcefa7c044f73ad3607c713c10f59bf0d939) — they verify the alert fires when the mean crosses the threshold and stays silent on isolated slow requests.

bbobrov:
We just talked about this in a call. We decide to split this into 2 parts - the buckets and the alerting. Alerts should not be added yet, because we are not sure about the exact behavior of them, and not sure that they will fire _before_ the incident and not after. We will look at the metrics during the next outage and see whether they would, and if yes, add them.

bbobrov (approved):

rajivmucheli (approved):

PR #12346, opened 2026-07-21
system/perses: fix persesDashboard alert label to use absolute URL

Frazia04:
## Problem
Some Slack alert  currently link to a relative path `observability/dashboards/perses-overview`, which Slack cannot resolve.

## Fix in this PR
Set `greenhouse.alertLabels.persesDashboard` in `system/perses/values.yaml` to an absolute URL, using the same `{{ .Values.global.region }}` / `{{ .Values.global.tld }}` template expressions that all other charts in this repo already use (see e.g. `openstack/hermes/alerts/openstack/api.alerts.tpl`).

### DIFF
--- a/system/perses/Chart.lock
+++ b/system/perses/Chart.lock
@@ -1,12 +1,12 @@
 dependencies:
 - name: perses
   repository: oci://ghcr.io/cloudoperators/greenhouse-extensions/charts
-  version: 0.24.0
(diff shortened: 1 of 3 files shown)


ibakshay:
Please also bump the chart.yaml :)

ibakshay (approved):

PR #11677, opened 2026-05-14
[kubernikus-rbac] remove abac-reader cluster role

s10:
* Remove now-unused `monsoon3:cluster-reader` ClusterRole
* Remove `monsoon3:abac-reader` ClusterRoleBinding

PR #12349, opened 2026-07-21
[octavia] convert db-migration job to regular resource

s10:
On first install the release SA is not applied yet, so the hook pod runs under `monsoon3/default`. The entrypoint endpointslices check then returns 403.

* drop hook annotations
* guard SA on `.Values.rbac.enabled`
* update api/worker/housekeeping DEPENDENCY_JOBS

### DIFF
--- a/openstack/octavia/templates/_helpers.tpl
+++ b/openstack/octavia/templates/_helpers.tpl
@@ -34,3 +34,20 @@ Create chart name and version as used by the chart label.
 {{- define "octavia.db_service" }}
   {{- include "utils.db_host" . }}
 {{- end }}
+
+{{- define "job_name" }}
+  {{- $name := index . 1 }}
+  {{- with index . 0 }}
+    {{- $all := list
+          (include (print .Template.BasePath "/octavia-etc-configmap.yaml") .)
+          (include (print .Template.BasePath "/secrets.yaml") .)
+          (include "utils.proxysql.job_pod_settings" .)
+          (include "utils.proxysql.volume_mount" .)
+          (include "utils.proxysql.container" .)
+          (include "utils.proxysql.volumes" .)
+          (tuple . (dict) | include "utils.snippets.kubernetes_entrypoint_init_container")
+      | join "\n" }}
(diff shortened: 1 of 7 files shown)


m-kratochvil (dismissed):

s10:
@velp @BenjaminLudwigSAP @notandy @fwiesel , could you please take a look at this PR?

openstack/octavia/templates/_helpers.tpl
```diff
@@ -34,3 +34,20 @@ Create chart name and version as used by the chart label.
...
+    {{- $all := list
+          (include (print .Template.BasePath "/octavia-etc-configmap.yaml") .)
+          (include (print .Template.BasePath "/secrets.yaml") .)
+          (include "utils.proxysql.job_pod_settings" .)
+          (include "utils.proxysql.volume_mount" .)
+          (include "utils.proxysql.container" .)
+          (include "utils.proxysql.volumes" .)
+          (tuple . (dict) | include "utils.snippets.kubernetes_entrypoint_init_container")
+      | join "\n" }}
+    {{- $hash := empty .Values.proxysql.mode | ternary "" $all | sha256sum }}
```
notandy: the job_name helper uses ternary "" $all on line 50, so when proxysql.mode is unset the hash is always sha256sum("") — a constant. This produces a double-dash in the job name and means config-only changes (to octavia-etc-configmap.yaml or secrets.yaml) won't trigger a new job. [Nova](https://github.com/sapcc/helm-charts/blob/fbe5b4a314d093642737a009687aea8fef199187/openstack/nova/templates/_helpers.tpl#L47-L51) avoids this by using $bin as the fallback instead of "".

notandy (changes requested):

openstack/octavia/templates/_helpers.tpl
s10: Fixed, octavia doesn't use `$bin` migration script, the migration is in Job body itself, so this fallback isn't needed at all

notandy (approved):

vssldmtrv (approved):

PR #12350, opened 2026-07-21
[manila] convert db-migration job to regular resource

s10:
On first install the release SA is not applied yet, so the hook pod runs under `monsoon3/default`. The entrypoint endpointslices check then returns 403.

* drop hook annotations
* update api/scheduler DEPENDENCY_JOBS

### DIFF
--- a/openstack/manila/Chart.yaml
+++ b/openstack/manila/Chart.yaml
@@ -10,7 +10,7 @@ maintainers:
 name: manila
 sources:
   - https://github.com/sapcc/manila
-version: 0.7.0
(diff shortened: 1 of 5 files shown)


chuan137:
On the first deployment, the migration job will run without being blocked?

s10:
> On the first deployment, the migration job will run without being blocked?

Yes, because after this PR it's applied as a normal resource and not as a hook, it has all the resources needed present at the moment of deployment, as they are rendered and applied together.

Carthaca (approved):

PR #12351, opened 2026-07-21
[barbican] convert db-migration job to regular resource

s10:
On first install the release SA is not applied yet, so the hook pod runs under `monsoon3/default`. The entrypoint endpointslices check then returns 403, and the deployment fails.

* drop hook annotations
* guard SA on `.Values.rbac.enabled`


rajivmucheli (approved):
Conflict resolved (chart bumped to 0.8.10 over master's 0.8.9). Fix is valid — the hook race on first install is a real issue and this matches the nova/cinder pattern.

rajivmucheli (approved):

JoJoPuppe (approved):

PR #12354, opened 2026-07-21
federation job for castellum

viennaa:
This enables castellum to entirely avoid the thanos query and shifts all load against prometheus-storage.


trouaux (approved):

majewsky (approved):
❤️

PR #12356, opened 2026-07-21
system/gatekeeper-sap-policies: add chart for SAP-specific gatekeeper policies

mikolajkucinski:


majewsky (changes requested):
As discussed in Slack, I would like to avoid code duplication between this chart and `system/gatekeeper`, since it is likely that the legacy chart will remain in use for quite a while. I have opened #12458 to move the existing policies into a reusable library chart. Once that is merged, please rebase this PR to use the new library chart.

system/gatekeeper-sap-policies/templates/constraint-outdated-mirrors.yaml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  enforcementAction: {{ .Values.policies.outdatedMirrors.enforcementAction }}
+  match:
+    kinds:
+      - apiGroups: [""]
+        kinds: ["Pod"]
+      - apiGroups: ["apps"]
+        kinds: ["DaemonSet", "Deployment", "ReplicaSet", "StatefulSet"]
+      - apiGroups: ["batch"]
+        kinds: ["CronJob", "Job"]
```
majewsky: This should use the existing helper function to make it easy to extend this match to new kinds in the future:
```suggestion
  match: {{ include "gatekeeper-policies.match-pods-and-pod-owners" . | indent 4 }}
```

system/gatekeeper-sap-policies/templates/constraint-prometheusrule-alert-labels.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if and .Values.policies.prometheusruleAlertLabels.enabled (ne .Values.cluster.type "postfix") }}
```
majewsky, line 2: This is dropping the important documentation comment about why the postfix clusters are excluded.
```suggestion
{{/* The postfix clusters do not have prometheus-operator at the moment. */}}
{{- if and .Values.policies.prometheusruleAlertLabels.enabled (ne .Values.cluster.type "postfix") }}
```

But arguably, the match on postfix clusters should be removed entirely, and the `policies.prometheusruleAlertLabels.enabled` value should be set accordingly instead.

And even that is not my actual recommendation. The reason that this constraint needs to be disabled if prometheus-operator is not deployed is that `kind: PrometheusRule` is not defined, and thus the match selector below would be invalid. Since prometheus-operator comes in via Greenhouse, this constraint (and its template) should be conditional on whether the Greenhouse plugin for prometheus-operator is deployed. (Disclaimer: I don't know whether Greenhouse can actually express such conditions.)

system/gatekeeper-sap-policies/templates/constraint-region-value-mismatch.yaml
```diff
@@ -0,0 +1,21 @@
...
+spec:
+  enforcementAction: {{ .Values.policies.regionValueMismatch.enforcementAction }}
+  match:
+    kinds:
+      - apiGroups: [""]
+        kinds: ["Secret"]
+    labelSelector:
+      matchExpressions:
+        - { key: "owner", operator: "In", values: [ "helm" ] }
+        - { key: "status", operator: "In", values: [ "pending-install", "pending-upgrade", "deployed" ] }
```
majewsky: This should use the existing helper function that documents all the specific choices that went into this expression:
```suggestion
  match: {{ include "gatekeeper-policies.match-active-helm-releases" . | indent 4 }}
```

system/gatekeeper-sap-policies/templates/constraint-region-value-missing.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  enforcementAction: {{ .Values.policies.regionValueMissing.enforcementAction }}
+  match:
+    kinds:
+      - apiGroups: [""]
+        kinds: ["Secret"]
+    labelSelector:
+      matchExpressions:
+        - { key: "owner", operator: "In", values: [ "helm" ] }
+        - { key: "status", operator: "In", values: [ "pending-install", "pending-upgrade", "deployed" ] }
```
majewsky: Same here.

system/gatekeeper-sap-policies/templates/constraint-restrict-monsoon3-namespace.yaml
```diff
@@ -0,0 +1,33 @@
...
+spec:
+  enforcementAction: {{ .Values.policies.restrictMonsoon3Namespace.enforcementAction }}
+  match:
+    kinds:
+      - apiGroups: [""]
+        kinds: ["Secret"]
+    labelSelector:
+      matchExpressions:
+        - { key: "owner", operator: "In", values: [ "helm" ] }
+        - { key: "status", operator: "In", values: [ "pending-install", "pending-upgrade", "deployed" ] }
```
majewsky: Same here.

system/gatekeeper-sap-policies/Chart.yaml
```diff
@@ -0,0 +1,17 @@
...
+keywords:
+  - gatekeeper
+  - opa
+  - policy
+  - admission-control
+sources:
+  - https://github.com/sapcc/helm-charts/tree/master/system/gatekeeper-sap-policies
+dependencies:
+  - name: gatekeeper-policies
+    repository: file://../../common/gatekeeper-policies # TODO: move this reference to ccloud-helm once the initial PR adding this chart is merged
```
majewsky: It is now merged:
```suggestion
    repository: oci://keppel.eu-de-1.cloud.sap/ccloud-helm
```
Actually, can you please also do the same change in `system/gatekeeper/Chart.yaml` (and then run `helm dep up` on both charts to regenerate the Chart.lock)?

majewsky (changes requested):
Last batch of comments, then this can be merged:

majewsky (approved):

PR #12381, opened 2026-07-22
bump prometheus-server version under openstack-prometheus

ashifnihalb:


ibakshay:
You also need to execute `helm dependency update` command, right?

ashifnihalb:
> You also need to execute `helm dependency update` command, right?

yes @ibakshay i missed it, updated now

ibakshay (approved):

PR #12397, opened 2026-07-23
cinder: support per-rule enabled flag for ProxySQL query cache

hemna:
## Problem
ProxySQL's query cache is **TTL-only with no write-invalidation**. The `volume_types` cache rules (added in b204d9d0fe, 2026-06-08) cause a volume-type **create-then-read to return 404**: the row is committed to MariaDB, but cinder's read-back `SELECT ... FROM volume_types WHERE name=...` is served from a stale cache snapshot until the TTL (5 min) expires.

This has broken **every cinder tempest volume-type admin test in qa-de-1** since 2026-06-08 (qa-de-1-tempest builds #207–#217, all failed). Verified by live reproduction: create returns 404 while `volume type list` shows the row present; flushing the ProxySQL query cache makes lookups correct; ProxySQL `Query_Cache_count_GET_OK` rises while `SET` stays flat (served from cache).

## Change
Add an optional `enabled: false` on each query rule in the shared ProxySQL template. Rules default to enabled (backward compatible — existing configs unchanged). This lets a region that creates volume types at runtime (e.g. tempest) omit the `volume_types` cache while keeping `attachment_specs`.

Rendered output verified: a rule with `enabled: false` is omitted; rules without the key are kept.

## Rollout
- This PR: template + default values documentation only. No behavior change in any region by default.
- Companion **cc/secrets** PR disables the two `volume_types` rules for **qa-de-1 only**.

Note: Helm replaces (not merges) lists, so the qa-de-1 override re-declares the full `query_rules` list with the two `volume_types` entries set `enabled: false`.

### DIFF
--- a/openstack/cinder/values.yaml
+++ b/openstack/cinder/values.yaml
@@ -166,11 +166,17 @@ proxysql:
   backup:
     max_connections_per_proc: 10  # backup pods get their own proxysql secret with max_connections = 10
   query_rules:
+    # Each rule supports an optional `enabled: false` to omit it (default: enabled).
+    # Note: Helm replaces lists, so a region override must re-declare the full list.
     # Cache attachment_specs lookups (table is empty, 0 rows ever, queried 1249/sec)
     # 1h TTL — table has never had data in any region, feature is unused
     - cache_ttl: 3600000
       match_pattern: "^SELECT.*FROM attachment_specs WHERE"
     # Cache volume_types full list (3 rows, static admin config, rarely changes)
+    # WARNING: ProxySQL query cache is TTL-only with no write-invalidation; a
+    # create/update/delete of a volume type is invisible until the TTL expires.
+    # Disable (enabled: false) in any region that creates volume types at runtime
+    # (e.g. tempest), otherwise create-then-read returns 404. See qa-de-1 override.
     - cache_ttl: 300000
       match_pattern: "^SELECT.*FROM volume_types.*ORDER BY"
     # Cache volume_types by ID or name lookup
--- a/openstack/utils/templates/snippets/_proxysql.cfg.tpl
+++ b/openstack/utils/templates/snippets/_proxysql.cfg.tpl
@@ -180,6 +180,7 @@ mysql_query_rules =
         match_pattern = "^SELECT 1$",
     },
 {{- range $i, $rule := .global.Values.proxysql.query_rules | default list }}
+{{- if ne $rule.enabled false }}
     {
         rule_id = {{ add 1 $i }},
         active = 1,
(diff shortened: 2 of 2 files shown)


fwiesel (approved):

openstack/utils/templates/snippets/_proxysql.cfg.tpl
```diff
@@ -180,6 +180,7 @@ mysql_query_rules =
         match_pattern = "^SELECT 1$",
     },
 {{- range $i, $rule := .global.Values.proxysql.query_rules | default list }}
+{{- if ne $rule.enabled false }}
```
s10, line 183: Could you please bump `openstack/utils` version with this change

joker-at-work (commented):
I don't like that the commit message talks about changes to "cinder:" while it's really a change to the utils chart (and some comments in the Cinder chart).

As Vlad pointed out, that change is not even used by Cinder, because there was no version bump of the utils chart and thus Cinder cannot even see the changes (charts will only get uploaded to keppel once per version).

openstack/cinder/values.yaml
```diff
@@ -166,11 +166,17 @@ proxysql:
   backup:
     max_connections_per_proc: 10  # backup pods get their own proxysql secret with max_connections = 10
   query_rules:
+    # Each rule supports an optional `enabled: false` to omit it (default: enabled).
+    # Note: Helm replaces lists, so a region override must re-declare the full list.
```
joker-at-work, line 170: Why does it have to re-declare the full list with `enabled: false` if the list gets replaced? Wouldn't it suffice to replace _this_ list with one having only a single query?

hemna: You are right — Helm replaces lists, so the qa-de-1 override can just declare the shorter list (with only the `attachment_specs` rule) and the volume_types rules are simply not rendered. That means no template change is needed at all. Closing this PR; the fix is now solely in cc/secrets (PR 21283, updated to declare only the `attachment_specs` rule).

openstack/utils/templates/snippets/_proxysql.cfg.tpl
hemna: Thanks — but per Vlad's comment on line 170, the template change turns out to be unnecessary: cc/secrets can just declare a shorter `query_rules` list (Helm replaces lists on override) and the volume_types rules simply won't be rendered. Closing this PR, so no version bump needed. The fix is now solely in cc/secrets PR 21283.

hemna:
Closing per @joker-at-work's review comment on line 170.

Since Helm replaces (not merges) lists on override, a region can drop rules from `proxysql.query_rules` simply by declaring a shorter list in its cc/secrets values. No template change or `enabled: false` flag is required.

The fix has moved entirely to **cc/secrets PR 21283** (`cinder/qa-de-1: drop ProxySQL volume_types query cache`), which now declares `query_rules` with only the `attachment_specs` rule — the two `volume_types` rules from the chart default simply won't be rendered.

Thanks @fwiesel for the approval, @s10 and @joker-at-work for the review.

PR #12411, opened 2026-07-27
keystone: Apache workers metrics

bbobrov:
- Worker metrics: a sampler loaded into each mod_wsgi daemon via WSGIImportScript reports its active_requests (0/1 with threads=1) to the existing statsd-exporter; summing across pids gives the pool-wide busy-worker count. Also fixes statsd-exporter.yaml being gated on watcher.enabled instead of api.metrics.enabled.
- Timing logs: adds the mod_wsgi request_start/queue_start/script_start timestamps to all four access-log formats (queue wait = script_start - queue_start), logged as strings so unset stays valid JSON.
- Double-scrape fix: removes the Service-level Prometheus scrape annotations so pod-sd is the sole scraper (was doubling every statsd metric); pod scrape annotation re-gated on api.metrics.enabled.

### DIFF
--- a/openstack/keystone/templates/bin/_wsgi_sampler.py.tpl
+++ b/openstack/keystone/templates/bin/_wsgi_sampler.py.tpl
@@ -1,0 +1,68 @@
+# SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company
+# SPDX-License-Identifier: Apache-2.0
+
+# WSGI worker-pool sampler.
+#
+# This module is loaded into every mod_wsgi daemon process via
+# WSGIImportScript. Because it runs *inside* the daemon interpreter, it can
+# call mod_wsgi.process_metrics() to read that process's own state. Each
+# process pushes its current "active_requests" (0 or 1, since keystone runs
+# with threads=1) to statsd as a gauge labelled by pid. The statsd-exporter
+# sidecar turns these into Prometheus series; summing across pids yields the
+# pool-wide busy worker count, and (processes - busy) yields the free count.
+
+import os
+import socket
+import threading
+import time
+
+try:
+    import mod_wsgi
+except ImportError:
+    mod_wsgi = None
+
+
+# How often (seconds) each process samples and reports its state.
+SAMPLE_INTERVAL = float(os.environ.get("WSGI_SAMPLER_INTERVAL", "2"))
+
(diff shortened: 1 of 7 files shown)


openstack/keystone/templates/bin/_wsgi_sampler.py.tpl
```diff
@@ -0,0 +1,79 @@
+# Copyright 2024 The Openstack-Helm Authors.
```
bbobrov: this needs to be fixed

bbobrov: fixed

openstack/keystone/templates/configmap-etc.yaml
```diff
@@ -24,13 +24,25 @@ data:
...
     defaults:
       timer_type: histogram
       buckets: [.025, .1, .25, 1, 2.5]
       match_type: glob
       glob_disable_ordering: false
-      ttl: 0 # metrics do not expire
+      ttl: 30s # worker gauges expire if a process stops reporting
 {{- end }}
   access_rules.json: |
 {{ include (print .Template.BasePath "/etc/_access_rules.json.tpl") . | indent 4 }}
```
tz3, line 49: `ttl` in `defaults` applies to every metric flowing through the exporter, not just the worker gauges — histograms for infrequent API routes will vanish after 30 s of silence. 

Move `ttl: 30s` onto the mapping entry itself and restore `ttl: 0` in `defaults` (see [Time series expiration](https://github.com/prometheus/statsd_exporter/blob/master/README.md#time-series-expiration)):

```yaml
- match: "...wsgi.worker.active_requests.*"
  name: "keystone_wsgi_worker_active_requests"
  match_metric_type: gauge
  labels:
    pid: "$1"
  ttl: 30s
defaults:
  timer_type: histogram
  buckets: [.025, .1, .25, 1, 2.5]
  match_type: glob
  glob_disable_ordering: false
  ttl: 0
```

bbobrov: Thank you, done

tz3 (dismissed):

BerndKue (dismissed):

rajivmucheli (approved):

tz3 (approved):

PR #12413, opened 2026-07-27
Update values.yaml

ivanlucaslozano:
Remove personal data


prometheus-exporters/aci-suite/values.yaml
```diff
@@ -8,7 +8,7 @@ owner-info:
   service: aci-exporter
   support-group: network-atm
   maintainers:
-    - Ivan Lucas Lozano
+    - D071722
```
ibakshay, line 11: should be your github handle name. 
```suggestion
    - ivanlucaslozano
```

PR #12415, opened 2026-07-27
add metal maintenance operator remote chart

asergeant01:


system/metal-maintenance-operator-remote/values.yaml
```diff
@@ -0,0 +1,74 @@
...
+    enable: false
+  serverReadinessNamespaces: []
+  manager:
+    enabled: true
+    image:
+      repository: keppel.global.cloud.sap/ccloud-ghcr-io-mirror/ironcore-dev/metal-maintenance-operator-controller-manager
+    args:
+      - "--leader-elect"
+      - "--leader-election-namespace=kube-system"
+      - "--sanitization-namespace=metal-servers"
```
stefanhipfel: do we want the same namespace for sanitization or use a separate one? it could make sense to keep it separated from the normal server claims.

asergeant01: @stefanhipfel I agree. Changed.

stefanhipfel (dismissed):

system/metal-maintenance-operator-remote/values.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - "--leader-elect"
+      - "--leader-election-namespace=kube-system"
+      - "--sanitization-namespace=metal-maintenance-sanitization"
+      - "--sanitization-image=keppel.global.cloud.sap/ccloud-ghcr-io-mirror/ironcore-dev/sanitizer@sha256:82f78c3d4c991a3b055415cb649cb371faaf63a64ed1407b7eeadf97680713e7"
+      - "--report-base-url=https://metal-maintenance-operator.garden.<region>.cloud.sap"
+    env:
+      - name: KUBECONFIG
+        value: "/var/run/remote-kubeconfig/kubeconfig"
+      - name: "KUBERNETES_SERVICE_HOST"
+        value: "<k8s-api-server-host>"
```
BerndKue: Is this on purpose and just a place holder?

BerndKue (dismissed):

system/metal-maintenance-operator-remote/values.yaml
```diff
@@ -0,0 +1,74 @@
...
+    extraVolumeMounts:
+      - name: remote-serviceaccount
+        mountPath: /var/run/secrets/kubernetes.io/remote-serviceaccount
+        readOnly: true
```
defo89:
```suggestion
      - name: remote-kubeconfig
        mountPath: /var/run/secrets/kubernetes.io/serviceaccount
        readOnly: true
```

system/metal-maintenance-operator-remote/values.yaml
```diff
@@ -0,0 +1,74 @@
...
+    extraVolumes:
+      - name: remote-serviceaccount
```
defo89:
```suggestion
      - name: remote-kubeconfig
```

system/metal-maintenance-operator-remote/values.yaml
defo89: can't we use same format/env as in https://github.com/sapcc/helm-charts/blob/master/system/boot-operator-remote/values.yaml#L40-L41 ? 

`KUBERNETES_SERVICE_HOST` env will be overridden in secrets anyway

system/metal-maintenance-operator-remote/values.yaml
```diff
@@ -0,0 +1,74 @@
...
+  ingress:
+    enable: false
+  serverReadinessNamespaces: []
+  manager:
+    enabled: true
+    image:
+      repository: keppel.global.cloud.sap/ccloud-ghcr-io-mirror/ironcore-dev/metal-maintenance-operator-controller-manager
+    args:
+      - "--leader-elect"
+      - "--leader-election-namespace=kube-system"
```
defo89: Is this required? Can't spot it in `boot-operator-remote`. 
Also do we need to ship RBAC for leader-election like it's done [here](https://github.com/sapcc/helm-charts/blob/master/system/boot-operator-remote/managedresources/crds-and-rbac.yaml#L696-L708)

defo89 (commented):
It seems that `helm dep up` run is needed as `system/metal-maintenance-operator-remote/Chart.lock` file is missing.

system/metal-maintenance-operator-remote/values.yaml
asergeant01: I only see the need for 1 replica currently. will remove

system/metal-maintenance-operator-remote/values.yaml
asergeant01: Changed to use same format

defo89 (approved):

BerndKue (approved):

PR #12420, opened 2026-07-27
[octavia] Add CPBMU feature flag support

BenjaminLudwigSAP:
The endpoint is not yet optimized, therefore not ready for production. Gate prod regions.
Feature flag code change in Octavia: https://github.com/sapcc/octavia/pull/63

### DIFF
--- a/openstack/octavia/templates/etc/_octavia.conf.tpl
+++ b/openstack/octavia/templates/etc/_octavia.conf.tpl
@@ -14,6 +14,7 @@ rpc_response_timeout = {{ .Values.rpc_response_timeout | default .Values.global.
 bind_host = 0.0.0.0
 bind_port = {{.Values.global.octavia_port_internal | default 9876}}
 healthcheck_enabled = True
+allow_cross_pool_batch_members_update = {{ if hasKey .Values "allow_cross_pool_batch_members_update" }}{{ .Values.allow_cross_pool_batch_members_update }}{{ else }}true{{ end }}
 
 # Enable/disable exposing API endpoints. By default, both v1 and v2 are enabled.
 api_v1_enabled = False


m-kratochvil (dismissed):

openstack/octavia/templates/etc/_octavia.conf.tpl
```diff
@@ -14,6 +14,9 @@ rpc_response_timeout = {{ .Values.rpc_response_timeout | default .Values.global.
 bind_host = 0.0.0.0
 bind_port = {{.Values.global.octavia_port_internal | default 9876}}
 healthcheck_enabled = True
+{{- if .Values.allow_cross_pool_batch_members_update }}
```
sebageek: If you check here for trueness™ of your value it means that you need to transport `allow_cross_pool_batch_members_update` as a string and not a bool in your values or you are never able to turn this option off. Should we keep this, check here if the value is defined or just render out the value with a default? Another option would be `{{- if ne .Values.allow_cross_pool_batch_members_update nil }}`.

BenjaminLudwigSAP: Oh... Of course I meant to check whether the value is defined. Thanks for pointing this out (:

openstack/octavia/templates/etc/_octavia.conf.tpl
```diff
@@ -14,6 +14,7 @@ rpc_response_timeout = {{ .Values.rpc_response_timeout | default .Values.global.
 bind_host = 0.0.0.0
 bind_port = {{.Values.global.octavia_port_internal | default 9876}}
 healthcheck_enabled = True
+allow_cross_pool_batch_members_update = {{ .Values.allow_cross_pool_batch_members_update | default True }}
```
sebageek: This won't work, as helm will try to call `True` as a function. It needs to either be `default true` or `default "True"`.

BenjaminLudwigSAP: Yes... Will `default true` render as `True` though? Or to `true`? I think oslo.config needs Pythonic `True`?
[This](https://github.com/sapcc/octavia/commit/24465ecb9c767b0b634f3618f5935f596b20d13e) is where this value is used btw.

I wish I could write something like `if defined` but I wasn't able to find out how that can be expressed. Looks like everyone ever is using `{{if .myvalue` and nobody ever uses booleans...

sebageek:
```
diff --git a/openstack/octavia/templates/etc/_octavia.conf.tpl b/openstack/octavia/templates/etc/_octavia.conf.tpl
index dd8d7a1712..ffc8110a23 100644
--- a/openstack/octavia/templates/etc/_octavia.conf.tpl
+++ b/openstack/octavia/templates/etc/_octavia.conf.tpl
@@ -14,7 +14,7 @@ rpc_response_timeout = {{ .Values.rpc_response_timeout | default .Values.global.
 bind_host = 0.0.0.0
 bind_port = {{.Values.global.octavia_port_internal | default 9876}}
 healthcheck_enabled = True
-allow_cross_pool_batch_members_update = {{ .Values.allow_cross_pool_batch_members_update | default True }}
+allow_cross_pool_batch_members_update = {{ .Values.allow_cross_pool_batch_members_update | default true }}
 
 # Enable/disable exposing API endpoints. By default, both v1 and v2 are enabled.
 api_v1_enabled = False
```

```
$ grep allow_cross_pool_batch_members_update /tmp/rendered-chart
    allow_cross_pool_batch_members_update = true
```

Jep, `default true` renders to `true` in the output.

sebageek: If you want to check if it is defined you can go with doing a nil check: `{{- if ne .Values.allow_cross_pool_batch_members_update nil }}`

BenjaminLudwigSAP: Oh, Go templating doesn't see `True` as a special value (like `true`), and thus treats it as a function identifier.
Also: I think `default` cannot be used, since it would also trigger when `.Values.allow_cross_pool_batch_members_update` is false.

Pushed an amending fix commit.

Edit: And of course GitHub didn't show me your last two comments before I wrote this one.........

sebageek (approved):
[optional] if you do a `hasKey` check and insert the default value anyway you could also leave the option out completely (by surrounding it with the hasKey-if), but in the end it doesn't matter. What you have should work

m-kratochvil (approved):

PR #12423, opened 2026-07-27
feat(shoots): exclude /32 and /128 host routes from BGP reject filter

defo89:
Anycast VIPs (host routes) in `externalServices` must be advertised via BGP, not rejected.

This adds a `hasSuffix` check to skip `/32` and `/128` CIDRs from the BGP export reject filter while keeping them in `serviceLoadBalancerIPs` and `IPPools`.

**Affected charts:**
- cc-shoots-controlplane (1.0.7 → 1.0.8)
- cc-shoots-storage (0.0.12 → 0.0.13)
- cc-shoots-mgmt (1.1.35 → 1.1.36)
- cc-shoots-compute (2.6.5 → 2.6.6)

**Test:** Added `1.1.1.1/32` to all CI test-values; verified it appears in `serviceLoadBalancerIPs` but NOT in the reject filter.

### DIFF
--- a/system/cc-shoots-compute/Chart.yaml
+++ b/system/cc-shoots-compute/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: cc-shoots-compute
 description: Converged Cloud Gardener shoots
 type: application
-version: 2.6.5
+version: 2.6.6
 appVersion: "0.1.0"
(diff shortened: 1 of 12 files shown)


goerangudat (dismissed):
lgtm

adziauho (dismissed):

kayrus (approved):

PR #12446, opened 2026-07-29
[ironic] limit workers for console nginx

sandzwerg:
The new nodes in the eu-de-3 region are baremetal and thus present all
80 virtual cpus to the containers instead of the 10 on the VMs in the
old setup. With the "worker_process auto;" setting in the nginx
container that means nginx will spawn a worker per CPU, breaking the
(small) memory limits and getting OOM killed. We could increase the
limit for the container but we probably never need so many workers. A
worker can have 1024 connections, as there is probably only one per
node we're never going to reach that. Instead we take the nginx.conf
from the container and just make the worker_process adjustable via
values. While we currently have 10 workers, 5 should be plenty for this
job as reverse proxy.
To make this easier to understand we rename an old mount from nginxconf
to default, because it is mounting the default.conf and not the
nginx.conf. This way the names better reflect what the variables do.

### DIFF
--- a/openstack/ironic/templates/_conductor-deployment.yaml.tpl
+++ b/openstack/ironic/templates/_conductor-deployment.yaml.tpl
@@ -167,11 +167,14 @@ spec:
             containerPort: 443
         volumeMounts:
           - mountPath: /etc/nginx/conf.d/default.conf
-            name: ironic-console-nginxconf
+            name: ironic-console-default
             subPath: default.conf
           - mountPath: /etc/nginx/conf.d/dhparam.pem
             name: ironic-console-dhparam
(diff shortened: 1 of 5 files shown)


BerndKue (approved):

xkonni (approved):
i liked the filename with nginx more, but the overall patch looks good.

sandzwerg:
Yeah but nginx has multiple config files. One is `/etc/nginx/nginx.conf` which is what I'm now adding. The other is `/etc/nginx/conf.d/default.conf` this secret was called nginx.conf before even when it isn't. As we now have a proper nginx.conf that would have been to confusing

PR #12448, opened 2026-07-30
[keystone] add  pod-level TLS termination

stefanhipfel:
Option 4 (direct BGP exposure) with key lifecycle option a (go-crypto + none + internal-k8s-secret) for 007-enabled regions.

### DIFF
--- a/openstack/keystone/templates/service-external.yaml
+++ b/openstack/keystone/templates/service-external.yaml
@@ -1,0 +1,27 @@
+{{- if .Values.global.keystone_external_ip }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ .Release.Name }}-external
+  labels:
+    app: {{ template "fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+    system: openstack
+    component: keystone
+    type: api-external
+  annotations:
+    {{- include "utils.linkerd.pod_and_service_annotation" . | indent 4 }}
+spec:
+  type: ClusterIP
+  selector:
+    name: {{ .Release.Name }}-api
+  ports:
+    - name: public
+      protocol: TCP
+      port: 5000
+      targetPort: 5000
+  externalIPs:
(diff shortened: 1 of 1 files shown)


BerndKue (dismissed):

openstack/keystone/templates/service-external.yaml
```diff
@@ -0,0 +1,27 @@
...
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+    system: openstack
+    component: keystone
+    type: api-external
+  annotations:
+    {{- include "utils.linkerd.pod_and_service_annotation" . | indent 4 }}
+spec:
+  type: ClusterIP
```
defo89:
```suggestion
  type: LoadBalancer
  externalTrafficPolicy: Local  
```
With Calico+BGP, `externalTrafficPolicy: Local` makes Calico advertise the route only from nodes with a local pod.

Otherwise externalIPs get announced from all nodes, add a SNAT hop, and loses source IP.

openstack/keystone/templates/service-external.yaml
```diff
@@ -0,0 +1,27 @@
...
+  labels:
+    app: {{ template "fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+    system: openstack
+    component: keystone
+    type: api-external
+  annotations:
+    {{- include "utils.linkerd.pod_and_service_annotation" . | indent 4 }}
```
defo89, line 15:
```suggestion
  annotations:
    {{- include "utils.linkerd.pod_and_service_annotation" . | indent 4 }}
    projectcalico.org/loadBalancerIPs: '["{{ .Values.global.keystone_external_ip }}"]'
```
externalIPs is [being deprecated](https://github.com/kubernetes/kubernetes/pull/137293) (KEP-5707). Migration hasn't started yet, but since this is a new service we can use `projectcalico.org/loadBalancerIPs` directly — one less thing to migrate later.

bbobrov:
This has been included into my branch where i work on wp4. This patch can be closed.

PR #12450, opened 2026-07-30
common/{pgmetrics,redis}: add support for inline templating

majewsky:
This idea has been in the back of my head for a long time. I hate our current values.yaml.erb mess in the secrets repo, esp. because most of it is just "this one region needs a single different value".

But Helm has the "tpl" function, which is basically an eval: It takes a template string and evaluates it as though it had been written inline. With this, we can write template strings inside the application chart's `values.yaml`, as long as the subchart that uses the value remembers to apply `tpl` to it.

As the docstring example indicates, I need this for Trivy, which usually goes in the scaleout cluster in Kubernikus regions, but needs to go to the controlplane cluster in Gardener regions. Since different cluster types have different Promethei, this necessitates templating. Usually I would put this in `values.yaml.erb`, but if you agree with the pattern as shown here, I would like to extend this to other value keys in the future in order to simplify the `values.yaml.erb` mess.


common/pgmetrics/templates/alerts.yaml
```diff
@@ -8,7 +8,7 @@ kind: PrometheusRule
 metadata:
   name: {{ $.Release.Name }}-pgmetrics-{{ $dbName }}-alerts
   labels:
-    prometheus: {{ required ".Values.alerts.prometheus missing" $.Values.alerts.prometheus }}
+    prometheus: {{ tpl $.Values.alerts.prometheus $ | quote }}
```
majewsky, line 11: FYI, I removed the `required` check here because it was ineffective: The key has a default value, so `required` will never see a null value.

common/redis/values.yaml
```diff
@@ -45,6 +45,7 @@ metrics:
   enabled: true
 
   # Name of the Prometheus by which the metrics will be scraped.
+  # This field can also accept a template string like '{{ eq .Values.global.clusterType "scaleout" | ternary "infra-frontend" "obs-metrics-product" }}'.
```
SuperSandro2000, line 48: Basically this:
```
if Values.global.clusterType == "scaleout" then
    result = "infra-frontend"
else
    result = "obs-metrics-product"
end
```

SuperSandro2000 (approved):
🤷🏼

PR #12470, opened 2026-07-31
Create wildcardcert.yaml

LimitNOFILE:


ronchi-oss:
linting failed:

```
[ERROR] templates/: template: persephone-runtime-seed/templates/wildcardcert.yaml:4:15: executing "persephone-runtime-seed/templates/wildcardcert.yaml" at <$cluster.wildcardCertificate.enabled>: nil pointer evaluating interface {}.enabled
```

ronchi-oss:
this will also require a chart version bump

LimitNOFILE:
Thanks. I created a nested IF, so we can keep the 
```
service: 
  enabled: true
```
  syntax that we have everywhere else. 
  I also bumped the chart version.

ronchi-oss (approved):

PR #12503, opened 2026-08-05
openstack/keppel: add multi storage driver

SuperSandro2000:


openstack/keppel/templates/_utils.tpl
```diff
@@ -63,7 +63,9 @@
 - name:  KEPPEL_DRIVER_STORAGE
   {{- if eq .Values.keppel.driver "ceph" }}
   value: '{"type":"swift","params":{"apply_workarounds_for_ceph":true,"service_type":"object-store-ceph","use_service_user_project":true}}'
-  {{- else }}{{/* TODO: if eq .Values.keppel.driver "multi", use that driver once it exists */}}
+  {{- if eq .Values.keppel.driver "ceph" }}
```
majewsky: Looks like a copy-paste error:
```suggestion
  {{- else if eq .Values.keppel.driver "multi" }}
```

majewsky (changes requested):

majewsky (approved):

PR #12516, opened 2026-08-07
metal-settings: make HPE iLO license setting optional

xkonni:
Skip the license POST setting and LicenseKey secretKeyRef variable when hpeLicenseSecretName is not set, preventing BMCSettingsSet validation failures on clusters without a license secret.

Also reverts the cloud.sap/inject-secrets annotation added in 0.2.42 as the webhook is opt-out and fires on all secrets by default.

### DIFF
--- a/system/metal-settings/bmc_settings/hpe-default.yaml
+++ b/system/metal-settings/bmc_settings/hpe-default.yaml
@@ -11,8 +11,10 @@
 #
 # Values sourced from: baremetal-config/hpe/*/bmc/defaultsettings.yml
 settings:
+  {{- if .Values.clusterConfig.hpeLicenseSecretName }}
   # License key — $(LicenseKey) fetched at reconcile time from the Secret named by clusterConfig.hpeLicenseSecretName.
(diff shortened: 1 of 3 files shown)


system/metal-settings/bmc_settings/hpe-default.yaml
```diff
@@ -11,8 +11,10 @@
 #
 # Values sourced from: baremetal-config/hpe/*/bmc/defaultsettings.yml
 settings:
+  {{- if .Values.clusterConfig.hpeLicenseSecretName }}
   # License key — $(LicenseKey) fetched at reconcile time from the Secret named by clusterConfig.hpeLicenseSecretName.
   "POST /redfish/v1/Managers/1/LicenseService/": '{"LicenseKey": "$(LicenseKey)"}'
+  {{- end }}
```
nagadeesh-nagaraja, line 17: if this templating actually works, then LGTM. please test if it works

nagadeesh-nagaraja (approved):

system/metal-settings/bmc_settings/hpe-default.yaml
xkonni: tested and it does work.

PR #12521, opened 2026-08-07
feat(audit-logs): adds mirror output for logstash export

olandr:
To send audit logs both to octobus and fortlogs we need to mirror the output to another endpoint. this is a very rudimentary copy of the message, not having any enriched fields (this is done downstream by parsing the message). a bit of double work, enriched is sent to octobus but message downstream. an argument could be that this is not the same log info.

---------

Signed-off-by: Simon Olander <simon.olander@sap.com>

### DIFF
--- a/system/audit-logs/vendor/logstash-audit-external/templates/_logstash.conf.tpl
+++ b/system/audit-logs/vendor/logstash-audit-external/templates/_logstash.conf.tpl
@@ -282,6 +282,14 @@ output {
       format => "json"
       http_method => "post"
     }
+    tcp {
+      id => "output_awx_tcp"
+      host => "logs-collector-external.eu-de-1.cloud.sap"
+      port => 514
+      codec => line {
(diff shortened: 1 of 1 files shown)


system/audit-logs/vendor/logstash-audit-external/templates/_logstash.conf.tpl
```diff
@@ -282,6 +282,14 @@ output {
       format => "json"
       http_method => "post"
     }
+    tcp {
+      id => "output_awx_tcp"
+      host => "logs-collector-external.eu-de-1.cloud.sap"
+      port => 514
+      codec => line {
+        format => "%{message} forwarded_by=octobus_logstash"
```
melvinsimmert: Can you adjust that to "forwarded_by=logstash-audit-external"?

melvinsimmert (changes requested):

melvinsimmert (approved):

PR #12531, opened 2026-08-07
keystone/saml: route transaction log to stdout, add SAML security event alert

tz3:
## Summary

Part of [#1180 — SAML Security Events Logging and Monitoring](https://github.wdf.sap.corp/sap-cloud-infrastructure/foundation-issues/issues/1180) (BSI SF.Fas.6, SF.Log.1).

**AC3 — transaction.log routing**

`Shibboleth-TRANSACTION` previously wrote only to a file inside the container (lost on restart, never reaches the pipeline). Added a `tran_stdout` appender in `_shibd-logger.tpl` that forwards all SAML auth events (success and failure) to `/proc/1/fd/1` (Apache stdout → Fluent Bit → OpenSearch).

**AC4 — SAML security event alert**

Added an `opensearch-logs-query-exporter` query file that counts shibd WARN/ERROR lines from `keystone-api` pods in `monsoon3` every 5 minutes. Added Prometheus alert `KeystoneSAMLSecurityEventLogged` that fires immediately when count > 0, routed to `support_group: identity`. Uses `no_alert_on_absence: "true"` so regions without SAML don't false-alert.

## Files changed

- `openstack/keystone/templates/etc/_shibd-logger.tpl` — add `tran_stdout` appender, include it in `Shibboleth-TRANSACTION` category
- `prometheus-exporters/opensearch-logs-query-exporter/files/opensearch_logs_keystone_saml.cfg` — new OpenSearch query
- `prometheus-exporters/opensearch-logs-query-exporter/alerts/opensearch-logs-query.alerts.tpl` — new `KeystoneSAMLSecurityEventLogged` alert

## Test plan

- [x] Deploy to qa-de-3, run a SAML login, verify `body: "transaction|"` entries appear in OpenSearch `logs-*`
- [x] Send a tampered SAML assertion, verify WARN line appears in OpenSearch and metric `opensearch_logs_keystone_saml_errors_doc_count` > 0 within 5 min
- [ ] Confirm alert fires in Alertmanager

### DIFF
--- a/prometheus-exporters/opensearch-logs-query-exporter/alerts/opensearch-logs-query.alerts.tpl
+++ b/prometheus-exporters/opensearch-logs-query-exporter/alerts/opensearch-logs-query.alerts.tpl
@@ -110,6 +110,7 @@ groups:
     expr: opensearch_logs_keystone_saml_errors_doc_count > 0
     for: 0m
     labels:
+      context: compliance
       severity: warning
       service: keystone
       support_group: identity
--- a/prometheus-exporters/opensearch-logs-query-exporter/alerts/opensearch-logs-query.alerts.tpl
+++ b/prometheus-exporters/opensearch-logs-query-exporter/alerts/opensearch-logs-query.alerts.tpl
@@ -105,3 +105,16 @@ groups:
     annotations:
       description: "Maillog entries older than 30 days found in {{`{{ $labels.region }}`}}."
       summary: "Maillog compliance violation"
+
+  - alert: KeystoneSAMLSecurityEventLogged
+    expr: opensearch_logs_keystone_saml_errors_doc_count > 0
+    for: 0m
+    labels:
+      severity: warning
+      service: keystone
+      support_group: identity
+      tier: os
+      no_alert_on_absence: "true"
+    annotations:
+      summary: "SAML security event logged in keystone-api ({{`{{ $labels.region }}`}})"
+      description: "mod_shib logged a WARN/ERROR for Shibboleth or XMLTooling.Signature in monsoon3/keystone-api in the last 5m. Check OpenSearch logs-* for 'Shibboleth' + 'WARN'/'ERROR'."
--- a/openstack/keystone/templates/etc/_shibd-logger.tpl
+++ b/openstack/keystone/templates/etc/_shibd-logger.tpl
@@ -18,7 +18,7 @@ log4j.ownAppenders.XMLTooling.Signature.Debugger=true
 
 # the tran log blocks the "default" appender(s) at runtime
 # Level should be left at INFO for this category
-log4j.category.Shibboleth-TRANSACTION=INFO, tran_log
+log4j.category.Shibboleth-TRANSACTION=INFO, tran_log, tran_stdout
 log4j.additivity.Shibboleth-TRANSACTION=false
 log4j.ownAppenders.Shibboleth-TRANSACTION=true
 
@@ -43,6 +43,14 @@ log4j.appender.warn_log.fileName=/proc/1/fd/1
 log4j.appender.warn_log.layout=org.apache.log4j.PatternLayout
 log4j.appender.warn_log.layout.ConversionPattern=%d{%Y-%m-%d %H:%M:%S} %p %c %x: %m%n
 log4j.appender.warn_log.threshold=WARN
+
+# [AC3] Route all SAML transaction log entries to container stdout.
+#   Every auth decision (success and failure) reaches Fluent Bit → OpenSearch
+#   via Apache stdout. Satisfies BSI SF.Fas.6 / SF.Log.1.
+log4j.appender.tran_stdout=org.apache.log4j.FileAppender
+log4j.appender.tran_stdout.fileName=/proc/1/fd/1
+log4j.appender.tran_stdout.layout=org.apache.log4j.PatternLayout
+log4j.appender.tran_stdout.layout.ConversionPattern=%d{%Y-%m-%dT%H:%M:%S}|transaction|%c|%m%n
 
 # [FIX L-01] maxFileSize increased from 1 MB to 10 MB.
 #   Transaction log is the SAML audit trail -- must not rotate prematurely.
--- a/prometheus-exporters/opensearch-logs-query-exporter/files/opensearch_logs_keystone_saml.cfg
+++ b/prometheus-exporters/opensearch-logs-query-exporter/files/opensearch_logs_keystone_saml.cfg
@@ -1,0 +1,50 @@
+[query_opensearch_keystone_saml_errors]
+QueryIntervalSecs = 300
+QueryTimeoutSecs = 15
+QueryOnMissing = zero
+QueryIndices = logs-*
+QueryJson = {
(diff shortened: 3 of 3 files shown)


prometheus-exporters/opensearch-logs-query-exporter/alerts/opensearch-logs-query.alerts.tpl
```diff
@@ -105,3 +105,16 @@ groups:
     annotations:
       description: "Maillog entries older than 30 days found in {{`{{ $labels.region }}`}}."
       summary: "Maillog compliance violation"
+
+  - alert: KeystoneSAMLSecurityEventLogged
+    expr: opensearch_logs_keystone_saml_errors_doc_count > 0
+    for: 0m
+    labels:
+      severity: warning
```
bbobrov, line 114: I don't exactly understand how it is going to work. But i would like the alerts like this to got to slack channel `#alert-identity-compliance` instead of to the warnings channel. Would it be possible?

tz3: Added `context: compliance` label to `KeystoneSAMLSecurityEventLogged` — matches the routing pattern used by `CronusLogsComplianceViolation` and `MaillogComplianceViolation`, both of which route to `#alert-identity-compliance` via Alertmanager.

AC3 tested in qa-de-3: `shibd.logger` configmap confirmed with `tran_stdout`.

bbobrov:
Could you please confirm that you got the alert to fire in any of the qa regions?

tz3:
Hi @bbobrov,

Tested in qa-de-3 — logging works. POSTing a malformed SAMLResponse to the ACS endpoint produces the `|transaction|` line in pod stdout immediately:

```
2026-09-04T13:32:01|transaction|Shibboleth-TRANSACTION.Login|||||||||urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST|xmltooling::XMLParserException||||curl/8.7.1|100.176.0.234
```

To trigger it yourself:

```bash
# terminal 1 — watch logs
kubectl logs -n monsoon3 -l app=keystone -c keystone-api -f | grep '|transaction|'

# terminal 2 — send invalid SAMLResponse
curl -sk -X POST https://identity-3.qa-de-3.cloud.sap/Shibboleth.sso/apollo/SAML2/POST \
  -d 'SAMLResponse=aW52YWxpZA==' \
  -H 'Content-Type: application/x-www-form-urlencoded'
```

For the `KeystoneSAMLSecurityEventLogged` alert — `opensearch-logs-query-exporter` isn't deployed in any QA region so we can't fire it there. But the log format is confirmed correct.

bbobrov:
> For the KeystoneSAMLSecurityEventLogged alert — opensearch-logs-query-exporter isn't deployed in any QA region so we can't fire it there. But the log format is confirmed correct.

I open the `#alert-identity-qa` and `#alert-identity-labs` channels. There i see many alerts that are definitely related to keystone, for example, `KeystoneMariaDatabaseIncBackupMissing`. Why is that alert visible and this one is not?

bbobrov:
I would still like to be sure that there is an actual alert firing when that log line appears. Claude tells me this. Is it wrong?

> 
> ## Why the `KeystoneSAMLSecurityEventLogged` alert does not work
> ### Summary
> The alert can never fire. The Prometheus rule waits for a metric named
> `opensearch_logs_keystone_saml_errors_doc_count`, but the exporter that runs this
> query never produces a metric by that name. Because the metric does not exist, the
> alert expression has no data to evaluate and stays permanently inactive — even in
> production, where the exporter runs, and even when a real SAML security event is
> logged.
> ### How these metrics are named
> The alert is fed by `prometheus-es-exporter` (the `opensearch-logs-query-exporter`).
> This tool reads each query definition file, runs the query against OpenSearch on a
> schedule, and turns the result into a Prometheus metric. The metric name is built
> from two parts:
> 1. A prefix taken directly from the query's section header, with the leading
>    `query_` removed and everything else kept exactly as written.
> 2. A suffix that depends on what the query returns:
>    - `_hits` — the total number of matching documents. Always available for any query.
>    - `<aggregation-name>_doc_count` — a per-bucket count. This only exists when the
>      query contains an aggregation, and the aggregation's own name is embedded in the
>      metric name.
> This behaviour is confirmed by the other alerts already in the same file:
> - The ESXi query section `[query_opensearch_esxi_exporter_login_error]` contains an
>   aggregation named `container_name`, and its metric is exactly
>   `opensearch_esxi_exporter_login_error_container_name_doc_count`.
> - The SIEM query section `[query_opensearch_siem]` with aggregation `error` yields
>   `opensearch_siem_error_doc_count`.
> - The octobus, cronus, and maillog alerts, whose queries have no usable aggregation,
>   all match on the `_hits` suffix.
> - The octobus query section was previously renamed to insert the word `logs` so that
>   its emitted metric name would match its alert.
> ### The problem: two independent mismatches
> The new query is defined as:
> - Section header: `[query_opensearch_keystone_saml_errors]`
> - Query body: counts matching log lines only (`"size": 0`), with no aggregation of
>   any kind.
> Applying the naming rules, the exporter publishes this metric:
>     opensearch_keystone_saml_errors_hits
> The alert is written against:
>     opensearch_logs_keystone_saml_errors_doc_count
> These differ in two separate ways, each sufficient on its own to break the alert:
> 1. Wrong prefix. The alert's name contains the extra word `logs`
>    (`..._logs_keystone_...`). The query's section header does not contain that word,
>    so the real prefix is `opensearch_keystone_saml_errors`.
> 2. Wrong suffix. The alert uses `_doc_count`, which the exporter only emits for
>    queries containing an aggregation. This query has none, so the only count it
>    produces carries the `_hits` suffix. Even if an aggregation were added, the
>    resulting name would embed the aggregation's own name before `_doc_count` (as in
>    the ESXi example) — so a bare `..._errors_doc_count` is a name the exporter would
>    never generate under any configuration.
> Because of these mismatches, the metric referenced by the alert does not exist. In
> Prometheus, an expression referencing a non-existent metric produces no result, so
> the `> 0` condition is never met and the alert never enters the firing state.

tz3:
@bbobrov Tested in qa-de-3. Alert fires and routes correctly to #alert-identity-qa:

> 5:55 PM — [WARNING] [QA-DE-3] KeystoneSAMLSecurityEventLogged appeared in Slack with correct labels (support_group: identity, severity: warning, service: keystone)
6:02 PM — [RESOLVED] on rule deletion

bbobrov (dismissed):

bbobrov:
@drochow @giobbe please let us know what the playbook for this alert is

rajivmucheli (dismissed):

tz3:
Rereview please @bbobrov @rajivmucheli 
Playbook is ready, no need to wait for SRC: https://operations.global.cloud.sap/docs/support/playbook/keystone/alerts/saml-security-event

bbobrov (changes requested):
Lets please wait for the meeting with Security Engineering Team before we get this in, because we are not in position to decide what to do when an alert happens - we don't even have access to the region where it happens.

PR #12539, opened 2026-08-10
fix(cc3test/cinder): reduce `OpenstackCinderKVMVolumeAttachFailed` sensitivity on alerts

iammati:
Require 3 actual failing runs in a 3h window instead of metric staying zero for 2h, and gate on setup phase == 1 to exclude SSH/network blips.


iammati:
Will be implemented in favor of [PR#12545](https://github.com/sapcc/helm-charts/pull/12545)

PR #12557, opened 2026-08-12
add-openstack-user-seeds

yahor-kurachkin:
Introduces a new openstack-user-seeds Helm chart that provisions users via an OpenstackSeed resource. The initial seed creates a read-only Glance image user in the Default domain with configurable name, password, and per-project role assignments.


adziauho (dismissed):

openstack/openstack-user-seeds/templates/user-seed.yaml
```diff
@@ -0,0 +1,21 @@
...
+spec:
+  requires:
+  - {{ .Values.global.keystoneNamespace }}/keystone-seed
+  - {{ .Values.global.keystoneNamespace }}/domain-default-seed
+  - {{ .Values.global.keystoneNamespace }}/domain-ccadmin-seed
+
+  domains:
+  - name: {{ .Values.name }}
+    users:
+    - name: {{ .Values.glance.users.name | quote }}
```
databus23: I think `.Values.glances.users` is a suboptimal choice.
First `.glance` does not carry and information as the user is fully specified in roleAssignments.
Also `users` indicates there should be more then one but because it not an array but an object there can be only one.

I would suggest to start with a generic value structure:

```
users:
  - name: glance-user
     description: 'Read-only access to Glance images'
     password: 
     roleAssigments:
     - role1
     - role2
```

If we range over `.Values.users` we support more then one user right from the start.

databus23 (changes requested):

openstack/openstack-user-seeds/templates/user-seed.yaml
yahor-kurachkin: Thx for your comment! I made changes.

adziauho (approved):

databus23 (approved):
Thanks LGTM

PR #12577, opened 2026-08-14
keystone: add CPU limit, memory/latency alerts

tz3:
Follows up on #12491 to complete the remaining items from [foundation-issues#630](https://github.wdf.sap.corp/sap-cloud-infrastructure/foundation-issues/issues/630).

After review, most of what I originally added got dropped:

- CPU limit — Boris pointed out keystone does crypto and should be able to burst freely, so no limit makes sense here
- OOMKilled alert — already covered by the platform-level alert in #cc-k8s-info
- OOMExceedingLimits alert — dropped after review feedback that predictive alerts aren't immediately actionable and create alert fatigue
- Latency alerts — deferred until we have a better handle on apache worker/processing time metrics

Nothing meaningful left in this PR at this point. Closing — remaining items (Grafana graphs, latency alerting) are tracked in the issue.


openstack/keystone/alerts/kubernetes/keystone.alerts
```diff
@@ -34,6 +34,37 @@ groups:
       description: 'The keystone ingress API endpoint is throwing 5xx for 15 min. Check if the keystone-api k8s pods are OK ASAP.'
       summary: 'Keystone API ingress endpoint is in Critical State'
 
+  - alert: OpenstackKeystonePodOOMKilled
```
bbobrov: We already have this alert, but from kubernetes, in #cc-k8s-info

openstack/keystone/alerts/kubernetes/keystone.alerts
```diff
@@ -34,6 +34,37 @@ groups:
...
+      tier: os
+      service: keystone
+      severity: info
+      context: memory
+      meta: "{{ $labels.pod_name }}"
+    annotations:
+      summary: Pod was oomkilled
+      description: The pod {{ $labels.pod_name }} was oomkilled recently.
+
+  - alert: OpenstackKeystonePodOOMExceedingLimits
```
bbobrov: Lets suppose that this is happening. I am an operator and i got this alert. What are my actions now?

openstack/keystone/alerts/kubernetes/keystone.alerts
```diff
@@ -51,3 +82,19 @@ groups:
       description: The container {{ $labels.pod }}/{{ $labels.container }} is being CPU-throttled
         constantly. This is probably impacting performance, so check if we can increase the number
         of replicas or the resource requests/limits.
+
+  - alert: OpenstackKeystoneApiNginxHighLatency
```
bbobrov: Please don't add this until we get an understanding how other metrics behave (with workers and processing times from apache)

openstack/keystone/alerts/openstack/keystone.alerts
```diff
@@ -91,3 +91,42 @@ groups:
       description: Api Requests to Keystone is {{ $value }}ops/s
       summary: Api Requests to Keystone is {{ $value }}ops/s
 
+  - alert: OpenstackKeystoneApiHighLatency
```
bbobrov: Please don't add this until we get an understanding how other metrics behave (with workers and processing times from apache)

openstack/keystone/alerts/openstack/keystone.alerts
```diff
@@ -91,3 +91,42 @@ groups:
...
+      service: keystone
+      severity: warning
+      tier: os
+      support_group: identity
+      playbook: 'docs/devops/alert/keystone/#openstackkeystoneapihighlatency'
+    annotations:
+      description: Keystone API service-level mean latency has been above 3s for 5 minutes.
+      summary: Keystone API high latency
+
+  - alert: OpenstackKeystoneApiCriticalLatency
```
bbobrov: Please don't add this until we get an understanding how other metrics behave (with workers and processing times from apache)

openstack/keystone/values.yaml
```diff
@@ -209,6 +209,7 @@ api:
       cpu: 3000m
     limits:
       memory: 6Gi
+      cpu: 6000m
```
bbobrov: There should not be a limit. Keystone should be able to burn through CPU as much as it wants, because it is doing crypto operations.

bbobrov (changes requested):

openstack/keystone/values.yaml
tz3: Makes sense for a crypto workload — dropping the limit. The issue says "check if that makes sense" for limits, so I think this is a valid answer to that check. Could you leave a note on [foundation-issues#630](https://github.wdf.sap.corp/sap-cloud-infrastructure/foundation-issues/issues/630) confirming no CPU limit is the right call for keystone?

openstack/keystone/alerts/kubernetes/keystone.alerts
tz3: Dropped — wasn't aware of the platform-level one in #cc-k8s-info.

openstack/keystone/alerts/kubernetes/keystone.alerts
tz3: Updated the description with operator actions — scale replicas if traffic-driven, restart and monitor if it looks like a leak.

openstack/keystone/alerts/kubernetes/keystone.alerts
tz3: Dropped for now — will revisit once we have a better understanding of apache worker and processing time metrics.

openstack/keystone/alerts/openstack/keystone.alerts
tz3: Dropped for now — will revisit once we have a better understanding of apache worker and processing time metrics.

openstack/keystone/alerts/openstack/keystone.alerts
tz3: Dropped for now — will revisit once we have a better understanding of apache worker and processing time metrics.

openstack/keystone/values.yaml
tz3: done: https://github.wdf.sap.corp/sap-cloud-infrastructure/foundation-issues/issues/630#issuecomment-19206870

bbobrov:
Lets not add it please. This alert is not actionable, and non-actionable alerts lead to alert fatigue.

tz3:
Nothing meaningful left in this PR at this point. Closing — remaining items (Grafana graphs, latency alerting) are tracked in the issue.

PR #12601, opened 2026-08-18
[nova][cross-hv] Add default vol type for adopted eph disks

grandchild:
The on eph-root-disk-VM migration to KVM, the conductor instructs cinder to adopt the root disk VMDK as a proper FCD cinder volume. This needs a volume type set.

Expect this setting in regional nova secrets config object in

    conductor.config_file.cross_hv

with the contained sub-key 'fcd_volume_type'.
The conductor has per-cell config, but this is a regional setting, so we propagate this into all per-cell-conductor configs.

Default to type "vmware".


openstack/nova/values.yaml
```diff
@@ -376,6 +376,8 @@ conductor:
       statsd_enabled: true
     workarounds:
       skip_hypervisor_version_check_on_lm: true
+    cross_hv:
```
anokfireball: If we don't add a default, we have an implicit feature gate. Which would be nice in terms of being able to turn it on once cinder is ready.

anokfireball (approved):

seb-kro (approved):

PR #12607, opened 2026-08-19
prometheus-server: disable local compaction (workaround for operator v0.93.0 bug)

ibakshay:
## What

Disables local compaction on Prometheus by setting `disableCompaction: true` in the Prometheus CR.

## Why

After upgrading to prometheus-operator v0.93.0, the new delay-compact-file integration https://github.com/prometheus-operator/prometheus-operator/pull/8694 causes the Thanos sidecar to crash in a loop.

This causes the sidecar to exit after 10 minutes with exit code 1, restarting the pod repeatedly and triggering `CC3TestPrometheusRegionDown` alerts.

## Fix

Setting `disableCompaction: true` skips the broken `compactionModeDelayed` code path and falls back to the legacy behavior. Thanos Compactor handles compaction globally and not the prometheus container itself. 

Created an ticket in upstream prometheus-operator repository -  https://github.com/prometheus-operator/prometheus-operator/issues/8763

### DIFF
--- a/common/prometheus-server/Chart.yaml
+++ b/common/prometheus-server/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Prometheus via operator.
 name: prometheus-server
-version: 8.2.6
+version: 8.2.7
 appVersion: v3.11.2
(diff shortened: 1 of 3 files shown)


ashifnihalb (approved):
LGTM!

trouaux (approved):

PR #12614, opened 2026-08-19
fix(hermes): silence absent-metric alerts for log-router in regions where it is not deployed

notque:
## Summary
- Gate the three `OpenstackHermesLogRouter*` alerts behind `{{- if .Values.logRouter.enabled }}` so they are not rendered in regions without log-router
- Add `no_alert_on_absence: "true"` to `OpenstackHermesPodNotRunning` — the metric only exists when a pod is in a waiting state (matches pattern of sibling alerts)

## Context
Log-router is only deployed in eu-de-1, eu-de-3, and qa-de-1. The alert rules were rendered unconditionally, so the absent-metrics-operator flagged missing metrics in all other regions. This produced ~30+ INFO alerts every scrape cycle across NA, AP, LA, and EU regions.


wiltongarcia (approved):

ibakshay (approved):
LGTM

PR #12627, opened 2026-08-21
Herems update logstash exporter

LeoZhangZXZ:


ibakshay:
Please also upgrade the version. It will be easier for us to rollback then.

PR #12636, opened 2026-08-24
feat(go-pmtud): remove init container and preStop hook 

rkschamer:
https://github.com/sapcc/go-pmtud/pull/79 moves the preStart and preStop scripts into Go code of go-pmtud to enable FROM SCRATCH runtime image.

This PR removes the script from the respective helm chart


.superpowers/sdd/2026-08-21-in-binary-firewall-lifecycle/task-5-review-verdict.md
```diff
@@ -0,0 +1,33 @@
...
+✓ **Commit message appropriate** - "feat(go-pmtud): remove init container and preStop hook (rule lifecycle now in binary)" clearly documents change rationale  
+✓ **Git history clean** - commit 9fe47af938 properly recorded  
+
+### Architecture Assessment
+
+The removal correctly delegates iptables rule lifecycle management entirely to the binary. The daemonset retains all necessary permissions (privileged: true, hostNetwork: true) and RBAC binding to support in-binary rule management.
+
+---
+
+**Verdict: APPROVED** — Specification fully met, code quality sound, no issues identified.
```
defo89: Do we need to check this in?

defo89:
nit: I'd prefer a safer rollout approach here. Instead of removing the init container and preStop hook outright, could we:

1. Add a values toggle (e.g. `.Values.pmtud.inBinaryLifecycle: false`) that conditionally skips the init container / preStop hook
2. Enable it only in QA values first to let it soak
3. Once confirmed stable, flip the default to true and remove the legacy code path in a follow-up

The iptables rules are crucial for PMTUD — if the new in-binary lifecycle has a bug in setup/teardown, we'd break path MTU discovery cluster-wide with no easy rollback other than reverting the chart bump.

.superpowers/sdd/2026-08-21-in-binary-firewall-lifecycle/task-5-review-verdict.md
rkschamer: Nope, this can be removed of course! Thanks for highlighting

rkschamer:
Yes...This makes sense. I'll address this.

Q: With the changes to the go-pmtud image (https://github.com/sapcc/go-pmtud/pull/79/changes#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R17) where will be no shell available once https://github.com/sapcc/go-pmtud/pull/79 is merged. So, even we have the flag in the helm chart the preStart/Stop scripts would only be work with an image version before https://github.com/sapcc/go-pmtud/pull/79.

Is this okay or should we also delay the `FROM scratch` runtime image...However, this brings me to another concern: having in the nftable rules set by the scripts and by the go binary might cause issues.

**Is there a way to prevent the promotion to production based on the chart version?** If not, we might need to add the  toggle also for go-pmtud.

rkschamer:
> ...
> **Is there a way to prevent the promotion to production based on the chart version?** If not, we might need to add the toggle also for go-pmtud.

I've double checked with Claude and it could answer all of my questions.  Based on the insights Claude provided, here is my proposal to make the rollout to QA first.

## Current Problem

Once the scratch image (go-pmtud#79) runs, there's no shell in the 
container. The preStop hook `exec ["/scripts/iptables-stop.sh"]` fails on 
scratch — it needs the legacy image. Plus the binary's `fw.Setup()` 
runs unconditionally, so having both the init container AND binary manage 
NFLOG rules creates conflict.

## Change Sequence 

1. DO NOT merge go-pmtud PR yet
2. `gh workflow run "Container Registry GHCR" --ref <pr-branch>` → produces 
   `sha-<sha>` image without merging master
3. Merge helm-charts PR (toggle, default false — prod unchanged)
4. Merge cc-kube-secrets QA override:
   ```yaml
   # values/helm/metal/qa-de-1/qa-de-1/go-pmtud.yaml
   images:
     pmtud:
       tag: sha-<from-step-2>
   pmtud:
     inBinaryLifecycle: true
     ```

defo89 (approved):

PR #12641, opened 2026-08-24
Add pod-level TLS termination for Keystone and Barbican - 007 WP-4

bbobrov:
Terminates TLS inside the API pod with Apache (mod_ssl) instead of at the
ingress. The API is exposed on port 443 through a LoadBalancer Service that
pins a configured public IP, with a matching DNS A record. All behavior is
gated behind `tls.enabled` and is off by default, so existing deployments
render unchanged.

- Keystone chart: 0.14.0
- Barbican chart: 0.10.0

## Changes

* Terminate TLS in Apache via read-only conf (mod_ssl, no runtime `a2enmod`),
  with TLS 1.2/1.3 cipher hardening included outside the VirtualHost and the
  packaged default site dropped.
* Add HTTP security response headers (HSTS, CSP, X-Frame-Options,
  X-Content-Type-Options, Referrer-Policy) on the TLS endpoint, and stop
  trusting client-supplied `Forwarded`/`X-Forwarded-*` and trust-key headers
  when the pod terminates TLS directly.
* Keystone only: reconstruct the x509 client-certificate headers (`cc_x509`)
  from the verified TLS session and mount the client-cert trust anchor, so
  client-certificate auth keeps working in the pod.
* Keystone only: store the OIDC federation session in the client cookie so it
  works across multiple replicas.
* Expose the API through a `type: LoadBalancer` Service on port 443 that pins a
  configured public IP (via `spec.externalIPs` by default, or the Calico
  `loadBalancerIPs` annotation when `calicoLoadBalancer: true`), plus a DNS A
  record. The public ingress renders only when TLS is disabled.
* Manage certificates with a bootstrap Job and a weekly rotation CronJob backed
  by cert-manager (issuer kind configurable, defaults to `ClusterIssuer`), with
  TLS-manager RBAC as a pre-install hook and `fsGroup` so the non-root tls-init
  container can read the Secret.
* Require explicit key generation/wrapping/storage choices when TLS is enabled,
  validated in-chart, with an `allowInsecureStorage` escape hatch for
  non-production.
* Add an optional blackbox endpoint Probe (off by default) and Prometheus alerts
  for certificate expiry, stale rotation, and rotation failures.

## Upgrade note

Keystone now stores the OIDC federation session in the client cookie so it works
across multiple replicas. On upgrade, existing federated users re-authenticate
once. Operators with unusually large OIDC claim sets should confirm the ingress
`large_client_header_buffers` is adequate.


rajivmucheli (dismissed):

openstack/barbican/templates/etc/_wsgi-barbican.conf.tpl
```diff
@@ -22,6 +22,36 @@ WSGIDaemonProcess barbican-api processes={{ .Values.api.processes | default 1 }}
...
+    WSGIApplicationGroup %{GLOBAL}
+    WSGIPassAuthorization On
+
+    <IfVersion >= 2.4>
+        ErrorLogFormat "%{cu}t %M"
+    </IfVersion>
+    ErrorLog /dev/stderr
+
+    KeepAliveTimeout 61
+</VirtualHost>
```
rajivmucheli, line 58: The TLS VirtualHost has no `CustomLog` directive; external HTTPS requests to Barbican produce no access log. Could we mirror the `CustomLog` block from the internal VirtualHost?

openstack/barbican/templates/etc/_wsgi-barbican.conf.tpl
```diff
@@ -22,6 +22,36 @@ WSGIDaemonProcess barbican-api processes={{ .Values.api.processes | default 1 }}
...
+    WSGIScriptAlias / /var/www/cgi-bin/barbican/barbican-wsgi-api
+    WSGIApplicationGroup %{GLOBAL}
+    WSGIPassAuthorization On
+
+    <IfVersion >= 2.4>
+        ErrorLogFormat "%{cu}t %M"
+    </IfVersion>
+    ErrorLog /dev/stderr
+
+    KeepAliveTimeout 61
```
rajivmucheli, line 57: The internal VirtualHost has `LimitRequestBody 114688` but the TLS VirtualHost doesn't — external callers could send unbounded request bodies. Should we add it here too?

openstack/barbican/values.yaml
```diff
@@ -489,6 +497,37 @@ ingress:
...
+  # keyStorage: internal-k8s-secret | k8s-secret | vault-secret
+  #   internal-k8s-secret (recommended): key written directly to the K8s API; stays in-cluster, never in Git
+  #   k8s-secret: Secret manifest emitted for Git commit + GitOps operator review
+  # Escape hatch: allowInsecureStorage: true also passes --allow-insecure-storage to the binary,
+  #   permitting wrapping=none + storage=k8s-secret (development only)
+  allowInsecureStorage: false
+  # TLS lifecycle manager binary for bootstrap Job and rotation CronJob
+  lifecycle:
+    image: keppel.eu-de-1.cloud.sap/ccloud/tls-lifecycle-manager
+    imageTag: latest
```
rajivmucheli: `imageTag: latest` makes the bootstrap Job and rotation CronJob non-reproducible — a bad image push could silently break the next rotation. Could this be pinned to a specific version?

openstack/barbican/templates/_helpers.tpl
```diff
@@ -39,3 +44,20 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+  {{- if not .Values.tls.keyGeneration }}
+    {{- fail "tls.keyGeneration is required when tls.enabled (options: go-crypto, hsm-entropy, hsm-full, tpm-entropy)" }}
+  {{- end }}
+  {{- if not .Values.tls.keyWrapping }}
+    {{- fail "tls.keyWrapping is required when tls.enabled (options: none, vault-transit, hsm, tpm)" }}
+  {{- end }}
+  {{- if not .Values.tls.keyStorage }}
+    {{- fail "tls.keyStorage is required when tls.enabled (options: internal-k8s-secret, k8s-secret, vault-secret)" }}
+  {{- end }}
+  {{- if and (eq .Values.tls.keyWrapping "none") (eq .Values.tls.keyStorage "k8s-secret") (not .Values.tls.allowInsecureStorage) }}
```
rajivmucheli, line 59: The guard only blocks `wrapping=none + storage=k8s-secret`; `wrapping=none + storage=internal-k8s-secret` stores an unwrapped private key in Kubernetes without requiring `allowInsecureStorage: true`. Should `internal-k8s-secret` be covered by the same check?

openstack/barbican/templates/rbac-tls-manager.yaml
```diff
@@ -0,0 +1,48 @@
...
+  labels:
+    app: {{ template "fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    component: barbican
+    type: tls-manager
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-10"
+    "helm.sh/hook-delete-policy": before-hook-creation
```
rajivmucheli, line 15: The SA/Role/RoleBinding use `before-hook-creation` delete policy, so they're briefly absent at the start of each upgrade. If the weekly rotation CronJob fires in that window it will fail with "ServiceAccount not found" — is this an acceptable risk?

openstack/barbican/templates/service-tls.yaml
```diff
@@ -0,0 +1,35 @@
...
+    type: api-tls
+    component: barbican
+  annotations:
+    {{- if .Values.calicoLoadBalancer }}
+    projectcalico.org/loadBalancerIPs: '["{{ $externalIP }}"]'
+    {{- end }}
+    {{- include "utils.linkerd.pod_and_service_annotation" . | indent 4 }}
+spec:
+  type: LoadBalancer
+  externalTrafficPolicy: Local
```
rajivmucheli, line 23: `externalTrafficPolicy: Local` with no PodDisruptionBudget means a single pod eviction drops all external HTTPS traffic immediately. Should we add a PDB with `minAvailable: 1`?

rajivmucheli:
just realised, should we also split the PR's per service ?

openstack/barbican/values.yaml
bbobrov: I actually pin it in per-region secrets. Do you think we should pin it here too?

openstack/barbican/templates/_helpers.tpl
bbobrov: No, it should not. The whole point of `internal-k8s-secret` is that it is safe because operator does not get it as an output. This option is the accepted behavior in the ADR.

bbobrov:
> just realised, should we also split the PR's per service ?

I don't see why this would be needed. I would rather leave it as is, but if you want i can do it.

openstack/barbican/templates/rbac-tls-manager.yaml
bbobrov: It should be fine. If the certificate is expiring, there will be alerts, and a human could manually run the job. In the followup PRs we could add more alerts to find out about issues earlier.

openstack/barbican/templates/etc/_wsgi-barbican.conf.tpl
bbobrov: thanks, done in a20d151

openstack/barbican/templates/etc/_wsgi-barbican.conf.tpl
bbobrov: thanks, done in a20d151

openstack/barbican/values.yaml
bbobrov: thanks, done in a20d151

openstack/barbican/templates/service-tls.yaml
bbobrov: thanks, done in a20d151

openstack/barbican/alerts/openstack/openstack-barbican.alerts
```diff
@@ -116,3 +116,17 @@ groups:
...
+  - alert: OpenstackBarbicanTLSRotationFailed
+    expr: kube_job_status_failed{job_name=~"barbican-tls-rotation.*"} > 0
+    for: 1h
+    labels:
+      context: security
+      dashboard: barbican
+      service: barbican
+      severity: warning
+      tier: os
+      support_group: identity
```
rajivmucheli, line 134: Missing `no_alert_on_absence: "true"` — the other two TLS alerts both have it. When `tls.enabled: false` the jobs do not exist so the metric is absent; consistent labelling avoids spurious absence alerts.

```suggestion
      support_group: identity
      no_alert_on_absence: "true"
```

openstack/barbican/alerts/kubernetes/openstack-barbican.alerts
```diff
@@ -17,3 +17,38 @@ groups:
...
+      playbook: 'docs/support/playbook/barbican/alerts/tls-certificate-expiry/#openstackbarbicantlscertificateexpirescritical'
+    annotations:
+      description: The TLS certificate served by Barbican at {{ $labels.instance }} expires in less than 7 days. External HTTPS access will break when it expires. Check the rotation CronJob and cert-manager.
+      summary: Barbican served TLS certificate expires within 7 days
+
+  - alert: OpenstackBarbicanTLSRotationStale
+    expr: >
+      time() - kube_cronjob_created{cronjob="barbican-tls-rotation"} >= 8*24*60*60
+      and on(cronjob) (time() - kube_cronjob_status_last_successful_time{cronjob="barbican-tls-rotation"}
+        or on(cronjob) kube_cronjob_status_last_schedule_time{cronjob="barbican-tls-rotation"}) >= 8*24*60*60
```
rajivmucheli, line 41: The PromQL expressions hard-code `cronjob="barbican-tls-rotation"`, but the CronJob is named `{{ .Release.Name }}-tls-rotation`. This only fires if the Helm release is named exactly `barbican`. Using a regex matcher makes the alert release-name-agnostic.

```suggestion
      time() - kube_cronjob_created{cronjob=~".*-tls-rotation"} >= 8*24*60*60
      and on(cronjob) (time() - kube_cronjob_status_last_successful_time{cronjob=~".*-tls-rotation"}
        or on(cronjob) kube_cronjob_status_last_schedule_time{cronjob=~".*-tls-rotation"}) >= 8*24*60*60
```

bbobrov: we have 2 jobs - barbican-tls-rotation and keystone-tls-rotation. Will the regex catch them both then? If yes, we should not do it to not get confused which job exactly failed.

bbobrov: We have 2 options what to do:
1. Just accept it as is. We already hardcode "barbican" in some places, for example, in https://github.com/sapcc/helm-charts/blob/master/openstack/barbican/alerts/openstack/openstack-barbican.alerts#L54.
2. Use helm templating for alerts, like maia does: https://github.com/sapcc/helm-charts/tree/master/openstack/maia/alerts/openstack

I myself would rather stay with 1, but i don't have a hard preference. What do you think?

openstack/barbican/alerts/openstack/openstack-barbican.alerts
bbobrov: thanks, done in ea666f2

openstack/barbican/alerts/kubernetes/openstack-barbican.alerts
rajivmucheli: fine with me to go with 1.

openstack/keystone/templates/deployment-api.yaml
```diff
@@ -54,12 +54,24 @@ spec:
         {{- if .Values.federation.saml.enabled }}
         federation-saml-hash: {{ include (print $.Template.BasePath "/federation-saml.yaml") . | sha256sum }}
         {{- end }}
+        {{- if .Values.tls.enabled }}
+        tls-secret-hash: {{ .Values.tls.secretName | default "keystone-tls" }}
```
tz3: Is this a `tls-secret-hash` or `tls-secret-name` ?

openstack/barbican/alerts/kubernetes/openstack-barbican.alerts
tz3, line 41: The `o`r fallback returns a raw Unix timestamp (~1.75B), not elapsed seconds. `1,750,000,000 >= 691,200` is always true, so this alert fires constantly whenever `last_successful_time` is absent. 

Fix: moves `time()` - outside so it applies to whichever branch the or picks.

openstack/keystone/templates/job-tls-bootstrap.yaml
```diff
@@ -0,0 +1,43 @@
...
+    app: {{ template "fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    component: keystone
+    type: tls-bootstrap
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": before-hook-creation
+spec:
```
tz3, line 17: What happens to `helm upgrade` if this Job hangs ?

openstack/keystone/templates/deployment-api.yaml
```diff
@@ -244,6 +282,28 @@ spec:
               mountPath: /etc/shibboleth/metadata
               readOnly: true
             {{- end }}
+            {{- if .Values.tls.enabled }}
```
tz3, line 287: `status-off.conf` is mounted unconditionally (line 247) but apache-`hardening.conf` is gated here on TLS. Shouldn't `ServerTokens Prod` and `TraceEnable Off` apply regardless of TLS mode?

openstack/barbican/alerts/kubernetes/openstack-barbican.alerts
```diff
@@ -17,3 +17,38 @@ groups:
     annotations:
       description: 'barbican-api pod {{ $labels.pod }} not ready; recreate pod. See playbook.'
       summary: 'barbican-api pod not ready'
+
+  - alert: OpenstackBarbicanTLSCertificateExpiresCritical
+    expr: (probe_ssl_earliest_cert_expiry{instance=~"https://keymanager-3\\..*"} - time()) / 86400 < 7
```
tz3, line 22: @rajivmucheli Does `keymanager-3\\..*` match all regions, or only the ones where the endpoint happens to start with `keymanager-3`? What happens in a region where it doesn't?

tz3 (commented):
A few things worth checking before this lands.

openstack/keystone/templates/deployment-api.yaml
bbobrov: Thanks, done in e634efd1db4a3a0ab55bf6c9d2a2a40373e2eb70

openstack/keystone/templates/job-tls-bootstrap.yaml
bbobrov: It is a pre-upgrade hook, so a hang blocks and then fails the upgrade. That's intended: we don't want pods coming up without a valid cert. The old pods keep running.

openstack/barbican/alerts/kubernetes/openstack-barbican.alerts
bbobrov: It is keymanager-3 everywhere.

openstack/keystone/templates/deployment-api.yaml
bbobrov: `status-off.conf` mirrors the existing disabling of mod_status, which we do unconditionally; it just changes how it's done: now via config rather than a command. The version banner and TRACE method are only reachable when Apache serves external traffic directly, which is the TLS path; without TLS, requests still go through the nginx ingress. "Hardening regardless of TLS" might be a good project, but a separate one.

openstack/barbican/alerts/kubernetes/openstack-barbican.alerts
bbobrov: The point of this alert is simple: warn us if Barbican's certificate has stopped being renewed.

To decide that, we look at two numbers Kubernetes reports about the rotation job:
- when it last finished successfully
- when it was last started

Normally we only need the first one: "did it succeed within the last 8 days?" If yes, everything is fine and the alert stays quiet.

The awkward case is a job that has *never* succeeded even once. Then the "last success" number doesn't exist at all, so there's nothing to measure against. To cover that, we fall back to the "last started" number. Because of how the arithmetic works out, that fallback always lands in the "too old" range, so a job that has never once succeeded is automatically treated as a problem — which is exactly what we want.

To avoid a false alarm right after a fresh install (before the first rotation has even had a chance to run), there's a guard that only starts checking once the job has existed for at least 8 days.

On moving `time()` inside the brackets: that would switch the fallback from "never succeeded = problem" to "how long since it was last started." The problem is that a rotation which runs every week but fails every time would then look recent and healthy, so we'd never be warned — and that's the exact situation we most want to hear about.

This follows the same pattern the cinder and nova cronjob alerts already use.

rajivmucheli (commented):
Did another review, found the below :

openstack/barbican/templates/rbac-tls-manager.yaml
```diff
@@ -0,0 +1,48 @@
...
+kind: RoleBinding
+metadata:
+  name: {{ .Release.Name }}-tls-manager
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-10"
+    "helm.sh/hook-delete-policy": before-hook-creation
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Release.Name }}-tls-manager
```
rajivmucheli, line 43: **[Critical] RoleBinding subjects missing `namespace:`**

Kubernetes can't resolve a ServiceAccount subject without a `namespace` field, so this RoleBinding is effectively a no-op right now — the bootstrap Job and rotation CronJob will both get HTTP 403 on every Secrets/CertificateRequests call. Same issue in the Keystone copy of this file.

```suggestion
    name: {{ .Release.Name }}-tls-manager
    namespace: {{ .Release.Namespace }}
```

openstack/barbican/templates/rbac-tls-manager.yaml
```diff
@@ -0,0 +1,48 @@
...
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-10"
+    "helm.sh/hook-delete-policy": before-hook-creation
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "create", "update"]
+  - apiGroups: ["cert-manager.io"]
+    resources: ["certificaterequests"]
+    verbs: ["create", "get", "delete", "watch"]
```
rajivmucheli, line 31: **[Critical] Missing `list` verb on `certificaterequests`**

Kubernetes informers always do a LIST first to get a `resourceVersion` before opening a WATCH stream. Without `list`, the initial call returns 403, the client can't prime its cache, and the lifecycle manager will crash-loop. Same in the Keystone RBAC file.

```suggestion
    verbs: ["create", "get", "delete", "watch", "list"]
```

openstack/keystone/templates/deployment-api.yaml
```diff
@@ -211,6 +242,12 @@ spec:
               mountPath: /etc/apache2/conf-enabled/wsgi-keystone.conf
               subPath: wsgi-keystone.conf
               readOnly: true
+            - name: keystone-etc
+              mountPath: /etc/apache2/conf-enabled/status-off.conf
+              subPath: status-off.conf
+              readOnly: true
+            - name: apache-sites
```
rajivmucheli, line 251: **[Critical] `apache-sites` emptyDir mount not gated on `tls.enabled`**

This mount replaces `/etc/apache2/sites-enabled` with an empty directory for *every* Keystone deployment on upgrade, including ones where `tls.enabled: false` (the default). Any Apache site config shipped in the container image (like `000-default.conf`) gets silently wiped, breaking existing deployments. The volume definition at line ~364 needs the same guard.

```suggestion
            {{- if .Values.tls.enabled }}
            - name: apache-sites
              mountPath: /etc/apache2/sites-enabled
            {{- end }}
```

openstack/keystone/templates/deployment-api.yaml
```diff
@@ -302,6 +361,8 @@ spec:
           emptyDir: {}
         - name: wsgi-keystone
           emptyDir: {}
+        - name: apache-sites
```
rajivmucheli, line 368: **[Critical, same as above] `apache-sites` emptyDir volume needs a `tls.enabled` guard too**

```suggestion
        {{- if .Values.tls.enabled }}
        - name: apache-sites
          emptyDir: {}
        {{- end }}
```

openstack/barbican/alerts/kubernetes/openstack-barbican.alerts
rajivmucheli, line 41: **[High] Alert fires immediately on fresh installs**

When the CronJob has never succeeded, `kube_cronjob_status_last_successful_time` is absent and the `or` fallback returns a raw epoch timestamp (~1.7e9). Comparing that directly against `8*24*60*60` (691200) is always true, so the alert fires from day one as a permanent false positive. The fallback needs to be wrapped in `time() - ...` just like the primary expression. Same bug in the Keystone alerts file.

```suggestion
        or on(cronjob) time() - kube_cronjob_status_last_schedule_time{cronjob="barbican-tls-rotation"}) >= 8*24*60*60
```

openstack/barbican/templates/_helpers.tpl
```diff
@@ -39,3 +44,24 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
```
rajivmucheli, line 59: **[High] Insecure storage guard misses `internal-k8s-secret`**

The check only blocks `keyWrapping=none + keyStorage=k8s-secret`. Using `internal-k8s-secret` (which is labeled as the recommended option in `values.yaml`) with `keyWrapping=none` slips through silently and writes an unencrypted private key to a plain-text Kubernetes Secret. Same gap in the Keystone helpers.

```suggestion
  {{- if and (eq .Values.tls.keyWrapping "none") (or (eq .Values.tls.keyStorage "k8s-secret") (eq .Values.tls.keyStorage "internal-k8s-secret")) (not .Values.tls.allowInsecureStorage) }}
```

openstack/barbican/templates/etc/_wsgi-barbican.conf.tpl
```diff
@@ -22,6 +22,52 @@ WSGIDaemonProcess barbican-api processes={{ .Values.api.processes | default 1 }}
...
 Listen 0.0.0.0:{{ .Values.api_port_internal }}
 
+{{- if .Values.tls.enabled }}
+# External HTTPS endpoint
+# mod_ssl is loaded via a conf-enabled snippet, which is parsed after
+# ports.conf, so its ssl_module-gated Listen 443 does not apply here.
+Listen 0.0.0.0:443
+
+# TLS hardening at server level (must be outside VirtualHost)
+Include /etc/apache2/conf-enabled/tls-hardening.conf
```
rajivmucheli: **[High] `tls-hardening.conf` loaded twice**

Apache auto-loads everything in `conf-enabled/` (via `IncludeOptional conf-enabled/*.conf` in `apache2.conf`), so `tls-hardening.conf` is already picked up before this file is even parsed. The explicit `Include` here causes `SSLSessionCache` to be re-initialized, which generates warnings and can prevent Apache from starting depending on the mod_ssl build. Same in the Keystone wsgi conf. The fix is to just remove this line — the comment above actually already explains *why* this snippet lives in `conf-enabled/`.

```suggestion

```

openstack/keystone/templates/etc/_federation-oidc.conf.tpl
```diff
@@ -16,6 +16,10 @@ OIDCClientSecret "{{ .Values.federation.oidc.client_secret | include "resolve_se
 OIDCPKCEMethod "S256"
 OIDCCryptoPassphrase "{{ .Values.federation.oidc.crypto_passphrase | include "resolve_secret_urlquery" }}"
 OIDCAuthRequestParams idp=#
+# Store the session in the client cookie (encrypted with the shared crypto
+# passphrase) instead of the default per-pod server cache, so the post-callback
+# request can be served by any keystone replica.
+OIDCSessionType client-cookie
```
rajivmucheli, line 22: **[Medium] `OIDCSessionType client-cookie` applied unconditionally**

I totally get why this is here for multi-replica Keystone — server-side sessions break across replicas, makes complete sense. Worth noting though: this line is not gated on `tls.enabled`, so on upgrade any Keystone with `federation.oidc.enabled: true` and `tls.enabled: false` will have all in-flight OIDC sessions immediately invalidated. Also, `mod_auth_openidc` requires HTTPS to set the `Secure` flag on the session cookie — without it some browsers will drop it entirely.

Might be worth guarding with `{{- if .Values.tls.enabled }}` or at least adding a note in `values.yaml` that client-cookie mode requires HTTPS at the pod level.

openstack/barbican/templates/job-tls-bootstrap.yaml
```diff
@@ -0,0 +1,43 @@
...
+metadata:
+  name: {{ .Release.Name }}-tls-bootstrap
+  labels:
+    app: {{ template "fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    component: barbican
+    type: tls-bootstrap
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
```
rajivmucheli, line 14: **[Medium] TLS validation only fires in hook context**

`barbican.tls.validate` is called at the top of this file, but this template only renders when Helm processes hooks. A `helm template --no-hooks` CI run (or any tooling that strips hooks) will skip validation entirely and happily render an insecure config without any error. Consider also calling `{{- include "barbican.tls.validate" . }}` in the main Deployment template (inside the `if .Values.tls.enabled` block) so it fires on every render path.

openstack/barbican/templates/cronjob-tls-rotation.yaml
```diff
@@ -0,0 +1,45 @@
+{{- if .Values.tls.enabled }}
```
rajivmucheli, line 1: **[Medium] CronJob doesn't call `barbican.tls.validate`**

The bootstrap Job has `{{- include "barbican.tls.validate" . }}` on line 2, but the CronJob doesn't. If someone introduces a broken TLS config after the initial install (e.g. via a `helm upgrade --set ...`), `helm upgrade` won't catch it — the CronJob renders fine and only fails at runtime. Adding the validate call here closes that gap.

```suggestion
{{- if .Values.tls.enabled }}
{{- include "barbican.tls.validate" . }}
```

openstack/barbican/templates/cronjob-tls-rotation.yaml
```diff
@@ -0,0 +1,45 @@
...
+metadata:
+  name: {{ .Release.Name }}-tls-rotation
+  labels:
+    app: {{ template "fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    component: barbican
+    type: tls-rotation
+spec:
+  schedule: "0 3 * * 1"
```
rajivmucheli: **[Minor] Hardcoded rotation schedule**

The schedule is hardcoded to `0 3 * * 1` (Monday 3am) with no values override. When barbican and keystone are deployed in the same cluster they'll fire simultaneously, which could hit cert-manager request rate limits. Making this configurable via a values key (with this as the default) would let operators stagger them easily.

```suggestion
  schedule: {{ .Values.tls.rotationSchedule | default "0 3 * * 1" | quote }}
```

openstack/keystone/templates/secret-x509-ca.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.services.ingress.x509.ca }}
```
rajivmucheli, line 1: **[Minor] x509-CA Secret rendered even when Ingress is disabled**

When `tls.enabled: true`, the Ingress is suppressed (correctly gated in `ingress-api.yaml`), but this Secret still renders unconditionally if `services.ingress.x509.ca` is set. The only consumer of this Secret is the nginx-ingress `auth-tls-secret` annotation, which won't exist in TLS-pod mode. The orphaned Secret could mislead operators into thinking mutual-TLS client-cert enforcement via the Ingress is still active when it isn't.

```suggestion
{{- if and .Values.services.ingress.x509.ca (not .Values.tls.enabled) }}
```

openstack/barbican/templates/rbac-tls-manager.yaml
bbobrov: Thanks. This actually already works today: we deployed it to qa-de-3 (keystone) and qa-de-1 (barbican) and the certificates were created. If the RoleBinding gave no permissions, the bootstrap Job could not create the CertificateRequest and no certificate would appear — but it did.

The reason it works: in a RoleBinding (which lives in one namespace), when a ServiceAccount subject has no `namespace` field, Kubernetes treats it as a ServiceAccount in the same namespace as the RoleBinding. Here the ServiceAccount, the RoleBinding and the Job are all in the same namespace, so the permission is granted. The `namespace` field is only required for a ClusterRoleBinding, not for a RoleBinding.

So it was not a no-op. That said, every other chart in this repo writes the namespace out explicitly, so I added it here too to match that style and make it clearer. Changed in 9e26fca652.

openstack/keystone/templates/deployment-api.yaml
bbobrov: Thanks. This does not break existing non-TLS deployments, for two reasons:

1. Keystone does not serve traffic from `sites-enabled`. It serves from the WSGI config in `conf-enabled` on port 5000. The `000-default` file in `sites-enabled` only listens on port 80, which nothing uses here, so emptying `sites-enabled` does not stop Keystone from working.
2. On master the start script never removed `000-default` either — it only switched off the status module — so the default site was simply unused there.

Still, blanking `sites-enabled` is only really needed when Apache terminates TLS inside the pod; it has no purpose when TLS is off. So I put both the mount and the volume behind the `tls.enabled` switch. With TLS off, `sites-enabled` is now left exactly as the image ships it, which fully removes your concern. Changed in 1d0ff4832f.

openstack/keystone/templates/deployment-api.yaml
bbobrov: Thanks — same fix as the mount above: the `apache-sites` volume is now created only when `tls.enabled` is true, so nothing changes for non-TLS deployments. Changed in 1d0ff4832f.

openstack/barbican/templates/rbac-tls-manager.yaml
bbobrov: Thanks. In this case `list` is not needed, and the qa deployments show it: in qa-de-3 and qa-de-1 the certificates were created successfully with only the current permissions (no `list`). If a `list` call had happened, it would have failed with 403 and no certificate would exist.

`list` and `watch` are two separate permissions for two separate API calls. Your point is correct for a long-running controller that keeps a local cache in memory — that kind of client does a `list` first and then a `watch`, so it needs both. But this program is not that kind of client. It is a small one-shot job: it creates one CertificateRequest, watches that single object until it is signed, and then exits. A direct `watch` on one object only needs the `watch` permission and never calls `list`. So the current permissions are enough, and adding `list` would only be an unused permission. I would prefer to keep it minimal, and we can add it later if we ever change this tool to the cache/controller style.

openstack/barbican/alerts/kubernetes/openstack-barbican.alerts
bbobrov: The point of this alert is simple: warn us if Barbican's certificate has stopped being renewed.

To decide that, we look at two numbers Kubernetes reports about the rotation job:
- when it last finished successfully
- when it was last started

Normally we only need the first one: "did it succeed within the last 8 days?" If yes, everything is fine and the alert stays quiet.

The awkward case is a job that has *never* succeeded even once. Then the "last success" number doesn't exist at all, so there's nothing to measure against. To cover that, we fall back to the "last started" number. Because of how the arithmetic works out, that fallback always lands in the "too old" range, so a job that has never once succeeded is automatically treated as a problem — which is exactly what we want.

To avoid a false alarm right after a fresh install (before the first rotation has even had a chance to run), there's a guard that only starts checking once the job has existed for at least 8 days.

On moving `time()` inside the brackets: that would switch the fallback from "never succeeded = problem" to "how long since it was last started." The problem is that a rotation which runs every week but fails every time would then look recent and healthy, so we'd never be warned — and that's the exact situation we most want to hear about.

This follows the same pattern the cinder and nova cronjob alerts already use.

openstack/barbican/templates/_helpers.tpl
bbobrov: Thanks. This is intentional and safe, and it was already discussed in an earlier round. `internal-k8s-secret` is different from `k8s-secret`: with `internal-k8s-secret` the private key is written straight into the Kubernetes API and stays inside the cluster — the operator never gets the key as an output. That is why it is the recommended option and why it does not need `allowInsecureStorage`. The guard only has to block the one truly unsafe combination (`wrapping=none` together with `k8s-secret`, where the key would be handed out as output). This is the behavior agreed in the ADR, so I would keep it as is.

openstack/keystone/templates/secret-x509-ca.yaml
bbobrov: Thanks, but this Secret is not orphaned when TLS is on — Apache uses it. When `tls.enabled` is true, Apache terminates TLS inside the pod and checks client certificates itself using this CA file: it is mounted at `/etc/apache2/x509-ca/ca.crt` (inside the `tls.enabled` block in the Deployment) and referenced by `SSLCACertificateFile` together with `SSLVerifyClient optional` in the Apache config. This is the same client-certificate check nginx used to do at the ingress; we moved it into Apache for the TLS path.

So adding `(not .Values.tls.enabled)` would remove the Secret exactly when Apache needs it and would break client-certificate checking. For that reason I would keep the current condition.

openstack/barbican/templates/etc/_wsgi-barbican.conf.tpl
bbobrov: Thanks, done in 785d9aeb12.

openstack/barbican/templates/job-tls-bootstrap.yaml
bbobrov: Thanks, done in 5377a998da.

openstack/barbican/templates/cronjob-tls-rotation.yaml
bbobrov: Thanks, done in 5377a998da.

openstack/barbican/templates/cronjob-tls-rotation.yaml
bbobrov: Thanks, done in 40e79ffbbe.

openstack/keystone/templates/etc/_federation-oidc.conf.tpl
bbobrov: Thanks. This is fine, I would like to leave it like this.

bbobrov:
I did this:
```
k create job keystone-tls-forcea \
  --from=cronjob/keystone-tls-rotation --dry-run=client -o yaml \
  | sed 's/--renewal-threshold-days=30/--renewal-threshold-days=9999/' \
  | k apply -f -
```
To trigger forced certificate rotation. And it worked, the certificate got reissued by the job.

openstack/barbican/templates/etc/_wsgi-barbican.conf.tpl
```diff
@@ -16,6 +16,49 @@ WSGIDaemonProcess barbican-api processes={{ .Values.api.processes | default 1 }}
 
 Listen 0.0.0.0:{{ .Values.api_port_internal }}
 
+{{- if .Values.tls.enabled }}
+# External HTTPS endpoint
+# mod_ssl is loaded via a conf-enabled snippet, which is parsed after
+# ports.conf, so its ssl_module-gated Listen 443 does not apply here.
+Listen 0.0.0.0:443
+
+<VirtualHost *:443>
```
tz3, line 25: Why in Barbican you didn't set the server name similar to [what has been done in keystone](https://github.com/sapcc/helm-charts/pull/12641/changes#diff-29878c1701849284715723edad04f9ced2ed25bf05adfa054364dc00072dd137R94)?

openstack/keystone/templates/cronjob-tls-rotation.yaml
```diff
@@ -0,0 +1,46 @@
...
+                - --common-name={{ .Values.services.public.host }}.{{ .Values.global.region }}.{{ .Values.global.tld }}
+                - --san-dns={{ .Values.services.public.host }}.{{ .Values.global.region }}.{{ .Values.global.tld }}
+                - --renewal-threshold-days=30
+                - --key-generation={{ required "tls.keyGeneration is required" .Values.tls.keyGeneration }}
+                - --key-wrapping={{ required "tls.keyWrapping is required" .Values.tls.keyWrapping }}
+                - --key-storage={{ required "tls.keyStorage is required" .Values.tls.keyStorage }}
+                {{- if .Values.tls.allowInsecureStorage }}
+                - --allow-insecure-storage
+                {{- end }}
+      backoffLimit: 3
```
tz3, line 45: No `activeDeadlineSeconds` on the job spec — if the lifecycle manager hangs for whatever reason, the pod runs forever and `concurrencyPolicy: Forbid` blocks all subsequent weekly runs until someone manually kills it. is this intentional?

openstack/barbican/templates/cronjob-tls-rotation.yaml
```diff
@@ -0,0 +1,46 @@
...
+                - --common-name={{ include "barbican_api_endpoint_host_public" . }}
+                - --san-dns={{ include "barbican_api_endpoint_host_public" . }}
+                - --renewal-threshold-days=30
+                - --key-generation={{ required "tls.keyGeneration is required" .Values.tls.keyGeneration }}
+                - --key-wrapping={{ required "tls.keyWrapping is required" .Values.tls.keyWrapping }}
+                - --key-storage={{ required "tls.keyStorage is required" .Values.tls.keyStorage }}
+                {{- if .Values.tls.allowInsecureStorage }}
+                - --allow-insecure-storage
+                {{- end }}
+      backoffLimit: 3
```
tz3, line 45: Same comment: https://github.com/sapcc/helm-charts/pull/12641/changes#r3932809451

openstack/keystone/alerts/openstack/keystone.alerts
```diff
@@ -91,3 +91,22 @@ groups:
       description: Api Requests to Keystone is {{ $value }}ops/s
       summary: Api Requests to Keystone is {{ $value }}ops/s
 
+  - alert: OpenstackKeystoneTLSRotationFailed
+    expr: >
+      kube_job_status_failed{job_name=~"keystone-tls-rotation.*"} > 0
```
tz3, line 96: `kube_job_status_failed` is a kube-state-metrics metric — `TLSRotationStale` sits in the `kubernetes/` alerts file for the same reason. Should this one move there too for consistency?

The same goes for barbican..

openstack/keystone/templates/service-api-tls.yaml
```diff
@@ -0,0 +1,33 @@
+{{- if and .Values.tls.enabled .Values.global.keystone_external_ip }}
```
tz3: When `tls.enabled: true` is deployed, the endpoint moves from the nginx ingress IP to the `keystone-public-tls` LoadBalancer IP and the Designate DNS record must be updated to match. When a helm revision without the flag is deployed on top (e.g. a rollback or a parallel deploy), this service is removed and the endpoint becomes unreachable until DNS is reverted.

I was testing some logs, alerts and observed that in qa-de-3: deploying a different chart revision over the TLS-enabled state dropped `keystone-public-tls`, leaving `identity-3.qa-de-3.cloud.sap` unresolvable until the rollback restored the service.

The PR should document: 

1. Which DNS record to update and to what IP when flipping `tls.enabled: true`.
2. Maybe also how to revert if rolling back.

openstack/keystone/values.yaml
```diff
@@ -57,6 +59,38 @@ osprofiler-redis:
   alerts:
     support_group: identity
 
+# Pod-level TLS termination in Apache
+tls:
```
tz3, line 63: When `tls.enabled: true` is flipped, `global.keystone_external_ip` becomes the public endpoint IP and the Designate record for `identity-3.<region>.cloud.sap` needs to be updated to match. The reverse is also true — if you roll back or deploy a revision without the flag, `keystone-public-tls` is gone but DNS still points to its IP until someone fixes it manually.

Hit this in qa-de-3 today: deployed a chart revision without the flag on top of a TLS-enabled state, `keystone-public-tls` disappeared, endpoint was unreachable until rollback.

Worth adding a note in the `tls:` block in `values.yaml` next to the other operator requirements — something like "when enabling, update the Designate record for the identity endpoint to `global.keystone_external_ip`; revert on rollback."

openstack/barbican/templates/etc/_wsgi-barbican.conf.tpl
bbobrov: Because in Barbican it is being set for the whole config on line 2 of this file, and virtualhosts inherit it.

openstack/keystone/values.yaml
bbobrov: This is a cool outage! Could you please describe how you recovered from it?

I am wondering though what to do with it. I doubt anybody will read that comment in helm-charts when switching true->false in the secrets repo.

openstack/keystone/alerts/openstack/keystone.alerts
bbobrov: I agree, let me fix it

openstack/barbican/templates/etc/_wsgi-barbican.conf.tpl
bbobrov: Because in Barbican it is being set for the whole config on line 2 of this file, and all the virtualhosts inherit it.

openstack/keystone/templates/cronjob-tls-rotation.yaml
bbobrov: yeah, sure, lets add it.

openstack/barbican/templates/cronjob-tls-rotation.yaml
bbobrov: done in b3e9980520

openstack/keystone/templates/cronjob-tls-rotation.yaml
bbobrov: done in b3e9980520

openstack/keystone/alerts/openstack/keystone.alerts
bbobrov: done in 35e91a9066

openstack/keystone/values.yaml
bbobrov: Now, if `tls.enabled` is set to false while the TLS LoadBalancer Service for the endpoint is still running, the Helm upgrade stops with an error before it changes anything in the cluster. It also will not make a release - a redeploy with the right flags will work. So this flag is not left on by mistake, the chart also refuses to start if `tls.allowDisable` is set while `tls.enabled` is true. Hopefully this will push an operator to research before proceeding.

This will not help in our case when an older version of the chart is being installed. We ourselves need to keep this in our head.

done in be04e53be8

tz3: I'm worried about it, the guard is useful but doesn't address the race between DNS propagation and Service deletion. Do you have a plan for that, or you think it will be a small outage window accepted as unavoidable?

bbobrov: So far this comment has been about accidental change, not intentional.

If we talk about intentional change, things become difficult. When the switch happens in either direction, there is downtime because of DNS caches, and where those caches are is hard to say.

We can indeed switch to a model where we run both the Service and the ingress at the same time, and switch the dns record between them.

tz3: Will market it as resolve, fine to merge if it is tracked.

bbobrov: done in 5566246783e2550fca16a599bee6bf0ef61516e6

bbobrov: i have unresolved this comment because what i have done doesn't seem to work, i am figuring it out

PR #12708, opened 2026-09-02
system/edith: add storage_awx agent configuration

sureshks25:
Expose AWX base URL and token for the storage_awx agent via Helm values and ConfigMap so region secrets can supply them at deploy time. for EDITH


soofff (approved):

martinpink (approved):
lgmt

PR #12709, opened 2026-09-02
[octavia] Override CPBMU endpoint policy with new role

BenjaminLudwigSAP:


sebageek (dismissed):

m-kratochvil (dismissed):

galkindmitrii (dismissed):

openstack/keystone/templates/etc/_policy.yaml.tpl
```diff
@@ -59,19 +59,11 @@
   'cloud_identity_viewer':%(target.role.name)s or
   'cloud_support_tools_viewer':%(target.role.name)s or
   'cloud_email_admin':%(target.role.name)s or
-  'cloud_inventory_viewer':%(target.role.name)s"
+  'cloud_inventory_viewer':%(target.role.name)s or
+  'lbaas_cpbmu':%(target.role.name)s"
```
bbobrov, line 63: In our today's setup, adding the role to the list of blocklisted roles means that only users from domain ccadmin and Default will be able to get an assignment with it. Is it what you want to do?

If you are wondering why - before it was not the case, but then September 1 happened.

BenjaminLudwigSAP: I want all users to be able to *have* the role, but only us cloud admins should be able to *give* the role to them (so not even their project admins). I was told this is how to implement that behavior. Did I misunderstand?

BenjaminLudwigSAP: Context: This is for enabling usage of the CPBMU endpoint project by project.

bbobrov: > I want all users to be able to have the role, but only us cloud admins should be able to give the role to them (so not even their project admins). I was told this is how to implement that behavior. Did I misunderstand?

You understanding was correct until September 1, became incorrect after September 1.

We currently don't do what you describe. We can start doing it again, for that we need "restricted_roles" or something.

tz3 (changes requested):
Hey @BenjaminLudwigSAP, the blocklist won't work here for eu-de-1. 

When `adminDomainIds` is active, `blocklisted_role_grant_ok` becomes a hard AND condition on `identity:create_grant` which means even cloud admins cannot assign a blocklisted role to a user from outside Default/ccadmin domains. 

Your customer project users would never be able to receive it.

Options:
- Wait for a `restricted_roles` mechanism in keystone (as @bbobrov mentioned)
- Or merge the octavia changes (seed + policy rule) now and hold the keystone blocklist line until that mechanism exists which is niether planned nor tracked.

bbobrov (approved):
I am good with the change.

I see that after the revert there is no special role for this operation at all. Which if fine by me, but we can also have that new role, which will be a usual role to be assigned by project admin.

tz3 (approved):

m-kratochvil (approved):

PR #12733, opened 2026-09-08
chore(pgmetrics): add database name sanitization for RFC 1123 compliance

ArtieReus:
## Summary
Add sanitization of database names to ensure Kubernetes resource names comply with RFC 1123 DNS subdomain naming requirements, enabling support for database names with underscores.

## Problem
When using database names with underscores (e.g., `monsoon-dashboard_production`), the pgmetrics chart fails to deploy because Kubernetes resource names must follow RFC 1123 subdomain rules:
- Only lowercase alphanumeric characters, hyphens, and dots allowed
- Must start and end with an alphanumeric character

Error message:
Deployment.apps "elektra-pgmetrics-monsoon-dashboard_production" is invalid:
metadata.name: Invalid value: a lowercase RFC 1123 subdomain must consist of
lower case alphanumeric characters, '-' or '.', and must start and end with
an alphanumeric character

## Changes
1. **Added helper function** `pgmetrics.sanitizeName` in `_helpers.tpl`:
   - Converts database names to RFC 1123 compliant format
   - Reusable across all templates
   - Documented with usage examples

2. **Updated templates** to use sanitized names for Kubernetes resources:
   - `deployment.yaml`: Deployment name, labels, selectors, ConfigMap file references
   - `alerts.yaml`: PrometheusRule name, alert group names, pod label selectors
   - `configmap-etc.yaml`: ConfigMap data keys

3. **Preserved original database name** for:
   - PostgreSQL connection strings (DATA_SOURCE_URI)
   - Prometheus query expressions (datname label)
   - SQL queries and database operations

4. **Bumped version** to 2.1.9 with changelog entry

### DIFF
--- a/common/pgmetrics/Chart.yaml
+++ b/common/pgmetrics/Chart.yaml
@@ -16,4 +16,8 @@ description: A Helm chart for Postgres Metrics
 #
 # 2.1.0
 # - Added support for inline templating in `.Values.alerts.prometheus`.
-version: 2.1.8 # this version number is SemVer as it gets used to auto bump
+#
+# 2.1.9
+# - Added sanitization of database names to ensure RFC 1123 compliance for Kubernetes resource names.
+#   Database names with underscores (e.g., `monsoon-dashboard_production`) are now supported.
+version: 2.1.9 # this version number is SemVer as it gets used to auto bump
--- a/common/pgmetrics/Chart.yaml
+++ b/common/pgmetrics/Chart.yaml
@@ -16,8 +16,4 @@ description: A Helm chart for Postgres Metrics
 #
 # 2.1.0
 # - Added support for inline templating in `.Values.alerts.prometheus`.
-#
-# 2.1.9
-# - Added sanitization of database names to ensure RFC 1123 compliance for Kubernetes resource names.
-#   Database names with underscores (e.g., `monsoon-dashboard_production`) are now supported.
 version: 2.1.9 # this version number is SemVer as it gets used to auto bump
--- a/common/pgmetrics/templates/_helpers.tpl
+++ b/common/pgmetrics/templates/_helpers.tpl
@@ -12,5 +12,5 @@ Converts to lowercase and replaces any non-compliant characters with hyphens.
 Usage: {{ include "pgmetrics.sanitizeName" "database_name" }}
 */}}
 {{- define "pgmetrics.sanitizeName" -}}
-{{- . | lower | regexReplaceAll "[^a-z0-9-]+" "-" -}}
+{{- regexReplaceAll "[^a-z0-9-]+" (lower .) "-" -}}
 {{- end -}}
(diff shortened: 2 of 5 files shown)


common/pgmetrics/templates/deployment.yaml
SuperSandro2000, line 1: I am not sure what helm is trying to tell us here 

```
rror: YAML parse error on keppel/charts/pgmetrics/templates/deployment.yaml: error converting YAML to JSON: yaml: line 18: block sequence entries are not allowed in this context
helm.go:92: 2026-09-08 13:18:36.756994432 +0200 CEST m=+2.322631635 [debug] error converting YAML to JSON: yaml: line 18: block sequence entries are not allowed in this context
```

common/pgmetrics/Chart.yaml
```diff
@@ -16,4 +16,10 @@ description: A Helm chart for Postgres Metrics
 #
 # 2.1.0
 # - Added support for inline templating in `.Values.alerts.prometheus`.
-version: 2.1.8 # this version number is SemVer as it gets used to auto bump
+#
+# 2.1.9
+# - Added sanitization of database names for RFC 1123 compliance in Kubernetes resource names.
+#   Database names with underscores, spaces, or dots are now supported (e.g., `monsoon-dashboard_production`).
+#   Uses chained replace functions (lower | replace "_" "-" | replace " " "-" | replace "." "-")
+#   instead of regex due to regexReplaceAll compatibility issues.
```
majewsky: This does not need an entry in the changelog, since users do not need to take action to either match backwards-incompatible changes (like in a major version) or take advantage of newly added features they are interested in (like in a minor version).

common/pgmetrics/templates/_helpers.tpl
```diff
@@ -5,3 +5,13 @@
...
 {{- end }}
+
+{{/*
+Sanitize database name to be RFC 1123 compliant for Kubernetes resource names.
+Converts to lowercase and replaces special characters (_, space, .) with hyphens.
+Uses chained replace functions instead of regex due to regexReplaceAll compatibility issues.
+Usage: {{ include "pgmetrics.sanitizeName" "database_name" }}
+*/}}
+{{- define "pgmetrics.sanitizeName" -}}
+{{- . | lower | replace "_" "-" | replace " " "-" | replace "." "-" -}}
```
majewsky: This is unnecessarily brittle. Please remove the last commit and fix the call instead:
```suggestion
{{- regexReplaceAll "[^a-z0-9-]+" (lower .) "-" -}}
```

majewsky (changes requested):
Please clean up the commit history. The commit that adds `trim` is reverted in the subsequent commit. Given the changes requested below, everything can probably be squashed into one commit:

common/pgmetrics/Chart.yaml
ArtieReus: ok, changes applied

common/pgmetrics/templates/_helpers.tpl
ArtieReus: thanks, suggestion applied

majewsky (approved):

PR #12741, opened 2026-09-09
chore(elektra): upgrade helm dependencies

ArtieReus:


ArtieReus:
not needed. It is automated

PR #12754, opened 2026-09-09
[barbican] add batched db-cleanup nanny container

rajivmucheli:
## Summary

Adds a `db-cleanup` sidecar container to the barbican-nanny Deployment that periodically runs `barbican-manage db clean` with the `--batch-size` flag introduced in sapcc/barbican#52. Without batching, the one-shot DELETE exhausts the InnoDB buffer pool on large deployments.

- New script template `scripts/_db-cleanup.sh.tpl` — loops on `$BARBICAN_DB_CLEANUP_INTERVAL`, passes `--batch-size`, `--min-days`, and optional `--clean-unassociated-projects` / `--soft-delete-expired-secrets` flags controlled by env vars
- `scripts-configmap.yaml` — gate changed from `db_secret_move.enabled` to `or db_secret_move.enabled db_cleanup.enabled`; `db-cleanup.sh` added conditionally
- `barbican-nanny-deployment.yaml` — outer gate updated to `or`; `move-secrets` and `db-cleanup` containers each independently conditional; `volumes:` block moved outside both container conditionals (latent bug in original)
- `values.yaml` — new `barbican_nanny.db_cleanup` block (all defaults off / safe)

New values:
```yaml
barbican_nanny:
  db_cleanup:
    enabled: false
    batch_size: 10000        # rows per committed batch
    min_num_days: 14         # threshold age for hard-deletes
    interval: 60             # minutes between cleanup runs
    clean_unassociated_projects: false
    soft_delete_expired_secrets: false
```

## Test plan

- [ ] `helm template` with `db_cleanup.enabled: false` produces identical output to current master (no regressions)
- [ ] `helm template` with `db_cleanup.enabled: true, db_secret_move.enabled: false` renders a single-container nanny Deployment with `db-cleanup` container and correct env vars
- [ ] `helm template` with both enabled renders both containers in the same pod

### DIFF
--- a/openstack/barbican/templates/scripts/_db-cleanup.sh.tpl
+++ b/openstack/barbican/templates/scripts/_db-cleanup.sh.tpl
@@ -34,12 +34,12 @@ while true; do
     if [ "$BARBICAN_DB_CLEANUP_ENABLED" = "True" ] || [ "$BARBICAN_DB_CLEANUP_ENABLED" = "true" ]; then
         date
         /var/lib/openstack/bin/barbican-manage db clean \
-            --min-num-days "$BARBICAN_DB_CLEANUP_MIN_NUM_DAYS" \
+            --min-days "$BARBICAN_DB_CLEANUP_MIN_NUM_DAYS" \
             --batch-size "$BARBICAN_DB_CLEANUP_BATCH_SIZE" \
             --verbose \
             $EXTRA_FLAGS
     fi
-    echo -n "INFO: waiting $BARBICAN_NANNY_INTERVAL minutes before starting the next loop run - "
+    echo -n "INFO: waiting $BARBICAN_DB_CLEANUP_INTERVAL minutes before starting the next loop run - "
     date
-    sleep $(( 60 * $BARBICAN_NANNY_INTERVAL ))
+    sleep $(( 60 * $BARBICAN_DB_CLEANUP_INTERVAL ))
 done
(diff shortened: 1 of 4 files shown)


openstack/barbican/templates/scripts/_db-cleanup.sh.tpl
```diff
@@ -0,0 +1,45 @@
...
+    EXTRA_FLAGS="$EXTRA_FLAGS --soft-delete-expired-secrets"
+fi
+
+echo "INFO: starting a loop to periodically run the barbican db cleanup"
+while true; do
+
+    if [ "$BARBICAN_DB_CLEANUP_ENABLED" = "True" ] || [ "$BARBICAN_DB_CLEANUP_ENABLED" = "true" ]; then
+        date
+        /var/lib/openstack/bin/barbican-manage db clean \
+            --min-num-days "$BARBICAN_DB_CLEANUP_MIN_NUM_DAYS" \
```
s10:
```
'--min-days', '-m'
```
https://github.com/sapcc/barbican/blob/stable/2025.2-m3/barbican/cmd/db_manage.py#L111

openstack/barbican/templates/scripts/_db-cleanup.sh.tpl
```diff
@@ -0,0 +1,45 @@
...
+            --batch-size "$BARBICAN_DB_CLEANUP_BATCH_SIZE" \
+            --verbose \
+            $EXTRA_FLAGS
+    fi
+    echo -n "INFO: waiting $BARBICAN_NANNY_INTERVAL minutes before starting the next loop run - "
```
s10: This runs every minute. Maybe for db cleanup task it would be better to have another interval?

s10 (changes requested):

s10 (approved):

dorneanu (approved):

PR #12763, opened 2026-09-10
[Keystone & Babican] add apache_exporter sidecar for Apache worker/scoreboard metrics

tz3:
## Summary

Adds `apache_exporter` sidecar to both **keystone** and **barbican** API deployments to expose Apache HTTP server metrics.

- Enables `mod_status` via a new `status.conf` mounted at `/etc/apache2/conf-enabled/status.conf`
- Scrapes `/server-status?auto` on localhost and exposes metrics on port 9117
- Wires up `prometheus.io/port_1: "9117"` annotation so the existing `pods_metric_port_1` Prometheus job picks it up
- Feature-gated behind `apacheMetrics.enabled` (defaults to `false`); keystone uses `api.apacheMetrics.enabled`

The `apache-exporter` image is already built and published by CI:
- **Pipeline**: https://ci2.eu-de-2.cloud.sap/teams/foundation/pipelines/apache-exporter
- **Image**: `keppel.eu-de-1.cloud.sap/ccloud/apache-exporter:latest`
- **Source**: https://github.wdf.sap.corp/sap-cloud-infrastructure/apache-exporter

## Metrics exposed

- `apache_workers` (busy/idle worker count)
- `apache_scoreboard` (per-state worker slots)
- `apache_accesses_total` (total request count, requires ExtendedStatus On)

## Test plan

- [x] Deployed to qa-de-3 (secrets PR: https://github.wdf.sap.corp/cc/secrets/pull/21834)
- [x] Keystone pods running 4/4 with apache-exporter sidecar
- [x] Barbican pods running 5/5 with apache-exporter sidecar
- [x] Metrics confirmed on port 9117: `apache_scoreboard`, `apache_workers`, `apache_accesses_total`
- [ ] Verify `pods_metric_port_1` Prometheus job scrapes the targets end-to-end

### DIFF
--- a/openstack/keystone/values.yaml
+++ b/openstack/keystone/values.yaml
@@ -207,6 +207,21 @@ api:
         memory: 150Mi
         cpu: 300m
 
+  apacheMetrics:
+    enabled: false
+    port: "9117"
+
+    image: 'shared-app-images/apache-exporter'
+    imageTag: '0.13.4-latest'
+
+    resources:
+      requests:
+        memory: 32Mi
+        cpu: 50m
+      limits:
+        memory: 64Mi
+        cpu: 100m
+
   terminationGracePeriodSeconds: 30
   minReadySeconds: 5
 
--- a/openstack/keystone/templates/configmap-etc.yaml
+++ b/openstack/keystone/templates/configmap-etc.yaml
@@ -25,6 +25,10 @@ data:
   watcher.yaml: |
 {{ include (print .Template.BasePath "/etc/_watcher.yaml.tpl") . | indent 4 }}
 {{- end }}
+{{- if .Values.api.apacheMetrics.enabled }}
+  status.conf: |
+{{ include (print .Template.BasePath "/etc/_status.conf.tpl") . | indent 4 }}
+{{- end }}
 {{- if .Values.api.metrics.enabled }}
   statsd-exporter.yaml: |
     mappings:
--- a/openstack/keystone/templates/deployment-api.yaml
+++ b/openstack/keystone/templates/deployment-api.yaml
@@ -54,9 +54,15 @@ spec:
         {{- if .Values.federation.saml.enabled }}
         federation-saml-hash: {{ include (print $.Template.BasePath "/federation-saml.yaml") . | sha256sum }}
         {{- end }}
+        {{- if and .Values.api.apacheMetrics.enabled (not .Values.api.metrics.enabled) }}
+          {{- fail "api.apacheMetrics.enabled requires api.metrics.enabled (prometheus.io/scrape annotation is set by api.metrics)" }}
+        {{- end }}
         {{- if .Values.api.metrics.enabled }}
         prometheus.io/scrape: "true"
         prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
+        {{- end }}
+        {{- if .Values.api.apacheMetrics.enabled }}
+        prometheus.io/port_1: {{ .Values.api.apacheMetrics.port | quote }}
         {{- end }}
         {{- include "utils.linkerd.pod_and_service_annotation" . | indent 8 }}
     spec:
@@ -248,6 +254,12 @@ spec:
               mountPath: /etc/apache2/mods-available/mpm_event.conf
               subPath: mpm_event.conf
               readOnly: true
+            {{- if .Values.api.apacheMetrics.enabled }}
+            - name: keystone-etc
+              mountPath: /etc/apache2/conf-enabled/status.conf
+              subPath: status.conf
+              readOnly: true
+            {{- end }}
             - name: keystone-etc
               mountPath: /etc/keystone/access_rules.json
               subPath: access_rules.json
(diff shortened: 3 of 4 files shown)


rajivmucheli:
Can this be implemented for barbican as well ?

rajivmucheli:
I think keystone and barbican are the only 2 openstack services running on apache, hence we could add this exporter to common charts at a later stage.

openstack/keystone/values.yaml
```diff
@@ -207,6 +207,21 @@ api:
         memory: 150Mi
         cpu: 300m
 
+  apacheMetrics:
+    enabled: false
+    port: "9117"
+
+    image: 'shared-app-images/apache-exporter'
```
bbobrov, line 214: where does it come from and who is the maintainer in our team?

tz3:
@rajivmucheli yes we can do the same for barbican, WDYT creating a dedicated pr and do the same?
Or just make a shared one with shared config for both?

rajivmucheli:
> @rajivmucheli yes we can do the same for barbican, WDYT creating a dedicated pr and do the same?
> 
> Or just make a shared one with shared config for both?

Shared config in common folder if there will be continuous development and image update. Else a dedicated PR for barbican.

bbobrov (commented):
Please explain why this change is needed.

openstack/keystone/values.yaml
tz3: > where does it come from and who is the maintainer in our team?

Image added to shared-app-images pipeline: https://github.wdf.sap.corp/cc/base-images/pull/105. Current Keppel image is manually pushed for QA testing only. Maintainer: identity team.

> Please explain why this change is needed.

Related task: https://github.wdf.sap.corp/sap-cloud-infrastructure/foundation-issues/issues/1696

bbobrov:
I am fine with the patch, but lets get the image with the exporter built by our CIs first.

tz3:
@rajivmucheli Updated the PR — it now includes the same apache_exporter sidecar for Barbican as well. Both keystone and barbican have been tested in qa-de-3 and metrics are confirmed on port 9117.

openstack/keystone/values.yaml
tz3: @bbobrov The `apache-exporter` image is managed by us in https://github.wdf.sap.corp/sap-cloud-infrastructure/apache-exporter — it has its own CI pipeline on ci2/foundation that builds and pushes to `keppel.eu-de-1.cloud.sap/ccloud/apache-exporter`: https://ci2.eu-de-2.cloud.sap/teams/foundation/pipelines/apache-exporter

openstack/barbican/templates/api-deployment.yaml
```diff
@@ -43,6 +43,9 @@ spec:
         prometheus.io/scrape: "true"
         prometheus.io/targets: {{ required ".Values.alerts.prometheus missing" .Values.alerts.prometheus | quote }}
         {{- end }}
+        {{- if .Values.apacheMetrics.enabled }}
```
rajivmucheli, line 49: Could we add a validation guard here similar to the one in keystone's `deployment-api.yaml`? In barbican, `prometheus.io/scrape: "true"` is only set when both `metrics.enabled` and `statsd.enabled` are true — so if someone enables `apacheMetrics` without those, the pod gets a `port_1` annotation but Prometheus never discovers it (no `scrape` annotation). A fail-fast check would prevent that silent misconfiguration:

```yaml
{{- if and .Values.apacheMetrics.enabled (not (and .Values.metrics.enabled .Values.statsd.enabled)) }}
  {{- fail "apacheMetrics.enabled requires metrics.enabled and statsd.enabled" }}
{{- end }}
```

openstack/barbican/values.yaml
```diff
@@ -66,6 +66,19 @@ statsd:
       memory: 64Mi
       cpu: 100m
 
+apacheMetrics:
+  enabled: false
+  port: "9117"
```
rajivmucheli: Minor nit: could this be `port: 9117` (integer) rather than a string? Kubernetes expects an integer for `containerPort`, and while Helm strips the quotes in template output so it works today, the values type would be more accurate as a number.

rajivmucheli (commented):
Barbican-specific comments only — the keystone side looks good.

openstack/keystone/values.yaml
```diff
@@ -207,6 +207,21 @@ api:
         memory: 150Mi
         cpu: 300m
 
+  apacheMetrics:
+    enabled: false
+    port: "9117"
+
+    image: 'apache-exporter'
+    imageTag: 'latest'
```
s10, line 215: This imageTag should be overriden with the specific version in secrets repo, with #lastest comment added in qa-de-2

openstack/barbican/templates/api-deployment.yaml
tz3: Done: [90d798c](https://github.com/sapcc/helm-charts/pull/12763/commits/90d798c62a1eb3c2b7d89e629ad8389df1095f57)

openstack/barbican/values.yaml
tz3: Done: [90d798c](https://github.com/sapcc/helm-charts/pull/12763/commits/90d798c62a1eb3c2b7d89e629ad8389df1095f57)

openstack/keystone/values.yaml
tz3: Done here: https://github.wdf.sap.corp/cc/secrets/pull/21892

s10 (approved):
Looks

rajivmucheli (approved):

PR #12769, opened 2026-09-11
fix(metal-operator-remote): exclude DaemonSet from shoot managed resources

defo89:
## Problem

`gardener-resource-manager` fails applying the metaldata DaemonSet to the worker-less metal-api shoot:

```
no matches for kind "DaemonSet" in version "apps/v1"
```

## Cause

The `yq` filter in `build-metal-operator-remote` excluded Service/Webhook kinds but not DaemonSet. The upstream chart renders the metaldata DaemonSet unconditionally, so it slipped into `crds-and-rbac.yaml` and got pushed to the shoot via ManagedResource.

## Fix

- Add `.kind != "DaemonSet"` to the yq filter in the Makefile
- Remove the stale DaemonSet block from crds-and-rbac.yaml
- Bump chart 0.6.47 → 0.6.48

Metaldata SA/Role/RoleBinding kept — needed on the shoot for remote kubeconfig RBAC.


xkonni (approved):

goerangudat (commented):
lgtm

goerangudat (approved):
lgtm

kayrus (approved):

PR #12780, opened 2026-09-15
feat(pmtud): set upd relay backend as default

rkschamer:
Related to https://github.com/sapcc/go-pmtud/pull/80


system/go-pmtud/values.yaml
```diff
@@ -14,19 +14,15 @@ global:
 
 images:
   pmtud:
-    tag: sha-6a36183cbe5955d0ee549890ee2ea295261a6558
-
-iptables:
```
rkschamer, line 19: Cleaning this up, since https://github.com/sapcc/go-pmtud/pull/79 this setting is not for an iptables image, but go-pmtud table is using and controlling this setting

system/go-pmtud/templates/go-pmtud-daemonset.yaml
```diff
@@ -26,7 +26,7 @@ spec:
           - /go-pmtud
           - --nodename=$(NODE_NAME)
           - --ttl={{ required ".Values.pmtud.ttl is required" .Values.pmtud.ttl }}
-          - --nflog_group={{ required ".Values.pmtud.nflogGroup is required" .Values.iptables.nflogGroup }}
+          - --nflog_group={{ required ".Values.pmtud.nflogGroup is required" .Values.pmtud.nflogGroup }}
```
rkschamer, line 29: Message and value location were not aligned

system/go-pmtud/values.yaml
```diff
@@ -14,19 +14,15 @@ global:
...
 pmtud:
   ttl: 10
   metricsPort: 30040
-  ifaceNames:
-  ifaceMtu:
+  nflogGroup: 33
   # relayBackend selects the ICMP replication transport: "l2" (default, raw Ethernet
   # over a shared L2 segment) or "udp" (unicast UDP, works across L3 boundaries).
-  relayBackend: "l2"
+  relayBackend: "udp"
```
defo89, line 25: do we want to use udp by default?

rkschamer: Yes, that's what we need to do

defo89 (dismissed):

system/go-pmtud/values.yaml
rkschamer: Good point. Let's go with l2 for now.


```suggestion
  relayBackend: "l2"
```

defo89 (approved):

PR #12786, opened 2026-09-15
update hermes and opensearch alerts

LeoZhangZXZ:


openstack/hermes/alerts/openstack/api.alerts.tpl
```diff
@@ -279,7 +283,7 @@ groups:
       dashboard: hermes-log-router
       persesDashboard: "https://perses.{{ .Values.global.region }}.{{ .Values.global.tld }}/projects/observability/dashboards/hermes-log-router"
       service: hermes
-      severity: critical
+      severity: warning
```
notque: I don't think we can change this to warning, it's a serious issue that has to be looked into if this is happening.

notque (dismissed):

notque (approved):

PR #12789, opened 2026-09-16
feat(identity): SAML SP cert expiry monitoring — per-region CronJob + alert rules

tz3:
## Summary

- Add `prometheus-exporters/keystone-saml-sp-cert-monitor` Helm chart: a K8s CronJob that runs per-region, filters certs to its own region via `$REGION`, and logs `ALERT: SP cert` / `WARN: SP cert` lines to regional OpenSearch
- Split the single `opensearch_logs_saml_sp_cert_expiry.cfg` query into `_alert` (matches `ALERT: SP cert`) and `_warn` (matches `WARN: SP cert`) so each feeds a distinct Prometheus metric
- Replace `KeystoneSAMLSPCertExpiringSoon` (single severity: warning rule) with two rules:
  - `KeystoneSAMLSPCertExpiryCritical` — `severity: critical`, fires when `_alert_doc_count > 0` (≤90 days / already expired)
  - `KeystoneSAMLSPCertExpiryWarning` — `severity: warning`, fires when `_warn_doc_count > 0` (≤180 days)

## Motivation

Alertmanager routes alerts to Slack channels using the `region` label from each regional Prometheus instance's `external_labels`. A centralised Concourse job (all logs in eu-de-2) cannot produce the correct `region` label for other regions. Per-region CronJobs log to each region's own OpenSearch → Prometheus fires with the correct `region` → correct channel routing without any custom label gymnastics.

Closes foundation-issues #1174 AC1.

## Test plan

- [ ] Deploy to qa-de-1, trigger CronJob manually, confirm ALERT/WARN lines in qa-de-1 OpenSearch
- [ ] Confirm `opensearch_logs_saml_sp_cert_expiry_alert_doc_count` metric appears in qa-de-1 Prometheus
- [ ] Confirm Alertmanager fires alert and routes to correct Slack channel
- [ ] Roll out to production regions via secrets PR Concourse pipeline


viennaa:
Move this tooling into a standalone exporter please.
Common pattern is to have a unix timestamp as a gauge, which shows the expiry date of the certificate.

You then can evaluate it with the current time:

Example:
https://github.com/sapcc/helm-charts/blob/b559e112368b9032e35094a9a916e561028ae77b/prometheus-exporters/k8s-secrets-certificate-exporter/templates/alerts/_cert-expiry.alerts.tpl#L4

This way it is flexible and can be reused.

tz3:
Superseded by # — pivoting to a proper Prometheus exporter per observability team feedback.

PR #12790, opened 2026-09-16
feat(cc-gardener): adopt garden-sap-project PrometheusRule for prometheus-garden

goasl:
## What

Adopt the `garden-sap-project` PrometheusRule (alert `ProjectCountReached`) into
the `cc-gardener` chart, mirroring the gardener-operator `garden-sap-*` rules, so
it is delivered as a `PrometheusRule` CR into the `garden` namespace (label
`prometheus: garden`) and collected by the operator-managed `prometheus-garden`.

This is the first of the `garden-sap-*` rules to be adopted; it doubles as an
end-to-end test of the delivery path before the remaining rules follow.

## How it's delivered

- New file `global/cc-gardener/managedresources/prometheusrules-garden.yaml`.
- Picked up by the existing `templates/managedresource.yaml`, which `tpl`s each
  `managedresources/*.yaml` and wraps every object into a `ManagedResource` +
  `Secret` applied to the `garden` namespace by gardener-resource-manager
  (resulting objects: `mr-pr-garden-sap-project`).
- Prometheus `{{ $value }}` is Helm-escaped (`{{ "{{ $value }}" }}`) so `tpl`
  leaves it intact in the rendered rule.

## Safety / rollout

- Gated behind a new **default-off** flag `garden.prometheusRules.enabled`
  (`values.yaml`), so nothing renders anywhere until it is explicitly enabled
  per cluster. Test-first on a single cluster.
- Chart version bumped `0.38.52` → `0.38.53`. Pinned clusters are unaffected
  until re-pinned.

## Validation

- `helm template` with the flag off → 0 objects; with the flag on → the
  `ManagedResource` + `Secret` pair, `namespace: garden`, label
  `prometheus: garden`, literal `{{ $value }}` preserved.
- `promtool check rules` on the rendered rule groups → SUCCESS, 1 rule.


databus23 (commented):
Thanks for adopting this rule into the chart. I have one architectural concern that blocks this as-is.

**PrometheusRule must not go through ManagedResource**

The file is placed under `managedresources/` and therefore picked up by `templates/managedresource.yaml`, which wraps it in a `ManagedResource` + `Secret` applied by gardener-resource-manager. That delivery path targets the **Gardener virtual API server** (the Garden cluster's shoot-like kube-apiserver), where `monitoring.coreos.com/v1` CRDs are not registered. The `PrometheusRule` CR will never be accepted there, so the `ManagedResource` will remain in an error state.

`PrometheusRule` resources need to land in the **runtime cluster** (the seed/base cluster where `prometheus-garden` actually runs), not in the virtual garden. They must be plain Helm manifests rendered directly into the runtime cluster - not wrapped in a `ManagedResource`.

**Fix:** Move the `PrometheusRule` out of `managedresources/` and into `templates/` as a regular Helm manifest (e.g. `templates/prometheusrule-garden.yaml`). The existing gate flag `garden.prometheusRules.enabled` can stay as-is.

databus23 (changes requested):

goasl:
> Thanks for adopting this rule into the chart. I have one architectural concern that blocks this as-is.
> 
> **PrometheusRule must not go through ManagedResource**
> 
> The file is placed under `managedresources/` and therefore picked up by `templates/managedresource.yaml`, which wraps it in a `ManagedResource` + `Secret` applied by gardener-resource-manager. That delivery path targets the **Gardener virtual API server** (the Garden cluster's shoot-like kube-apiserver), where `monitoring.coreos.com/v1` CRDs are not registered. The `PrometheusRule` CR will never be accepted there, so the `ManagedResource` will remain in an error state.
> 
> `PrometheusRule` resources need to land in the **runtime cluster** (the seed/base cluster where `prometheus-garden` actually runs), not in the virtual garden. They must be plain Helm manifests rendered directly into the runtime cluster - not wrapped in a `ManagedResource`.
> 
> **Fix:** Move the `PrometheusRule` out of `managedresources/` and into `templates/` as a regular Helm manifest (e.g. `templates/prometheusrule-garden.yaml`). The existing gate flag `garden.prometheusRules.enabled` can stay as-is.



> Thanks for adopting this rule into the chart. I have one architectural concern that blocks this as-is.
> 
> **PrometheusRule must not go through ManagedResource**
> 
> The file is placed under `managedresources/` and therefore picked up by `templates/managedresource.yaml`, which wraps it in a `ManagedResource` + `Secret` applied by gardener-resource-manager. That delivery path targets the **Gardener virtual API server** (the Garden cluster's shoot-like kube-apiserver), where `monitoring.coreos.com/v1` CRDs are not registered. The `PrometheusRule` CR will never be accepted there, so the `ManagedResource` will remain in an error state.
> 
> `PrometheusRule` resources need to land in the **runtime cluster** (the seed/base cluster where `prometheus-garden` actually runs), not in the virtual garden. They must be plain Helm manifests rendered directly into the runtime cluster - not wrapped in a `ManagedResource`.
> 
> **Fix:** Move the `PrometheusRule` out of `managedresources/` and into `templates/` as a regular Helm manifest (e.g. `templates/prometheusrule-garden.yaml`). The existing gate flag `garden.prometheusRules.enabled` can stay as-is.

Thanks — fixed in 13c3edfbae. Moved the rule out of `managedresources/` into `templates/prometheusrules-garden.yaml` as a plain Helm manifest, so it renders directly into the runtime cluster where `prometheus-garden` runs, instead of being wrapped in a `ManagedResource` for the virtual garden API server. The `garden.prometheusRules.enabled` gate is unchanged.

Verified end-to-end on `prt-q-eu-de-1`: applied the rendered `PrometheusRule` into ns `garden` → the operator picked it up via `ruleSelector {prometheus: garden}`, generated it into `prometheus-garden-rulefiles-0`, and `prometheus-garden` served it (`ProjectCountReached`, state=inactive/health=ok — inert, threshold 4500 >> ~300 projects). Deleted it again → operator removed it from the rulefiles ConfigMap and the running Prometheus, cluster back to baseline.

PR #12794, opened 2026-09-17
feat(identity): add opensearch query + alert rules for SAML SP cert expiry

tz3:
## Summary

- Adds `opensearch_logs_saml_sp_cert_expiry_alert.cfg` — queries OpenSearch for `ALERT: SP cert` log lines within the last 25h, exposes metric `opensearch_logs_saml_sp_cert_expiry_alert_doc_count`
- Adds `opensearch_logs_saml_sp_cert_expiry_warn.cfg` — same for `WARN: SP cert`, exposes `opensearch_logs_saml_sp_cert_expiry_warn_doc_count`
- Adds alert rules `KeystoneSAMLSPCertExpiryCritical` and `KeystoneSAMLSPCertExpiryWarning` with `support_group: identity`, `for: 1h`

## Context

These files are the observability layer for the SAML SP cert expiry monitoring CronJob (separate PR). The CronJob runs per-region, clones the federation repo, and prints `ALERT:`/`WARN:`/`OK:` lines to stdout. This PR wires those logs into Prometheus alerts.


olandr (approved):

viennaa:
Thanks for bringing up the idea to monitor this via OpenSearch. We should however make this more generic and add the alerts to the (to be built) exporter:

https://github.com/sapcc/helm-charts/pull/12789#issuecomment-5711844653

Please don't merge, the mechanism needs to change. I think this pretty much can be closed.

PR #12797, opened 2026-09-17
[thanos-global] addded ScrapeTargetDown alert for per-support-group target for health monitoring

ashifnihalb:
Fires once per support group (across all regions/clusters) when one or more of their scrape targets are down. One alert per team


ibakshay (approved):
LGTM!

PR #12809, opened 2026-09-21
refactor: update BMC image handling to use imagePath instead of imageURI

davidgrun:
### Problem

Every hardware model entry in `values.yaml` hard-codes the **full**
firmware image URL, including the repo server:

```yaml
bmc:
  imageURI: "https://repo.eu-de-1.cloud.sap/hw-firmware/dell/poweredge-r7615/iDRAC-with-Lifecycle-Controller_..."
```

`repo.eu-de-1.cloud.sap` is baked into **all ~50 model entries**, regardless
of which region the cluster actually runs in. As a result, a cluster in
e.g. `qa-de-1` would fetch BMC firmware from the **wrong (eu-de-1) repo
server** instead of its own regional one (`repo.qa-de-1.cloud.sap`).
Fixing this by hand would mean editing every model entry per cluster
values file — error-prone and not maintainable across 20+ clusters.

### Solution: global vars + URL composition in the template

**1. New global values** (already used elsewhere, e.g. cluster type/region),
overridable per cluster:

```yaml
global:
  region: eu-de-1
  tld: cloud.sap

firmwareRepo:
  scheme: https
  host: repo
  hostOverride: ""   # escape hatch, see below
```

**2. Values now only store the path fragment**, not the full URL
(done for BMC so far):

```yaml
bmc:
  imagePath: "/hw-firmware/dell/poweredge-r7615/iDRAC-with-Lifecycle-Controller_..."
```

**3. The template composes the URL at render time**
(`templates/bmc_versions.yaml`):

```gotemplate
{{- $firmwareHost := $.Values.firmwareRepo.hostOverride | default (printf "%s.%s.%s" $.Values.firmwareRepo.host $.Values.global.region $.Values.global.tld) }}
...
URI: {{ printf "%s://%s%s" $.Values.firmwareRepo.scheme $firmwareHost $entry.bmc.imagePath | quote }}
```

For a cluster with `global.region: qa-de-1`, this automatically resolves to:

```
https://repo.qa-de-1.cloud.sap/hw-firmware/dell/poweredge-r7615/...
```

**4. Escape hatch for special cases:** if a cluster ever needs a
completely different firmware repo server (not just a different region),
`firmwareRepo.hostOverride` can be set to fully override the composed
host — without requiring Go template syntax inside values files.
`firmwareRepo.scheme` is applied independently of `hostOverride`, so both
can be combined freely (e.g. `http` + a custom host).

### Benefits

- A single source of truth (`global.region` / `global.tld`) — already
  maintained per cluster for other purposes — now also drives the correct
  firmware repo server, with zero extra config in the common case.
- Cluster values in `foundation-deployment` don't need to touch
  `firmwareRepo` at all for the default case.
- Special cases remain fully supported via `hostOverride` / `scheme`.

### Testing

- Added/updated unit tests in `tests/bmc_versions_test.yaml` covering:
  default composition, `global.region` override, `firmwareRepo.hostOverride`,
  and `firmwareRepo.scheme` override. Tests use a synthetic
  `bmc.imagePath` value so they stay independent of the real firmware
  data in `values.yaml`.
- Verified end-to-end against a local Kind cluster: `helm install` the
  chart and confirmed the rendered
  `BMCVersionSet.spec.bmcVersionTemplate.image.URI` matches the expected
  composed URL.

### Scope / current status

- ✅ **BMC** (`bmc_versions.yaml` + `bmc.imagePath`) fully migrated and
  verified.
- ⏳ **BIOS** (`bios_versions.yaml` + `bios.imageURI`) is **not yet migrated**
  and still uses the old, fully hard-coded URL. This is the natural next
  step, following the same pattern as BMC.


system/metal-settings/README.md
```diff
@@ -173,13 +173,17 @@ machines:
...
         bmc:
           version: "7.00.00.183"   # Target BMC firmware version gate.
           settingsFileName: dell-default.yaml  # File under bmc_settings/ to use for BMCSettingsSet.
-          imageURI: "https://..."  # Firmware image download URL for BMCVersionSet.
+          imagePath: "/hw-firmware/..."  # Path appended to the composed firmwareRepo URL for BMCVersionSet.
 ```
 
+The full image URL is composed at render time from `firmwareRepo.scheme`,
+`firmwareRepo.host` + `global.region` + `global.tld` (or `firmwareRepo.hostOverride`
+if set), and this `imagePath`. See [`firmwareRepo`](#firmware-repository-firmwarerepo) below.
```
atd9876, line 185: This appears to link to nowhere

system/metal-settings/templates/bmc_versions.yaml
```diff
@@ -1,13 +1,14 @@
...
+{{- $firmwareHost := $.Values.firmwareRepo.hostOverride | default (printf "%s.%s.%s" $.Values.firmwareRepo.host $.Values.global.region $.Values.global.tld) }}
 {{- range $vendor, $vendorConfig := .Values.machines }}
 {{- if $vendorConfig }}
 
 {{/* Vendor filtering */}}
 {{- if include "metal-settings.shouldInclude" (dict "item" $vendor "filters" $.Values.bmcVersions.filters.vendors) }}
 {{- $processedModels := dict }}
 {{- range $entryName, $entry := $vendorConfig.models }}
-{{- if and $entry.bmc $entry.bmc.imageURI }}
+{{- if and $entry.bmc $entry.bmc.imagePath }}
```
atd9876, line 11: Silent no-op for stale imageURI overrides

Entries that still set imageURI instead of imagePath are now skipped without any error.

atd9876:
Is the statement in description valid?

⏳ BIOS (bios_versions.yaml + bios.imageURI) is not yet migrated
and still uses the old, fully hard-coded URL. This is the natural next
step, following the same pattern as BMC.

system/metal-settings/templates/bmc_versions.yaml
nagadeesh-nagaraja: ya, this is important as the kube secret repo and other places might still be hardcoding the imageURI.

nagadeesh-nagaraja (approved):
with nitpick of addressing the older imageURI being hardcoded in other repo. we need to throw some error or handle it smoothly.

